PR #275, opened 2014-12-08
CP-10204: XenCenter: guest RDP control

cheng--zhang:
This code change is to add changes as bellow:
1. The option "Enable Remote Desktop console scanning"change to"Enable Remote Desktop console scanning (XenServer 6.5 and earlier)"
2. Remove connection between "Automatically switch to the Remote Desktop console when it becomes available" and "Enable Remote Desktop console scanning"
3. Add "Turn on Remote desktop"

One thing need to modify is that "CreedenceorGreater" funtion check should be change.


cheng--zhang:
Hi @stephen-turner & @huizh 
I already change the API interface to vm_call_plugin.(Why using this name can refer to https://github.com/xapi-project/xapi-project.github.io/pull/12#issuecomment-67524375)
Also about the windows firewall and watch RDP status issue, I already made pull request to win-guestagent, and the code change already merged into master.

So, @stephen-turner do you think we can start the code review now?

stephen-turner:
Thank you, Cheng. So just to be clear, with your changes to the guest agent, will XenCenter now get notified even if RDP is turned on or off directly in the guest?

If so, we will proceed with reviewing the pull request.

cheng--zhang:
Hi @stephen-turner,
Yes, with the change of windows pv tool which I mentioned as guest agent, RDP status change can be notified.

stephen-turner:
We're planning to review this on Thursday. Sorry, we don't have any free slots today or tomorrow with all the right people.

XenModel/Messages.resx
```diff
@@ -11406,6 +11409,9 @@ To learn more about the XenServer Dynamic Workload Balancing feature or to start
   <data name="VNC_RDESKTOP" xml:space="preserve">
     <value>Sw&amp;itch to Remote Desktop</value>
   </data>
+  <data name="VNC_RDESKTOP_TRUN_ON" xml:space="preserve">
```
stephen-turner: Spelling mistake: should be TURN not TRUN

XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -768,7 +779,7 @@ private string GuessNativeConsoleLabel(VM source)
                     if (osString != null)
                     {
                         if (osString.Contains("Microsoft"))
-                            label = UseRDP;
+                            label = (Helpers.CreedenceOrGreater(targetHost) && !RDPEnabled) ? enableRDP : UseRDP;
```
stephen-turner: 1) We now have different version numbers on trunk so we can use CreamOrGreater() instead of CreedenceOrGreater()
2) The construction "Helpers.CreedenceOrGreater(host) && !RDPEnabled" occurs several times, and it might be better to extract it into a separate function

XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -1133,7 +1153,25 @@ private void toggleConsoleButton_Click(object sender, EventArgs e)
...
+                            ThreeButtonDialog d = new ThreeButtonDialog(
+                                new ThreeButtonDialog.Details(System.Drawing.SystemIcons.Warning, Messages.FORCE_ENABLE_RDP),
+                                   "EnableRDPonVM",
+                                new ThreeButtonDialog.TBDButton(Messages.YES, DialogResult.Yes),
+                                new ThreeButtonDialog.TBDButton(Messages.NO, DialogResult.No));
+                            if (d.ShowDialog(Program.MainWindow) == DialogResult.Yes)
+                            {
+                                Session session = source.Connection.DuplicateSession();
+                                Dictionary<string, string> _arguments = new Dictionary<string, string>();
+                                XenAPI.VM.call_plugin(session, source.opaque_ref, "guest-agent-operation", "request-rdp-on", _arguments);
```
stephen-turner: The design asked for us to switch RDP on, wait for it to become enabled, and then connect via RDP. I don't see that happening.

XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -1133,7 +1153,25 @@ private void toggleConsoleButton_Click(object sender, EventArgs e)
...
                     vncScreen.UseVNC = !vncScreen.UseVNC;
                     vncScreen.UserWantsToSwitchProtocol = true;
-                    
+
+                    if (Helpers.CreedenceOrGreater(targetHost))
+                    {
+                        if (!RDPEnabled)
+                        {
+                            ThreeButtonDialog d = new ThreeButtonDialog(
+                                new ThreeButtonDialog.Details(System.Drawing.SystemIcons.Warning, Messages.FORCE_ENABLE_RDP),
```
stephen-turner: Should have a question mark icon, not a warning icon.

XenModel/XenAPI/VM.cs
```diff
@@ -3798,6 +3798,20 @@ public virtual vm_power_state power_state
         private vm_power_state _power_state;
 
         /// <summary>
+        /// Call a XenAPI plugin on this vm
```
stephen-turner: Here and in Proxy.cs, please use the SDK that has been built by the build system. That will ensure that the calls don't move later.

XenAdmin/TestResources/api-version.xml
```diff
@@ -160,5 +160,6 @@
     <call name="secret_set_value" />
     <call name="secret_get_all" />
     <call name="secret_get_all_records" />
+    <call name="vm_call_plugin" />
```
stephen-turner: This file and callVersions.csv need to be autogenerated: I think you have a bug here that the call is in the wrong server version. But in any case, they are very out of date, so I suggest just ignoring them and not making any changes to them.

XenAdmin/ConsoleView/XSVNCScreen.cs
```diff
@@ -686,10 +686,15 @@ private bool InDefaultConsole()
...
+            {
+                parentVNCTabView.EnableToggleVNCButton();
+            }
+            else if (InDefaultConsole())
+            {
                 parentVNCTabView.DisableToggleVNCButton();
             }
+            
 
             //Start the polling again
```
stephen-turner: The design asks for polling not to happen for Cream and later servers. Now that you have modified the guest agent to notice changes to RDP, we should never need to poll.

XenAdmin/ConsoleView/VNCTabView.cs
cheng--zhang: I still cannot find CreamOrGreater in master branch. Do I still need to wait for some time?

XenModel/XenAPI/VM.cs
cheng--zhang: Hi @stephen-turner 
Do you mean we can just copy VM.cs and Proxy.cs from build system?
But currently we do not have a release version which include these interface, do we still need to wait for a specific version?

stephen-turner: You can use the one which was built on trunk, or on the trunk-ring3 branch if it's not yet on trunk. We will update it closer to release time anyway.

stephen-turner:
You asked about CreamOrGreater(): feel free to write this function.

Let us know when you want us to review it again.

cheng--zhang:
Hi @stephen-turner , thank you for you reminder.
I doing following changes to meet your conditons:
When RDP status is on and currently the VM using VNC, these two condition you mentioned will be checked to make decision about try to connect RDP or not.
Those logic is also in guestMetrics_PropertyChanged

Current, all rework jobs are been done based on all your comments, and I think we can start the review process again. Thanks a lot.

stephen-turner:
Sorry I didn't notice before, but this pull request accidentally introduces a new file, XenModel/XenAPI-Extensions/Proxy.cs. Please remove it. (It's large, so I don't even want it added and then deleted again; please make a new, squashed changeset without it). Thanks.

XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -741,6 +761,11 @@ public void Unpause()
                 vncScreen.Unpause();
         }
 
+        private bool RDPDiabledOnCreamOrGreater(IXenConnection conn)
```
stephen-turner: Typo in function name: should be Disabled not Diabled.

XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -1133,7 +1168,23 @@ private void toggleConsoleButton_Click(object sender, EventArgs e)
 
                     vncScreen.UseVNC = !vncScreen.UseVNC;
                     vncScreen.UserWantsToSwitchProtocol = true;
```
stephen-turner: This variable is never set false, so once it becomes true, it stays true for ever.

XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -501,7 +508,17 @@ private void guestMetrics_PropertyChanged(object sender, PropertyChangedEventArg
         {
             if (e.PropertyName == "other")
             {
+                if (RDPEnabled)
+                {
+                    if (vncScreen.UseVNC && (tryToConnectRDP || (!vncScreen.UserWantsToSwitchProtocol && Properties.Settings.Default.AutoSwitchToRDP)))
```
stephen-turner: Could you explain in words what all the variables on this line mean? I'm wondering if it should just read 
`if (vncScreen.UseVNC && (tryToConnectRDP || Properties.Settings.Default.AutoSwitchToRDP))`
but I'm not clear what the meaning of all the variables are.

XenAdmin/ConsoleView/VNCTabView.cs
cheng--zhang: Yes, I think it is because user already made choice of using which kind of connection, if the decision has already been made, change of the status of RDP shall not affect the connection status. Am I right?
Same logic is also here:  https://github.com/xenserver/xenadmin/blob/f806228923acd7a20656e759c661040a8f0c763a/XenAdmin/ConsoleView/VNCTabView.cs#L1082

cheng--zhang: Yes, I think it is because user made choice of using which kind of connection already. It will affect the behavior of "Automatically switch to the Remote Desktop console when it becomes available" which is parameter "Properties.Settings.Default.AutoSwitchToRDP".
If this UserWantsToSwitchProtocol has been set, we will not try to automatically switch to Remote Desktop console.

XenAdmin/ConsoleView/VNCTabView.cs
cheng--zhang: I Add some comments on code, and I can explain the logic here:
If current connection is VNC (which is "vncScreen.UseVNC == true" ), will switch to RDP connection as following conditions:
1. Click "Turn on Remote Desktop" which will set tryToConnectRDP = true
2. RDP status changed by turned on RDP in-guest and with "Automatically switch to the Remote Desktop console when it becomes available" is on (which is Properties.Settings.Default.AutoSwitchToRDP = true). But if user click button to chose connection type before (which set "vncScreen.UserWantsToSwitchProtocol" = true), we will not do the automatic switch.

cheng--zhang:
Hi @stephen-turner I already check on all comments and thanks for your great work which help me a lot.
The second round of reword is done. Please help to check the comment and code again. 
Thanks a lot.

Zhengchai:
Thank you @stephen-turner Thank you for your review and @cheng--zhang for the discussion, do we have more comments for the codes change to be merged?

stephen-turner:
Sorry, I still need to do another review, but it's been a really busy week with the train planning.

stephen-turner:
I have scheduled this for Monday. Apologies for the delay.

XenAdmin/ConsoleView/VNCTabView.cs
stephen-turner: Ah yes, thank you, I understand now. I was confused because I thought that UserWantsToSwitchProtocol was a new variable in your changeset, but in fact it was there previously.

stephen-turner:
The code looks OK to me now, but I wanted to do some smoke testing too. Could you confirm which branch and build has all the changes (xapi and guest agent) on it? I tried with trunk-ring3 build 90554, but it didn't work right: I saw "switch to RDP" when RDP was turned off; no button when RDP was turned on; and I never saw "turn on RDP". But maybe I just haven't got the guest agent changes.

cheng--zhang:
Hi @stephen-turner, I am so sorry about the test on trunk-ring3. Because this feature also need code change for Windows PV tool, the commit is not merged in trunk-ring3, so the RDP status in trunk-ring3 will not be the correct one. You can use trunk branch build as a test environment. The trunk branch included all changes for this feature.
I also equipped my DT box dt65 with the latest trunk build for your reference.
Thanks.

stephen-turner:
OK, thanks, I'll try that.

stephen-turner:
This still isn't working well for me. It doesn't appear to be responding to changes made via the in-guest UI. And when I press "Turn on XenDesktop", the button gets greyed out, but it never connects. I have also seen a state where the remote desktop option is set, but with a warning that I need to configure the Windows firewall too, which I thought is supposed to happen automatically.

I think we should have a meeting to discuss this, rather than continuing to do it in the pull request. I'll set one up for tomorrow.

stephen-turner:
It turns out that the problems were caused by not having all the guest agent changes on the server I was testing on. It looks good now, so I'm accepting this pull request. There are some minor bugs, but we will take them as separate items.

PR #283, opened 2014-12-19
CA-146821: Accept IPv6 address when configuring WLB

kaifengz:
The WLB address should be either an IPv4/IPv6 address or a valid host name.

Signed-off-by: Kaifeng Zhu <kaifeng.zhu@citrix.com>


stephen-turner:
Looks good to me, merging.

PR #297, opened 2015-01-21
CP-10833: New licensing text in XenCenter

MihaelaStoica:


stephen-turner:
Suggest some more tweaks to text in the pool join UI:

FROM: This server does not have the same XenServer license edition as the pool master.
TO: This server does not have the same license as the pool master.

FROM: You are attempting to add the server '{0}' to a pool that has had its license upgraded.
TO: You are attempting to add the server '{0}' to a licensed pool.
[4 similar instances]

stephen-turner:
FROM: The license for the edition you requested is not available.
TO: The license type you requested is not available.

stephen-turner:
We have agreed with PM that we can remove the superfluous "Citrix" at the start of all license names.

stephen-turner:
All looks fine to me now, as far as I can see from a bit of dev testing. Merging.

PR #298, opened 2015-01-22
CP-10204 and CP10955 For Windows RDP feature

cheng--zhang:

Change message from "Would you like to turn on Remote Desktop in this VM, and then connect to it over Remote Desktop?" to "Would you like to configure this VM to accept Remote Desktop connections, and then connect to it over Remote Desktop?"

### DIFF
--- a/XenAdmin/ConsoleView/VNCTabView.cs
+++ b/XenAdmin/ConsoleView/VNCTabView.cs
@@ -84,6 +84,8 @@ namespace XenAdmin.ConsoleView
 
         private bool RDPControlEnabled { get { return source != null ? source.RDPControlEnabled : false; } }
 
+        public bool IsRDPControlEnabled() { return RDPControlEnabled; }
+
         public VNCTabView(VNCView parent, VM source, string elevatedUsername, string elevatedPassword)
         {
             Program.AssertOnEventThread();
@@ -543,7 +545,7 @@ namespace XenAdmin.ConsoleView
                 {
                     log.DebugFormat("'{0}' console: Starting RDP polling. (RDP polling is enabled in settings.)", source);
                     toggleConsoleButton.Visible = true;
-                    if(Helpers.CreamOrGreater(source.Connection))
+                    if(Helpers.CreamOrGreater(source.Connection) && RDPControlEnabled)
                         toggleConsoleButton.Enabled = true;
                     else
                         toggleConsoleButton.Enabled = false;
@@ -1192,7 +1194,7 @@ namespace XenAdmin.ConsoleView
                         }
                     }
 
-                    if (vncScreen.rdpIP == null && vncScreen.UseVNC && Properties.Settings.Default.EnableRDPPolling && (!Helpers.CreamOrGreater(source.Connection) || tryToConnectRDP))
+                    if (vncScreen.rdpIP == null && vncScreen.UseVNC && Properties.Settings.Default.EnableRDPPolling && (!(Helpers.CreamOrGreater(source.Connection) && RDPControlEnabled) || tryToConnectRDP))
                     {
                         toggleConsoleButton.Enabled = false;
                     }
--- a/XenAdmin/ConsoleView/VNCTabView.cs
+++ b/XenAdmin/ConsoleView/VNCTabView.cs
@@ -82,6 +82,8 @@ namespace XenAdmin.ConsoleView
 
         private bool RDPEnabled { get { return source != null ? source.RDPEnabled : false; } }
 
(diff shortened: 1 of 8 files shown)


stephen-turner:
This change has improved the button updating, but it seems to have broken the following use case:
1. The "automatically switch" option should be off.
2. RDP should be turned off in the guest.
3. Press the "Turn on RDP" button and accept the confirmation.
4. Once RDP is on, we should automatically switch to using it, but we appear not to do so: instead the button just changes to "Switch to RDP"

cheng--zhang:
Hi @stephen-turner 
If user click "Turn on RDP", it will need some time to complete the RDP enable, so if we switch to RDP imminently, the RDP connection may not be switched successfully because the operation may have not been complete.
Instated synchronize operation, we do it asynchronously. We will just mark the parameter "tryToConnectRDP" as "true" and switch to RDP when RDP status change been notified to XenCenter.

stephen-turner:
Yes, I understand that, but it didn't seem to be working with this new code. It never switched, even after the button had changed to "Switch to RDP".

Zhengchai:
Hi Stephen, sorry for this, which PV driver are you using with the test? the trunk only contains the Win XP PV driver codes, PV driver for Win7 is not there yet, and we are asking Ben about how to line up the PV driver with same branch. We will double confirm this with Cheng on Monday morning here.

stephen-turner:
I was using Cheng's dt box still.

stephen-turner:
If I'm wrong, let me know, and I'll retest it.

cheng--zhang:
Hi @stephen-turner 
I retry this test case as you provide. And I got below results:
1. After click "Turn on RDP", the result will be the same as you described. Why RDP do not connect directly is because when XenCenter try to use RDP the first time, it will do "PollPort" for RDP port, which will use some time to check the port for all network interface.
2. After PollPort, which in my test environment is 15 seconds, connection will switch to RDP, button will be enable again and shows "Switch to Default".
3. Why there is such a long time waiting is because in my test environment DT65, VM "windows 7 with PV", I configured two network interfaces: network0 and network1. network1 is a local network, it cannot connect with RDP. But we will try to connect to RDP use network1 first time. I disable network1 in my test VM, and the switch can go more fluently(less than 2 seconds).

I just reproduce this case by your steps, if anything I missed, please let me know.

stephen-turner:
Thanks for your comments, Cheng. I'm off work today, but I'll look at them later.

cheng--zhang:
Hi @stephen-turner 
I add a commit for below reasons:
1.Add xapi version check in cream since Cream and Creedence have the same platform version
2.Check feature-ts2 to see if turn on RDP feature can use on specific VM PV tool
Could you please help to review.
For my self I also provide environment DT65, so you can have a smoke test if you want.
There are 3 VMs on DT65 as following configuration:
1. Win 7 installed with Creedence PV tools
2. Win 7 installed with Cream PV tools
3. Windows XP installed with Cream PV tools.

Any problem please let me know. Thanks.

stephen-turner:
I'm trying to look at this, but dt65 is empty.

stephen-turner:
I haven't done any dev testing yet because dt65 is not ready. But I have read the code, and I have one comment. You have three more uses of CreamOrGreater(), apart from the one in CanEnableRDPOnCreamOrGreater(). I assume that some or all of them should change to use the new feature-ts2? For example, I'm fairly sure that RDP port scanning should still be done for VMs with an old guest agent.

stephen-turner:
When you set up dt65, could you kindly include two networks on the Creedence Win 7 VM, so that I can retest the problem I found before? Thank you.

cheng--zhang:
Hi @stephen-turner 
Thanks a lot for your comment.
1. I already change the logic on each CreamOrGreater so you can check it now.
2. About the environment empty issue, I send you a email with screenshot to make sure we are using the same DT box. 
3. To reproduce the problem you found before, I create each Win 7 VMs with 2 network interfaces. To make sure you can reproduce, this problem happens only the first time switch to RDP, after that, you need to disconnect XenServer and reconnect it to reproduce again.
Thanks.

stephen-turner:
Yes, I'm seeing dt65 now. I've no idea what was wrong yesterday.

I'm interviewing this morning, but I'll try and look at it this afternoon. Thank you.

cheng--zhang:
@stephen-turner 
Have you try the DT65 already? If you done, please let me know, I may use DT65 to do other testing.
Thanks.

stephen-turner:
Thanks, I'm happy with this now.

stephen-turner:
Merged.

PR #301, opened 2015-01-23
CAR-1732: Supp Pack installation through XenCenter

MihaelaStoica:
1. Select and upload a supp pack in the hotfix wizard:
    Select a supp pack in the hotfix wizard; 
    Recognise that it's a supp pack not a hotfix; 
    Check if there's enough space on a SR to contain it: (1) default SR if shared or (2) any shared SR or (3) local SR on each host in pool; 
    Create a VDI to contain it; 
    Upload the supp pack to the VDI.
2. Install the supp pack from the VDI
3. Delete the temporary VDI after installation
4. Fix error in the wizard: Update the Patch property even when the selected update type is not "Existing patch". This will be set it to null if the selected type is supplemental pack
5. Changed the CreamOrGreater function to use the API version
6. Make the feature only available for Cream or greater hosts (other hosts are disabled on the Select Server page)
7. Add other_config entry for the temporary VDI


GaborApatiNagy:
Review findings:

Agreed to be fixed:
-HTTP put should use default timeout
-UploadSupplementalPlanAction: RemoveVDI function should just throw, there is no need to encapsulate the current Exception in a new
-GetUpdateName: cosmetic changes
-UploadSupplementalPackAction: Add other_config key before uploading to make sure it has this flag from the beginning of its lifetime. This flag is there for further use to mark these vdis (should we fail to remove them by XC)

Candidates for further improvements:
-Upload progress bar can move from 0 to 100% multiple times (number of hosts) per row - this can be confusing. If we aim to show the progress by row, the progressbar should reach 100% only once per row.
-Changing 'Hide successful tests' checkbox causes the progress bar to move from 0 to 100% again. It seems this checkbox is causing the page to reload or the pre-checks to be re-executed. This setting/checkbox should only change the displayed items above. 
-Cancel button is greyed out when installing a patch or a supp pack, but users can still click on the Close [X] button of the wizard. Do we want to show them a confirmation dialog at least or just cancel the Close event?

MihaelaStoica:
Added a new commit that includes changes following code review. 
The bits not covered here are:

_-Changing 'Hide successful tests' checkbox causes the progress bar to move from 0 to 100% again. It seems this checkbox is causing the page to reload or the pre-checks to be re-executed. This setting/checkbox should only change the displayed items above.
-Cancel button is greyed out when installing a patch or a supp pack, but users can still click on the Close [X] button of the wizard. Do we want to show them a confirmation dialog at least or just cancel the Close event?_

These are generic improvements to the wizard and will deal with them separately.

PR #303, opened 2015-01-28
CP-10825: Supp Pack installation through XenCenter: Display installed supplemental packs on host's General tab

MihaelaStoica:
The 'Updates' section on host's General tab displays which packs have been installed onto a host, including the version number


GaborApatiNagy:
Looks good

PR #313, opened 2015-02-05
CP-10988: Fix the issue where VMs running containers appear twice in the...

MihaelaStoica:
... treeview (and at the wrong level) if the host is in a pool


GaborApatiNagy:
great

PR #315, opened 2015-02-05
CP-10797: Ensuring sufficient disk space for service pack installation: ...

MihaelaStoica:
...In the Install Update Wizard check if there is enough space to upload a hotfix

- Before starting the upload to the master hosts, check if there is enough disk space (check only performed for Cream or greater hosts)
- If enough space available the upload starts automatically; otherwise an error is displayed
- If we can free up enough disk space then we offer the option to Clean up. Otherwise we provide the user with the information on required and available space and the user will have to manually free up required space.

- Also Disable the Upload page for oem updates


GaborApatiNagy:
Looks good

PR #317, opened 2015-02-06
CP-10900: Add functionality for enabling/disabling/querying the containe...

MihaelaStoica:
...r monitoring on a per-VM-basis to XenCenter

New Page: Enlightenment on VM Properties Dialog, visible only for:
- VMs on Cream Or Greater hosts
- that can be enlightened, regardless of their power state

A VM can be enlightened if the key "xscontainer-monitor" exists in other_config.
A VM is enlightened if other_config:xscontainer-monitor if true.


MihaelaStoica:
These conditions are also checked before adding the page to the Properties dialog, so the page should not be displayed if these conditions are not met. But as all the settings panels can be reused, we want to make sure that they are not reused for the wrong objects. 
If these assertions fail, they fail because we are trying to use the page in a wrong way, and the issue will be caught during development or QA.

GaborApatiNagy:
That makes sense. I thought about displaying a not user friendly error
message without logging it might not be what we want here, but I see now
that cannot happen here. Thanks

GaborApatiNagy:
Merging

PR #320, opened 2015-02-09
CA-159907: Fix refresh issues on containers

MihaelaStoica:
- when we update the containers, we should update existing objects (if they exist), instead of always creating new ones.


GaborApatiNagy:
Tested, seems to work fine

PR #321, opened 2015-02-10
CP-10929: On the General tab, Update 'Properties' button visibility for ...

MihaelaStoica:
...all object types

This fixes the following issue: when displaying the General tab of a container, the Properties button is hidden, but not made visible again when switching to another object type (e.g. VM)


GaborApatiNagy:
Seems correct

PR #322, opened 2015-02-11
Normalizing the line endings and another fix in the tree refresh for containers

MihaelaStoica:
- Changing line endings back to Windows type
- Fix another refresh issue on containers


GaborApatiNagy:
Looks fine

PR #323, opened 2015-02-11
CP-10927: Add container restart operation and add all lifetime operations to the toolbar

MihaelaStoica:


GaborApatiNagy:
looks good

PR #324, opened 2015-02-11
CP-11052: Add per-container "Processes"-Tab to XC

huizh:
This tab will show docker process id, command and CPU time.

### DIFF
--- a/XenAdmin/Help/HelpManager.resx
+++ b/XenAdmin/Help/HelpManager.resx
@@ -1041,6 +1041,9 @@
   <data name="TabPageGPU" xml:space="preserve">
     <value>9901</value>
   </data>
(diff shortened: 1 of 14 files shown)


MihaelaStoica:
This looks good. Some minor issues I've observed: 
1. On the DockerProcess page, please make sure that all columns are visible (without horizontal scroll) when the window is resized to minimum size. We can also make the columns resizable.
2. Please leave a bit more space between the grid and Refresh button (see SrStoragePage for an example)

huizh:
Thanks for the review and comments.
Here I upload the new commit to address the comments.
Could you please help review it?
Much appreciate.

MihaelaStoica:
Looks good. Merging.

huizh:
Thanks a lot.

PR #325, opened 2015-02-12
CP-11053: Add per-container "Details"-Tab

cheng--zhang:
- Showing all detailed information from a docker_inspect call
- The information refresh/polls every 20s while it's open. Also shows time when the last refresh happens.
- Button “Refresh” add, so user can refresh the result. Also shows time when the last refresh happens.
- Add TabPageDetails as value 9903 in HelpManager.resx

### DIFF
--- a/XenAdmin/Help/HelpManager.resx
+++ b/XenAdmin/Help/HelpManager.resx
@@ -426,6 +426,9 @@
   <data name="TabPageCPUMemory" xml:space="preserve">
     <value>1040</value>
   </data>
+  <data name="TabPageDockerDetails" xml:space="preserve">
+    <value>9903</value>
+  </data>
   <data name="TabPageHA" xml:space="preserve">
     <value>1095</value>
   </data>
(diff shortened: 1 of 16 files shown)


MihaelaStoica:
1. Could we call the new page DockerDetailsPage (instead of DetailsTabPage, which I think is too generic, and also to be consistent with the other tab page recently added, DockerProcessPage)?

2. On the Details page, the refresh button (and label) should be located below the tree (to be consistent with the Processes tab page).

3. The resource strings LAST_REFRESH_FAIL and LAST_REFRESH_SUCCESS have been added as part of the work for the "Processes" tab (now merged into master), so please exclude changes to the Messages file from this commit.

Everything else looks fine. Thanks.

cheng--zhang:
Hi @MihaelaStoica 
I already change the code based on your comments, could you please help to review again?
Thanks a lot.

MihaelaStoica:
Thanks Cheng for the rework. Merging

PR #326, opened 2015-02-12
CP-11039: Change VM.CanHaveVGpu to use the new allow-vgpu flag

MihaelaStoica:
-  A VM can have vGPU if it is HVM and the the allow-vgpu restriction is absent or is non-zero
- allow-vgpu is 0 for HVM Linux VMs


GaborApatiNagy:
Looks good

PR #331, opened 2015-02-13
A-160189: Change RDP button display on Console tab

cheng--zhang:
Remove “Remote access is not enabled on this guest” in cream in below condition:
1.RDP status is on
2.PV driver is cream or greater


stephen-turner:
I don't like the way the tooltip is done here (putting it in the Designer, and then removing it when it doesn't apply). However, that's not a new problem and your changeset looks fine. Merging.

PR #332, opened 2015-02-13
CA-161001: Display Created field as DateTime on Docker General Tab.

huizh:


MihaelaStoica:
Looks good. Merging

PR #333, opened 2015-02-13
CP-10926:  Add dialog for customizing the config drive in XenCenter after VM creation; +CA-161002; + CP-11159

MihaelaStoica:
- The cloud config parameters page is only visible on a VM that already has a config drive.
- The cloud config parameters can only be changed when the VM is halted.
- In the New VM wizard, if a template is selected that already has a config drive, them the wizard will display the existing configuration, not the default one.

Also included in this pull request: 
- CA-161002: Internationalize strings on the Cloud Config Parameters page and CreateVMAction
- CP-11159: Only display the Cloud Config page if the "xscontainer" supp-pack is installed


XenAdmin/Wizards/NewVMWizard/Page_CloudConfigParameters.cs
```diff
@@ -129,11 +152,85 @@ private void GetDefaultParameters()
...
+
+        private void GetDefaultParameters()
+        {
+            GetCloudConfigParameters(null);
+        }
+
+        public void ShowHideButtonsAndWarnings(bool inNewVmWizard)
+        {
+            // IncludeConfigDriveCheckBox and reloadDefaults only visible in the New VM Wizard
+            IncludeConfigDriveCheckBox.Visible = reloadDefaults.Visible = inNewVmWizard;
```
GaborApatiNagy: I would suggest IncludeConfigDriveCheckBox to be visible, ticked, but greyed out

XenModel/Utils/Helpers.cs
```diff
@@ -1934,5 +1934,11 @@ public static bool VLAN0Allowed(IXenConnection connection)
            // For Creedence or later on the vSwitch backend, we allow creation of VLAN 0
            return master != null && CreedenceOrGreater(master) && master.vSwitchNetworkBackend;
        }
+
+       public static bool ContainerCapability(IXenConnection connection)
```
GaborApatiNagy: This looks good, but is it sufficient to check only for the plugin or do we need a Cream or greater check here as well? (Or are we sure this plugin cannot be installed on any pre-Cream host?)

MihaelaStoica:
Following code review, I've added a check for CreamOrGreater to the container capability check

GaborApatiNagy:
Looks good now

PR #334, opened 2015-02-14
CA-161097: Pause refresh on DockerProcessPage when page not visible

robertbreker:
Also rename timer1 to RefreshTimer to improve readability.

Signed-off-by: Robert Breker <robert.breker@citrix.com>

### DIFF
--- a/XenAdmin/MainWindow.cs
+++ b/XenAdmin/MainWindow.cs
@@ -1870,6 +1870,11 @@ namespace XenAdmin
             else
                 SearchPage.PanelHidden();
 
+            if (t == TabPageDockerProcess)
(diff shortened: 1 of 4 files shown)


MihaelaStoica:
Gabor is right here. The call to the plugin can freeze the UI, so we shouldn't call the plugin on the UI thread. We should instead put the call in an action (or use existing ExecutePluginAction), execute the action async and on completion update the UI. I missed this in the initial code review, but I think we should fix it now.
Same for the Details page.

robertbreker:
Agreed. It makes sense to not have any of this in the UI thread.
Let's get this done properly.

PR #339, opened 2015-02-18
CA-161097: Pause the automatic refresh on Container processes page when the page is not visible

MihaelaStoica:
- Instead of calling the plugin on the UI thread, we call it through an action that we execute asynchronously  and on completion update the UI.
- Pause the refresh timer on leaving the page and resume it when entering the page again.
- We do this for both Processes and Details page.

Also fixing a Null reference exception on General tab.

### DIFF
--- a/XenAdmin/TabPages/DockerDetailsPage.cs
+++ b/XenAdmin/TabPages/DockerDetailsPage.cs
@@ -44,17 +44,17 @@ namespace XenAdmin.TabPages
     {
         private const int REFRESH_INTERVAL = 20000;
 
-        private IXenObject _xenObject;
-        private VM _vmResideOn;
-        private Host _hostResideOn;
-        private string _resultCache;
+        private IXenObject xenObject;
+        private VM parentVM;
+        private Host host;
+        private string cachedResult;
 
         public IXenObject XenObject
         {
             get
             {
                 Program.AssertOnEventThread();
-                return _xenObject;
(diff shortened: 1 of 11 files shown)


GaborApatiNagy:
Looks good, but we found the following in code review for both Processes and Details tabs:
-Refresh can be clicked when a refresh is already in progress, that would execute multiple actions. I think disabling the Refresh button would be good (eg. as in check for updates)
-If the user switches to an other container when an action from the previously visited container is in progress, the new container's page will be updated with the results that belong to the previous container.

XenAdmin/TabPages/DockerDetailsPage.cs
```diff
@@ -41,47 +43,76 @@ namespace XenAdmin.TabPages
...
 
                 if (value == null)
                     return;
 
-                if (_xenObject != value)
+                if (xenObject != value)
                 {
-                    _xenObject = value;
-                    if (_xenObject is DockerContainer)
+                    xenObject = value;
```
GaborApatiNagy: I know this was existing code, but I have just now found the following that would be nice to be fixed.

xenObject will be assigned even if it is not DockerContainer. I think an assertion would be good to have here.

GaborApatiNagy:
Everything else looks good, I am fine with the whole changeset.

GaborApatiNagy:
Good stuff

PR #342, opened 2015-02-18
CP-10772: Display the read caching status

MihaelaStoica:
- Show a “Read Caching” section on the General Tab of a VM when the VM is running on a Cream or greater host
- When the Read Caching is enabled, the “Read Caching” section will contain two entries: Status and Disks (list of disks with RC enabled)
- When the Read Caching is disabled, the “Read Caching” section will contain two entries: Status and Reason

(Also fixed the sorting of Messages.resx)


GaborApatiNagy:
:+1:

stephen-turner:
Looks good now, although not dev tested with a real server because the storage-side code is not complete yet.

PR #343, opened 2015-02-20
CA-161460: XenCenter should fail gracefully if patch download fails

stephen-turner:


stephen-turner:
NB It's probably easiest to review each changeset in this pull request separately.

MihaelaStoica:
Looks good

PR #345, opened 2015-02-23
CA-160996: XenCenter: Subscribe to fewer events in DockerContainers

GaborApatiNagy:
-ServerXenObject-PropertyChanged eventhandler now returns immediately for a non-VM sender
-Changed IXenObject to VM wherever possible to explicitly state that those methods are expecting VMs
-The UpdateDockerContainersForVM method sets the dockerContainersChanged flag only if there are changes

### DIFF
--- a/XenModel/DockerContainers.cs
+++ b/XenModel/DockerContainers.cs
@@ -106,20 +106,20 @@ namespace XenAdmin.Model
         {
             InvokeHelper.AssertOnEventThread();
 
-            IXenObject ixmo = e.Element as IXenObject;
-            if (ixmo == null)
+            VM vm = e.Element as VM;
+            if (vm == null)
                 return;
 
             switch (e.Action)
             {
                 case CollectionChangeAction.Add:
-                    ixmo.PropertyChanged += ServerXenObject_PropertyChanged;
-                    UpdateDockerContainer(ixmo);
+                    vm.PropertyChanged += ServerXenObject_PropertyChanged;
+                    UpdateDockerContainer(vm);
                     break;
 
                 case CollectionChangeAction.Remove:
-                    ixmo.PropertyChanged -= ServerXenObject_PropertyChanged;
-                    RemoveObject(ixmo);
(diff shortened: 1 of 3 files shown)


XenModel/DockerContainers.cs
```diff
@@ -106,20 +106,20 @@ private static void CollectionChanged(Object sender, CollectionChangeEventArgs e
         {
             InvokeHelper.AssertOnEventThread();
 
-            IXenObject ixmo = e.Element as IXenObject;
-            if (ixmo == null)
+            VM vm = e.Element as VM;
+            if (vm == null)
```
stephen-turner: Is it the case that only VMs can ever reach here? If so, it would be better if this were an Assert?

XenModel/Network/Cache.cs
```diff
@@ -564,22 +565,23 @@ public void CheckDockerContainersBatchChange()
...
         private bool dockerContainersChanged = false;
-        public void UpdateDockerContainersForVM(IEnumerable<DockerContainer> containers, VM vm)
+        public void UpdateDockerContainersForVM(IList<DockerContainer> containers, VM vm)
         {
-            dockerContainersChanged = true;
-         
+            Trace.Assert(vm != null);
+
             //updating existing, adding new containers
+            dockerContainersChanged = dockerContainersChanged || containers.Count > 0;
```
stephen-turner: ||= would be more idiomatic here (and eleven lines below)

stephen-turner:
I think I was wrong before about which events are subscribed to, but the new code makes it a lot clearer.

XenModel/Network/Cache.cs
GaborApatiNagy: It definitely would, but it is not there in C#. 
I was not sure, so did a bit of research and it seems the bitwise operator |= could be used instead, although that one is not short circuiting as || is, so I would say I would leave it as is.

stephen-turner: Weird, why is that not in the language? Other languages have it. OK then...

PR #349, opened 2015-02-24
CA-161460: Correct help text in the case that the patch has been deleted without being installed

stephen-turner:
Plus line ending fixes in the first changeset


MihaelaStoica:
looks good

PR #353, opened 2015-02-26
CP-10926: Make the textbox on Cloud-Config Parameters page read-only (instead of disabled) if the config drive parameters cannot be changed

MihaelaStoica:


GaborApatiNagy:
:+1:

PR #354, opened 2015-02-26
CP-10926: Handle exception when running the action that gets the config drive parameters

MihaelaStoica:
- If failed to retrieve the config parameters, then make the ConfigDriveTemplate textbox read-only and display the reason why it is not editable (similar with what we do for running VMs).
- In the New VM Wizard, if the action on "Reset to Default" button fails, then popup a message box with an error saying that we were unable to retrieve the default parameters. In this case, the ConfigDriveTemplate textbox remains unchanged.


GaborApatiNagy:
New VM Wizard: I would suggest we should keep the cloud config parameters editable even if the default config could not be retrieved.

GaborApatiNagy:
Good stuff

PR #357, opened 2015-02-27
CA-161620: If the supplemental pack installation fails, then XenCenter will display a message suggesting the user refers to the users guide for installing via the CLI.

MihaelaStoica:
- When the install fails, log the full error description
- Added a specific exception for the supp pack installation failure
- Better action title (depending on the number of servers it applies to)
- Added a "More Info" button on the "Update" page, which displays a message with more information on the supp pack installation failure

### DIFF
--- a/XenModel/Actions/SupplementalPack/InstallSupplementalPackAction.cs
+++ b/XenModel/Actions/SupplementalPack/InstallSupplementalPackAction.cs
@@ -31,6 +31,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using XenAPI;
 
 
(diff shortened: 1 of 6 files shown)


XenModel/Actions/SupplementalPack/InstallSupplementalPackAction.cs
```diff
@@ -73,15 +78,20 @@ private void InstallSuppPack(Host host, VDI vdi)
             }
             catch (Failure failure)
             {
-                log.WarnFormat("Plugin call install-supp-pack.install({0}) on {1} failed with {2}", vdi.uuid, host.Name,
-                               failure.Message);
-                throw;
+                log.WarnFormat("Plugin call install-supp-pack.install({0}) on {1} failed with {2}", vdi.uuid, host.Name, failure.Message);
```
GaborApatiNagy: I think this and the following log entries made more sense as Errors, not just warnings.

GaborApatiNagy:
Looks good, merging.

PR #358, opened 2015-03-03
CP-10144: Support CIFS as an SR type in XenCenter

GaborApatiNagy:
-Added support for CIFS (SR type, SR Wizard)
-Added some Debug.Asserts to the Create SR action to watch for clear text paswords


GaborApatiNagy:
Thanks for the comments, I have fixed all of these. Could you have a look on the latest commit?

MihaelaStoica:
Looks good now. Will do the api bindings and the cifs errors as part of another work item.

PR #359, opened 2015-03-04
CA-162966: Fix the bug where the GPU value could not be updated via VM Properties for Linux HVM

MihaelaStoica:
- The problem was that we didn't create a vgpu if the vgpu_type was null. But for Linux HVM, this is gpu passthrough case.


stephen-turner:
After some discussion of alternative approaches, this looks like the simplest solution. Merging.

PR #360, opened 2015-03-05
CA-163138: Include the pre-check  for VM migration status for supplemental pack installation, as the guidance for these is to restart the host after install.

MihaelaStoica:


stephen-turner:
Looks good, merging.

PR #362, opened 2015-03-05
CA-163129: Default file-filter in Tool -> Install Update -> Add should i...

GaborApatiNagy:
...nclude supp-packs

-Now there are three filers in the open file dialog: updates&supp. packs, updates, supp. packs
-Also removed xsoem extension from the filter, because it is not supported anymore


MihaelaStoica:
Looks good

PR #364, opened 2015-03-05
CA-162989: Container Management GUI usability/homogeneity fixes

GaborApatiNagy:
Implemented changes as follows (copied from ticket):
"
    Combine "Docker Version" and "Docker Information" on the VM-General-tab into "Container Management - Docker Status" with the following fields only:
        API version
        Version
        Git Commit
        Driver
        Index Server Address
        Execution Driver
        IPv4 Forwarding
    In the "Processes" tab, change the name from "Docker Processes" to "Container Processes"
    In the "Details" tab, change the name from "Docker Detail" to "Container Details"
    In the "Details" tab, drop the top level element "docker_inspect" (XML requires a single root-node, afaik the Windows form treenode does not), or alternatively open the root node by default and rename it to "Inspect Result"
    In the "Details" tab, add the "Details"-headline in black on white - just like on the "Processes"-tab

Also, on the container's General tab, show Properties button disabled, instead on hiding it (to be consistent to other cases, e.g. disconnected servers)
"


MihaelaStoica:
Looks good

PR #365, opened 2015-03-05
CP-11372: XenCenter: Update the model with the latest api bindings, including the new cifs specific errors

GaborApatiNagy:


MihaelaStoica:
looks fine

PR #366, opened 2015-03-06
CA-162992: Better display of container network port mapping information on General tab

MihaelaStoica:
- Show ports as: Address: 0.0.0.0; Public port: 8088; Private port: 8088; Protocol: tcp (multiple lines if more than one set)


GaborApatiNagy:
same as #367

PR #367, opened 2015-03-09
CA-162992: Better display of container network port mapping information on General tab

MihaelaStoica:
- Show ports as: Address: 0.0.0.0; Public port: 8088; Private port: 8088; Protocol: tcp (multiple lines if more than one set)


GaborApatiNagy:
Looks good, but when a container is stopped the Ports section shows <None>, but when the container is running it shows the list of ports correctly. I would suggest we do not show it or show an "unavailable" message.

GaborApatiNagy:
OFF: Did you know we need to escape "\<" and "\>" characters in this comment editor?

PR #370, opened 2015-03-09
XenModel/XenAPI-Extensions/SR.cs IntelliSense-typo

GaborApatiNagy:


MihaelaStoica:
Why changing the existing function ParseSRListXML?
With this change, XenCenter throws an XmlException: Data at the root level is invalid.

GaborApatiNagy:
Hmm, I should have tried it before suggesting a change... 
I see now why it's like this. Let me close this.

PR #380, opened 2015-03-11
CA-164176: Make sure that the Install Update does not get stuck on precheck stage after cleanup on slave

MihaelaStoica:
- override the Equals function on HostOutOfSpaceProblem to check the disk space requirements, not the Description.

### DIFF
--- a/XenAdmin/Diagnostics/Problems/HostProblem/HostOutOfSpaceProblem.cs
+++ b/XenAdmin/Diagnostics/Problems/HostProblem/HostOutOfSpaceProblem.cs
@@ -115,5 +115,13 @@ namespace XenAdmin.Diagnostics.Problems.HostProblem
                 return false;
             }
         }
+
+        public override bool Equals(object obj)
+        {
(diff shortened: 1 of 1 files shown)


XenAdmin/Diagnostics/Problems/HostProblem/HostOutOfSpaceProblem.cs
```diff
@@ -115,5 +115,11 @@ public override bool IsFixable
                 return false;
             }
         }
+
+        public override bool Equals(object obj)
+        {
+            var problem = obj as HostOutOfSpaceProblem;
+            return problem != null && diskSpaceReq.Equals(problem.diskSpaceReq);
```
GaborApatiNagy: I am afraid there is a chance for getting a null reference exception. (diskSpaceReq)

MihaelaStoica:
Updated the pull request with a new commit where I've added more checks for null and also implemented Equals function in DiskSpaceRequirements class.

GaborApatiNagy:
Cool :+1:

PR #381, opened 2015-03-11
CA-164176: Make sure that the Install Update does not get stuck on precheck stage after cleanup on slave (for creedence-sp branch)

MihaelaStoica:
- override the Equals function on HostOutOfSpaceProblem to check the disk space requirements, not the Description.


MihaelaStoica:
Updated the pull request with a new commit where I've added more checks for null and also implemented Equals function in DiskSpaceRequirements class.

GaborApatiNagy:
looks good

PR #388, opened 2015-03-16
CA-164678: Got incorrect guest screen resolution with user's first switch to Remote Desktop from XenCenter Console

cheng--zhang:
If user want to switch to RDP, but the RDP IP is empty, original logic will try to connect VNC again, within the VNC re-connection, the Desktop Size lost.

In this situation, we should not switch RDP.


MihaelaStoica:
Looks good. Merging

PR #389, opened 2015-03-16
CA-164186: NFSv4: end-to-end test and clean-up

GaborApatiNagy:


MihaelaStoica:
We need to make sure that we don't pass the NFS version parameter to the reatach function, only to create.

MihaelaStoica:
looks ok

PR #393, opened 2015-03-17
CP-11032: Remove storage motion constraints: Allow cross-pool move of stopped and suspended VMs

MihaelaStoica:
- With these changes, a stopped or suspended VM can be moved across pools; this is performed as a vm migrate operation.
- The intra pool move will continue to be a vdi.copy + destroy operation for stopped VMs, but for suspended VMs we need to do it a as vm migrate operation.
- We use the Cross Pool migrate wizard for both intra and cross pool move operations, but introduced a wizard mode parameter in order to adapt the wizard to the specific operation we want to perform (migrate, move or copy)

### DIFF
--- a/XenAdmin/Commands/MoveVMCommand.cs
+++ b/XenAdmin/Commands/MoveVMCommand.cs
@@ -67,23 +67,20 @@ namespace XenAdmin.Commands
         {
             VM vm = (VM)selection[0].XenObject;
 
-            var cmd = new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection);
-            if (cmd.CanExecute())
-                cmd.Execute();
+            if (CrossPoolMoveVMCommand.CanExecute(vm, null))
+                new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).Execute();
             else
                 MainWindowCommandInterface.ShowPerXenModelObjectWizard(vm, new MoveVMDialog(vm));
         }
 
         protected override bool CanExecuteCore(SelectedItemCollection selection)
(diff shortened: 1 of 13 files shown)


XenAdmin/Commands/MoveVMCommand.cs
```diff
@@ -67,26 +67,23 @@ protected override void ExecuteCore(SelectedItemCollection selection)
...
+            var cmd = new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection);
+            if (cmd.CanExecute())
+                cmd.Execute();
+            else
+                MainWindowCommandInterface.ShowPerXenModelObjectWizard(vm, new MoveVMDialog(vm));
         }
 
         protected override bool CanExecuteCore(SelectedItemCollection selection)
         {
+            if (new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).CanExecute())
```
GaborApatiNagy: I am afraid this (creating new objects) will run too many times. 
This could be solved by a static method if it fits here

GaborApatiNagy:
I think we should show the Network Selection page to let the user to select a migration network and so we don't need to default to any network. (performance on management network)

GaborApatiNagy:
Looks good, but found two issues we are going to consider separately:
-while moving is in progress we see the same green icon (just like for live migration) that would suggest that this VM is running, but it is not. I think we need the same icon, but in red. (And maybe blue for suspended ones...)
-when the move is in progress, Start operation is allowed and it fails (but at least the migration is not affected)

PR #400, opened 2015-03-20
CA-111770: Version check messages confusing because the target version i...

stephen-turner:
...s a regexp.

Also changes to the FriendlyErrorNames.Designer.cs, which it appears was lagging the FriendlyErrorNames.resx.


GaborApatiNagy:
seems good

PR #402, opened 2015-03-23
CA-164800: New Storage wizard: hide CIFS option for old servers and not licensed pools

GaborApatiNagy:


MihaelaStoica:
Looks great

PR #407, opened 2015-03-26
CA-164819: On board VGA appears on GPU list

GaborApatiNagy:
-We used to check host.RestrictVgpu instead of checking the API version when populating the dropdown on the GPU Settings page. This is not sufficient anymore as it depends on the license too, so changed the code to check version (vGPU was introduced in Clearwater SP1).


MihaelaStoica:
Looks good

PR #410, opened 2015-03-26
CA-164816: After disabling HA on a pool, it should be possible to apply vGPU/GPU Passthrough to previously protected VMs

MihaelaStoica:
- When deciding whether to allow GPU configuration on a VM, check if the HA is enabled, not just the the VM's restart priority setting


GaborApatiNagy:
Looks good

PR #412, opened 2015-03-27
CA-165096: solve XenCenter slow to find RDP interface

cheng--zhang:
To solve this long latency issue, we will only scan IPs whose PIF link status is connected.


stephen-turner:
Looks good, merging.

PR #427, opened 2015-03-31
CA-165712: Show the GPU tab when GPU is licensed, even when there are no...

GaborApatiNagy:
... supported GPUs


MihaelaStoica:
Looks good

PR #429, opened 2015-04-08
CA-149867: Invoke on MainWindow instead of various controls

GaborApatiNagy:
We observed that some threads can reach deadlock-ish state after they have Invoked into a control's UI thread. When it happens they are all in a waiting for join or in sleep state for very long time, although there should not be any deadlock situations.
It seems this has something to do with multiple parent controls and with which control we invoked on. This should not make a difference, because we have got one UI thread (for MainWindow) they should wait for, but we have seen it does.

The solution that fixed this issue was to invoke on the MainWindow instead of various controls (see a4fe507adf68302124bfc732ac11abde946e52cd ).

This changeset is changing all our Invokes to invoke into MainWindow
instead of a control itself. (MainWindow's UI thread is the only UI thread
all Control is using in XenCenter)
This changeset should be in place until we have found the root cause or the exact reason for the above.

### DIFF
--- a/XenAdmin/ConsoleView/VNCGraphicsClient.cs
+++ b/XenAdmin/ConsoleView/VNCGraphicsClient.cs
@@ -239,10 +239,10 @@ namespace XenAdmin.ConsoleView
             Program.AssertOffEventThread();
 
             // Set the remote clipboard based on the current contents.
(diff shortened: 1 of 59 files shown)


MihaelaStoica:
Looks fine.

PR #435, opened 2015-04-16
CA-150890: Enter Maintenance Mode dialog should offer to shut down a VM ...

MihaelaStoica:
...that cannot be suspended.

- Added the "shutdown" solution and used it for VMs that cannot be suspended.
- Changed the text on the dialog from "This operation will migrate or suspend all VMs running on this server and transition it into maintenance mode" to "This operation will migrate all VMs running on this server and transition it into maintenance mode. If a VM can not be migrated you need to suspend it or shut it down"


GaborApatiNagy:
Looks good

PR #440, opened 2015-04-21
CP-11292: Remove storage motion constraints: Wording changes in the cros...

MihaelaStoica:
...s pool migrate wizard

- changes to the text to be suitable for migrate and copy operations, where possible
- added different text for copy on some pages
- also different text for single and multiple selection
- changes to text alignment, tab order, hotkeys on the copy pages
- removed Finish page from the intra-pool copy wizard


GaborApatiNagy:
Looks good, but before merging we will have a demo

PR #450, opened 2015-04-30
CA-68578: Shutting down a Linux guest with VNC console causes password d...

MihaelaStoica:
...ialog to pop up

- store the empty vnc password after a successful login, to avoid displaying the login dialog box on a subsequent reconnection attempt
- abort the vnc connection attempt if we notice that vncIP is null and switch back to the default console


GaborApatiNagy:
Looks good

PR #451, opened 2015-05-05
Creedence sp ln

vetrivelc:
Fixed L10N Issues and translation updates


stephen-turner:
Vetri,

Thank you for your diff. But I'm sorry, I can't take it as it is, because you have made a lot of changes to the English files, and in Creedence we had regressions from that. Please submit another diff without the English changes.

You seem to have changed "&gt;" to ">" in several places (in all three languages), and I'm not sure why it's necessary (or even whether it's correct). If it's not necessary, I would rather keep the diff as small as possible.

Also, could you please describe the new files you have added to the dotNetInstaller directory, and what effect they will have on the installer and on the installed product for a customer.

Thank you.

vetrivelc:
@stephen-turner , I reverted the changes to EN files. For language files I retained the changes because we are in process of automating the localization process. This time files were handled in that system and merged manually. File structure changes were updated as per the system file output to reduce the overheads in future releases when full automation is in place.

I have validated the changes by generating local build in VS IDE and it got successful. So, I kindly request you to approve and merge these changes.

stephen-turner:
Thanks, Vetri. Could you tell me about the new files in the dotNetInstaller directory too? Thanks.

vetrivelc:
@stephen-turner : In dotNetInstaller folder, XenCenterSetupBootstrapper_l10n.xml is a multilingual file which cannot be handled by the localization system. Hence I split up the file content into language based files. But the logic remains same. Just added 3 new files. Nothing else changed in workflow.

GaborApatiNagy:
Hi Vetri, Stephen,

My 2 cents on the &gt; characters.

Resx files are 1.0 xml files, so 
http://www.w3.org/TR/xml/#syntax
in 2.4 applies: 
" The right angle bracket (>) may be represented using the string " &gt; ", and must, for compatibility, be escaped using either " &gt; " or [gabor: removed text]"

It looks like ">" is not allowed. Actually, have you tried the designer? Would it keep them or change all back? I am not against this, but I am a bit inconvenient seeing this.

I would like to see the "&gt;" ones back sooner or later. I am not sure how the localisation system works, but an xml writer should really support this.

GaborApatiNagy:
It looks like the 3 new files in dotnetinstaller directory are used as an internal step only. The installer will use the existing XenCenterSetupBootstrapper_l10n.xml file only.
Vetri, do we need to have these new files in the repo?

stephen-turner:
Gabor, it doesn't exactly say that. It says:

The ampersand character (&) and the left angle bracket (<) must not appear in their literal form, except \[some exceptions]. If they are needed elsewhere, they must be escaped using either numeric character references or the strings " \&amp; " and " \&lt; " respectively. The right angle bracket (>) may be represented using the string " \&gt; ", and must, for compatibility, be escaped using either " \&gt; " or a character reference when it appears in the string " ]]> " in content, when that string is not marking the end of a CDATA section.

stephen-turner:
So what do the three new files actually do? Gabor, what do you mean by an "internal step"?

GaborApatiNagy:
I think the Localisation team uses them for their internal process. It looks like they create the "en" one to start with and then they create the other two for CN an JP. Then the contents from those two are merged/copied back to the file we use.

stephen-turner:
Hmmm, I'm not very comfortable with that. The same information should only be stored in one place in the repository, otherwise there's confusion over which is the master copy. If we want to go down that route, we would need to change the installer to use separate ones, or else construct the combined one programmatically at build time.

GaborApatiNagy:
The installer cannot really use separate ones for each language, because it consumes one file for each MSI and we have the L10N one with two languages in it at the time. That is why we have one file _l10n.xml there.

Considering localising this part of the installer would happen rarely, I would recommend just not adding those three new files to the repository so there would be no confusion later on.

XML: I went through it again. Stephen's right. So it should be fine this way.

stephen-turner:
I experimenting with changing \&gt;\&gt; to >> in the English files, and the Visual Studio Designer changed them back on next edit. I don't think we can have the Designer fighting with the localisation tool, because we will end up with constant edits.

vetrivelc:
Okay. Thanks @stephen-turner & @GaborApatiNagy . Can I revert the changes related to "<"?

stephen-turner:
That would be our preferred solution, if your tool can handle it.

vetrivelc:
@stephen-turner  & @GaborApatiNagy : I did the changes as per the discussion. Please do the needful.

stephen-turner:
Thanks for that, Vetri. I'm pulling the changes now.

PR #453, opened 2015-05-06
CA-169170: “Switch to Remote desktop” button should not be disabled in Cream when Windows VM installed with latest PV tools.

cheng--zhang:
Check current XenServer Version, if current version is greater than Cream and PV tool installed, do not disable control button.


MihaelaStoica:
Looks good. Merging.

PR #456, opened 2015-05-06
CA-164686: Host got removed from XC when it is removed from a pool

MihaelaStoica:
- On receiving a HOST_IS_SLAVE exception, only change connection.Hostname property if we need to connect to the pool master. Leave it unchanged if we are already connected to the pool, as no further action is required. This will ensure that we don't change the Hostname of a slave back to the master, if the user tries to connect to it immediately after it has been removed from the pool.


GaborApatiNagy:
Looks fine

PR #460, opened 2015-05-07
Creedence sp ln

vetrivelc:
Fixed Cream LN Localization bugs.


GaborApatiNagy:
Hi Vetri,

XenAdmin/Wizards/PatchingWizard/PatchingWizard_UploadPage.zh-CN.resx
seems to be broken, are you sure those lines are not needed?

GaborApatiNagy:
I tried it again, it is good now
merging

PR #461, opened 2015-05-08
Creedence sp ln

vetrivelc:
Fixed Fix: CA-169132 and CA-169268


GaborApatiNagy:
Thanks Vetri, Merging

PR #470, opened 2015-05-18
CA-161524: Resource report menu item is only shown when you're on the…

MihaelaStoica:
… pool node - Fixed


GaborApatiNagy:
I think this should be available for multiple selection inside a pool (same as vApps).

GaborApatiNagy:
Looks fine

PR #471, opened 2015-05-18
CA-147164: Filter by status doesn't notice changes of status

GaborApatiNagy:
Fixed the bug by subscribing to events of all actions and not only to the
ones that are displayed in the grid.
Besides this, the visibility of the row is being used instead of removing and
re-adding rows on status changes.


MihaelaStoica:
Looks good. Merging.

PR #472, opened 2015-05-18
CA-120187: Manifestation of "anonymous" action when repairing multiple storages

GaborApatiNagy:
Passing suppressHistory=false in the MultipleAction's constructor (so no action item will be displayed in the Event view for the MultipleAction itself)


MihaelaStoica:
Looks good.

PR #473, opened 2015-05-18
CA-116505: Popup when adding CD drive to running HVM guest without tools does not get focus

GaborApatiNagy:
Using ShowDialog() instead of Show()
Also removed unnecessary code that displayed the popup again


MihaelaStoica:
Looks good.
and it also fixes CA-116506 (Instructed to shutdown and restart already halted HVM guest when adding CD)

PR #474, opened 2015-05-19
CP-12019: Allow virtual disk migration for vdis attached to stopped a…

MihaelaStoica:
…nd suspended VMs


GaborApatiNagy:
Looks good

PR #477, opened 2015-05-25
CP-12157: Create XenServerHealthCheck framework

cheng--zhang:
1.Create a service named XenServerHealthCheck
2.Read the XenCenter Server List from configuration
3.Try to connect to these servers
4.Modify protection of XenCenter config from CurrentUset to LocalMachine

### DIFF
--- a/XenCenterLib/EncryptionUtils.cs
+++ b/XenCenterLib/EncryptionUtils.cs
@@ -65,18 +65,22 @@ namespace XenAdmin.Core
         public static string Protect(string data)
         {
             byte[] dataBytes = Encoding.Unicode.GetBytes(data);
-
+#if CALL_HOME_DEV
+            byte[] protectedBytes = ProtectedData.Protect(dataBytes, GetSalt(), DataProtectionScope.LocalMachine);
+#else
             byte[] protectedBytes = ProtectedData.Protect(dataBytes, GetSalt(), DataProtectionScope.CurrentUser);
-
+#endif
             return Convert.ToBase64String(protectedBytes);
         }
 
         public static string Unprotect(string protectedstring)
         {
             byte[] protectedBytes = Convert.FromBase64String(protectedstring);
-
+#if CALL_HOME_DEV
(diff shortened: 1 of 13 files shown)


XenCenterLib/EncryptionUtils.cs
```diff
@@ -66,7 +66,7 @@ public static string Protect(string data)
         {
             byte[] dataBytes = Encoding.Unicode.GetBytes(data);
 
-            byte[] protectedBytes = ProtectedData.Protect(dataBytes, GetSalt(), DataProtectionScope.CurrentUser);
+            byte[] protectedBytes = ProtectedData.Protect(dataBytes, GetSalt(), DataProtectionScope.LocalMachine);
```
GaborApatiNagy: This would change how XenCenter stores data.
While I see why this change was made, this would require users to know that they will need to trust not only administrators, but all the local users on a machine they are using.

cheng--zhang: Yes, without the scope change, we cannot read the config from a service. because service runing in system account.

stephen-turner: We couldn't change it without an upgrade path from the CurrentUser encryption. But actually, I think this may not be necessary in the end.

MihaelaStoica: We are currently investigating other ways of passing the server list to the service, without having to modufy the encryption level
For the time being, we can keep these changes, so that we can progress with the implementation. But we plan to revert them later when a different solution is agreed.
Cheng, could you please include this change into a #if directive, so that the change is only in place when compiling for this testing configuration.

cheng--zhang:
Hi @MihaelaStoica 
Please help to review the code again, I add a parameter called: CALL_HOME_DEV there.

MihaelaStoica:
This looks good Cheng. Just one note, the timer interval is set to 10000 (10 seconds). I think it should be 30*60000 (30 minutes).

cheng--zhang:
Hi @MihaelaStoica 
Sorry about the wrong time. I use it for my test, and just update it, please help to check and merge it if possible.
Thanks

MihaelaStoica:
Looks good. Merging.

PR #478, opened 2015-05-27
CP-12282: Semantics of Move / Migrate / Copy for suspended VMs

MihaelaStoica:
Changed the name of the operation in XenCenter to "Migrate VM" for suspended VMs, instead of "Move VM".
This clarifies the distinction between copy/move and migration:
- running and suspended VMs can only be migrated
- stopped VMs can be moved and copied


GaborApatiNagy:
Looks good

PR #479, opened 2015-05-27
CP-12153: Call Home Authentication

MihaelaStoica:

### DIFF
--- a/XenModel/Actions/CallHome/CallHomeAuthenticationAction.cs
+++ b/XenModel/Actions/CallHome/CallHomeAuthenticationAction.cs
@@ -1,25 +1,153 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using Microsoft.Win32;
+using XenAdmin.Network;
 using XenAPI;
+using System.Web.Script.Serialization;
 
 namespace XenAdmin.Actions
 {
     public class CallHomeAuthenticationAction : PureAsyncAction
     {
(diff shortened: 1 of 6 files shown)


XenModel/Actions/CallHome/CallHomeAuthenticationAction.cs
```diff
@@ -1,25 +1,153 @@
...
+
+                SetUploadTokenSecret(Connection, newConfig, uploadToken);
+            }
+            catch (Exception e)
+            {
+                Exception = e;
+            }
+            finally
+            {
+                Pool.set_gui_config(Connection.Session, pool.opaque_ref, newConfig);
```
GaborApatiNagy: I think this should be at the end of the try block.

XenModel/Actions/CallHome/CallHomeAuthenticationAction.cs
```diff
@@ -1,25 +1,153 @@
...
+            {
+                string identityToken = GetIdentityToken();
+                string uploadGrantToken = GetUploadGrantToken(identityToken);
+                string uploadToken = GetUploadToken(uploadGrantToken);
+
+                SetUploadTokenSecret(Connection, newConfig, uploadToken);
+            }
+            catch (Exception e)
+            {
+                Exception = e;
```
GaborApatiNagy: Looks like logging is missing here

GaborApatiNagy:
It is not clear (to me) from the code how you delete/forget tokens when needed to do so. I think this needs to be addressed later.

GaborApatiNagy:
I think we need better error messages later on. Especially for 403 Forbidden that what we get for invalid credentials.

GaborApatiNagy:
Merging, some of the comments will be addressed later on.

PR #480, opened 2015-05-27
CA-165366 Better phrasing of boolean groupings in search

stephen-turner:


stephen-turner:
Note: This change partially breaks saving the search from an old version of XenCenter and loading it into this version. The search results will still be correct, but the group names will be shown old-style, and the group will not appear in the "Group By" drop down.

It would be fixable by using if(T is bool) switches in PropertyGrouping<T>, rather than subclassing it, but I decided I preferred cleaner code to preserving perfect behaviour for this unlikely use case.

GaborApatiNagy:
Looks good, I think this is a good solution

PR #483, opened 2015-06-01
CA-153347 [...]: fixing regression caused by CA-160189

GaborApatiNagy:
-Reverted previous fix attempt, because
-managed to track back to root cause, so
-fixed the original problem


MihaelaStoica:
Looks good

PR #486, opened 2015-06-04
CA-169167: Screen resolution is less than the default value after swi…

GaborApatiNagy:
Fixing unused field as warning as error build issue


MihaelaStoica:
Good!

GaborApatiNagy:
No problem, I like these kind of easy fixes :smile:

PR #487, opened 2015-06-04
CP-11985: Implement RPU pre-check phase for safe repartitioning

GaborApatiNagy:
testSafe2Upgrade pugin function in prepare_host_upgrade.py is called as part of the RPU pre-checks


MihaelaStoica:
Looks good

PR #491, opened 2015-06-08
CP-7644: Write XenCenter product version and build number

rosslagerwall:
Write out the XenCenter version and build number into a version file so
that other tools can determine them from the build artifacts.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/mk/archive-push.sh
+++ b/mk/archive-push.sh
@@ -54,7 +54,7 @@ then
 fi
 
 #update local xenadmin-ref.hg repository
-cp ${OUTPUT_DIR}/{manifest,latest-*-build} ${ROOT}/xenadmin-ref.hg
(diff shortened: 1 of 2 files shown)


MihaelaStoica:
I think we also need to change archive-push.sh to copy the xcversion file into xenadmin-ref repository.

rosslagerwall:
OK, done.

GaborApatiNagy:
Thanks Ross, this looks good.
Merging.

(I think we still need to add this new file using "hg add xcversion", I am happy to do it using the Jenkins server, probably that's the easiest way.)

rosslagerwall:
OK, please do add it. Thanks!

PR #492, opened 2015-06-08
CP-12154: Offer Call Home enrollment when connecting to Call-Home cap…

MihaelaStoica:
…able hosts

- Check call home enrollment status on connecting to a pool: If status is unknown it means that the pool is not enrolled (enabled) and has never been (disabled). If that's the case, then show the Health Check Overview dialog with the pool selected
- "Enroll now" on Health Check Overview will try enroll the selected pool into call home using existing token authentication. If this is not possible, then a dialog will be presented for the user to perform the initial authentication.

### DIFF
--- a/XenAdmin/Dialogs/CallHome/HealthCheckOverviewLauncher.cs
+++ b/XenAdmin/Dialogs/CallHome/HealthCheckOverviewLauncher.cs
@@ -1,0 +1,193 @@
+﻿/* Copyright (c) Citrix Systems Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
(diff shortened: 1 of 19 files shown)


XenAdmin/Dialogs/CallHome/HealthCheckOverviewLauncher.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        private void LaunchIfRequired(bool nag, IEnumerable<IXenObject> selectedObjects)
+        {
+            lock (healthCheckLock)
+            {
+                if (!healthCheckOverviewVisible)
+                {
+                    LoadHealthCheckOverviewDialog();
+                    if (nag && TimeSinceLastClose < TimeSpan.FromSeconds(10))
```
GaborApatiNagy: I am just thinking about a case when connecting to multiple hosts would take minutes, that would mean too many pop ups that have to be clicked - some would find it annoying (they are modal...).

GaborApatiNagy:
This is still in progress, I think it is fine for merging.

PR #495, opened 2015-06-09
CP-11985: Implement RPU pre-check phase for safe repartitioning

GaborApatiNagy:
Added check for the "not_enough_space" case


MihaelaStoica:
Looks good.

PR #500, opened 2015-06-10
CA-172444: Consistency should be followed for behavior of all buttons in same page

GaborApatiNagy:


MihaelaStoica:
looks good

GaborApatiNagy:
:balloon::balloon::balloon: 500 :balloon::balloon::balloon:

PR #503, opened 2015-06-16
CP-12626: Use health_check_config field for Call Home enrollment sett…

MihaelaStoica:
…ings

- update XenModel with the api bindings containing xapi fields and methods related to this task


GaborApatiNagy:
:+1:

PR #507, opened 2015-06-18
CP-12627: Use new xapi field health_check_config for Call Home

cheng--zhang:
use health_check_config to replace original gui_config


MihaelaStoica:
Looks good. Merging

PR #513, opened 2015-06-25
CP-12777: FCoE status in NICs tab

MihaelaStoica:
- Added "FCoE Capable" column, showing FCoE capability as Yes/No values in the NICs grid
- This column is only visible for Dundee or higher hosts


GaborApatiNagy:
Looks good

PR #514, opened 2015-06-25
CA-12164: Server Status Report wizard: Add option to upload the serve…

MihaelaStoica:
…r status report to CIS

- add the option on the destination page of the wizard
- for authentication purposes, MyCitrix credentials are required.
- these credentials are only used to obtain an upload token that will be used for the upload.
- the upload token expires in 24 hours and is not saved in the xapi database.

### DIFF
--- a/XenAdmin/Wizards/BugToolWizardFiles/BugToolPageDestination.cs
+++ b/XenAdmin/Wizards/BugToolWizardFiles/BugToolPageDestination.cs
@@ -38,9 +38,11 @@ using System.IO;
 using System.Text;
 using System.Windows.Forms;
 using Microsoft.Win32;
+using XenAdmin.Actions;
 using XenAdmin.Controls;
 using XenAdmin.Controls.Common;
 using XenAdmin.Core;
+using Registry = XenAdmin.Core.Registry;
 
 
 namespace XenAdmin.Wizards.BugToolWizardFiles
@@ -48,6 +50,8 @@ namespace XenAdmin.Wizards.BugToolWizardFiles
     public partial class BugToolPageDestination : XenTabPage
     {
         private bool m_buttonNextEnabled;
+        
+       private const int TokenExpiration = 86400; // 24 hours
 
         public BugToolPageDestination()
         {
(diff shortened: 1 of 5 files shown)


XenAdmin/Wizards/BugToolWizardFiles/BugToolPageDestination.cs
```diff
@@ -48,6 +50,8 @@ namespace XenAdmin.Wizards.BugToolWizardFiles
     public partial class BugToolPageDestination : XenTabPage
     {
         private bool m_buttonNextEnabled;
+        
+       private const int TokenExpiration = 86400; // 24 hours
```
GaborApatiNagy: I would suggest to have something like TOKEN_EXPIRATION_IN_SECONDS as this is a const value.

XenModel/Actions/CallHome/CallHomeAuthenticationAction.cs
```diff
@@ -63,31 +83,41 @@ protected override void Run()
...
+            get { return uploadToken; }
+        }
+
         public static void SetUploadTokenSecret(IXenConnection connection, Dictionary<string, string> config, string uploadToken)
         {
-            if (uploadToken == null)
+            SetTokenSecret(connection, config, CallHomeSettings.UPLOAD_TOKEN_SECRET, uploadToken);
+        }
+        
+        public static void SetTokenSecret(IXenConnection connection, Dictionary<string, string> config, string tokenKey, string tokenValue)
```
GaborApatiNagy: I think it would be good to return when the tokenKey is null. (public static, may be reused later)

XenAdmin/Wizards/BugToolWizardFiles/BugToolPageDestination.cs
```diff
@@ -144,27 +159,91 @@ private bool CheckPathValid(out string error)
...
+            var action = new CallHomeAuthenticationAction(null, usernameTextBox.Text.Trim(), passwordTextBox.Text.Trim(),
+                Registry.CallHomeIdentityTokenDomainName, Registry.CallHomeUploadGrantTokenDomainName, Registry.CallHomeUploadTokenDomainName,
+                false, TokenExpiration, false);
+
+            try
+            {
+                action.RunExternal(null);
+            }
+            catch
+            {
```
GaborApatiNagy: Should not the next few lines be inside the catch block?

MihaelaStoica:
Added a new commit with changes following code review.
I will rename the constants in a different commit.

GaborApatiNagy:
Looks fine, merging

PR #515, opened 2015-06-26
CP-12770: use server connection details received from XenCenter

cheng--zhang:
1. Create named pipe: “HealthCheckServicePipe”
2. Receive credential send from XenServer and decrypt it
3. Update credential setting and save encrypted credential using current user scope system user.
4. Create work thread to handle the pipe communication.
5. If HealthCheckServicePipe already existed before service start, stop service and output service error log
6. If sign off information is received for a host (XenserverName or IP without username and password) then the existing saved credentials will be deleted
7. If the HealthCheckService connects to a host that is no longer enrolled, then it will delete the existing saved credentials


XenAdmin/Dialogs/CallHome/CallHomeSettingsDialog.cs
```diff
@@ -46,12 +46,16 @@ public partial class CallHomeSettingsDialog : XenDialogBase
...
+        private string password;
 
         public CallHomeSettingsDialog(Pool pool)
         {
             this.pool = pool;
-            callHomeSettings = pool.CallHomeSettings;
-            authenticationToken = callHomeSettings.GetExistingUploadToken(pool.Connection);
+            callHomeSettings = pool.CallHomeSettings;
+            authenticationToken = callHomeSettings.GetExistingSecretyInfo(pool.Connection, CallHomeSettings.UPLOAD_TOKEN_SECRET);
+            usetName = callHomeSettings.GetExistingSecretyInfo(pool.Connection, CallHomeSettings.UPLOAD_CREDENTIAL_USER_SECRET);
```
MihaelaStoica: For userName and Password, we need to call GetSecretyInfo so that we get the value from the settings of this connection only. GetExistingSecretyInfo will also try to get the values from other connections, which shouldn't be done this case.

XenModel/XenAPI-Extensions/Pool.cs
```diff
@@ -552,13 +560,30 @@ public static int GetDefaultTime()
...
 
-        public string GetUploadToken(IXenConnection connection)
+        public string GetSecretyInfo(IXenConnection connection, string secretType)
         {
-            if (connection == null || string.IsNullOrEmpty(UploadTokenSecretUuid))
+            string UUID = string.Empty;
+            switch (secretType)
+            {
+                case CallHomeSettings.UPLOAD_CREDENTIAL_USER_SECRET:
+                    UUID = PasswordSecretUuid;
```
MihaelaStoica: I think this should be: UUID = UserNameSecretUuid;
And three lined below:  UUID = PasswordSecretUuid;

XenModel/Actions/CallHome/CallHomeAuthenticationAction.cs
```diff
@@ -86,41 +88,41 @@ protected override void Run()
...
-            }
-            else if (config.ContainsKey(tokenKey))
+        }
+
+        public static void SetSecretInfo(IXenConnection connection, Dictionary<string, string> config, string infoKey, string infoValue)
+        {
+            SetInfoSecret(connection, config, infoKey, infoValue);
+        }
+
+        public static void SetInfoSecret(IXenConnection connection, Dictionary<string, string> config, string infoKey, string infoValue)
```
MihaelaStoica: I think these two functions can be combined into one, SetSecretInfo.

XenAdmin/Dialogs/CallHome/CallHomeSettingsDialog.cs
```diff
@@ -149,9 +157,10 @@ private void okButton_Click(object sender, EventArgs e)
                     (int) (frequencyNumericBox.Value * 7),
                     (DayOfWeek) dayOfWeekComboBox.SelectedValue, 
                     (int) timeOfDayComboBox.SelectedValue,
-                    CallHomeSettings.DefaultRetryInterval);
-
-                new SaveCallHomeSettingsAction(pool, newCallHomeSettings, authenticationToken, false).RunAsync();
+                    CallHomeSettings.DefaultRetryInterval);
+
+                new SaveCallHomeSettingsAction(pool, newCallHomeSettings, authenticationToken, usetName, password, false).RunAsync();
+                new TransferCallHomeSettingsAction(pool, newCallHomeSettings, usetName, password, false).RunAsync();
```
MihaelaStoica: I think we should use the TransferCallHomeSettingsAction with suppressHistory = true (the last parameter) so it does not show in the Events page.

MihaelaStoica:
Looks good. Merging.

GaborApatiNagy:
Hi Cheng,

I noticed you added back XenAdmin as a project reference to XenServerHealthCheck.
Please do not reference this, because that will bring all the XenCenter files into the Health Service which is not really what we want. (Also the service itself has no reason to contain or has access to XenCenter UI code...)

I have seen you added it in order to have
```c
XenAdminConfigManager.Provider = new WinformsXenAdminConfigProvider();
```
While XenAdminConfigManager and IXenAdminConfigProvider is defined in XenModel, the WinformsXenAdminConfigProvider is part of XenAdmin. 

If you need to use XenAdminConfigManager, you will need to create a new implementation of IXenAdminConfigProvider (like there is one WinformsXenAdminConfigProvider for XenAdmin) for the Health Check service and pass an instance of that to Provider.

Thanks!

cheng--zhang:
Hi @GaborApatiNagy 
Thank you for your advice, revert the code and add the new Provider code.

PR #516, opened 2015-06-26
CP-12154: Check whether the Health Check feature is available before…

MihaelaStoica:
… displaying the overview dialog when connecting to a host

- also the "Health Check" menu item will only be enabled if the feature is available


GaborApatiNagy:
Looks fine

PR #519, opened 2015-06-29
CP-12314: Add New SR type for FCoE in the SR wizard 

MihaelaStoica:


GaborApatiNagy:
We reviewed it, looks fine.
Not merging it yet, because we are waiting for other merges to happen first.

PR #525, opened 2015-07-01
CP-12769: Send server connection details to the Health Check Service

MihaelaStoica:
- Call TransferCallHomeSettingsAction with suppressHistory=true so the action doesn't show on the Events page
- This change also fixes LogsTabTests test failure.


GaborApatiNagy:
Looks good

PR #526, opened 2015-07-02
CP-12769: Fix issue when integration test with CIS

cheng--zhang:
1.	Using en-us to transfer upload request time
2.	Check the pipe before close it.
3.	On demand upload will not have 24 hour due problem since it only check once in half an hour
4.	Avoid get upload token by empty session.


MihaelaStoica:
Looks good. Merging.

PR #529, opened 2015-07-03
CA-173358: Change the unit of vbd-io-throughput data sources to MiB

kaifengz:
The unit of VBD IO throughput is changed to MiB after merged all
VBD related data sources from xcp-rrdd to rrdp-iostat,

Signed-off-by: Kaifeng Zhu <kaifeng.zhu@citrix.com>


MihaelaStoica:
Looks good. Merging.

PR #537, opened 2015-07-13
CA-176636: Transfer VM with manually assigned IP doesn't work.

huizh:
The subnet mask is incorrectly set as the static IP.
Update it to correct address.


MihaelaStoica:
Looks good.

huizh:
Thanks.

PR #541, opened 2015-07-13
CP-12966 GPU drop-down to behave correctly when there is one vGPU typ…

stephen-turner:
…e and no passthrough


MihaelaStoica:
Looks good. Merging.

PR #543, opened 2015-07-14
CP-12769: Send credential only when HealthCheckService is running.

cheng--zhang:


cheng--zhang:
@MihaelaStoica Could you please help to check the solution here?
Thanks.

MihaelaStoica:
Looks good. Merging

PR #544, opened 2015-07-14
CP-13071: Make the CIS upload domain and upload timer interval config…

cheng--zhang:
…urable in service


MihaelaStoica:
Looks good. Merging.

PR #553, opened 2015-07-16
CP-13217: Make the Call Home service optional in the XenCenter installer

GaborApatiNagy:
-Added new page to the XenCenter installer to include a feature selector control where users can choose to install the XenServer Health Check Service.
-Removed the InstallDir page, because the new customize page does provide folder selection (per feature)


MihaelaStoica:
Looks good

PR #557, opened 2015-07-17
CP-13217: Make the Call Home service optional in the XenCenter installer

GaborApatiNagy:
Line endings in patch file


GaborApatiNagy:
Trivial change, merging

PR #559, opened 2015-07-20
CP-13245: Check if the update is applicable to the servers on the Sel…

agimofcarmen:
…ect Servers Page.

CA-157785: Hotfix wizard forgets which servers should be greyed out.
When clicking Download and Install update in the Updates Tab, instead of opening a new dialog, it opens the Updates Wizard on the Select Servers Page and download the file in the Upload Page.
Fixes issue regarding the upload of the update every time that the user gets to Upload Page. Now it only uploads it if it has not been uploaded before.
Fixes issue regarding the download of the update every time the user goes back to the Select Patch Page. If the user goes back, but does not change the choice of the update, the download will not be done again.

### DIFF
--- a/XenAdmin/TabPages/ManageUpdatesPage.cs
+++ b/XenAdmin/TabPages/ManageUpdatesPage.cs
@@ -390,45 +390,31 @@ namespace XenAdmin.TabPages
             if (string.IsNullOrEmpty(patchUri))
                 return;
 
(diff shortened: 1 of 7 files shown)


MihaelaStoica:
This commit is missing the declaration of PatchingWizard_ModePage.SelectedUpdateAlert, 
but I see that it is added in a new pull request (#560), so it's OK

PR #563, opened 2015-07-20
CP-13248

GaborApatiNagy:
Check what ALLUSERS install actually does in XenCenter installer, investigate how can this align with the Call Home Service


MihaelaStoica:
:+1:

PR #565, opened 2015-07-20
CP-13310: Installer: In change mode, the installer shows All Users/Current User option, but it should not.

GaborApatiNagy:


MihaelaStoica:
Looks good. Merging.

PR #572, opened 2015-07-24
CP-13120: User-friendly errors when Health Check authorization fails

MihaelaStoica:


GaborApatiNagy:
Looks good

We agreed Mihaela adds a Show Progress Dialog for authentication so users will see the progress and also they cannot close the dialog early.

PR #574, opened 2015-07-24
CA-173497: Thin provisioning: Repair of SR of thin-provisioning on SR hang without return

GaborApatiNagy:
-When detaching an SR, we detach the slaves first
-When reattaching the SR (SR.plug), we attach the master first

This fix was required by thin provisioning, but it does worth to use the same logic regardless of the provisioning method.


MihaelaStoica:
Looks good. Merging.

PR #577, opened 2015-07-24
Fix an error in the tests where the tests can select a default templa…

MihaelaStoica:
…te which is not shown in the tree

(e.g. XenSource P2V Server, which is an internal template, not displayed in the tree)


GaborApatiNagy:
:+1:

PR #583, opened 2015-07-27
Fix XenServerPatchAlert and XenServerUpdateAlert tests

MihaelaStoica:
- setup mock connections for the unit tests with a non-null Cache property
- this will fix the test failure due to Alert.IsDismissed() causing NullReferenceException


GaborApatiNagy:
:+1:

PR #590, opened 2015-07-30
CA-178338: XenServerHealthCheckService: bundle uploading update.

huizh:
1. Add logs to track the uploading process.
2. Catch the potential exception.


MihaelaStoica:
Looks good.

PR #593, opened 2015-08-03
CA-178243: When only Automatic check for XenServer updates was checke…

agimofcarmen:
…d in Updates Options, when opening XC, the grid appeared empty in the Updates Tab. This was because to get the list of XS updates, we need the list of XS versions. Created a list of XenServer versions that is used only to get the XenServer updates.


GaborApatiNagy:
We agreed some refactoring is needed, eg. naming of the new list and to eliminate the duplicated list from the DownloadUpdatesXmlAction class.

GaborApatiNagy:
The change looks good, so once that is committed this can be merged

GaborApatiNagy:
:+1:

PR #595, opened 2015-08-03
CP-13519: If more options are selected in Updates Options, refresh th…

agimofcarmen:
…e updates list.

### DIFF
--- a/XenAdmin/Dialogs/OptionsPages/UpdatesOptionsPage.cs
+++ b/XenAdmin/Dialogs/OptionsPages/UpdatesOptionsPage.cs
@@ -37,6 +37,7 @@ using System.Data;
 using System.Text;
 using System.Windows.Forms;
 using XenAdmin.Properties;
+using XenAdmin.Core;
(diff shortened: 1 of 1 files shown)


XenAdmin/Dialogs/OptionsPages/UpdatesOptionsPage.cs
```diff
@@ -85,6 +88,22 @@ public void Save()
...
+            {
+                Updates.CheckForUpdates(false);
+            }
+        }
+
+        /// <summary>
+        /// Returns true if at least one other box has been checked in Updates Options. Otherwise returns false.
+        /// </summary>
+        /// <returns></returns>
+        private bool AtLeastOneOptionHasBenChecked()
```
GaborApatiNagy: it looks good, but I would rename this, because the naming is misleading. I think what it does is more like IsCheckForUpdateRequired()

GaborApatiNagy:
:+1:

PR #600, opened 2015-08-05
CA-173695: Reclaim freed space is failing with "Index and length must refer to a location with in the string" for AD users

GaborApatiNagy:
This changeset fixes the mock to return "True" for host_call_plugin-->trim calls


stephen-turner:
Good, merging.

PR #602, opened 2015-08-06
CP-13390: Update XenCenter installer to require .NET 4.5

GaborApatiNagy:


stephen-turner:
Looks fine, but we will do a final dev test after it's built.

PR #608, opened 2015-08-07
CP-13592: Add reg to store upload url and time interval

cheng--zhang:


XenServerHealthCheck/Registry.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace XenServerHealthCheck
+{
+    public class Registry
+    {
+        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+
+        /// <summary>
+        /// Reads a key from XENCENTER_LOCAL_KEYS\k.
+        /// </summary>
+        private static string ReadKey(string k, string def)
```
GaborApatiNagy: I think it would be better to either have something like 
```c
bool TryReadKey(string k, out value)
```
that could return true or false based on whether it succeeded or not.

Or just have 
```c
int ReadKey() 
```
and return null.

I am suggesting this, because I think it is not this method's responsibility to consume a default value and return it. Its scope should only be to look up a value and either return that or fail somehow.

XenServerHealthCheck/Registry.cs
```diff
@@ -0,0 +1,121 @@
...
+                return def;
+            }
+        }
+
+        /// <summary>
+        /// Read a true/false key from XENCENTER_LOCAL_KEYS\k.
+        /// </summary>
+        /// <returns>False if the value is "false" (case insensitive), true if the key is present but not
+        /// "false", def otherwise.</returns>
+        private static Int32 ReadLong(string k, Int32 def)
```
GaborApatiNagy: I am suggesting the same change here as at ReadKey()

XenServerHealthCheck/Registry.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                log.DebugFormat(@"Failed to read {0}\{1} from registry; assuming {1} is {2}.", HEALTH_CHECK_LOCAL_KEYS, k, def);
+                log.Debug(e, e);
+                return def;
+            }
+        }
+
+        public static string HealthCheckUploadDomainName
+        {
+            get { return ReadKey(HEALTH_CHECK_UPLOAD_DOMAIN_NAME, "https://rttf-staging.citrix.com/feeds/api/"); }
```
GaborApatiNagy: This could invoke ReadKey and then decide what to do (take result or use a default)

GaborApatiNagy:
I think this looks good, so I am merging.

GaborApatiNagy:
@cheng--zhang: Cheng, I have merged your changes, they look good. If you have some time for some refactoring -and if you agree- please submit that later as a seperate pull request. Thanks

PR #610, opened 2015-08-10
Changes Tab Indexing and disables refresh and restore button while ch…

agimofcarmen:
…eck for updates is in progress.


GaborApatiNagy:
Looks good

PR #611, opened 2015-08-10
Build script: Removed switches from mkisofs cmd

GaborApatiNagy:
These are no longer needed


GaborApatiNagy:
I am merging this, because we have tried this fix with @ssbarnea.

PR #612, opened 2015-08-11
CA-179369: Change CIS default setting

cheng--zhang:
Change the default from http://cis-daily.citrite.net to "https://rttf-daily.citrite.net".
Also make workaround for rttf cannot accept expiration = 0.


GaborApatiNagy:
Looks fine

PR #616, opened 2015-08-17
CP-13072: HealthCheckService: Use proxy for bundle upload.

huizh:


MihaelaStoica:
Looks good. Merging

PR #618, opened 2015-08-17
CP-13165: Make quiescing into a licensed feature

MihaelaStoica:
- Add code to check license before offering quiescing on the VM Snapshot dialog
- If the feature is not available, the message displayed is: "This feature is disabled due to license restrictions on the server."


stephen-turner:
Thanks, looks good.

PR #619, opened 2015-08-19
CP-13374 XenCenter can now turn on and off ssl-legacy

stephen-turner:
The main functionality is working, but the ExpectDisruption flag isn't working properly yet: but suspect a pre-existing bug.


GaborApatiNagy:
Looks good, merging

PR #629, opened 2015-08-20
CP-13074: Finalize the CIS domain names and product key used for retr…

MihaelaStoica:
…ieving the authentication tokens and upload

- change the URLs to the production site
- for development and testing these settings should be changed to the staging site via registry keys


stephen-turner:
Merged.

PR #636, opened 2015-08-26
CP-13174: Change units used for initial_allocation and allocation_qua…

agimofcarmen:
…ntum to MB or GB instead of percentages.

Changes allocation type from "dynamic" to "xlvhd".


MihaelaStoica:
merging

PR #638, opened 2015-08-26
CP-12156: Obtain and display the result of the CIS analysis of the last uploaded report

MihaelaStoica:
(plus some refactoring)


XenModel/Actions/HealthCheck/GetHealthCheckAnalysisResultAction.cs
```diff
@@ -0,0 +1,154 @@
...
+            httpWebRequest.Headers.Add("Authorization", authorizationHeader);
+            httpWebRequest.Method = "GET";
+
+            string result;
+            var httpResponse = (HttpWebResponse)httpWebRequest.GetResponse();
+            using (var streamReader = new StreamReader(httpResponse.GetResponseStream()))
+            {
+                result = streamReader.ReadToEnd();
+            }
+            TemplateResponse response = new JavaScriptSerializer().Deserialize<TemplateResponse>(result);
```
GaborApatiNagy: I think some check is needed here, because the serialize might throw an exception

GaborApatiNagy: But this is an action, so it is fine. Did not think it before 
:walking: to merge

PR #640, opened 2015-08-27
CA-180577: "Not applied" updates list is now removed from the Updates…

agimofcarmen:
… section in General Tab. The "Install pending updates" context menu entry is also removed.


MihaelaStoica:
:+1:

PR #641, opened 2015-08-27
CA-180581: Modified the table columns size in order for the year in t…

agimofcarmen:
…he Date column to be visible.


MihaelaStoica:
:+1:

PR #644, opened 2015-08-28
CP-13100: Fixed unit tests and fixed a bug in code

GaborApatiNagy:
Fixing the unit tests pointed me at the command's CanExecuteCore where the check for Dundee or higher was missing. Fixed that as well.


MihaelaStoica:
:+1:

PR #645, opened 2015-08-28
CP 13902 - Refactors Provisioning page in SR wizards to use the new control…

agimofcarmen:
… for allocations.

Adds tab indexing to the control for allocation and changes label names.
Corrects mistake that led to exception being thrown when the Disk Size is decreased in the the New Disk Dialog.


MihaelaStoica:
Looks good. Merging.

PR #646, opened 2015-09-01
CA-177783: Create SR using ISCSI type in XenCenter cannot go back and…

agimofcarmen:
… choose to format after reattach.

Solves issue related to error message that was not visible in the Location Page for the ISCSI type.


MihaelaStoica:
Looks good

PR #651, opened 2015-09-03
CP-13077: Changes to the Health Check feature to work properly with RBAC

MihaelaStoica:
Changed all the health check actions to perform the necessary role checks
- HealthCheckAuthenticationAction: changed so it does not save the tokens, so no role checks required in this action (all the checks are performed in SaveHealthCheckSettingsAction)
- GetHealthCheckAnalysisActions: added role checks

On the Health Check Overview dialog, a pool is displayed as read-only if connected as a user with roles that don't permit changing the health check settings:
- the following actions are not available: enroll, edit settings, disable health check, request upload, get analysis result

### DIFF
--- a/XenAdmin/Core/HealthCheck.cs
+++ b/XenAdmin/Core/HealthCheck.cs
@@ -44,21 +44,17 @@ namespace XenAdmin.Core
         public static event Action<bool> CheckForAnalysisResultsCompleted;
 
         /// <summary>
-        /// Checks the analysis results for all connections
+        /// Checks the analysis results for all connections.
+        /// Will only perform the check for the connection that have sufficient rights; will not show the action progress.
         /// </summary>
         public static void CheckForAnalysisResults()
         {
             List<AsyncAction> actions = new List<AsyncAction>();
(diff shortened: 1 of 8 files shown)


XenAdmin/Core/HealthCheck.cs
```diff
@@ -75,22 +71,49 @@ public static void CheckForAnalysisResults()
...
 
             if (healthCheckSettings.Status == HealthCheckStatus.Enabled && !healthCheckSettings.HasAnalysisResult)
             {
-                var action = new GetHealthCheckAnalysisResultAction(pool, Registry.HealthCheckDiagnosticDomainName, true);
-                action.Completed += actionCompleted;
-                action.RunAsync();
+                var action = new GetHealthCheckAnalysisResultAction(pool, Registry.HealthCheckDiagnosticDomainName, suppressHistory);
+
+                if (PassedRbacChecks(pool.Connection))
+                    return action;
```
agimofcarmen: I think we should create the action only if the connection passes the Rbac checks.

agimofcarmen:
Looks good.

PR #655, opened 2015-09-04
CA 181442: Cancelling the wizard while upload is in progress does not cause an unhandled exception anymore.

agimofcarmen:


MihaelaStoica:
Looks good

PR #657, opened 2015-09-07
CA-181445: Changes message and page title on the Upload Page in the U…

agimofcarmen:
…pdates Wizard. If we are actually downloading  an update, then the message will mention that. Otherwise, it will only mention the upload.


MihaelaStoica:
Looks good.

PR #666, opened 2015-09-09
CP-13786: PV drivers onto Windows update: Check all uses of GetVirtualisationStatus

GaborApatiNagy:
Extended the enum, aimed for minimal changes, also using Flags on the enum now

### DIFF
--- a/XenAdmin/Commands/RebootVMCommand.cs
+++ b/XenAdmin/Commands/RebootVMCommand.cs
@@ -198,16 +198,19 @@ namespace XenAdmin.Commands
                 case vm_power_state.unknown:
                     return base.GetCantExecuteReasonCore(item);
             }
-            
-            if (vm.virtualisation_status == VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED || vm.virtualisation_status == VM.VirtualisationStatus.UNKNOWN)
+
+            if (vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED) || vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.UNKNOWN))
             {
                 return FriendlyErrorNames.VM_MISSING_PV_DRIVERS;
             }
             
-            if (vm.virtualisation_status == VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE)
+            if (vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE))
             {
                 return FriendlyErrorNames.VM_OLD_PV_DRIVERS;
             }
+
+            if (!vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
+                return Messages.VM_MISSING_IO_DRIVERS;
 
             return base.GetCantExecuteReasonCore(item);
         }
--- a/XenAdmin/Commands/RebootVMCommand.cs
+++ b/XenAdmin/Commands/RebootVMCommand.cs
@@ -199,20 +199,7 @@ namespace XenAdmin.Commands
                     return base.GetCantExecuteReasonCore(item);
             }
 
-            if (vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED) || vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.UNKNOWN))
-            {
-                return FriendlyErrorNames.VM_MISSING_PV_DRIVERS;
-            }
-            
-            if (vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE))
-            {
-                return FriendlyErrorNames.VM_OLD_PV_DRIVERS;
-            }
-
-            if (!vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
-                return Messages.VM_MISSING_IO_DRIVERS;
-
-            return base.GetCantExecuteReasonCore(item);
+            return GetCantExecuteNoToolsOrDriversReasonCore(item) ?? base.GetCantExecuteReasonCore(item);
         }
 
         protected override AsyncAction BuildAction(VM vm)
--- a/XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
+++ b/XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
@@ -50,7 +50,7 @@ namespace XenAdmin.Controls.Ballooning
         {
             InitializeComponent();
         }
-
+        
         protected override void OnPaint(PaintEventArgs e)
         {
             if (vms == null || vms.Count == 0)
@@ -101,7 +101,7 @@ namespace XenAdmin.Controls.Ballooning
                     if (identical)
                     {
                         var status = vm0.GetVirtualisationStatus;
-                        if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED) && status.HasFlag(VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
+                        if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
                             labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTSUPPORTED_PLURAL;
                         else if (!status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
                             labelDMCUnavailable.Text = vm0.HasNewVirtualisationStates ? Messages.DMC_UNAVAILABLE_NO_IO_NO_MGMNT_PLURAL : Messages.DMC_UNAVAILABLE_NOTOOLS_PLURAL;
@@ -123,7 +123,7 @@ namespace XenAdmin.Controls.Ballooning
                 {
                     var status = vm0.GetVirtualisationStatus;
 
-                    if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED) && status.HasFlag(VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
+                    if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
                             labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTSUPPORTED;
                     else if (!status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
                         labelDMCUnavailable.Text = vm0.HasNewVirtualisationStates ? Messages.DMC_UNAVAILABLE_NO_IO_NO_MGMNT : Messages.DMC_UNAVAILABLE_NOTOOLS;
--- a/XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
+++ b/XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
@@ -100,21 +100,15 @@ namespace XenAdmin.Controls.Ballooning
                     }
                     if (identical)
                     {
-                        switch (vm0.GetVirtualisationStatus)
-                        {
-                            case VM.VirtualisationStatus.OPTIMIZED:
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTSUPPORTED_PLURAL;
-                                break;
-                            case VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED:
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTOOLS_PLURAL;
-                                break;
-                            case VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE:
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_OLDTOOLS_PLURAL;
-                                break;
-                            default:  // VM.VirtualisationStatus.UNKNOWN
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VMS;
-                                break;
-                        }
+                        var status = vm0.GetVirtualisationStatus;
+                        if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED | VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
+                            labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTSUPPORTED_PLURAL;
+                        else if (status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED))
+                            labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTOOLS_PLURAL;
+                        else if (status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE))
+                            labelDMCUnavailable.Text = vm0.HasNewVirtualisationStates ? Messages.DMC_UNAVAILABLE_NO_IO_NO_MGMNT_PLURAL : Messages.DMC_UNAVAILABLE_OLDTOOLS_PLURAL;
+                        else
+                            labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VMS;
                     }
                     else
                         labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VMS;
@@ -127,21 +121,17 @@ namespace XenAdmin.Controls.Ballooning
                 }
                 else
                 {
-                    switch (vm0.GetVirtualisationStatus)
-                    {
-                        case VM.VirtualisationStatus.OPTIMIZED:
+                    var status = vm0.GetVirtualisationStatus;
+
+                    if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED | VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
                             labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTSUPPORTED;
-                            break;
-                        case VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED:
-                            labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_NOTOOLS;
-                            break;
-                        case VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE:
+                    else if (status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED))
+                        labelDMCUnavailable.Text = vm0.HasNewVirtualisationStates ? Messages.DMC_UNAVAILABLE_NO_IO_NO_MGMNT : Messages.DMC_UNAVAILABLE_NOTOOLS;
+                    else if (status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE))
                             labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_OLDTOOLS;
-                            break;
-                        default:  // VM.VirtualisationStatus.UNKNOWN
-                            labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VM;
-                            break;
-                    }
+                    else
+                        labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VM;
+                    
                     linkInstallTools.Visible = InstallToolsCommand.CanExecute(vm0);
                 }
             }
--- a/XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
+++ b/XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
@@ -101,12 +101,12 @@ namespace XenAdmin.Controls.Ballooning
                     if (identical)
                     {
                         var status = vm0.GetVirtualisationStatus;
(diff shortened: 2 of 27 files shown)


XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -678,12 +679,21 @@ public DateTime BodgeStartupTime
...
         }
 
         private Timer VirtualizationTimer = null;
 
         [Flags]
-        public enum VirtualisationStatus { UNKNOWN = 0, OPTIMIZED = 1, PV_DRIVERS_NOT_INSTALLED = 2, PV_DRIVERS_OUT_OF_DATE = 4 };
+        public enum VirtualisationStatus
+        {
+            UNKNOWN                     = 1,
+            OPTIMIZED                   = 1 << 1,
```
stephen-turner: I don't think we need to retain OPTIMIZED. We should use IO_DRIVERS_INSTALLED|MANAGEMENT_INSTALLED, and when we need to test for the presence of Tools, we should use whichever of those two we actually need.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -693,6 +703,28 @@ public VirtualisationStatus virtualisation_status
             }
         }
 
+        public VirtualisationStatus virtualisation_status_for_search
```
stephen-turner: I think the logic for extracting the correct flag should go in the search code.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -756,9 +790,13 @@ public VirtualisationStatus GetVirtualisationStatus
                     if (vm_guest_metrics != null && vm_guest_metrics.PV_drivers_installed)
                     {
                         if (vm_guest_metrics.PV_drivers_up_to_date)
-                            return VirtualisationStatus.OPTIMIZED;
+                        {
+                            return VirtualisationStatus.OPTIMIZED | VirtualisationStatus.IO_DRIVERS_INSTALLED | VirtualisationStatus.MANAGEMENT_INSTALLED;
```
stephen-turner: This seems wrong, because it's not looking up GetNewVirtualisationFlags for new VMs.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -766,20 +804,55 @@ public VirtualisationStatus GetVirtualisationStatus
...
                 else if (!vm_guest_metrics.PV_drivers_up_to_date)
                 {
-                    return VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE;
+                    return VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE | GetNewVirtualisationFlags();
                 }
                 else
                 {
-                    return VirtualisationStatus.OPTIMIZED;
+                    if (!IsNewVM)
+                        return VirtualisationStatus.OPTIMIZED;
```
stephen-turner: There is a bug here. You need IO_DRIVERS_INSTALLED and MANAGEMENT_INSTALLED too, otherwise some of the tests elsewhere in the code will wrongly fail.

stephen-turner: Maybe you need a bit more dev testing, which could have caught that some old-style VMs didn't get some operations they should have got.

XenModel/XenSearch/Common.cs
```diff
@@ -1367,23 +1367,19 @@ public static Object GetImagesFor(PropertyNames p)
                             return Icons.XenCenter;
                         };
 
-                case PropertyNames.virtualisation_status:
+                case PropertyNames.virtualisation_status_for_search:
```
stephen-turner: This didn't need to change from a switch to an if. However, if you use virtualisation_status instead of virtualisation_status_for_search, it might be necessary to make the change.

XenAdmin/Dialogs/VmSnapshotDialog.cs
```diff
@@ -151,7 +151,7 @@ private void pictureBoxQuiesceInfo_Click(object sender, EventArgs e)
                 tt = Messages.FIELD_DISABLED;
             else if (_VM.power_state != vm_power_state.Running)
                 tt = Messages.INFO_QUIESCE_MODE_POWER_STATE.Replace("\\n", "\n");
-            else if (_VM.virtualisation_status != VM.VirtualisationStatus.OPTIMIZED)
+            else if (!_VM.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.OPTIMIZED))
```
stephen-turner: I'm not sure why this changed from virtualisation_status to GetVirtualisationStatus.

XenAdmin/XenSearch/Columns.cs
```diff
@@ -94,8 +94,8 @@ protected bool CheckVMTools(IXenObject o, out GridItemBase item)
             {
                 VM.VirtualisationStatus status = vm.virtualisation_status;
                 if (vm.power_state != vm_power_state.Running ||
-                    status == VM.VirtualisationStatus.OPTIMIZED ||
-                    status == VM.VirtualisationStatus.UNKNOWN)
+                    vm.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.OPTIMIZED) ||
```
stephen-turner: This changed from status to GetVirtualisationStatus, which results in calling the function a second time.

XenAdminTests/CommandTests/VMLifecycleCommandTest.cs
```diff
@@ -188,7 +188,7 @@ public void TestCommandErrorDialog()
                           {
                               var curRow = row;//closure
                               VM vm = curSelection.AsXenObjects<VM>().Find(v => v.Name == curRow.Cells[1].Value.ToString());
-                              Assert.IsFalse(vm.virtualisation_status == VM.VirtualisationStatus.OPTIMIZED, "PV drivers installed on " + vm + " but it couldn't suspend.");
+                              Assert.IsFalse(vm.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.OPTIMIZED), "PV drivers installed on " + vm + " but it couldn't suspend.");
```
stephen-turner: Again, this changed from virtualisation_status to GetVirtualisationStatus.

XenModel/Messages.resx
```diff
@@ -2167,7 +2167,8 @@ Deleting this bond will disrupt traffic through the secondary interface on the b
     <value>It is not possible to activate this virtual disk on VM '{0}' as the VM is not running.</value>
   </data>
   <data name="CANNOT_ACTIVATE_VD_VM_NEEDS_TOOLS" xml:space="preserve">
-    <value>You must install XenServer Tools on VM '{0}' before you can activate this virtual disk for the VM.</value>
+    <value>You must install XenServer Tools or the I/O drivers on VM '{0}' before you can activate this virtual disk for the VM.</value>
```
stephen-turner: This is a bit ugly, but it's probably the best solution. But let's change it to "XenServer Tools or I/O drivers" (remove "the").

stephen-turner: The language in this and subsequent comments gets a bit convoluted. Reluctantly, I think we need separate comments for old and new VMs.

GaborApatiNagy:
This PR is ready for review again - I have committed my code changes that previous comments suggested to be made. Thanks

XenModel/XenSearch/Common.cs
```diff
@@ -646,7 +650,7 @@ private static IComparable MemoryTextProperty(IXenObject o)
                         vm.power_state != vm_power_state.Running)
                         return null;
 
-                    if (vm.virtualisation_status != VM.VirtualisationStatus.OPTIMIZED)
+                    if (!vm.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED | VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
```
stephen-turner: I was hoping that all the instances like this would need only one of IO_DRIVERS_INSTALLED and MANAGEMENT_INSTALLED. Why would it need both?

XenModel/XenSearch/GroupingTypes.cs
```diff
@@ -277,7 +277,7 @@ public override String GroupingName
 
         public override String GetGroupName(Object group)
         {
-            if (!(group is T) || i18ns == null)
+            if (!(group is T) || i18ns == null || !i18ns.ContainsKey((T)group))
```
stephen-turner: Was this just added as a precaution, or was there a case where it caused an exception? I would rather make sure we don't get in here if this condition is violated, and redesign the calling function if necessary. Maybe an Assert would be appropriate?

stephen-turner:
I haven't done a complete review. I've made a couple of quick comments for things I noticed immediately.

XenModel/XenSearch/Common.cs
GaborApatiNagy: Good catch, I fixed these. There had been unnecessary checks for the presence of the Management Agent here.

XenModel/XenSearch/GroupingTypes.cs
GaborApatiNagy: It can cause exception being thrown anytime when the i18ns does not contain the group we are looking at. Despite this should not happen, it might, so I think this check is needed to be on the safe side while running in Release build.

For debug build I have added Debug.Assert() as you suggested - this will be a great way to point developers' attention here should they forgot to define the right i18ns entries (as I had managed to do so - answering your first question here)

GaborApatiNagy:
Thanks for your first comments, I have just submitted a commit with the necessary changes.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -729,10 +736,13 @@ public string GetVirtualisationWarningMessages()
                     }
                     else
                         return Messages.PV_DRIVERS_OUT_OF_DATE_UNKNOWN_VERSION;
-
-                default:  // VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED:
-                    return Messages.PV_DRIVERS_NOT_INSTALLED;
             }
+
+            if (virtualisation_status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED)
```
stephen-turner: This "if" is no longer necessary: it will be caught by the next case.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -756,9 +766,13 @@ public VirtualisationStatus GetVirtualisationStatus
                     if (vm_guest_metrics != null && vm_guest_metrics.PV_drivers_installed)
                     {
                         if (vm_guest_metrics.PV_drivers_up_to_date)
-                            return VirtualisationStatus.OPTIMIZED;
+                        {
+                            return HasNewVirtualisationStates ? GetNewVirtualisationFlags() : VirtualisationStatus.IO_DRIVERS_INSTALLED | VirtualisationStatus.MANAGEMENT_INSTALLED;
```
stephen-turner: We seem to call GetNewVirtualisationFlags() in each state, sometimes logical-ORed with the other flags, and sometimes on its own. It would probably be neater to extract
if (HasNewVirtualisationStates)
   return GetNewVirtualisationFlags();
higher up.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -774,13 +772,86 @@ public VirtualisationStatus GetVirtualisationStatus
...
+        ///   2 = Out of date
+        ///  12 = Tools installed (Optimized)
+        ///  
+        /// For Windows VMs on Dundee or higher:
+        ///    0 = Not installed
+        ///    1 = Unknown
+        ///    4 = I/O Optimized
+        ///   12 = I/O and Management installed
+        /// </remarks>
+        public VirtualisationStatus GetVirtualisationStatus
```
stephen-turner: I think this function is correct now. So we now just need to make sure we have the right state for all the usages.

XenAdmin/Commands/RebootVMCommand.cs
```diff
@@ -198,17 +198,20 @@ protected override string GetCantExecuteReasonCore(SelectedItem item)
                 case vm_power_state.unknown:
                     return base.GetCantExecuteReasonCore(item);
             }
-            
-            if (vm.virtualisation_status == VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED || vm.virtualisation_status == VM.VirtualisationStatus.UNKNOWN)
+
+            if (vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED) || vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.UNKNOWN))
```
stephen-turner: This is a bug, which occurs a few times. PV_DRIVERS_NOT_INSTALLED == 0, and HasFlag(0) will always return true.

If you want to test for zero, you should use == 0 or .Equals(0). But before doing that, you should think about whether you really want to test for both I/O and Management to be absent, as 0 would do, or whether you just want to test for one or other of them missing.

XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
```diff
@@ -100,21 +100,15 @@ protected override void OnPaint(PaintEventArgs e)
...
-                                break;
-                            case VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE:
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_OLDTOOLS_PLURAL;
-                                break;
-                            default:  // VM.VirtualisationStatus.UNKNOWN
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VMS;
-                                break;
-                        }
+                        var status = vm0.GetVirtualisationStatus;
+                        if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED | VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
```
stephen-turner: Do you really need both I/O optimized and Management Installed for this, or is one or the other sufficient? My guess is that you only need one. But if you really need both, it might be better to make it more explicit with
HasFlag(IO_DRIVERS_INSTALLED) && HasFlag(MANAGEMENT_INSTALLED)

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -683,7 +683,14 @@ void VirtualizationTimer_Elapsed(object sender, ElapsedEventArgs e)
         private Timer VirtualizationTimer = null;
 
         [Flags]
-        public enum VirtualisationStatus { UNKNOWN = 0, OPTIMIZED = 1, PV_DRIVERS_NOT_INSTALLED = 2, PV_DRIVERS_OUT_OF_DATE = 4 };
+        public enum VirtualisationStatus
+        {
+            PV_DRIVERS_NOT_INSTALLED    = 0,
```
stephen-turner: Because PV_DRIVERS_NOT_INSTALLED has the value zero, it isn't really a flag like the others, and it might be better just to delete it. I can see that it's sometimes nice to use a symbolic name, but I think it is likely lead to bugs like the one above, and it would be better for people to test for the absence of IO_DRIVERS_INSTALLED and/or MANAGEMENT_INSTALLED.

XenAdmin/Commands/RebootVMCommand.cs
GaborApatiNagy: Oh, good finding, I fixed these

XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
GaborApatiNagy: Agreed, I changed them to use && to make it more explicit.

In this file: We do not need both, but this is the method that is trying to guess what has gone wrong, so it can be a bit different - I think. From this point of view I would leave this basically as is (after a small change I have made). It will suggest to install I/O drivers on NewVMs when failed. (More specifically, XenBus is needed, but having seen the presence of IO drivers or Management agent implies we have that. Although as this is an error message, I decided to refer to missing IO drivers only - for new VMs)
For CanExecute we check the feature-balloon - that is done the right way.

In other files I made some more changes, please find them in my last commit.

XenModel/XenAPI-Extensions/VM.cs
GaborApatiNagy: Yes, I agree with this - removed confusing PV_DRIVERS_NOT_INSTALLED from the enum.

XenAdmin/Dialogs/VmSnapshotDialog.cs
```diff
@@ -151,8 +151,8 @@ private void pictureBoxQuiesceInfo_Click(object sender, EventArgs e)
                 tt = Messages.FIELD_DISABLED;
             else if (_VM.power_state != vm_power_state.Running)
                 tt = Messages.INFO_QUIESCE_MODE_POWER_STATE.Replace("\\n", "\n");
-            else if (_VM.virtualisation_status != VM.VirtualisationStatus.OPTIMIZED)
-                tt = Messages.INFO_QUIESCE_MODE_NO_TOOLS.Replace("\\n", "\n");
+            else if (!(_VM.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED) && _VM.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.MANAGEMENT_INSTALLED)))
```
stephen-turner: I think this should just check for Management, not I/O too.

XenAdmin/Controls/Ballooning/VMMemoryControlsBasic.cs
```diff
@@ -100,21 +100,15 @@ protected override void OnPaint(PaintEventArgs e)
...
-                                break;
-                            case VM.VirtualisationStatus.PV_DRIVERS_OUT_OF_DATE:
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_OLDTOOLS_PLURAL;
-                                break;
-                            default:  // VM.VirtualisationStatus.UNKNOWN
-                                labelDMCUnavailable.Text = Messages.DMC_UNAVAILABLE_VMS;
-                                break;
-                        }
+                        var status = vm0.GetVirtualisationStatus;
+                        if (status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED) && status.HasFlag(VM.VirtualisationStatus.MANAGEMENT_INSTALLED))
```
stephen-turner: I think this line should only test for I/O drivers, otherwise you will get the wrong error message in the case you have I/O drivers but no management agent.

XenAdminTests/CommandTests/VMLifecycleCommandTest.cs
```diff
@@ -188,7 +188,7 @@ public void TestCommandErrorDialog()
                           {
                               var curRow = row;//closure
                               VM vm = curSelection.AsXenObjects<VM>().Find(v => v.Name == curRow.Cells[1].Value.ToString());
-                              Assert.IsFalse(vm.virtualisation_status == VM.VirtualisationStatus.OPTIMIZED, "PV drivers installed on " + vm + " but it couldn't suspend.");
+                              Assert.IsFalse(vm.virtualisation_status.HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED) && vm.GetVirtualisationStatus.HasFlag(VM.VirtualisationStatus.MANAGEMENT_INSTALLED), "PV drivers installed on " + vm + " but it couldn't suspend.");
```
stephen-turner: This should only test for I/O drivers, not management agent too.

XenModel/Messages.resx
```diff
@@ -2166,6 +2166,9 @@ Deleting this bond will disrupt traffic through the secondary interface on the b
   <data name="CANNOT_ACTIVATE_VD_VM_HALTED" xml:space="preserve">
     <value>It is not possible to activate this virtual disk on VM '{0}' as the VM is not running.</value>
   </data>
+  <data name="CANNOT_ACTIVATE_VD_VM_NEEDS_IO_DRIVERS" xml:space="preserve">
+    <value>You must install the I/O drivers on VM '{0}' before you can activate this virtual disk for the VM.</value>
```
stephen-turner: "the I/O drivers" sounds OK, but in most other places we just refer to "I/O drivers", so let's drop the "the" here to make it consistent.

XenModel/Messages.resx
```diff
@@ -6326,6 +6341,9 @@ Click Configure HA to alter the settings displayed below.</value>
   <data name="INFO_QUIESCE_MODE" xml:space="preserve">
     <value>Quiesced snapshots need the VSS service to be enabled on the VM</value>
   </data>
+  <data name="INFO_QUIESCE_MODE_NO_IO_MGMNT" xml:space="preserve">
+    <value>Quiesced snapshots need I/O drivers and Management Agent to be installed</value>
```
stephen-turner: Only Management Agent.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -729,24 +734,16 @@ public string GetVirtualisationWarningMessages()
                     }
                     else
                         return Messages.PV_DRIVERS_OUT_OF_DATE_UNKNOWN_VERSION;
-
-                default:  // VirtualisationStatus.PV_DRIVERS_NOT_INSTALLED:
-                    return Messages.PV_DRIVERS_NOT_INSTALLED;
             }
+
+            return HasNewVirtualisationStates ? Messages.VIRTUALIZATION_STATE_VM_MANAGEMENT_AGENT_NOT_INSTALLED : Messages.PV_DRIVERS_NOT_INSTALLED;
```
stephen-turner: I think this will give the wrong result when it's used in EvacuateHostDialog, where only I/O drivers are needed. I'm not quite sure what the right solution is: do we need to split it into more cases?

GaborApatiNagy:
Thanks for the comments @stephen-turner, I have committed the fix.

XenAdmin/XenSearch/Columns.cs
```diff
@@ -161,6 +161,33 @@ protected bool CheckVMTools(IXenObject o, out GridItemBase item)
 
             return false;
         }
+
+        public string GetVMToolsInstallMessage(VM vm)
```
stephen-turner: This function is unused.

stephen-turner:
Thanks, I think that fixes everything. Merging now.

PR #672, opened 2015-09-14
CA-182881: Use local time zone to do the schedule check

cheng--zhang:


MihaelaStoica:
Looks fine. Merging.

PR #673, opened 2015-09-14
Use the Enum functions for string-to-enum and enum-to-string conversions

MihaelaStoica:

### DIFF
--- a/XenModel/HealthCheckSettings.cs
+++ b/XenModel/HealthCheckSettings.cs
@@ -377,28 +377,15 @@ namespace XenAdmin.Model
 
         internal static DiagnosticAlertSeverity StringToDiagnosticAlertSeverity(string severity)
         {
-            switch (severity)
-            {
-                case "Error":
-                    return DiagnosticAlertSeverity.Error;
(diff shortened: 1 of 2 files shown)


XenModel/HealthCheckSettings.cs
```diff
@@ -377,28 +377,15 @@ private static bool TryGetExistingTokensFromOtherConnections(IXenConnection curr
...
-            switch (severity)
-            {
-                case DiagnosticAlertSeverity.Error:
-                    return "Error";
-                case DiagnosticAlertSeverity.Warning:
-                    return "Warning";
-                default:
-                    return "Info";
-            }
+            return severity.ToString();
```
GaborApatiNagy: I think this method could be just removed as it is only calling the base ToString()
That would also eliminate the NullReference exception.

GaborApatiNagy: Ops, probably I have been working too much with Nullable enums recently. So just ignore my NullReferenceException-related statement please :)

GaborApatiNagy:
Looks good

PR #675, opened 2015-09-15
CP-13985: Add "Test credentials" button on the Health Check dialog

MihaelaStoica:


GaborApatiNagy:
I think the button should be disabled when the test is in progress

GaborApatiNagy:
Looks fine

PR #676, opened 2015-09-17
Removes the extra line in the Disk Space Allocation Message in the Ge…

agimofcarmen:
…neral Tab.

Creates a new function MemorySizeStringSuitableUnits that takes one more parameter that specifies the units that must be used when the number of bytes is 0.


MihaelaStoica:
Looks good.

PR #682, opened 2015-09-22
CP-13983: Fix the SR convert action

MihaelaStoica:
- use SR.opaque_ref instead of uuid


GaborApatiNagy:
:+1:

PR #687, opened 2015-09-25
CP-12776: Increase VM memory to 1TB in anticipation of increased limi…

stephen-turner:
…t in CAR-1874


stephen-turner:
NB *NOT* dev tested, because this value is only used as a backstop: and even when I forced it to be used, the visible values are also constrained by the host memory.

MihaelaStoica:
Looks good

PR #688, opened 2015-09-28
CA-134554: Re-enable WLB Disk Read/Write configuration since WLB VPX is ready

huizh:
Revert commit 32c945e104d755b0c172760846ffe57de6d3ad25 which was used to
"Hide Disk Read/Write configuration until the backend server side is ready."


stephen-turner:
Thanks, merging. Glad to see this back!

huizh:
Thank you Stephen.

PR #689, opened 2015-09-28
CA-183958: Committed percentage for thinly provisioned SRs should not…

MihaelaStoica:
… be displayed as negative value

### DIFF
--- a/XenModel/XenAPI-Extensions/SR.cs
+++ b/XenModel/XenAPI-Extensions/SR.cs
@@ -1135,7 +1135,7 @@ namespace XenAPI
         {
             get
             {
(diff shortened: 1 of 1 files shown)


stephen-turner:
I realise we do this in another place, but it looks wrong to me. I think we should try and reproduce it, and understand how the number got negative and what it means. Is it a xapi or SM bug? Or are we (XenCenter) messing up the value somewhere along the line?

MihaelaStoica:
We reproduced this, and it appears to be a bug on the server-side (xapi or sm). I've created a ticket for it. 
I will close this pull request, as the virtual-allocation value shouldn't be negative and therefore we don't need to use Abs().

PR #692, opened 2015-09-30
CA-184303: Adds condition so that the "Allow XenCenter to delete upda…

agimofcarmen:
…te installation files" checkbox would be hidden if the any of the selected servers is older than 6.1.


MihaelaStoica:
looks good

PR #694, opened 2015-10-02
CA-173102: Validation issues with address field in connection proxy settings

GaborApatiNagy:
Improved validation for the proxy address


MihaelaStoica:
Looks good

PR #697, opened 2015-10-06
CA-180128: Make RelatedTask property setter aware of ExpectDisruption flag

MihaelaStoica:
When the pool's ssl_legacy settings are changed, we may temporarily lose connection. The action generally copes well, because PollToCompletion is aware of the ExpectDisruption flag.
But when the action's RelatedTask is assigned, we perform more api calls, which may fail with WebException. When disruption is expected, we need to ignore it and try again on a new session (CancellingAction.DoWithSessionRetry).


stephen-turner:
The change looks good, and fixes the common failure of a spurious error message. I have still once seen the rare but worse failure, where we disconnect for 120s, so we may need to revisit that one.

PR #699, opened 2015-10-07
CA-185020: Search by UUID

stephen-turner:
I found I wanted this when debugging, but it could be more generally useful. Note that we now expose the UUID in the General panel, which we didn't when we wrote the search mechanism.


MihaelaStoica:
Looks good.

PR #703, opened 2015-10-13
CA-185019: HealthCheckService should retrieve the server capabilities

huizh:
HealthCheckService retrieves system capabilities, filter them with
the reports included in Health Check.
Since Dundee, add verbosity for host status report request.

### DIFF
--- a/XenServerHealthCheck/XenServerHealthCheckBugTool.cs
+++ b/XenServerHealthCheck/XenServerHealthCheckBugTool.cs
@@ -37,7 +37,9 @@ using XenAdmin.Core;
 using XenAPI;
 using XenAdmin.Actions;
 using XenAdmin;
+using System.Linq;
 using System.Globalization;
+using System.Xml;
 
 namespace XenServerHealthCheck
 {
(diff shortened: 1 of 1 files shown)


huizh:
Thanks for the comment.
Since there might be other capabilities from server (for example, the capabilities in "currently not used" section) and they are not explicitly mentioned to be included, if we just use exclude list they will be included in the report. So I use the fixed include capability list which are defined in the requirement. That's my understanding.
Could you please let me know your opinion? Thanks.

MihaelaStoica:
The capabilities listed in the requirement are the ones available in 6.5. We probably have new ones already and it's OK to include them by default. As for the ones listed in "currently not used", we should add them to the exclusion list.
So, let's go for the exclusion list and I will update the design and the list of capabilities with all the capabilities available on Dundee hosts.

huizh:
OK, I understand the requirement now. Thanks for the details.
Code is updated accordingly.
Please could you help review?

MihaelaStoica:
Looks good. Thanks.

huizh:
Thanks.

PR #704, opened 2015-10-13
CA-184563: Do not show Provisioning method for SRs other than iSCSI and LVM

MihaelaStoica:
- display the Provisioning entries in the General Tab only for iSCSI and LVM SRs
- also changed the checks in the ConvertToThinSRCommand to use SR.IsThinProvisioned property instead of SR.Provisioning

### DIFF
--- a/XenAdmin/Commands/ConvertToThinSRCommand.cs
+++ b/XenAdmin/Commands/ConvertToThinSRCommand.cs
@@ -98,7 +98,7 @@ namespace XenAdmin.Commands
             {
                 var sr = item.XenObject as SR;
 
-                if (sr != null && Helpers.DundeeOrGreater(sr.Connection) && sr.Provisioning == SrProvisioning.Thick && (sr.type == "lvmohba" || sr.type == "lvmoiscsi"))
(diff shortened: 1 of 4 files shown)


XenAdmin/Commands/ConvertToThinSRCommand.cs
```diff
@@ -98,7 +98,7 @@ private SR GetFirstThickSRFromSelection(SelectedItemCollection selection)
             {
                 var sr = item.XenObject as SR;
 
-                if (sr != null && Helpers.DundeeOrGreater(sr.Connection) && sr.Provisioning == SrProvisioning.Thick && (sr.type == "lvmohba" || sr.type == "lvmoiscsi"))
+                if (sr != null && Helpers.DundeeOrGreater(sr.Connection) && (sr.type == "lvmohba" || sr.type == "lvmoiscsi") && !sr.IsThinProvisioned)
```
stephen-turner: I suggest removing the SR.Provisioning method, which is no longer used, to avoid confusion in future.

stephen-turner:
Thanks, merging now.

PR #706, opened 2015-10-14
Increased timeout of checkUploadLock test, as it usually takes more t…

MihaelaStoica:
…han a minute to run


stephen-turner:
:clock1030:

PR #709, opened 2015-10-19
CP-12617: Port CP-12191 to trunk

GaborApatiNagy:
Notes:
Cherry picked all the three commits from cream-bugfix and resolved the conflicts in the first commit, but it also had one file removed and added back (instead of just moved and edited) so I soft reset it and committed the changes as new commit to avoid losing history.


GaborApatiNagy:
@MihaelaStoica: This is now ready for review/merge. It had to be two commits at the end - I could not even squash them without the add/remove file problem came in.

XenModel/Messages.resx
```diff
@@ -8782,6 +8782,12 @@ However, there is not enough space to perform the repartitioning, so the current
   <data name="OPTIMIZED" xml:space="preserve">
     <value>Optimized</value>
   </data>
+  <data name="OPTIONS_SEARCH" xml:space="preserve">
```
MihaelaStoica: OPTIONS_SEARCH and OPTIONS_SEARCH_DESC are not used

MihaelaStoica:
:+1:

PR #713, opened 2015-10-21
CA-184560: Fix the issue when the RPU wizard cannot check the host version after reboot

MihaelaStoica:
In the RPU wizard, after we notice that the host has been rebooted we try to resolve the host from cache, but sometimes the cache is not populated at that point.
Then, because we cannot resolve the host, we cannot check the host version and we assume is a newer version (which might not be true).
This commit adds a call to retry to resolve the host (with timeout).

### DIFF
--- a/XenAdmin/Wizards/RollingUpgradeWizard/PlanActions/AutomaticBackgroundThread.cs
+++ b/XenAdmin/Wizards/RollingUpgradeWizard/PlanActions/AutomaticBackgroundThread.cs
@@ -81,6 +81,11 @@ namespace XenAdmin.Wizards.RollingUpgradeWizard.PlanActions
                                 if (planAction is UpgradeHostPlanAction)
                                 {
                                     Host hostAfterReboot = host.Connection.Resolve(new XenRef<Host>(host.opaque_ref));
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/RollingUpgradeWizard/PlanActions/AutomaticBackgroundThread.cs
```diff
@@ -81,6 +81,11 @@ private void BworkerDoWork()
                                 if (planAction is UpgradeHostPlanAction)
                                 {
                                     Host hostAfterReboot = host.Connection.Resolve(new XenRef<Host>(host.opaque_ref));
+                                    if (hostAfterReboot == null)
+                                    {
+                                        log.InfoFormat("Host '{0}' cannot be resolved. Try to resolve it again with timeout", host.Name);
+                                        hostAfterReboot = (planAction as UpgradeHostPlanAction).Host;
```
stephen-turner: We don't need to resolve it once with XenConnection.Resolve and then again with UpgradeHostPlanAction.Host if it fails. Just use the latter.

stephen-turner:
Thanks, taking the change now.

PR #719, opened 2015-10-28
CP-14085: Feature leveling not needed on pool-join

MihaelaStoica:
Feature leveling should "just work" for Dundee and higher hosts, without the need for us to level them explicitly

### DIFF
--- a/XenModel/Actions/Pool/PoolAction.cs
+++ b/XenModel/Actions/Pool/PoolAction.cs
@@ -86,6 +86,10 @@ namespace XenAdmin.Actions
             // Mask the CPUs, and reboot the hosts (simultaneously, as they must all be on separate connections)
             foreach (Host host in hostsToCpuMask)
             {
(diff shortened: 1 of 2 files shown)


XenModel/Actions/Pool/PoolAction.cs
```diff
@@ -86,6 +86,10 @@ protected static bool FixCpus(Pool pool, List<Host> hostsToCpuMask, Func<HostAbs
             // Mask the CPUs, and reboot the hosts (simultaneously, as they must all be on separate connections)
             foreach (Host host in hostsToCpuMask)
             {
+                // No CPU masking is needed for Dundee or greater hosts 
+                if (Helpers.DundeeOrGreater(host))
```
stephen-turner: I think we should Assert here, rather than (or perhaps, as well as) just continuing, because Dundee hosts shouldn't get here.

stephen-turner:
Thanks, merging now.

PR #722, opened 2015-10-30
CP-14641: Add warnings on pool join if the pool and the joining host have different CPU features

MihaelaStoica:


stephen-turner:
Looks good now, merging.

PR #724, opened 2015-11-02
CA-186232: Fix the "Test credentials" button in the Health Check dialog

MihaelaStoica:

### DIFF
--- a/XenAdmin/Dialogs/HealthCheck/HealthCheckSettingsDialog.cs
+++ b/XenAdmin/Dialogs/HealthCheck/HealthCheckSettingsDialog.cs
@@ -320,7 +320,7 @@ namespace XenAdmin.Dialogs.HealthCheck
 
         private void CheckXenServerCredentials()
         {
-            if (!CheckCredentialsEntered())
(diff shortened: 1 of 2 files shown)


XenAdmin/Dialogs/HealthCheck/HealthCheckSettingsDialog.cs
```diff
@@ -320,7 +320,7 @@ private void testCredentialsButton_Click(object sender, EventArgs e)
 
         private void CheckXenServerCredentials()
         {
-            if (!CheckCredentialsEntered())
+            if (string.IsNullOrEmpty(textboxXSUserName.Text) || string.IsNullOrEmpty(textboxXSPassword.Text))
```
GaborApatiNagy: could we add Trim() to user name fields?

MihaelaStoica: Thanks Gabor, I'll add trim() to all instances where we use the username

GaborApatiNagy:
Looks good

PR #726, opened 2015-11-03
CA-180576: Provide a option to give case number when uploading a repo…

MihaelaStoica:
…rt from the Server Status Report wizard

- added Case Number edit box on the Destination page of the wizard
- valid case number needs to be 9 or less digits


XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
```diff
@@ -66,10 +66,12 @@ public XenServerHealthCheckUpload(string token, int verbosity, string uploadUrl,
         }
 
         // Request an upload and fetch the uploading id from CIS.
-        public string InitiateUpload(string fileName, long size)
+        public string InitiateUpload(string fileName, long size, string caseNumber)
         {
             // Request a new bundle upload to CIS server.
             string FULL_URL = UPLOAD_URL + "bundle/?size=" + size.ToString() + "&name=" + fileName;
```
GaborApatiNagy: Not really becasue of this change set, but in general I think special characters should be URLencoded. Like & should be %26 (this character is allowed in file names)

MihaelaStoica: Thanks Gabor, I will raise a separate ticket for the url encoded filename

GaborApatiNagy:
:+1:

XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
GaborApatiNagy: Cool, thanks @MihaelaStoica

PR #728, opened 2015-11-04
CA-187150: Dundee XenCenter cannot create iSCSI SRs on older hosts

GaborApatiNagy:
Fixed regression that was caused by hiding the Storage Provisioning page


MihaelaStoica:
:+1:

PR #729, opened 2015-11-04
CP-14470: UI: Show Management Agent not installed when guest agent does not support upgrading

GaborApatiNagy:
CP-14470: UI: Show Management Agent not installed when guest agent does not support upgrading


MihaelaStoica:
looks fine

PR #732, opened 2015-11-05
CA-187698: When restoring dismissed updates, only do it on the connec…

MihaelaStoica:
…tions with sufficient privileges

- This fixes the uncaught exception raised when an unauthorized user tries to restore dismissed updates.
- Also added the roles check when dismissing updates.


GaborApatiNagy:
:+1:

PR #736, opened 2015-11-09
 CA-149229: Add VDI and VIF mappings when calling "assert_can_migrate"

MihaelaStoica:


GaborApatiNagy:
:+1:

PR #739, opened 2015-11-09
CA-188017: When checking credentials on the Health Check dialog, disp…

MihaelaStoica:
…lay a better error message when the user is not authorised to connect to the server


GaborApatiNagy:
:+1:

PR #744, opened 2015-11-23
CA-188554: XenCenter Error Uncaught exception System.NullReferenceException

GaborApatiNagy:
Fixed the exception that was caused by a missing null reference check and also the fact that the license status had to be watched only for pools and hosts, so added a check before starting watching the license status of the IXenObject (on GeneralTabPage)

### DIFF
--- a/XenAdmin/TabPages/GeneralTabPage.cs
+++ b/XenAdmin/TabPages/GeneralTabPage.cs
@@ -102,20 +102,18 @@ namespace XenAdmin.TabPages
             GeneralTabLicenseStatusStringifier ss = new GeneralTabLicenseStatusStringifier(licenseStatus);
             Program.Invoke(Program.MainWindow, () => 
             {
-                if (pdSectionLicense != null)
-                    pdSectionLicense.UpdateEntryValueWithKey(
-                        FriendlyName("host.license_params-expiry"),
-                        ss.ExpiryDate, 
-                        ss.ShowExpiryDate);
+                pdSectionLicense.UpdateEntryValueWithKey(
+                    FriendlyName("host.license_params-expiry"),
+                    ss.ExpiryDate, 
+                    ss.ShowExpiryDate);
             });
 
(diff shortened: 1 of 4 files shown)


XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -179,13 +191,21 @@ public IXenObject XenObject
...
+        private void SetupAndStartLicenseStatus(IXenObject xo)
         {
+            System.Diagnostics.Debug.Assert(xo is Host || xo is Pool);
+
             licenseStatus = new LicenseStatus(xo);
             licenseStatus.ItemUpdated += licenseStatus_ItemUpdated;
             licenseStatus.BeginUpdate();
         }
 
+        private void UnregisterLicenseStatusUpdater()
```
stephen-turner: I think we should combine SetupAndStartLicenseStatus and UnregisterLicenseStatusUpdater into one function to do the unregistering and registering, so that they're obviously next to each other in the code flow.

XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -96,19 +96,27 @@ public GeneralTabPage()
...
                 return;
 
             GeneralTabLicenseStatusStringifier ss = new GeneralTabLicenseStatusStringifier(licenseStatus);
-            Program.Invoke(Program.MainWindow, () => pdSectionLicense.UpdateEntryValueWithKey(
-                                            FriendlyName("host.license_params-expiry"),
-                                            ss.ExpiryDate, 
-                                            ss.ShowExpiryDate));
+            Program.Invoke(Program.MainWindow, () => 
+            {
+                if (pdSectionLicense != null)
```
stephen-turner: Is this check needed? I don't see how pdSectionLicense can be null.

stephen-turner:
Looks fine now, merging.

PR #745, opened 2015-11-24
CA-186244: Cross-pool copy of a stopped VM allows selection of the current pool

GaborApatiNagy:
CrossPoolMigrateCanMigrateFilter will accept a parameter to decide whether to allow same target pool or not -- if allowSameTargetPool is false, it won't.


stephen-turner:
The user has already selected Cross-pool copy, so I think the pool it's in should be invisible, not just disabled.

stephen-turner:
While you're in there, please change "cannot migrate to resident host or pool" to "cannot migrate to the same host"

stephen-turner:
Looks good now: merging.

PR #746, opened 2015-11-25
CA-178985: Removed ova.xml import from XenCenter

GaborApatiNagy:
because it has been broken and we decided not to keep it


XenAdmin/Wizards/ImportWizard/ImportSourcePage.cs
```diff
@@ -244,9 +242,7 @@ private bool GetDiskCapacityXva(out string error)
 				FileInfo info = new FileInfo(m_textBoxFile.Text);
 				ImageLength = info.Length > 0 ? (ulong)info.Length : 0;
 
-				DiskCapacity = IsXvaVersion1
-				               	? GetTotalSizeFromXmlGeneva() //Geneva style
-				               	: GetTotalSizeFromXmlXva(GetXmlStringFromTarXVA()); //xva style
+				DiskCapacity = GetTotalSizeFromXmlXva(GetXmlStringFromTarXVA()); //xva style
```
stephen-turner: The function GetTotalSizeFromXmlGeneva() is (I assume) no longer called and can be removed.

Also applyVersionOneFiles() from ImportVmAction.cs.

stephen-turner:
While you're in there, please change "cannot migrate to resident host or pool" to "cannot migrate to the same host"

stephen-turner:
Thanks, merging now.

PR #749, opened 2015-12-01
CA-188553: [PAR-80] Creating a VLAN with MTU 9000 should be disabled …

GaborApatiNagy:
…if the base interface's MTU is < 9000

Fixed how we set the Maximum for the dropdown controls on:
* Creating External network,
* Creating Bonds
pages on the New Network Wizard


stephen-turner:
The code looks OK. However,
1. It needs to be dev tested with NICs that support higher than 1500.
2. It needs to be dev tested when there are three or more NICs.
3. If the field is fixed at MTU = 1500 and can't be edited, it should be greyed out with an explanatory message.

GaborApatiNagy:
The explanatory message is not final, but all the discussed changes are implemented and dev-tested.

stephen-turner:
Looks good now, merging.

PR #750, opened 2015-12-01
CA-189674: Move VM shows an irrelevant (VIF mapping) error, but it sh…

GaborApatiNagy:
…ouldn't

Fixed filters that are used to determine whether a vm can be migrated to a destination or not.


MihaelaStoica:
Looks good. Merging

PR #753, opened 2015-12-08
CP-14874:Enable XC SSH support for Netscaler VPX & all Linux VMs

GaborApatiNagy:
Upgraded logic in VM.IsWindows getter


MihaelaStoica:
Looks good.

PR #754, opened 2015-12-08
CA-189084: Control domain memory alerts not getting generated in Xencenter.

GaborApatiNagy:
Dom0 memory alerts have to be configured in the other_config of the Control Domain (Dom0) and not of the the host. This changeset implements this exception.

### DIFF
--- a/XenModel/Actions/Perfmon/PerfmonDefinitionAction.cs
+++ b/XenModel/Actions/Perfmon/PerfmonDefinitionAction.cs
@@ -115,9 +115,6 @@ namespace XenAdmin.Actions
                 }
             }
 
-            if (dom0_memory_usage != null && perfmonDefinitions.Contains(dom0_memory_usage))
-                perfmonDefinitions.Remove(dom0_memory_usage);
-
             if (perfmonDefinitions == null || perfmonDefinitions.Count == 0)
             {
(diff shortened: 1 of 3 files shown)


XenModel/Actions/Perfmon/PerfmonDefinitionAction.cs
```diff
@@ -64,6 +65,59 @@ protected override void Run()
...
+                    int found = dom0PerfmonDefinitions.RemoveAll(d => d.IsDom0MemoryUsage);
+                    if (found > 0)
+                    {
+                        string dom0PerfmonDefinitionsXml = PerfmonDefinition.GetPerfmonDefinitionXML(dom0PerfmonDefinitions);
+                        Helpers.SetOtherConfig(Session, dom0Vm, PerfmonDefinition.PERFMON_KEY_NAME, dom0PerfmonDefinitionsXml);
+                    }
+                }
+            }
+
+            if (dom0_memory_usage != null && perfmonDefinitions.Contains(dom0_memory_usage))
```
MihaelaStoica: I think this is not needed, as the item has been removed from the list already in the previous 'if' block.

MihaelaStoica:
:+1:

PR #757, opened 2015-12-14
CA-190928: Disable "Disconnect Workload Balancing Server" menu option…

MihaelaStoica:
… if the feature is restricted by license


GaborApatiNagy:
Looks fine

PR #758, opened 2015-12-14
CA-148491: XenServer breaks if any XenAPI client leaks sessions

GaborApatiNagy:
Fixing XenServerHealthCheck to use the modern overload of session.login_with_password. With this change XC and all its components are now sending an originator field (with some version added to it).


MihaelaStoica:
Looks fine, Merging

PR #759, opened 2015-12-15
CA-161457: Hide network if PIF.managed==false

MihaelaStoica:


GaborApatiNagy:
Looks good

PR #761, opened 2015-12-16
CA-191315: Storage motion from XenCenter should no longer be blocked …

MihaelaStoica:
…on WLB-enabled pools

The restriction that Storage Motion was not possible in the presence of WLB has been removed in Dundee. This commit removes the check in XenCenter.


GaborApatiNagy:
:+1:

PR #763, opened 2015-12-17
CA-160922: When a exception occurs in an http PUT call, set a timeout…

MihaelaStoica:
… on the function that polls for task result

- set a timeout on this function so that, after 2 minutes, we give up on waiting for the task status to change and just report the error


GaborApatiNagy:
Neat

PR #764, opened 2015-12-17
CP-14822: UI: New VM Creation: update code to apply recommendations

GaborApatiNagy:
This changeset implements the followings:

On creation of a new VM,
* XenCenter will use licensed state of the host to determine whether it should obey/apply the template-recommendation regarding the vendor device.
* If the host is licensed, and the selected template's recommendations suggest that has_vendor_device can be set/should be set to true, as part of VM creation, XenCenter will set the vm's has_vendor_device based on the recommendation using set_has_vendor_device API call (after it has cloned the template and before calling vm.provision).
* XenCenter will not expose details whether 1G or 2G VM is going to be created.


MihaelaStoica:
:+1:

PR #767, opened 2016-01-04
CA-194940: WLB - Hide host disk IO settings for XenServer previous to Dundee.

huizh:
1. Check the Xenserver platform version and display the host disk read/write
threshold and metric weighting for Dundee or later XenServer.
2. Update the metric weighting page to make it autosize-able.

### DIFF
--- a/XenAdmin/Dialogs/Wlb/WlbConfigurationDialog.cs
+++ b/XenAdmin/Dialogs/Wlb/WlbConfigurationDialog.cs
@@ -80,9 +80,11 @@ namespace XenAdmin.Dialogs.Wlb
                 }
 
                 verticalTabs.Items.Add(wlbThresholdsPage);
+                wlbThresholdsPage.Connection = _pool.Connection;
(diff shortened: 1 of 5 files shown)


stephen-turner:
@huizh, can you explain why this feature is decided according to the PlatformVersion in the WlbConfiguration? Why not just use the pool's platform version directly? It seems to me that writing the PlatformVersion into the WlbConfiguration will cause trouble when the pool is upgraded (because it won't pick up new features). And it leads to duplicated code: if you used the pool platform version here, you could use the standard DundeeOrGreater() function, and we wouldn't need to keep the logic in two places.

Maybe there's something I've overlooked, so please tell me if I'm wrong.

huizh:
@stephen-turner, thanks a lot for the suggestion. Totally agree that we should use the pool's platform version directly. I will update the commit accordingly.

stephen-turner:
Thanks, @huizh. I haven't tested it myself, but it looks fine to me now. Merging.

huizh:
Thanks, @stephen-turner.

PR #768, opened 2016-01-07
CA-187681: Improvement for VM.IsWindows property

GaborApatiNagy:
VM.IsWindows  will try to use guest_metrics referenced by the last_booted_record when the guest_metrics on the VM is not available.
This gives some extra chance to detect a Non-Windows VM.


MihaelaStoica:
Looks good

PR #769, opened 2016-01-08
CP-15396: Use xenadmin-branding and change the location of icons

cheng--zhang:
As build script will use the temp branch to do the build and the replacement already existed in current solution, we can just do the rebranding directly in the code repo and will not impact the next build since the next build will create other code branch with latest code commit.

Build script will:
1. Copy file from git://hg.uk.xensource.com/carbon/${XS_BRANCH}/XenAdmin-Branding.git
2. Replace the branding information to xenadmin/XenAdmin_Branding/

Solution will:
Change the usage of the local icons and BMPs to use xenadmin/XenAdmin_Branding/ one.

Parameter: XC_BRANDING
If define XC_BRANDING before run build.sh, then we can copy XC_BRANDING directory in XenAdmin_Branding to our solution and generate XC_BRANDING branded build.
By default, it will copy citrix directory.

What we need to do further:
Add XenAdmin-Branding.git with:
    /citrix
        /Image/...


cheng--zhang:
Hi @MihaelaStoica 
Could you please help to review the code change and merge if possible?
Thanks.

MihaelaStoica:
Looks OK. Merging

PR #771, opened 2016-01-12
CA-159790: [XSO-208] MeddlingActions don't show location information

GaborApatiNagy:
Updated the GetLocation extension method for ActionBase objects to return action.Connection.FriendlyName should it have no Host or Pool set up. This way we see location for MeddlingActions.
Also changed some messages (Task Names) to include "VM" in them too, because xe command line initiated VM lifecycle actions looked really misleading after the above fix in XenCenter (Eg. restarting a VM would show up as "Restarting", "server" which would suggest that not just the VM, but the host is being restarted.)


MihaelaStoica:
:+1:

PR #772, opened 2016-01-13
CA-195662: Remove signing as a post build event from splash project

MihaelaStoica:


GaborApatiNagy:
Oh so this was it!
:+1:

PR #773, opened 2016-01-13
CA-187681: xs-tools ISO doesn't get mounted on dlvm xva(Centos 7 HVM)

GaborApatiNagy:
There were times when "HasUpgradeSupportInManagementAgent" was needed, but not anymore, so this commit is changing code to use only VirtualisationStatus.MANAGEMENT_INSTALLED again.


MihaelaStoica:
Looks good. Merging.

PR #774, opened 2016-01-14
CA-192640: Don't log exceptions when HealthCheck service is not insta…

robhoes:
…lled


robhoes:
Hmm... I think that there is something wrong with the line endings...

robhoes:
Updated patch: fixed the line endings.

GaborApatiNagy:
Interesting, this file seems to have and has had Unix line endings. Several files have Unix and there may still be mixed ones around... 
Probably the easiest is not to allow Visual Studio & git to convert them automatically.

GaborApatiNagy:
:+1:

PR #776, opened 2016-01-14
CP-14472: Remove the lock in the License Manager and ensure that all the updates are done on the main thread.

MihaelaStoica:
- changed the licenseStatus_ItemUpdated event handler to invoke on the main thread, as this is the only case when a background thread needs to update the storedRows list..All the other reads and writes are taken place on the main (UI) thread
- removed the storedRowsLock as it is not needed since we do all the updates to storedRows on one thread.


stephen-turner:
:clap::clap::clap:

GaborApatiNagy:
:fingers_crossed:

PR #781, opened 2016-01-18
CP-15408: Add license restrictions

MihaelaStoica:
Port changes from cream-indigo for the additional restrictions


XenAdmin/Dialogs/VmSnapshotDialog.cs
```diff
@@ -175,7 +175,7 @@ private void CheckpointInfoPictureBox_Click(object sender, EventArgs e)
         {
             string tt;
             if (Helpers.FeatureForbidden(_VM, Host.RestrictCheckpoint))
-                tt = string.Format(Messages.FEATURE_NOT_AVAILABLE_NEED_ENTERPRISE_OR_PLATINUM_PLURAL.Replace("\\n", "\n"), Messages.DISKMEMORY_SNAPSHOTS);
+                tt = Messages.FIELD_DISABLED;
```
cheng--zhang: Can we add more information here like:
string.Format(Messages.FIELD_DISABLED.Replace("\\n", "\n"), Messages.DISKMEMORY_SNAPSHOTS);
So the detail reason can also go with the message.

MihaelaStoica:
@cheng--zhang More information about the license type could be useful for the user, but we would need to make sure that the correct information is displayed for each XenServer version that XenCenter supports, as the license editions for which the feature is available can be different from one server version to the other. Given all this, I think the generic message is OK here (better than the wrong message, anyway)

XenAdmin/Commands/Controls/MigrateVMToolStripMenuItem.cs
```diff
@@ -89,6 +89,10 @@ public MigrateVMCommand2(IMainWindow mainWindow, IEnumerable<SelectedItem> selec
 
             private static bool CanExecute(VM vm)
             {
+                if (vm.Connection != null && vm.Connection.IsConnected)
+                    if (Helpers.FeatureForbidden(vm.Connection, Host.RestrictInterPoolMigrate))
```
GaborApatiNagy: Could we use Intra- or Cross- instead of Inter- to have the same text everywhere?

GaborApatiNagy:
Looks fine

PR #784, opened 2016-01-20
CA-192640: Don't log exceptions when HealthCheck service is not insta…

robhoes:
…lled (part 2)


MihaelaStoica:
Looks good

PR #787, opened 2016-01-25
Removing possible null object reference

koushikcgit:
Signed-off-by: Koushik Chakravarty <Koushik.Chakravarty@citrix.com>

### DIFF
--- a/XenAdmin/Program.cs
+++ b/XenAdmin/Program.cs
@@ -821,6 +821,9 @@ namespace XenAdmin
             {
                 if (!IsExiting(c))
                 {
+                    if (c == null)
+                        return;
+
                     if (c.InvokeRequired)
                     {
                         MethodInvoker exceptionLogger = () =>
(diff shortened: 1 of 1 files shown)


XenAdmin/Program.cs
```diff
@@ -821,7 +821,7 @@ public static void Invoke(Control c, MethodInvoker f)
             {
                 if (!IsExiting(c))
                 {
-                    if (c.InvokeRequired)
+                    if (c != null && c.InvokeRequired)
```
GaborApatiNagy: I think it would be better to keep this as it was and just have a 

if (c == null)
   return;

so we don't get to the else to call f() when c is "accidentally" null at this point

koushikcgit:
@GaborApatiNagy : Changed accordingly

GaborApatiNagy:
Thanks @koushikcgit 
merging

PR #791, opened 2016-01-26
CP-15409: Add read-only mode to License manager

MihaelaStoica:


GaborApatiNagy:
It looks like this way is the feasible way to solve this, to avoid bigger changes that would be harder to test.
:+1:

PR #796, opened 2016-01-28
 CA-196887: New function for reading registry key written by the installer

MihaelaStoica:
- When the installer writes HiddenFeatures registry entry, it specifies registry key HKMU, that means that it will be written in HKLM if the user selected "All Users" for installation, or to HKCU if the user selected "Just Me".
- Added a new function to read an installed key by trying CurrentUser first, then Local Machine.
This function specifies a registry view (RegistryView.Registry32) when it opens the key (On the 64-bit version of Windows, portions of the registry are stored separately for 32-bit and 64-bit applications. There is a 32-bit view for 32-bit applications and a 64-bit view for 64-bit applications).

+
CP-15401: Move HiddenFeatures to a new class


cheng--zhang:
Looks good!

PR #797, opened 2016-01-29
CA-192539: Adding iNode host to XenCenter should be restricted

cheng--zhang:
Merge code from indigo and also make it rebrandable.


MihaelaStoica:
:+1:

PR #801, opened 2016-01-29
CP-15711: Modify Select Update Page

GaborApatiNagy:
Added Automatic option, but the option itself is disabled for now.


MihaelaStoica:
Looks good.

PR #804, opened 2016-02-01
CA-196231: Don't confuse start-on with migration

robhoes:
The "Start on Server" VM context menu incorrectly enabled hosts that a VM can
only be started on after a storage migration.  This was due to the fact that
the VMOperationToolStripMenuItem class is used for both "Start on Server" as
well as "Migrate to Server".

This change ensures that start checks are done when a start is requested, and
migration checks otherwise.


stephen-turner:
@robhoes, do we understand how this happened? I assume it's a regression?

robhoes:
@stephen-turner I assume it is due to the expansion of the scope of SXM, which is now able to migrate VMs that are not running. I think that this has changed the outcome of `cpmCmd.CanExecute` for shutdown VMs, cause this issue.

stephen-turner:
Ah yes, that makes sense. Thanks.

MihaelaStoica:
Looks good. Merging.

PR #806, opened 2016-02-02
CP-15404: Make help file rebrandable

cheng--zhang:
This pull request is to make the help file rebrandable.

I also removed 2 file that haven't been use in XenCenter.
XenAdmin/InvisibleMessages.ja.resx
XenAdmin/InvisibleMessages.zh-CN.resx


stephen-turner:
@cheng--zhang, why are those two files not used? My understanding is that they allow us to supply different URLs for different languages if we want to.

cheng--zhang:
@stephen-turner 
The files I remove is in XenAdmin directory. This dir only contains CN and JA resources without EN resx and any designer to use it.
I think it was add by mistake and remove these files.

There is InvisibleMessages in XenModle that we actually use.
See: https://github.com/xenserver/xenadmin/blob/master/XenModel/InvisibleMessages.resx

stephen-turner:
Ah, I understand now. We must have forgotten to remove the localised versions when we moved it from XenAdmin to XenModel. Thank you,.

MihaelaStoica:
Looks fine. Merging.

PR #811, opened 2016-02-03
CP-15399: Move check for updates url to Branding.cs

MihaelaStoica:
- InvisibleMessages.resx contains the default url for check for updates, which can be overwritten in Branding.cs
- Added an extra parameter to DownloadUpdatesXmlAction for the updates url. If this is not specified (or null), then the default one (from InvisibleMessages) is used


GaborApatiNagy:
:+1:

PR #813, opened 2016-02-04
Avoid sed cygwin replace CRLF to IF by using -b

cheng-z:
This pull request is to:
Using -b in sed to make sure file will keep windows ending.


MihaelaStoica:
looks fine. merging.

PR #815, opened 2016-02-04
CP-15825: Fix error where the About dialog pops up when a connection …

MihaelaStoica:
…status changes


GaborApatiNagy:
looks good

PR #820, opened 2016-02-05
CA-194025: Variable $date-of-software replaced with actual year.

stephen-turner:
I used the year the code was added to XenCenter: it may be older than that,
but I couldn't easily find out the exact date.

I also made the LegalNoticesDialog resizable, because it's silly trying to
read the legal notices in a teeny tiny window.


MihaelaStoica:
Looks good. Merging.

PR #821, opened 2016-02-08
CA-197388: Fix error where License Manager displays "Standard feature…

MihaelaStoica:
…s only" for Premium edition


GaborApatiNagy:
:+1:

PR #822, opened 2016-02-08
CA-196315: XS65ESP1018 not showing up in XenCenter Updates (on staging)

GaborApatiNagy:
Fixed code to treat uuids of patches case-insensitive when consuming, but not changing them (eg. to lowercase) to keep compatibility with older XAPIs.


MihaelaStoica:
Lokks good. Merging.

PR #825, opened 2016-02-09
CP-15399: Changed Branding.cs so that only the constants get rebranded

MihaelaStoica:


GaborApatiNagy:
I see why we need to have it this way, so :+1:

PR #826, opened 2016-02-10
Fix null reference exception that can occur when running the tests

MihaelaStoica:


GaborApatiNagy:
:+1:

PR #827, opened 2016-02-10
CP-16012: Rename branding variables

MihaelaStoica:


GaborApatiNagy:
I could not review all lines for good reason (too much), but it looks good and let's see it on Jenkins

PR #830, opened 2016-02-11
CP-15402: New Home Page (html based)

GaborApatiNagy:
Replaced existing to a new, html-based Home Page


GaborApatiNagy:
Note, that this is not the final version, it doesn't need to be merged yet.

stephen-turner:
Gabor, I haven't looked at the code in detail, but I'm wondering why you deleted HomePage.cs and added HomePageWeb.cs. I would prefer to reuse HomePage.cs.

Also, can we call the directory with the files in HomePage rather than Home, please. Thank you.

MihaelaStoica:
Gabor, can we please put all the test from the License file in the Legal Notices dialog, and than we won't need to include this file.

GaborApatiNagy:
@stephen-turner @MihaelaStoica Thanks for the comments, I've done all the changes needed. I dropped some commits in order to keep the history of the HomePage.cs (and related) files.

XenAdmin/HomePage/License
```diff
@@ -0,0 +1,23 @@
+    Blueprint CSS Framework License
```
stephen-turner: I don't think we need this file if we have it in Legal Notices.

XenAdmin/Properties/Resources.resx
```diff
@@ -127,9 +127,6 @@
   <data name="sl_add_storage_system_16" type="System.Resources.ResXFileRef, System.Windows.Forms">
     <value>..\Images\sl_add_storage_system_16.png;System.Drawing.Bitmap, System.Drawing, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
   </data>
-  <data name="homepage_background" type="System.Resources.ResXFileRef, System.Windows.Forms">
```
stephen-turner: If you're going to remove these images from the resources, you can remove them from the source tree too.

WixInstaller/XenCenter.wxs
```diff
@@ -80,6 +80,8 @@
                             <File Id="DiscUtils" Source="..\XenAdmin\bin\Release\DiscUtils.dll" />
                             <File Id="XenOvf" Source="..\XenAdmin\bin\Release\XenOvf.dll" />
                             <File Id="XenOvfTransport" Source="..\XenAdmin\bin\Release\XenOvfTransport.dll" />
+							<!-- Homepage -->
+	                        <File Id="" Source="..\XenAdmin\bin\Release\HomePage\index.mht" />
```
stephen-turner: To think about: is this the right location to pick this file up from?

stephen-turner:
Apart from these small comments, the changeset looks good: very neat.

GaborApatiNagy:
Updated PR, please see my latest commit for the changes. Thank you

stephen-turner:
@GaborApatiNagy, I'm pulling these changes, but you still haven't addressed the request to remove no-longer-used images from the source tree. Please do that in a separate PR. Thanks.

PR #835, opened 2016-02-12
CP-15576: Hide containers options that aren't available for Windows S…

MihaelaStoica:
…erver

- Container pause is not available if the the container is on a Windows VM.
- Processes tab is not available if the the container is on a Windows VM.

### DIFF
--- a/XenAdmin/Commands/PauseDockerContainerCommand.cs
+++ b/XenAdmin/Commands/PauseDockerContainerCommand.cs
@@ -63,7 +63,7 @@ namespace XenAdmin.Commands
 
         private static bool CanExecute(DockerContainer dockerContainer)
         {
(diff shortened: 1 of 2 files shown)


stephen-turner:
@MihaelaStoica The ticket suggests looking at vm->other_config['docker_info']['OperatingSystem'], whereas you've looked at our usual vm.IsWindows test. Could you explain why you've chosen that one?

MihaelaStoica:
@stephen-turner I thought that, as we already have a function to determine if a VM is Windows, we should use it, instead of writing another one that would need to parse the OperatingSystem string. I had also checked with Robert and he was OK with using our usual IsWindows test.

stephen-turner:
Thanks. If Robert doesn't have any reason why we should use his variable, I'm happy. Merging.

PR #839, opened 2016-02-15
Scheduled Snapshots Feature

koushikcgit:
This PR has all the commits for the VMSS feature (REQ-277)


stephen-turner:
I have noted on the xapi pull request, https://github.com/xapi-project/xen-api/pull/2577, that there are terminology problems in the SDK -- places where "scheduled_snapshot" should be "snapshot_schedule". This will affect the function names in this pull request too.

XenModel/XenAPI-Extensions/IVMPolicy.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        //string test { get; }
+        string Name { get; }
+        string name_description {get; }
+        string uuid { get; }
+        List<XenRef<VM>> VMs { get; }
+        IXenConnection Connection { get; }
+        bool is_enabled();
+        bool is_running();
+        bool is_archiving();
```
stephen-turner: It would be more idiomatic to make is_enabled, is_running and is_archiving into getters, not functions.

XenModel/XenAPI-Extensions/VMSS.cs
```diff
@@ -0,0 +1,271 @@
...
+    public partial class VMSS : IVMPolicy
+    {
+        public bool is_enabled()
+        {
+            return this.enabled;
+        }
+
+        public bool is_running()
+        {
+            return false;
```
stephen-turner: Why is is_running always false for VMSS? Is it the case that we have this info for VMPP but not for VMSS for some reason?

XenModel/Messages.resx
```diff
@@ -11496,6 +11572,14 @@ To learn more about the [XenServer] Performance Reporting capabilities feature o
 
 To learn more about the [XenServer] Role Based Access Control feature or to start a [XenServer] trial, click the button below.</value>
   </data>
+  <data name="UPSELL_BLURB_VMSS" xml:space="preserve">
+    <value>Upgrade your [XenServer] license to enable VM Snapshot Schedule. VM Snapshot Schedule allows you to create automatic snapshots for your critical VMs.</value>
```
stephen-turner: The feature is called "VM Scheduled Snapshots". (This affects five lines below this too, and maybe other places).

XenModel/Messages.resx
```diff
@@ -1874,12 +1874,18 @@ Archive destination: {0}</value>
   <data name="ASSIGNED_PROTECTION_POLICY" xml:space="preserve">
     <value>Protection policy assigned.</value>
   </data>
+  <data name="ASSIGNED_VMSS_POLICY" xml:space="preserve">
+    <value>Snapshot schedule policy assigned.</value>
```
stephen-turner: The phrase "snapshot schedule policy" sounds clumsy to me. I would rather just call it a "snapshot schedule", i.e. without the "policy". (This affects lots of strings).

stephen-turner:
We still have a lot of places where we test (typeof(T) == typeof(VMPP)) or (typeof(T) == typeof(VMSS)). The purpose of having the IVMPolicy interface is to do away with all (or almost all) of these. Something like archiving, or quiescing, where the features genuinely differ may be an exception. But for anything where the features are essentially the same, but are using different classes or variable names in the API, the IVMPolicy interface should abstract these differences away leaving the class that's using them clean.

stephen-turner:
@koushikcgit, I haven't done a full review of the logic, but above are some higher-level comments on terminology and code design. I think we should address these first and then go into a more detailed review. I realise some of them are in Sharad's code and he's away for three weeks -- is anyone deputising for him?

XenModel/XenAPI-Extensions/IVMPolicy.cs
koushikcgit: I kept these as functions as this was what it was earlier in the VMPP.cs (XAPI-Extensions) and in the usage as well. Do we want to change that existing behaviour?

XenModel/XenAPI-Extensions/VMSS.cs
koushikcgit: Yes. We do not have is_running in the VMSS object. Hence this is by default 'false'

XenModel/Messages.resx
koushikcgit: Ah! So there seems to be a disconnect in what we would want to call this feature after all. I recall, from our discussion, that we wanted to have this as `VM Snapshot Schedule` and looks like we want to different now.
I would then prefer to have `VM Scheduled Snapshot` be consistent, in the XC, XAPI, SDK, etc. Does this sound acceptable?

koushikcgit:
@stephen-turner : I think the use of typeof(T) to differentiate VMPP and VMSS is present in areas where the distinction is required either for UI controls, or when different apis need to be invoked in some other objects (like vm, etc.). Let's discuss this in detail to see which ones should be removed.

XenModel/XenAPI-Extensions/IVMPolicy.cs
stephen-turner: I hadn't realised this was existing behaviour. But I think we should change it anyway, as it's a small change.

XenModel/XenAPI-Extensions/VMSS.cs
stephen-turner: OK

XenModel/Messages.resx
stephen-turner: No, sorry, I don't think that works.

The feature name is "Scheduled Snapshots" (actually, that's better than the "VM Scheduled Snapshots" I used above).
A single VMSS is a "Snapshot Schedule".
A "scheduled snapshot" is a single snapshot, created according to the schedule.

I have made the same comment on xapi-project/xen-api#2577

stephen-turner:
@koushikcgit I still think we can get rid of almost all of them. Do we need a GTM to go through some examples?

XenModel/XenAPI-Extensions/IVMPolicy.cs
koushikcgit: Indeed it's a small change. Will make the modifications.

koushikcgit:
@stephen-turner : I've updated the PR. Please have a look

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotFrequencyPage.cs
```diff
@@ -45,11 +45,12 @@ namespace XenAdmin.Wizards.NewPolicyWizard
 {
     public partial class NewPolicySnapshotFrequencyPage : XenTabPage, IEditPage
     {
-        public NewPolicySnapshotFrequencyPage()
+        public NewPolicySnapshotFrequencyPage(bool isVMSS)
```
stephen-turner: This file still needs to have the duplicate code removed.
_copy and _copyVMSS should be declared as IVMGroup. Then all the other instances of if(isVMSS) can be removed. (IVMGroup.frequency will have to return a new enum rather than vmss_frequency or vmpp_backup_frequency).

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            if (VMGroup<T>.isVMPolicyVMPP)
+                _clone.backup_type = BackupType;
+            else
+                _cloneVMSS.type = BackupTypeVMSS;
+            
+            return null;
+        }
+
+        private VMPP _clone;
```
stephen-turner: This class also needs duplicate code removed.

XenAdmin/Wizards/NewPolicyWizard/NewPolicyWizardSpecific.cs
```diff
@@ -0,0 +1,303 @@
...
+        private static string FormatBackupType(vmpp_backup_type backupType)
+        {
+            if (backupType == vmpp_backup_type.snapshot)
+                return Messages.DISKS_ONLY;
+            else if (backupType == vmpp_backup_type.checkpoint)
+                return Messages.DISKS_AND_MEMORY;
+            throw new ArgumentException("wrong argument");
+        }
+
+        private static string FormatBackupTypeVMSS(vmss_type backupType)
```
stephen-turner: This could be combined with the previous function by defining a new enum for backup type in IVMGroup.

XenAdmin/Wizards/NewPolicyWizard/NewPolicyWizardSpecific.cs
```diff
@@ -0,0 +1,303 @@
...
+                    name_description = xenTabPagePolicy.PolicyDescription,
+                    type = xenTabPageSnapshotType.BackupTypeVMSS,
+                    frequency = (vmss_frequency)xenTabPageSnapshotFrequency.Frequency,
+                    schedule = xenTabPageSnapshotFrequency.Schedule,
+                    retained_snapshots = xenTabPageSnapshotFrequency.BackupRetention,
+                    enabled = xenTabPageVMsPage.SelectedVMs.Count == 0 ? false : true,
+                    Connection = Pool.Connection
+                };
+
+                var action = new CreateVMSS(vmss, xenTabPageVMsPage.SelectedVMs, xenTabPageFinish.RunNow);
```
stephen-turner: I think the CreateVMPP and CreateVMSS blocks can also be combined, via a VMGroup<T>.CreateObjectAction() function. We will pass in the archive data even for VMSS, and it will get ignored later.

XenModel/Actions/VMPP/AssignVMsToPolicyAction.cs
```diff
@@ -36,57 +36,64 @@
...
         private List<XenRef<VM>> _selectedVMs;
+        private IVMPolicy _policy;
 
-        public RemoveVMsFromPolicyAction(VMPP vmpp, List<XenRef<VM>> selectedVMs)
-            : base(vmpp.Connection, selectedVMs.Count == 1 ?
-            string.Format(Messages.REMOVE_VM_FROM_POLICY, vmpp.Connection.Resolve(selectedVMs[0]), vmpp.Name)
-            : string.Format(Messages.REMOVE_VMS_FROM_POLICY, vmpp.Name))
+        public RemoveVMsFromPolicyAction(IVMPolicy policy, List<XenRef<VM>> selectedVMs)
+            : base(policy.Connection, selectedVMs.Count == 1 ?
+            (typeof(T) == typeof(VMPP) ? string.Format(Messages.REMOVE_VM_FROM_POLICY, policy.Connection.Resolve(selectedVMs[0]), policy.Name) :
```
stephen-turner: I'm happy to have the typeof(T) == typeof(VMPP) here, but let's move it inside the string.Format for conciseness.

XenModel/Actions/VMPP/ChangePolicyEnabledAction.cs
```diff
@@ -39,24 +39,39 @@
...
-            bool value = !_vmpp.is_policy_enabled;
-            Description = value ? string.Format(Messages.ENABLING_VMPP, _vmpp.Name) : string.Format(Messages.DISABLING_VMPP, _vmpp.Name);
-            
-            VMPP.set_is_policy_enabled(Session, _vmpp.opaque_ref, !_vmpp.is_policy_enabled);
-            Description = value ? string.Format(Messages.ENABLED_VMPP, _vmpp.Name) : string.Format(Messages.DISABLED_VMPP, _vmpp.Name);
-            
+            bool value = !_policy.is_enabled;
+            if (typeof(T) == typeof(VMPP))
+            {
+                Description = value ? string.Format(Messages.ENABLING_VMPP, _policy.Name) : string.Format(Messages.DISABLING_VMPP, _policy.Name);
```
stephen-turner: Again in this area, let's move the typeof(T) == typeof(VMPP) inside the string.Format.

XenModel/Actions/VMPP/RunPolicyNowAction.cs
```diff
@@ -36,20 +36,21 @@
...
-            Pool = Helpers.GetPool(vmpp.Connection);
+            _policy = policy;
+            Pool = Helpers.GetPool(policy.Connection);
         }
 
         protected override void Run()
         {
-            Result = VMPP.protect_now(Session, _vmpp.opaque_ref);
-            Description = string.Format(Messages.RUN_POLICY_STARTED, _vmpp.Name);
+            var _vmpp = _policy as VMPP;
```
stephen-turner: _vmpp seems to be an unused variable. This won't compile in Release build.

XenModel/Actions/VMSS/CreateVMSS.cs
```diff
@@ -0,0 +1,75 @@
+﻿/* Copyright (c) Citrix Systems Inc.
```
stephen-turner: I think the CreateVMSS action and the GetVMSSAlertsAction should not be needed: we should template CreateVMPP and GetVMPPAlertsAction instead.

XenModel/FriendlyNames.resx
```diff
@@ -1214,6 +1403,24 @@
   <data name="Message.body-vmpp_xapi_logon_failure" xml:space="preserve">
     <value>Unable to log in to XAPI session.</value>
   </data>
+  <data name="Message.body-vmss_license_error" xml:space="preserve">
+    <value>The XenServer license you are using does not allow you to use the Scheduled Snapshot feature.</value>
```
stephen-turner: Scheduled Snapshots feature (in the plural).

XenModel/FriendlyNames.resx
```diff
@@ -546,18 +582,45 @@
   <data name="Label-host.edition-advanced" xml:space="preserve">
     <value>[XenServer product] Advanced Edition</value>
   </data>
+  <data name="Label-host.edition-desktop" xml:space="preserve">
```
stephen-turner: There are a lot of new strings in this file, and they don't appear to be all new or recent features. Where did they come from? We need to understand this before we can merge them.

XenModel/Messages.resx
```diff
@@ -2409,6 +2427,16 @@ Do you want to assign these VMs to the policy '{0}' instead?</value>
     
 Do you want to assign it to the policy '{2}' instead?</value>
   </data>
+  <data name="CHANGE_VMS_SNAPSHOT_SCHEDULE_WARNING" xml:space="preserve">
+    <value>Some of the VMs you selected are already assigned to a different snapshot schedule policy.
```
stephen-turner: Delete "policy"

XenModel/Messages.resx
```diff
@@ -1874,12 +1874,18 @@ Archive destination: {0}</value>
   <data name="ASSIGNED_PROTECTION_POLICY" xml:space="preserve">
     <value>Protection policy assigned.</value>
   </data>
+  <data name="ASSIGNED_VMSS_POLICY" xml:space="preserve">
+    <value>Snapshot Schedule assigned.</value>
```
stephen-turner: There's a lot of inconsistency whether we write Snapshot Schedule with or without upper case letters. I recommend standardising on lower case.

XenModel/Messages.resx
```diff
@@ -8430,6 +8485,12 @@ It is strongly recommended that you Cancel and apply the latest version of the p
   <data name="NEW_UPDATE_AVAILABLE" xml:space="preserve">
     <value>New Update Available - {0}</value>
   </data>
+  <data name="NEW_VMPP_PAGE_TEXT" xml:space="preserve">
+    <value>Use this wizard to define a VM protection policy that will ensure your critical VMs are protected via automatic scheduled snapshot jobs.</value>
+  </data>
+  <data name="NEW_VMSS_PAGE_TEXT" xml:space="preserve">
+    <value>Use this wizard to define a VM Snapshot Schedule that will create automatic schedule snapshot jobs for your VMs.</value>
```
stephen-turner: "automatic scheduled" (not "schedule").

XenModel/Messages.resx
```diff
@@ -3253,6 +3284,9 @@ Please reconnect the host and try again.</value>
   <data name="CREATED_VMPP" xml:space="preserve">
     <value>VM protection policy '{0}' created.</value>
   </data>
+  <data name="CREATED_VMSS" xml:space="preserve">
+    <value>VM Snapshot Schedule '{0}' created.</value>
```
stephen-turner: I think we should always refer to just "Snapshot schedule" not "VM snapshot schedule"

XenModel/XenAPI-Extensions/VMSS.cs
```diff
@@ -0,0 +1,305 @@
...
+                var min = Convert.ToInt32(backup_schedule_min);
+                return GetWeeklyDate(time, hour, min, new List<DayOfWeek>(DaysOfWeekBackup));
+            }
+            return new DateTime();
+
+        }
+
+        private static DateTime GetDailyDate(DateTime time, int min, int hour)
+        {
+            var nextDateTime = new DateTime(time.Year, time.Month, time.Day, hour, min, 0);
```
stephen-turner: In this file, any function which is static (or could be static like TryGetKey), we can just call the VMPP version directly, we don't need to make another copy of the code.

stephen-turner:
Thank you for your updates, @koushikcgit. This looks much better now. I hope you agree that this approach makes the code much cleaner.

I think we're close now. There are still a couple more places where we could avoid code duplication, which I've mentioned above, along with a couple of other things.

We haven't tried out the code yet. I'll try and find time to do that tomorrow.

stephen-turner:
What is the plan for merging this? I guess we need to wait until the xapi changes are in trunk?

GaborApatiNagy:
It looks like github could not merge this PR  because there are conflicts,
could you rebase/merge master into your branch as well, please? Thanks.

XenAdmin/Dialogs/VMProtectionRecovery/VMProtectionPoliciesDialog.cs
```diff
@@ -123,8 +88,8 @@ private class PolicyRow : DataGridViewRow
             private DataGridViewTextBoxCell _status = new DataGridViewTextBoxCell();
             private DataGridViewTextBoxCell _nextArchiveRuntime = new DataGridViewTextBoxCell();
             private DataGridViewTextAndImageCell _lastResult = new DataGridViewTextAndImageCell();
-            public readonly VMPP VMPP;
-            public PolicyRow(VMPP policy)
+            public readonly IVMPolicy _Policy;
```
cheng-z: _Policy -> _policy will be better I think.

koushikcgit:
Stephen, this can be merged to trunk without the xapi changes as the feature depends on the `restrict_vmss` license param.

stephen-turner:
@koushikcgit The code looks good now, thank you.

I did find a couple of bugs while I was trying it out on your test server:

1. Select a VM that already has a snapshot schedule; "assign to snapshot schedule" from the VM menu and choose another schedule; get error "Object has been deleted.VMSS:"
2. The New VMSS Wizard has "policy" in lots of places. First, even before you get into the wizard, the menu item is called "New policy"; four times on the first tab of the wizard; on the Snapshot Type tab (where it says "Protection Policy").

stephen-turner:
For the merge strategy, it does leak the name "snapshot schedule" in the UI, even if it's greyed out, so I would prefer to wait to merge until the feature is ready end-to-end. We can either keep this PR open, or merge it onto a branch.

XenModel/XenAPI-Extensions/VMPP.cs
```diff
@@ -166,6 +167,95 @@ public override string Description
...
+
+        public long policy_retention 
+        {
+            get { return backup_retention_value; }
+            set { backup_retention_value = value; }
+        }
+
+        public policy_backup_type policy_type 
+        {
+            get { return (policy_backup_type)backup_type; }
```
stephen-turner: This type of code is not really safe. It only happens to work because the policy_backup_type enum is in the same order as the vmpp_backup_type enum, with the additional items not present in vmpp_backup_type at the end. But you shouldn't make such assumptions about enums in case they get extended in a different way later.

I realise it's longer code, but it would be more idiomatic not to convert between different types of enums and to use a switch statement to enumerate through the backup types and return the right policy_backup type for each one (and vice versa in the setter).

koushikcgit:
@stephen-turner : Thanks for the review. I've updated the PR and things should now converge.
On the bugs you uncovered:
1. I've already seen this earlier and have filed a CA for this. Needs to be looked into from xapi side.
2. I've fixed these texts in the update to the PR

On the merge strategy:
1. Yes, I agree that it will leave the name "snapshot schedule" in the UI. Out of the suggestions you have, what is that you prefer?

stephen-turner:
Great, thanks. I'll look at it later today.

Re merging, I think my preference is to create a new branch and merge it onto there, and then merge it into master when the xapi work is done.

stephen-turner:
@koushikcgit, there was just one more little thing I noticed. I'm not sure we need the "Quiescing Supported" column in the New VMSS Wizard. Feel free to disagree if you think I'm wrong, but it seems to me to give it too much prominence. It's already covered on the next page by virtue of the option being greyed out or not.

stephen-turner:
I have created a "VMSS" branch which is currently identical to master. Could you raise the PR against that branch instead? Then we can keep it on there until we're ready to merge it into master. Thank you.

koushikcgit:
Creating new PR for merging into VMSS branch - https://github.com/xenserver/xenadmin/pull/850
Closing this one and referenced this into the new PR for the comments

PR #840, opened 2016-02-16
CA-198895: Have parameter Languages back in xencenter.l10n.diff

cheng-z:
This fix is to have CN version wizard back in XenCenter installer
Without Language, the l10n installer can only show EN wizard in CN environment.


GaborApatiNagy:
This looks good, but we need to investigate what else is needed before merging this

GaborApatiNagy:
Let's see

PR #843, opened 2016-02-16
CP-15402: Improvements to the new HomePage

GaborApatiNagy:


XenModel/InvisibleMessages.resx
```diff
@@ -138,6 +138,9 @@
   <data name="HOMEPAGE" xml:space="preserve">
     <value>http://www.citrix.com/xenserver</value>
   </data>
+  <data name="HOMEPAGE_FILENAME" xml:space="preserve">
+    <value>HomePage.mht</value>
```
stephen-turner: It would be good to add the equivalent line to the InvisibleMessages.(cn|jp).resx files, so the translators don't have to figure it out.

stephen-turner: Also, please add HomePage.ja.mht and HomePage.zh-CN.mht files (just copies of the English ones at the moment), so the translators have something to translate.

stephen-turner: And change the build scripts to bundle those files in the XenCenterResources.zip.

stephen-turner:
The layout looks better now, but I would like to
1. restore the centering
2. get rid of the vertical scrollbar when the window is narrow (which should just be a matter of losing whitespace from the bottom)

GaborApatiNagy:
I updated this PR with changes fixing issues you found.

stephen-turner:
Looks great now, thanks. Merging.

PR #844, opened 2016-02-17
CA-198962: Fix compile error of rebranding

cheng-z:
After this code change, we also need to modify branding.sh for non-citrix build to use:
REBRANDING_WEB_LATEST_BUILD
instead of 
WEB_LATEST_BUILD

### DIFF
--- a/mk/package-and-sign.sh
+++ b/mk/package-and-sign.sh
@@ -39,7 +39,8 @@ REPO=${XENADMIN_DIR}
 OUTPUT_DIR=${ROOT}/output
 DOTNETINST=${REPO}/dotNetInstaller
 	
-source ${REPO}/Branding/branding.sh
+BRANDING_BRAND_CONSOLE=[XenCenter]
+BRANDING_COMPANY_NAME_SHORT=[Citrix]
 
 WIX=${REPO}/WixInstaller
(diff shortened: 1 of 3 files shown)


mk/package-and-sign.sh
```diff
@@ -39,7 +39,8 @@ REPO=${XENADMIN_DIR}
 OUTPUT_DIR=${ROOT}/output
 DOTNETINST=${REPO}/dotNetInstaller
 	
-source ${REPO}/Branding/branding.sh
+BRANDING_BRAND_CONSOLE=[XenCenter]
```
GaborApatiNagy: It looks like it is not like

> BRANDING_COMPANY_NAME_SHORT=$(cat ${SCRATCH_DIR}/globals | grep -w COMPANY_NAME_SHORT | sed -e 's/COMPANY_NAME_SHORT=//g' -e 's/"//g')

from branding.sh, but a hard-coded value instead. Is this really what we want here?

cheng-z: Hi Gabor
Yes, it is not a hard code value, it is with [] which will be rebranded when build in rebranding.sh

GaborApatiNagy:
Taking this now, thanks @cheng-z

PR #848, opened 2016-02-18
CA-199083: Fix display issue in CN manageUpatePage

cheng-z:
This code change is to change the display to make sure all information can be displayed correctly.


cheng-z:
Hi @GaborApatiNagy 
Could you please help to check this commit?
I already verified this in my CN environment.

PR #854, opened 2016-02-19
CA-198767: Add only-physical-PIFs pool-join constraint

GaborApatiNagy:
There is a new pool join rule in xapi: the joining host may only have physical PIFs, i.e. no bonds, VLANs and tunnels (cross-server private networks). This is checked by asserting that for all PIFs of the joining host, PIF.physical = true. When this occurs, people should be encouraged to manually remove any existing bonds, VLANs or tunnels and try again.


stephen-turner:
Looks good, merging.

stephen-turner:
Actually, although I pulled this, I suddenly wondered: Does PIF.IsPhysical cover the case of bonds?

GaborApatiNagy:
It works fine for Bonds, because we've got 3 PIFs for a Bond, from which two have IsPhysical==false set

stephen-turner:
Great, thanks for checking.

PR #863, opened 2016-02-25
CA-199145: Home page missing the page icons that are used as list bullets

GaborApatiNagy:
Now the Mhtml file contains the missing image as well. I used IE to save it and apparently (like other browsers) it does not save css-specified background images, so had to add a 0x0px sized img into the html to make IE to save it. Easy workaround and no change to the layout.


stephen-turner:
I was skeptical about this hack, but I've changed my mind: it's a good hack. Pulling.

PR #865, opened 2016-02-26
CA-196643: Check the CPU feature compatibility in StartDiagnosisForm

MihaelaStoica:
- The StartDiagnosisForm dialog shows the reasons why a VM cannot be started or resumed. XenCenter tries to get these reasons from a call to VM.assert_can_boot_here, but his function does not perform a CPU feature compatibility check.
- This commit adds the CPU feature compatibility check for resume operations.


GaborApatiNagy:
looks ok

PR #866, opened 2016-02-26
Fix CA-20888 and CA-200532

stephen-turner:


XenModel/Alerts/PerfmonDefinition.cs
```diff
@@ -179,6 +179,7 @@ private PerfmonDefinition(XmlNode parentNode)
                 else if (node.Name.Equals(ALARM_TYPE_ELEMENT_NAME))
                 {
                     name = node.Attributes[ALARM_COMMON_ATTR_NAME].Value;
+                    success = true;
```
stephen-turner: NB This line is present so that we don't write an erroneous log line at the bottom of this function.

GaborApatiNagy:
Looks good, merging

PR #871, opened 2016-03-01
CA-175498: Better error message when patching fails

GaborApatiNagy:


MihaelaStoica:
Looks good. And I've tested it with supplemental pack installation to verify that the behaviour has not changed.

GaborApatiNagy:
Thanks

PR #872, opened 2016-03-01
CA-201414: Remove the background worker, which seemed to cause the He…

MihaelaStoica:
…alth Check dialog to be shown blank at startup.


GaborApatiNagy:
Looks fine :snowflake:

PR #873, opened 2016-03-02
CA-202295: Hotkeys should not be part of the branding tag

GaborApatiNagy:


MihaelaStoica:
Looks good. Merging.

PR #875, opened 2016-03-02
CA-201576: Windows Update capable text does not fit into the control

GaborApatiNagy:
Resized them accordingly


MihaelaStoica:
Looks good. Merging.

PR #876, opened 2016-03-03
CA-201645: The Install Update wizard should delete patch installation…

MihaelaStoica:
… file if manual mode is selected

- Added PoolPatchCleanAction which is called for each pool after the patch has been applied on all selected hosts


GaborApatiNagy:
Looks good

PR #886, opened 2016-03-07
CA-202427: Correction: we do still need to check the vendor, so movin…

stephen-turner:
…g the test below that


MihaelaStoica:
Thank you. Merging.

PR #887, opened 2016-03-07
CA-187132: XenCenter loses XenServer address when retrying connection

MihaelaStoica:


GaborApatiNagy:
Looks fine

PR #893, opened 2016-03-10
CA-202417: New VM Wizard Memory Modify button is not working reliably…

GaborApatiNagy:
… when specified in MB

Now we set Increment exactly as on other pages
when the unit is MB: the increment it will be calculated dynamically (1, 2, 4,...)
when GB, increment will be 1 GB for values >=10GB, 0.1GB otherwise


MihaelaStoica:
Looks good. Merging.

PR #895, opened 2016-03-11
CA-202477: XenCenter enables the Enter Maintenance mode button for a …

MihaelaStoica:
…host  even when the VM suspend is in progress


GaborApatiNagy:
Looks good

PR #899, opened 2016-03-14
CA-204172: Switch Health Check URLs to the production site

MihaelaStoica:
Also use 0 (never expire) for the diagnostic token.
And line endings.


GaborApatiNagy:
Looks good, but because of the line-ending changes, we'd make it harder to use the blame functionality

MihaelaStoica:
Thanks @GaborApatiNagy .
I've replaced the commit with a new one that doesn't change the line endings.

PR #900, opened 2016-03-14
CA-204188: Select "Optimize manually" when the shadow multiplier valu…

MihaelaStoica:
…e changes (not when the control is entered)


GaborApatiNagy:
:+1:

PR #905, opened 2016-03-17
CA-204191: Trying to avoid Uncaught exception System.InvalidOperation…

MihaelaStoica:
…Exception: Object is currently in use elsewhere.


GaborApatiNagy:
I agree this is making it safer - merging

PR #910, opened 2016-03-21
CA-204212: XenCenter doesn't warn on creating a vif with duplicate ma…

MihaelaStoica:
…c address on a different network


GaborApatiNagy:
:+1:

PR #911, opened 2016-03-23
L10N:Dundee:Xencenter:ITR002 transltion check-ins

sriharshak:


stephen-turner:
@sriharshak The changes to the HOMEPAGE_FILENAME variable appear to be wrong: they should refer to the localised .mht files, and then those files should be translated.

stephen-turner:
Thanks, merging now.

PR #914, opened 2016-03-24
CA-205204: Uncaught exception during RPU

MihaelaStoica:
When an exception is reported during the RPU, XenCenter tries to re-enable the host. But when this fails, we should just ignore it and report the original exception.


GaborApatiNagy:
Looks good :sunglasses:

PR #916, opened 2016-03-29
CA-198956: VM.set_snapshot_schedule throws exception "Object deleted: VMSS" in xapi

geosharath:
XAPI expects "OpaqueRef:NULL" in the value field to unassign a VM from
a particular policy, however XC is sending an empty string "". As a fix
the empty string is replaced with Helper.NullOpaqueRef.

### DIFF
--- a/XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
+++ b/XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
@@ -72,6 +72,31 @@ namespace XenAdmin.Dialogs.VMProtectionRecovery
         }
 
         private IVMPolicy _policy;
+        public void StartRefreshTab()
+        {
+            /* hoursFromNow has 3 possible values:
+                1) 0 -> top 10 messages (default)
+                2) 24 -> messages from past 24 Hrs
+                3) 7 * 24 -> messages from lst 7 days */
+
+            var hoursFromNow = 0;
+            switch (comboBox1.SelectedIndex)
+            {
+                case 0: /* default value*/
+                    break;
+                case 1:
+                    hoursFromNow = 24;
+                    break;
+                case 2:
+                    hoursFromNow = 7 * 24;
+                    break;
+            }
+
+            PureAsyncAction action = _policy.getAlertsAction(_policy, hoursFromNow);
+            action.Completed += action_Completed;
+            action.RunAsync();
+        }
+
         public void RefreshTab(IVMPolicy policy)
         {
             _policy = policy;
@@ -83,23 +108,7 @@ namespace XenAdmin.Dialogs.VMProtectionRecovery
             else
             {
                 comboBox1.Enabled = true;
-                int hoursFromNow = 0;
-                switch (comboBox1.SelectedIndex)
-                {
-                    case 0: /* default value*/
-                        break;
-                    case 1:
-                        hoursFromNow = 24;
-                        break;
-                    case 2:
-                        hoursFromNow = 7 * 24;
-                        break;
-                }
-        
-                PureAsyncAction action = _policy.getAlertsAction(_policy, hoursFromNow);
-                action.Completed += action_Completed;
-                action.RunAsync();
-                RefreshGrid(_policy.PolicyAlerts);
+                StartRefreshTab();
             }
 
         }
@@ -184,30 +193,7 @@ namespace XenAdmin.Dialogs.VMProtectionRecovery
         {
             if (_policy != null)
             {
-                /* hoursFromNow has 3 possible values:
-                    1) 0 -> top 10 messages (default)
-                    2) 24 -> messages from past 24 Hrs
-                    3) 7 * 24 -> messages from lst 7 days */
-
-                int hoursFromNow = 0 ;
-
-                switch (comboBox1.SelectedIndex)
-                {
-                    case 0: /* default value*/
-                        break;
-                    case 1:
-                        hoursFromNow = 24;
-                        break;
-                    case 2:
-                        hoursFromNow = 7 * 24;
-                        break;
-                }
-
-                dataGridView1.Rows.Clear();
-                panelLoading.Visible = true;
-                PureAsyncAction action = _policy.getAlertsAction(_policy, hoursFromNow);                
-                action.Completed += action_Completed;
-                action.RunAsync(); 
+               StartRefreshTab();  
             }
         }
 
--- a/XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
+++ b/XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
@@ -83,6 +83,22 @@ namespace XenAdmin.Dialogs.VMProtectionRecovery
             else
             {
                 comboBox1.Enabled = true;
+                int hoursFromNow = 0;
+                switch (comboBox1.SelectedIndex)
+                {
+                    case 0: /* default value*/
+                        break;
+                    case 1:
+                        hoursFromNow = 24;
+                        break;
+                    case 2:
+                        hoursFromNow = 7 * 24;
+                        break;
+                }
+        
+                PureAsyncAction action = _policy.getAlertsAction(_policy, hoursFromNow);
+                action.Completed += action_Completed;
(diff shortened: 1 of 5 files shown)


XenModel/XenAPI/VM.cs
```diff
@@ -3590,7 +3590,7 @@ public static void set_protection_policy(Session session, string _vm, string _va
         /// <param name="_value">The value</param>
         public static void set_snapshot_schedule(Session session, string _vm, string _value)
         {
-            session.proxy.vm_set_snapshot_schedule(session.uuid, (_vm != null) ? _vm : "", (_value != null) ? _value : "").parse();
+            session.proxy.vm_set_snapshot_schedule(session.uuid, (_vm != null) ? _vm : "", (_value != null) ? _value : Helper.NullOpaqueRef).parse();
```
MihaelaStoica: This file is auto generated from the xen-api, and we shouldn't change it, except by updating it with the generated one.
If we need to pass "OpaqueRef:NULL" to XAPI instead of the empty string, then we should make sure we don't call this function with _value==null (and explicitly use "OpaqueRef:NULL" when we want to unassign a VM from a policy)

XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
```diff
@@ -169,7 +169,13 @@ private void comboBox1_SelectedIndexChanged(object sender, EventArgs e)
             if (_policy != null)
             {
                 if (comboBox1.SelectedIndex == 0)
-                    RefreshGrid(_policy.PolicyAlerts);
+		{
```
MihaelaStoica: I think we still need to call RefreshGrid(_policy.PolicyAlerts) for VMPP, because in this case the PolicyAlerts contains the most recent 10 alerts and there is no need for a xapi call (VMPP.get_alerts) in GetVMPPAlertsAction.

XenModel/Actions/VMSS/GetVMSSAlertsAction.cs
```diff
@@ -54,16 +54,38 @@ protected override void Run()
...
+                2) 24 -> messages from past 24 Hrs
+                3) 7 * 24 -> messages from lst 7 days */
 
-            /*for VMSS: Populate the alerts from Messages by filtering out the alerts for this schedule
-                This is not required in VMPP as the VMPP record itself has the recentAlerts */
-            foreach (var message in messages)
+            if (_hoursFromNow == 0)
             {
-                if (message.cls == cls.VMSS && message.obj_uuid == VMSS.uuid)
+                int messageCounter = 0;
```
MihaelaStoica: The list of messages is not sorted and first 10 items are not necessarly the most recent 10 messages.

XenModel/XenAPI/VM.cs
stephen-turner: To be explicit, if this file is changed, the changes will be reverted whenever we next update it from the autogenerated one.

In any case, I'm confused about this fix. Why does VMPP not need it? Does the API behave differently for VMSS and VMPP? If so, the VMSS API should be fixed instead.

XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
```diff
@@ -169,7 +169,13 @@ private void comboBox1_SelectedIndexChanged(object sender, EventArgs e)
             if (_policy != null)
             {
                 if (comboBox1.SelectedIndex == 0)
-                    RefreshGrid(_policy.PolicyAlerts);
+		{
+                    dataGridView1.Rows.Clear();
```
MihaelaStoica: This sequence (ln 173-177) is repeted in each if block with the only difference on the last argument of the getAlertsAction() function. 
It could be optimised to just use a different value for that parameter, depending on comboBox1.SelectedIndex.

XenAdmin/Dialogs/VMProtectionRecovery/VMProtectionPoliciesDialog.cs
```diff
@@ -201,43 +201,54 @@ private void LoadPolicies()
...
+            /* populate the dictionary with policy uuid */
+
+            foreach (var policy in policyList)
+            {
+                policy.PolicyAlerts.Clear();
+                List<XenAPI.Message> messageList = new List<XenAPI.Message>();
+                policyMessage.Add(policy.uuid, messageList);
+            }
+       
+            /* iterate through all messages and populate the dictionary with message list */
```
MihaelaStoica: We should try to reuse the code in GetVMSSAlertsAction to populate the PolicyAlerts for each polocy with the last 10 messages.

MihaelaStoica: Also we need to make sure that the filter on the PolicyHistory is respected when we refresh the policies with LoadPolicies()

MihaelaStoica: But, thinking more of it, using GetVMSSAlertsAction for each policy is not the best solution, as it will itterate through messages for each policy. While the current solution only iterates through the messages once.

XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
geosharath: Currently were not respecting the filter hence Policy alerts would have most recent 10 alerts but now we have modified the code to respect the filter so there is a possibility that Policy Alerts would not hold 10 alerts but also 24 hrs and 7 days alerts.

XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
geosharath: Thanks for this suggestion, I have used a switch case as discussed.

XenAdmin/Dialogs/VMProtectionRecovery/VMProtectionPoliciesDialog.cs
geosharath: Thanks for you suggestion, retaining this committed code.

XenModel/Actions/VMSS/GetVMSSAlertsAction.cs
geosharath: Thanks a lot for pointing this out, I have sorted the items now.

XenModel/XenAPI/VM.cs
geosharath: Sharad is working on this and has found the root cause and will be fixing it from XAPI, hence I have retained the original code here.

XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
```diff
@@ -83,6 +83,22 @@ public void RefreshTab(IVMPolicy policy)
...
+                        break;
+                    case 2:
+                        hoursFromNow = 7 * 24;
+                        break;
+                }
+        
+                PureAsyncAction action = _policy.getAlertsAction(_policy, hoursFromNow);
+                action.Completed += action_Completed;
+                action.RunAsync();
                 RefreshGrid(_policy.PolicyAlerts);
```
MihaelaStoica: RefreshGrid() call (which populates the grid with the last 10 messages- not always the correct filter) is not needed here because it is done action_Completed()

XenAdmin/Dialogs/VMProtectionRecovery/PolicyHistory.cs
```diff
@@ -168,30 +184,30 @@ private void comboBox1_SelectedIndexChanged(object sender, EventArgs e)
         {
             if (_policy != null)
```
MihaelaStoica: All the code in this function can be reused in RefreshTab(), for example by extracting it into a function StartRefreshGrid() and calling this function in here and in RefreshTab() at line 86.

geosharath: Thanks for this suggestion, I have modified the commit.

MihaelaStoica:
Thanks @geosharath, this looks good now. Merging.

PR #917, opened 2016-03-30
CA-205118: RPU Wizard Fails Due to Host CPU Incompatibility Error

MihaelaStoica:
Added a Precheck to the RPU wizard, to check if the hosts in the pool have different CPU feature sets (using the old feature check for pre-Dundee pools) and display an error, offering to shut down all VMs.


GaborApatiNagy:
Looks good

PR #918, opened 2016-03-30
CA-205199: Reattaching an SR fails with error "The SR is not attached"

MihaelaStoica:
Set the name and description of the SR after the PBDs have been created and plugged.


GaborApatiNagy:
:+1:

PR #921, opened 2016-04-04
CA-202374: XC greyed out 'Install Tools' menu for imported VM from Cream

GaborApatiNagy:
Check different feature flag for Windows VMs on Dundee to detect Dundee guest agent. For earlier versions nothing is changed.
This will cause XC to report "Management not installed" for Windows VMs on Dundee hosts with Cream guest agent (This is the case when a host has been upgraded from Cream to Dundee).

### DIFF
--- a/XenModel/XenAPI-Extensions/VM.cs
+++ b/XenModel/XenAPI-Extensions/VM.cs
@@ -830,7 +830,7 @@ namespace XenAPI
             {
                 Debug.Assert(HasNewVirtualisationStates);
 
-                var flags = !Helpers.DundeeOrGreater(Connection) && HasRDP || HasStaticIP
+                var flags = HasStaticIP
                     ? VirtualisationStatus.MANAGEMENT_INSTALLED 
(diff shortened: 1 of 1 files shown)


XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -818,7 +830,9 @@ private VirtualisationStatus GetVirtualisationStatusNewVM
             {
                 Debug.Assert(HasNewVirtualisationStates);
 
-                var flags = HasRDP ? VirtualisationStatus.MANAGEMENT_INSTALLED : 0;
+                var flags = !Helpers.DundeeOrGreater(Connection) && HasRDP || HasStaticIP
```
MihaelaStoica: I think that DundeeOrGreater is not needed here, as this function will only be called for Dundee hosts

MihaelaStoica:
:+1:

GaborApatiNagy:
Thanks

PR #923, opened 2016-04-04
CA-205693: Add a check to prevent Index out of range error

MihaelaStoica:


GaborApatiNagy:
Looks fine

PR #925, opened 2016-04-05
CA-205116: Add scripts to push latest successful build to the ref rep…

MihaelaStoica:
…ository and mark build as keep-forever


stephen-turner:
Looks lovely.

PR #926, opened 2016-04-06
CA-202269: Hide "Subscription advantage" link on the Updates page bas…

MihaelaStoica:
…ed on the registry key


GaborApatiNagy:
:+1:

PR #930, opened 2016-04-08
CA-206761: Health Check shows "No issues found" when the report analysis is not yet available

MihaelaStoica:
- Try to get the analysis progress from cis (as an integer value representing percentage 0 - 100). If this function returns a valid progress less than 100, then we don't try to get the report analysis and continue to display "Analysis not yet available"
- If the progress is 100 then we get the analysis result and display it
- If the progress cannot be obtained, then we try to get the analysis result; If an empty result is obtained within 10 minutes from the upload completion, then we assume that the analysis is still in progress and continue to display "Analysis not yet available"

### DIFF
--- a/XenAdminTests/UnitTests/HealthCheckAnalysisProgressTest.cs
+++ b/XenAdminTests/UnitTests/HealthCheckAnalysisProgressTest.cs
@@ -39,11 +39,14 @@ namespace XenAdminTests.UnitTests
     [TestFixture, Category(TestCategories.Unit)]
     public class HealthCheckAnalysisProgressTest
     {
-        private Dictionary<string, int> expectedResults = new Dictionary<string, int>()
+        private Dictionary<string, double> expectedResults = new Dictionary<string, double>()
                                          { 
                                             { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 0 }", 0 },
                                             { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 10 }", 10 },
                                             { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 100 }", 100 },
+                                            { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 100.0 }", 100 },
+                                            { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 0.01 }", 0.01 },
+                                            { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 99.99 }", 99.99 },
                                             { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : 10,  \"131-435\" : 12 }", 10 },
                                             { "{ \"30253b07-138a-1b17-80a1\" : 12, \"30253b07-138a-1b17-80a1-117317ded3ca\" : 10 }", 10 },
                                             { "{ 30253b07-138a-1b17-80a1-117317ded3ca : 10 }", 10 },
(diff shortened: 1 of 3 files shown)


XenModel/Actions/HealthCheck/GetHealthCheckAnalysisResultAction.cs
```diff
@@ -118,6 +138,61 @@ private int GetDistinctIssueCount(List<AnalysisResult> issues)
...
+        /// <returns>the analysis progress as pecentage, or -1 if the JSON object is invalid</returns>
+        public static int ParseAnalysisProgress(string jsonString, string uploadUuid)
+        {
+            if (string.IsNullOrEmpty(jsonString))
+                return -1;
+            try
+            {
+                var serializer = new JavaScriptSerializer();
+                Dictionary<string, object> result = serializer.DeserializeObject(jsonString) as Dictionary<string, object>;
+                return result != null && result.ContainsKey(uploadUuid) ? (int) result[uploadUuid] : -1;
```
stephen-turner: What happens if the result is a decimal not an integer? What if it's 99.6?

stephen-turner: Could also return -1 here if the result isn't in the expected range [0, 100].

XenModel/Actions/HealthCheck/GetHealthCheckAnalysisResultAction.cs
```diff
@@ -81,8 +82,27 @@ protected override void Run()
...
+                    Description = Messages.COMPLETED;
+                    return;
+                }
+
                 var analysisResult = GetAnalysisResult(diagnosticToken, Pool.HealthCheckSettings.UploadUuid);
-                log.Info("Saving analysis result");
+
+                if (analysisResult.Count == 0 && analysisProgress == -1 && DateTime.Compare(Pool.HealthCheckSettings.LastSuccessfulUploadTime.AddMinutes(10), DateTime.UtcNow) > 0)
+                {
+                    log.InfoFormat("Diagnostic result for {0} is empty. Maybe analysis result is not yet available", Pool.Name);
```
stephen-turner: I think this should be .DebugFormat not .InfoFormat.

XenModel/Actions/HealthCheck/GetHealthCheckAnalysisResultAction.cs
```diff
@@ -81,8 +82,27 @@ protected override void Run()
                     Description = Messages.ACTION_GET_HEALTH_CHECK_RESULT_FAILED;
                     return;
                 }
+
+
+                int analysisProgress = GetAnalysisProgress(diagnosticToken, Pool.HealthCheckSettings.UploadUuid);
+
+                if (analysisProgress >= 0 && analysisProgress < 100) // check if the progress is a valid value less than 100
+                {
+                    log.InfoFormat("Analysis for {0} not completed yet", Pool.Name);
```
stephen-turner: I think this should be .DebugFormat not .InfoFormat.

XenAdminTests/UnitTests/HealthCheckAnalysisProgressTest.cs
```diff
@@ -0,0 +1,69 @@
...
+                                            { "{ \"30253b07-138a-1b17-80a1\" : 12, \"30253b07-138a-1b17-80a1-117317ded3ca\" : 10 }", 10 },
+                                            { "{ 30253b07-138a-1b17-80a1-117317ded3ca : 10 }", 10 },
+                                            { "{ 30253b07-138a-1b17-80a1-117317ded3ca : 10, 2345: 25 }", 10 },
+                                            { "{ 2345: 25, 30253b07-138a-1b17-80a1-117317ded3ca : 10 }", 10 },
+                                            { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : not a number }", -1 },
+                                            { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : \"not a number\" }", -1 },
+                                            { "{ \"30253b07-138a-1b17-80a1-117317ded3ca\" : \"10\" }", -1 },
+                                            { string.Empty, -1 },
+                                            { " ", -1 },
+                                            { "{ }", -1 }
```
stephen-turner: We could test some decimals, and some numbers outside the range [0, 100], here.

stephen-turner:
Looks good now, thank you.

PR #948, opened 2016-04-18
CA-137301: Double-click should choose item in Move VM/VDI dialogs

stephen-turner:


stephen-turner:
NB This pull request should *not* go on Dundee, so we may choose to take it later when the branches have already diverged.

XenAdmin/Controls/CustomTreeView.cs
```diff
@@ -529,6 +532,8 @@ protected override void OnMouseDoubleClick(MouseEventArgs e)
                 Resort();
                 Refresh();
             }
+
+            DoubleClickOnRow(this, e);
```
MihaelaStoica: I think that a null check is needed here, as not all usages of CustomTreeView would have this event handler assigned.
Same for srListBox_DoubleClickOnRow() in SrPicker.cs

stephen-turner:
@MihaelaStoica Thanks, fixed.

PR #949, opened 2016-04-20
CA-208495: XenCenter does not specify the NFS version when reattaching an existing SR

MihaelaStoica:
- Re-position the version radio-buttons, so that is clear that the selected version applies to both Create and Reattach
- Pass the selected version to the attach command in the device_config parameter


GaborApatiNagy:
:+1:

PR #966, opened 2016-04-26
CA-208808: The items of “Day of the week” drop-down list are not localized

MihaelaStoica:

### DIFF
--- a/XenAdmin/Core/HelpersGUI.cs
+++ b/XenAdmin/Core/HelpersGUI.cs
@@ -503,6 +503,23 @@ namespace XenAdmin.Core
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


stephen-turner:
I think this is wrong, because it will be localised to the language of the OS, not the language of the program (so you will end up with "foreign" day names in the middle of English text).

MihaelaStoica:
@stephen-turner This function will actually return the day of the week localized to the language of the program, because the DateTimeFormatInfo.CurrentInfo object is based on the CultureInfo object for the current thread, which in this case is the UI thread (and the curent culture of the UI thread is the programe's culture, defined in InvisibleMessages.LOCALE).
However, as this wasn't clear from the code, I've created a function in HelpersGUI, similar to DateTimeToString()

stephen-turner:
Ah yes, makes sense. Thanks. Merging now.

PR #983, opened 2016-05-12
CA-210732: XenCenter VM properties GPU always shows "None"

MihaelaStoica:
The error occurs when there is a gpu group with only one vgpu type: When setting the selected item in the vGPU combobox, XenCenter tries to select the group header instead of the vGPU type; because the group header is not selectable, we then default to "None".
With this fix, we ensure that two items are not equal if one is a group header and the other is a subitem.

### DIFF
--- a/XenAdmin/Controls/ComboBoxes/VgpuComboBox.cs
+++ b/XenAdmin/Controls/ComboBoxes/VgpuComboBox.cs
@@ -195,6 +195,9 @@ namespace XenAdmin.Controls
                     if (!VgpuTypes[i].Equals(other.VgpuTypes[i]))
                         return false;
                 }
+
+                if (IsGpuHeaderItem && other.IsVgpuSubitem)
+                    return false;
             }
 
             return result;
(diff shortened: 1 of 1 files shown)


XenAdmin/Controls/ComboBoxes/VgpuComboBox.cs
```diff
@@ -195,6 +195,9 @@ public bool Equals(GpuTuple other)
                     if (!VgpuTypes[i].Equals(other.VgpuTypes[i]))
                         return false;
                 }
+
+                if (IsGpuHeaderItem && other.IsVgpuSubitem)
```
stephen-turner: In order to make a valid Equals function, we need (a == b) => (b == a). I.e., we need the reverse case here as well: this.IsVgpuSubitem && other.IsGpuHeaderItem.

GaborApatiNagy: If we are changing this method, could we make sure other == null would not throw a null reference exception?

MihaelaStoica:
Thanks @stephen-turner. I've updated the function.

stephen-turner:
Thanks, merging.

PR #986, opened 2016-05-12
CA-209502: Exception while deleting multiple folders

GaborApatiNagy:
No longer running delete folder actions parallel


stephen-turner:
What is the performance impact of this change?

MihaelaStoica:
Closig this PR, as it is replaced by #1012

PR #997, opened 2016-05-18
CP-15719: UI: Implement error handling/reporting on Upload & Install page

GaborApatiNagy:
Implemented error handling.
If there has been an error in a pool, we stop executing actions in that pool immediately, and report the error once every other actions have finished. Error in one pool will not affect the upgrade process of other pools.


MihaelaStoica:
Looks good. Merging.

PR #1002, opened 2016-05-23
CA-208707: [SCTX-2316] XenCenter Incorrect pop-up error message displayed if SSL secure channel cannot be created

GaborApatiNagy:
My commit fixes the error message so the dialog will show "Could not create SSL/TLS secure channel." instead of "An unknown error occurred". This text is localised.
Unfortunately we see only a WebException with Status=SecureChannelFailure with no other details - couldn't provide an improved error message


MihaelaStoica:
:+1:

PR #1007, opened 2016-05-25
CA-211418: Exception in XenCenter while attaching vdi to a VM on a p…

Frezzle:
…ool with an SR with 20,000 VDIs

All event handlers are now unsubscribed when AttachDiskDialog is closed, allowing its memory to be garbage collected.
Image for each VDI now references an existing image instead of creating a new object.

Signed-off-by: Frederico Mazzone <fredericom@citrite.net>


MihaelaStoica:
Looks good, merging.

PR #1008, opened 2016-05-31
CP-15714: UI: Change Select Servers Page

GaborApatiNagy:
DownloadUpdatesXmlAction is now called on PageLeave and not when the user selects Automatic Mode. This way we get the error message at the right time, also showing a progress dialog what makes the UX much better.
New tooltips for servers on the Select Server Page to to cover two more cases


MihaelaStoica:
Looks good. Merging.

PR #1010, opened 2016-06-02
CA-187187: Enable AD on pool failing when one of the hosts have failed

GaborApatiNagy:
Calling disable on a best effort basis. (We always call disable AD before enabling AD -  CA-48122)


MihaelaStoica:
Looks good. Merging.

PR #1012, opened 2016-06-06
CA-209502: Exception while deleting multiple folders

MihaelaStoica:
- Changed the FolderAction class so that it can be used to delete or move multiple folders across connections, instead of having multiple actions (one for each holder) which can not run in parallel.
- The api calls that update the EMPTY_FOLDERS entry in pool's other_config are preformed last for all connections, avoiding cross-thread exceptions when two threads try to add/remove the same key.

### DIFF
--- a/XenModel/Actions/Folders/FolderAction.cs
+++ b/XenModel/Actions/Folders/FolderAction.cs
@@ -65,12 +65,17 @@ namespace XenAdmin.Actions
         //   leave extra empty folders in the list, so we just catch the RBAC error and fail silently.
         //
         // * For safety, destructive operations are called last: duplication is not as bad as destructive failure.
+        //
+        // * When this action is used to move or delete multiple objects across multiple connections, 
+        //   the action can not be sudo'ed as it does not have a single session that can be escalated. In this the action
+        //   would fail if any of the connections does not have sufficient roles.
+        //
 
         private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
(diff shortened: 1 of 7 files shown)


XenModel/Actions/Folders/FolderAction.cs
```diff
@@ -192,9 +227,68 @@ public override void RecomputeCanCancel()
...
+            {
+                AddFoldersToEmptyList(con, cancelling, emptyFolders[con].ToArray());
+            }
+            RemoveFoldersFromEmptyList(nonEmptyFolders.ToArray(), cancelling);
+        }
+
+        internal void DeleteOrMove(List<IXenObject> objects, Folder folder, Func<bool> cancelling)
+        {
+            IXenConnection connection = null;
+            if (objects.Count > 0 && objects[0] != null)
```
kc284: Is it likely to have null objects in the list? If so we need to add a null check in the other DeleteOrMove overload.
Alternatively we can filter out null objects at the beginning of this method - also return if the list is empty.

kc284:
Apart from the comment above, the changes look ok to me.

Not strictly related to this ticket, but overall the FolderAction could use some subclassing to get rid of the Kind enum, simplify the code and be more object oriented instead of having comments on the use of constructors etc.

XenModel/Actions/Folders/FolderAction.cs
MihaelaStoica: The null check is not needed, we should never have null objects in the list.

MihaelaStoica:
Thanks @kc284 for the suggestion. I've refactored the FolderAction file, with subclasses for each action kind.

kc284:
Look better now. Merging.

PR #1013, opened 2016-06-07
CA-212078: Do not check for CPU compatibility feature when the VM's r…

kc284:
…esident host is not dundee or greater.

### DIFF
--- a/XenAdmin/Commands/VMOperationHostCommand.cs
+++ b/XenAdmin/Commands/VMOperationHostCommand.cs
@@ -185,8 +185,8 @@ namespace XenAdmin.Commands
 
         public static bool VmCpuFeaturesIncompatibleWithHost(Host targetHost, VM vm)
         {
-            // check the CPU feature compatibility for Dundee and higher hosts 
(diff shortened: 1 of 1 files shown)


XenAdmin/Commands/VMOperationHostCommand.cs
```diff
@@ -185,8 +185,8 @@ protected override string GetCantExecuteReasonCore(SelectedItem item)
 
         public static bool VmCpuFeaturesIncompatibleWithHost(Host targetHost, VM vm)
         {
-            // check the CPU feature compatibility for Dundee and higher hosts 
-            if (!Helpers.DundeeOrGreater(targetHost) || !Helpers.DundeeOrGreater(vm.Connection))
+            // check the CPU feature compatibility for Dundee and higher hosts
+            if (!Helpers.DundeeOrGreater(targetHost) || !Helpers.DundeeOrGreater(vm.Connection.Resolve(vm.resident_on)))
```
MihaelaStoica: We also need to consider the case when resident_on is null.

MihaelaStoica:
Looks good. Merging.

PR #1019, opened 2016-06-10
CA-211369: focus of the machine/console switching back to the vm post snapshot with disk and memory is performed incorrectly/misleading

Frezzle:
ConsolePanel.setCurrentSource() being called in ConsolePanel.Snapshot() was causing the console view to switch to the target snapshotted VM's console, but only if the user is already on a console tab. The screenshot now happens before the snapshot begins, instead of after; this way, the screenshot always occurs when not on the console tab.


MihaelaStoica:
Looks goog. Merging.

PR #1023, opened 2016-06-13
[CA-207306] Select box is not in line with pGPU bar on GPU tab of host

Frezzle:
The last pGPU's row was stretching to the bottom, causing the lowest select box to fill its space and go down too much.
Changed the select boxs' dock style and margin.

Signed-off-by: Frederico Mazzone <fredericom@citrite.net>


MihaelaStoica:
:ballot_box_with_check:

PR #1024, opened 2016-06-14
CA-213236: Wrong license edition reported at pool level

MihaelaStoica:
Fix the order in the Edition enumeration


GaborApatiNagy:
looks fine

PR #1025, opened 2016-06-14
CA-212147: RPU failing with insufficient license error when host lice…

MihaelaStoica:
…nse removed

If the license does not allow VM migration, then cause the VMs to be shut down / suspended on the pre-checks page


XenAdmin/Diagnostics/Checks/AssertCanEvacuateCheck.cs
```diff
@@ -78,8 +80,18 @@ protected List<Problem> CheckHost()
                     problems.Add(new LocalCD(this, residentVM));
                     VMsWithProblems.Add(residentVM.opaque_ref);
                 }
+
+                if (restrictMigration && residentVM.is_a_real_vm && !VMsWithProblems.Contains(residentVM.opaque_ref))
```
kc284: For performance reasons, I believe this should be the first check-and-if-true-continue block. If the licence restricts migration, do we care about the other problems?

MihaelaStoica: I think we still care about the other problems, even when migration is restricted: We require to switch off VM's auto power, so that VMs don't start automatically while in the middle of an upgrade. And we require to eject the tools ISO before RPU, because after the upgrade, the old tools iso is guaranteed to be missing.

kc284:
Agreed.

PR #1026, opened 2016-06-14
CP-17752: make unit tests use a generic class of policy wizard and

geosharath:
added support for New Policy Wizard help document.

### DIFF
--- a/XenAdmin/Wizards/NewPolicyWizard/NewPolicyWizardSpecific.cs
+++ b/XenAdmin/Wizards/NewPolicyWizard/NewPolicyWizardSpecific.cs
@@ -268,7 +268,7 @@ namespace XenAdmin.Wizards.NewPolicyWizard
 
             else if (!VMGroup<T>.isVMPolicyVMPP)
             {
-                return "NewPolicyWizardSpecific_" + CurrentStepTabPage.HelpID + "Pane";    
+                return "NewPolicyWizardVMSS_" + CurrentStepTabPage.HelpID + "Pane";    
             }
 
             return "NewPolicyWizard_" + CurrentStepTabPage.HelpID + "Pane";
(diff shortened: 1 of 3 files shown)


XenAdminTests/WizardTests/NewPolicyWizardTest.cs
```diff
@@ -38,9 +38,9 @@ namespace XenAdminTests.WizardTests.cowleyPolicies_xml
 {
 
     [TestFixture, Category(TestCategories.UICategoryB)]
-    class NewPolicyWizardTest : WizardTest<NewPolicyWizard>
```
MihaelaStoica: This won't solve the generics problem we have in the tests, because the NewPolicyWizard is generic and we need to use NewPolicyWizardSpecific<XenAPI.VMPP> instead. This would work:
 `class NewPolicyWizardTest : WizardTest<NewPolicyWizardSpecific<XenAPI.VMPP>>`

XenAdmin/Wizards/NewPolicyWizard/NewPolicyWizardSpecific.cs
```diff
@@ -261,8 +261,18 @@ protected override void FinishWizard()
...
         {
-            return CurrentStepTabPage is RBACWarningPage ? FormatHelpId("Rbac") : base.WizardPaneHelpID();
+            if (CurrentStepTabPage is RBACWarningPage)
+            {
+                return FormatHelpId("Rbac");
+            }
+
+            else if (!VMGroup<T>.isVMPolicyVMPP)
+            {
+                return "NewPolicyWizardSpecific_" + CurrentStepTabPage.HelpID + "Pane";
```
MihaelaStoica: I propose to use the prefix _NewPolicyWizardVMSS_ instead of _NewPolicyWizardSpecific_, to make it clear that we are refering to the VMSS wizard

MihaelaStoica:
Looks good now. Merging.

PR #1031, opened 2016-06-20
CP-17231: Implement proxy authentication setting in XenCenter

Frezzle:
Basic and Digest authentication implemented. Tested with FreeProxy (Basic+Digest), CCProxy (Basic) and AnalogX Proxy (no authentication); each were tested with correct credentials, incorrect credentials and missing credentials; all gave expected results.


XenModel/XenAPI/HTTP.cs
```diff
@@ -150,45 +163,90 @@ private static string ReadLine(Stream stream)
...
-        private static bool ReadHttpHeaders(ref Stream stream, IWebProxy proxy, bool nodelay, int timeout_ms)
+        private static bool ReadHttpHeaders(ref Stream stream, IWebProxy proxy, bool nodelay, int timeout_ms, List<string> headers = null)
         {
-            string response = ReadLine(stream);
-            int code = getResultCode(response);
+            // read headers/fields
+            string line = ReadLine(stream), initialLine = line, transferEncodingField = null;
+            if (string.IsNullOrEmpty(initialLine)) // sanity check
+                return false;
+            while (!string.IsNullOrEmpty(line) && !string.IsNullOrWhiteSpace(line))
```
kc284: IsNullOrWhiteSpace checks for empty string too, so the first check is not necessary (you could add it as a reminder/comment at the end of the line because the method name doesn't sound like it).

XenModel/XenAPI/HTTP.cs
```diff
@@ -299,6 +357,39 @@ public static Uri BuildUri(string hostname, string path, params object[] args)
...
+            return null;
+        }
+
+        private static string GetPartOrNull(string str, int partIndex)
+        {
+            string[] parts = str.Split(new char[] { ' ' }, partIndex + 2, StringSplitOptions.RemoveEmptyEntries);
+            return partIndex < parts.Length - 1 ? parts[partIndex] : null;
+        }
+
+        private static void RemoveNewLineEnding(ref string str)
```
kc284: You can use str.TrimEnd('\r', '\n'), in which case you don't need a new method either, you can call it inline.

XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.resx
```diff
@@ -358,14 +358,233 @@
...
+    <value>0, 5, 3, 0</value>
+  </data>
+  <data name="BypassForServersCheckbox.Size" type="System.Drawing.Size, System.Drawing">
+    <value>276, 17</value>
+  </data>
+  <data name="BypassForServersCheckbox.TabIndex" type="System.Int32, mscorlib">
+    <value>8</value>
+  </data>
+  <data name="BypassForServersCheckbox.Text" xml:space="preserve">
+    <value>Bypass proxy server for &amp;all [XenServer] connections</value>
```
kc284: The shortcut key A is already used for the Address textbox.

XenModel/XenAPI/HTTP.cs
```diff
@@ -38,6 +38,7 @@
 using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
 using System.Runtime.Serialization;
+using XenAdmin.Core;
```
kc284: This file/class comes from the C# SDK, which knows nothing about XenCenter, hence we cannot use references to XC namespaces. If there are methods useful for the SDK, consider moving them to that level, otherwise consider moving the code using them from this class to a XC only class extension.

XenModel/XenAPI/HTTP.cs
```diff
@@ -299,6 +357,39 @@ public static Uri BuildUri(string hostname, string path, params object[] args)
             return uri.Uri;
         }
 
+        private static List<string> GetFields(List<string> lines, string field)
```
kc284: Do we need these new get methods? Why not use linq FindAll, FirstOrDefault etc. instead?

kc284:
In HTTP.cs it may also be worth investigating why there is a lot of legwork reading/writing lines and handling line endings instead of using a StreamReader/Writer. Best effort only, as this is old code and not strictly related to this ticket.

Frezzle:
@kc284 I've made the suggested changes. Also, I tried out using StreamReader and it gets stuck reading; it doesn't seem to be able to recognize that it's the end of the stream. I Googled the issue and this seems to be a common problem; the solution of using ReadByte to implement a ReadLine function (as we have) seems to be a common workaround. Shall I make a ticket for investigating this further?

kc284:
Nope, I just wanted to know whether there was a reason for this. Thanks for researching it.

stephen-turner:
@Frezzle I haven't looked at the code, but is there a place we could add a code comment to avoid us researching the same question again on a future occasion?

XenModel/XenAPI/HTTP.cs
```diff
@@ -469,11 +456,11 @@ public static Stream ConnectStream(Uri uri, IWebProxy proxy, bool nodelay, int t
         private static void AuthenticateProxy(ref Stream stream, Uri uri, IWebProxy proxy, bool nodelay, int timeout_ms, List<string> initialResponse, string header)
         {
             // perform authentication only if proxy requires it
-            List<string> fields = GetFields(initialResponse, "Proxy-Authenticate:");
+            List<string> fields = initialResponse.FindAll(delegate(string str) { return str.StartsWith("Proxy-Authenticate:"); });
```
kc284: More compact syntax using a lambda: initialResponse.FindAll(str => str.StartsWith(""Proxy-Authenticate:""))

kc284:
Looks ok now.

PR #1038, opened 2016-06-24
CP-17470: Warn users if they attempt rebooting or shutting down a host running a...

kc284:
... VM with the hci-warn-before-shutdown flag. Minor changes to make the warnings for
rebooting and shutting down more consistent (preselect the No button for rebooting;
extended the Yes button text; modified warning text)

### DIFF
--- a/XenAdmin/Commands/RebootHostCommand.cs
+++ b/XenAdmin/Commands/RebootHostCommand.cs
@@ -31,6 +31,7 @@
 
 using System.Collections.Generic;
 using System.Drawing;
+using System.Linq;
 using System.Windows.Forms;
 using XenAdmin.Actions;
 using XenAdmin.Properties;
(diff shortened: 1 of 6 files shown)


XenModel/Messages.resx
```diff
@@ -2705,11 +2719,14 @@ Are you sure you want to export VM {2}?</value>
...
   </data>
   <data name="CONFIRM_REBOOT_SERVER_NO_VMS" xml:space="preserve">
-    <value>Are you sure you want to reboot the selected server?</value>
+    <value>Are you sure you want to reboot '{0}'?</value>
   </data>
   <data name="CONFIRM_REBOOT_SERVER_TITLE" xml:space="preserve">
     <value>Reboot Server</value>
   </data>
+  <data name="CONFIRM_REBOOT_SERVER_YES_BUTTON_LABEL" xml:space="preserve">
+    <value>Yes, Reboot</value>
```
MihaelaStoica: This button text is missing the hotkey.

XenAdmin/Commands/RebootHostCommand.cs
```diff
@@ -120,12 +121,32 @@ protected override string ConfirmationDialogText
...
-                    return (hasRunningVMs ? Messages.CONFIRM_REBOOT_SERVERS : Messages.CONFIRM_REBOOT_SERVERS_NO_VMS);
+                bool hasRunningVMs = false;
+                var hciHosts = new List<Host>();
+
+                foreach (Host h in hosts)
+                {
+                    if (h.HasRunningVMs)
+                        hasRunningVMs = true;
+
+                    if (h.Connection.ResolveAll(h.resident_VMs).FindAll(v => v.HciWarnBeforeShutdown).Count > 0)
```
MihaelaStoica: I think Exists() would be better (instead of FindAll and Count>0), because it returns when a match is found ant that's all we need here.

MihaelaStoica:
Looks good now. Merging.

PR #1040, opened 2016-06-24
CA-213414: Check feature-flag for ssl-legacy control switch

MihaelaStoica:
- For Dundee and newer hosts: the feature is restricted only if the "restrict_ssl_legacy_switch" flag exists and it is set to true
- For pre-Dundee hosts: the feature is restricted if the "restrict_ssl_legacy_switch" is absent or it is present and set to true


GaborApatiNagy:
Looks good

PR #1041, opened 2016-06-24
CP-17866: Grey out the "Move" button on the VM storage tab if the storage does not allow mirroring.

kc284:


MihaelaStoica:
Looks good. Merging.

PR #1042, opened 2016-06-27
Latest API bindings, including definition of dundee-plus release...

kc284:
...plus an unrelated refactoring which I did while investigating something on the vdi copy dialog, and a credits update.


MihaelaStoica:
Looks good. Merging.

PR #1043, opened 2016-06-27
[CA-213410] XenCenter ignores proxy settings in some places

Frezzle:
Assigned Proxy property in all WebClient, WebRequest and HttpWebRequest instances.
Removed "proxy not supported" message from Download Appliance dialog, since the download does work with a proxy.


XenAdmin/Plugins/Features/TabPageFeature.cs
```diff
@@ -961,6 +961,7 @@ private void writeXML(object context)
                 request.ContentType = "xml";
                 request.ContentLength = Encoding.UTF8.GetBytes(jsCallbackAndData[1]).Length;
                 request.UserAgent = Branding.BRAND_CONSOLE + "\\Plugin";
+                request.Proxy = XenAdminConfigManager.Provider.GetProxyFromSettings(connection, false);
```
MihaelaStoica: I think the second parameter in GetProxyFromSettings() needs to be "true", because this is a call to the server (See SessionUrl = connection.Session.Url;)

MihaelaStoica:
Looks good now. Merging.

PR #1045, opened 2016-06-29
CP-17689: Restrict maximum host failures to tolerate to the number sp…

kc284:
…ecified by the other-config:hci-limit-fault-tolerance field.


MihaelaStoica:
Looks good. Merging.

PR #1049, opened 2016-06-30
CA-116547: During Installation of Hotfix requiring restart of PV gues…

MihaelaStoica:
…ts displays Dom0 in the list of VMs to be rebooted in XenCenter.


kc284:
Looks ok to me. Merging.

PR #1051, opened 2016-07-01
Case sensitivity fixes 

kallisti5:
Fix case differences. Windows guys need to be more careful :-)


kallisti5:
These fixes make XenAdmin compile up to a "Inner exception: Member 'ListViewItemCount' was not found." on Mono 5.10 under Linux

kc284:
I've pulled the patch and tried it out on my windows machine. It compiles ok.Merging.

PR #1053, opened 2016-07-04
Some refactoring and small issues fixed.

kc284:


Frezzle:
@kc284 Some of the controls' size/padding in BaseTabPage has changed (by more than I'm used to seeing in 'designer-made' changes). Could you please confirm that everything is still laid out correctly?

kc284:
These controls are docked so the initial size does not really matter. In any case I had one more look and they seem to display and resize as expected.

Frezzle:
Great, thanks; merging.

PR #1054, opened 2016-07-04
CP-17756: use new field Host.control_domain to find dom0

kc284:
Better have this reviewed before pushing further changes. Plus some preliminary changes for CP-18028.

### DIFF
--- a/XenAdmin/Controls/MultipleDvdIsoList.cs
+++ b/XenAdmin/Controls/MultipleDvdIsoList.cs
@@ -32,8 +32,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.Drawing;
-using System.Data;
 using System.Linq;
 using System.Text;
 using System.Windows.Forms;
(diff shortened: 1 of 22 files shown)


XenAdmin/Controls/MultipleDvdIsoList.cs
```diff
@@ -51,6 +51,8 @@ public partial class MultipleDvdIsoList : UserControl
         public MultipleDvdIsoList()
         {
             InitializeComponent();
+            labelSingleDvd.ForeColor = Program.HeaderGradientForeColor;
```
MihaelaStoica: By doing this, the ForeColor for these labels will be set to HeaderGradientForeColor (white) for all usages of the MultipleDvdIsoList control, including one on VM Storage page, where the background is also white.

kc284: Aah, right, there it's placed on white and not on the teal bit, I missed that - I'll revert the change, thanks for spotting it.

MihaelaStoica:
Looks good. Merging.

PR #1055, opened 2016-07-04
CP-17933: Add feature flag for proxy authentication

Frezzle:
Added Registry.ProxyAuthenticationEnabled boolean property to enable/disable proxy authentication UI controls in Connection Options page.
Cleaned up event handler logic in Connection Options page to return from multiple/unnecessary calls to event handlers when manually changing controls through code.

### DIFF
--- a/XenAdmin/Core/Registry.cs
+++ b/XenAdmin/Core/Registry.cs
@@ -80,6 +80,14 @@ namespace XenAdmin.Core
             }
         }
 
+        internal static bool ProxyAuthenticationEnabled
+        {
+            get
+            {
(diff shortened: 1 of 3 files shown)


GaborApatiNagy:
I can see what you tried to achieve here, but I think it would be nicer to keep this a bit less abstract and just have something like:

in the class
bool eventsdisabled = false;

and then in the beginning of the event handlers:
if (eventsdisabled)
   return;

while you set this flag to true and then to false wherever you need to. It would be less abstract but much more readable (especially find all references in an IDE would list them nicely)

Frezzle:
Thanks @GaborApatiNagy, I've gotten rid of the CallWithFreezeAwareness wrapper function; should be more readable now.

PR #1056, opened 2016-07-05
CA-215093: Batch updates: Update progress failed in a pool of mixed state hosts

GaborApatiNagy:
Where (in code removed by this commit) we iterated through the planActionsByHost dictionary, the hosts were iterated through in the order how the dictionary was initialized. Despite the actions were added in the right order, the result was screwed up, because the Cache.Hosts list can have the hosts randomly while we want the master to be first.

To fix the bug, I changed the data structure from a dictionary to a list that has the actions exactly in the good order, the order they were added, eliminating the risk of the actions being retrieved in the wrong order.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutoUpdatingPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutoUpdatingPage.cs
@@ -134,19 +134,15 @@ namespace XenAdmin.Wizards.PatchingWizard
                 return;
             }
 
-            Dictionary<Host, List<PlanAction>> planActionsPerPool = new Dictionary<Host, List<PlanAction>>();
-
             foreach (var pool in SelectedPools)
             {
                 var master = Helpers.GetMaster(pool.Connection);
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutoUpdatingPage.cs
```diff
@@ -387,12 +377,7 @@ private void UpdateDelayedAfterPatchGuidanceActionListForHost(List<PlanAction> d
             else
             {
                 //remove delayed action of the same kind for this host (because it is mandatory and will run immediately)
-                delayedGuidances.RemoveAll(dg => actions.Any(ma => ma.GetType() == dg.GetType())); 
-
-                //if it is a restart, clean delayed list
-                if (patch.after_apply_guidance == after_apply_guidance.restartHost)
```
MihaelaStoica: Why has this block (clean delayed list) been removed?

MihaelaStoica:
Looks good. Merging.

PR #1059, opened 2016-07-07
CA-203345: Exporting snapshot jams and cant be cancelled (when export…

MihaelaStoica:
…ing 2 snapshots to the same file)

Replaced the call to Task.destroy() with the DestroyTask() method, which additionally sets RelatedTask to null. This ensures that we stop polling the task for its progress and report the error.

### DIFF
--- a/XenModel/Actions/VM/ExportVmAction.cs
+++ b/XenModel/Actions/VM/ExportVmAction.cs
@@ -131,7 +131,7 @@ namespace XenAdmin.Actions
                     // which probably means there was an exception in the GUI code before the
                     // action got going. Kill the task so that we don't block forever on
                     // taskThread.Join(). Brought to light by CA-11100.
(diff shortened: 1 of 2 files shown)


kc284:
This looks ok to me.

There are a few more places in XenCenter where we call Task.destroy on the action's RelatedTask and don't always set it to null afterwards, e.g. ImportVmAction. Could we audit them to ensure they're not suffering from the same issue?

MihaelaStoica:
Thanks @kc284. I've applied the same fix to WlbReportAction. ImportVmAction does not require it; in fact we shouldn't set it to null cecause we use it in the wizard to find the imported VM in the cache (by matching RelatedTask.opaque_ref with VM.other_config["import_task"])

kc284:
Ok, thanks for checking it. Looks ok. Merging.

PR #1060, opened 2016-07-07
CA-214953: Multiple connection dialogs while attempting to connect

Frezzle:
ConnectingToServerDialog now keeps track of the dialogs for each connection and focuses on them if they already exist, instead of creating new ones.

### DIFF
--- a/XenAdmin/Dialogs/ConnectingToServerDialog.cs
+++ b/XenAdmin/Dialogs/ConnectingToServerDialog.cs
@@ -31,6 +31,7 @@
 
 using System;
 using System.Windows.Forms;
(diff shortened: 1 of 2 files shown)


MihaelaStoica:
I found two issues with this fix: 
- when the existing connection dialog is minimized, the second connection attempt does not restore it.
- when XenCenter cannot connect to the host, the error message is displayed twice, because there are have two ConnectionResult event handles registered.

Please have a look at the ShowPerConnectionWizard() function in MainWindow.cs which does something similar - maybe we can use it.

MihaelaStoica:
Looks good. Merging.

PR #1062, opened 2016-07-08
CA-107240: VDI move window shows incorrect message for broken storage…

MihaelaStoica:
… as ‘Local storage must belong to resident host’

Also fixed a similar issue, where the broken SR was selectable (Create new virtual disk from the SR Storage tab)

### DIFF
--- a/XenAdmin/Controls/SrPickerItem.cs
+++ b/XenAdmin/Controls/SrPickerItem.cs
@@ -131,8 +131,10 @@ namespace XenAdmin.Controls
                     return Messages.CURRENT_LOCATION;
                 if (LocalToLocalMove())
                     return Messages.LOCAL_TO_LOCAL_MOVE;
-                if (!SrIsLocalToTheHostOnForExsistingVDIs())
+                if (TheSR.IsLocalSR && !SrIsLocalToTheHostOnForExsistingVDIs())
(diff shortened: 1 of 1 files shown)


XenAdmin/Controls/SrPickerItem.cs
```diff
@@ -131,8 +131,10 @@ protected override string CannotBeShownReason
                     return Messages.CURRENT_LOCATION;
                 if (LocalToLocalMove())
                     return Messages.LOCAL_TO_LOCAL_MOVE;
-                if (!SrIsLocalToTheHostOnForExsistingVDIs())
+                if (TheSR.IsLocalSR && !SrIsLocalToTheHostOnForExsistingVDIs())
                     return Messages.SRPICKER_ERROR_LOCAL_SR_MUST_BE_RESIDENT_HOSTS;
+                if (!TheSR.CanBeSeenFrom(Affinity))
+                    return string.Format(Messages.SR_CANNOT_BE_SEEN, Affinity == null ? Helpers.GetName(TheSR.Connection) : Helpers.GetName(Affinity));
```
Frezzle: In Helpers.GetName(TheSR.Connection), can the connection here be null? If so, it will return empty string and an unclear message will be shown.

Frezzle:
Looks good. Merging.

PR #1064, opened 2016-07-11
CP-16921: Update XenCenter bindings

MihaelaStoica:

### DIFF
--- a/XenModel/XenAPI/FriendlyErrorNames.resx
+++ b/XenModel/XenAPI/FriendlyErrorNames.resx
@@ -861,6 +861,18 @@
   <data name="PROVISION_ONLY_ALLOWED_ON_TEMPLATE" xml:space="preserve">
     <value>The provision call can only be invoked on templates, not regular VMs.</value>
   </data>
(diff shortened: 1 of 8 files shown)


kc284:
The methods in the downloader are not needed because for servers after boston we do not call DownloadObjects. We need to check why the SDK engine has generated them. In the meantime you could perhaps strip manually this chunk from the submitted code.

MihaelaStoica:
Thanks @kc284. I've reverted the changes to the XenOnjectDownloader file.

kc284:
One more thing that I forgot earlier - sorry!
The SR_BACKEND_FAILURE_457 seems to be absent in the xapi branch used for the generation of the above bindings, if it was purposefully added on trunk for some past feature we should restore it.

MihaelaStoica:
The SR_BACKEND_FAILURE_457 has been removed from trunk xapi branch as well.

kc284:
Ok, as discussed, it seems SR_BACKEND_FAILURE_457 is only left over in XC and can be removed.
Looks good, Merging.

PR #1068, opened 2016-07-13
CP-16921: Add new PVS tab

MihaelaStoica:
- tab visible on pool or standalone server if the feature is not restricted and the pools has at least one farm


XenModel/Messages.resx
```diff
@@ -9471,6 +9471,15 @@ Press OK to continue the wizard and return to the server and follow the instruct
...
     <value>Proxy servers are not supported.</value>
   </data>
+  <data name="PVS_CACHE_MEMORY_AND_DISK" xml:space="preserve">
+    <value>Memory and disk</value>
+  </data>
+  <data name="PVS_CACHE_MEMORY_ONLY" xml:space="preserve">
+    <value>Memory only</value>
+  </data>
+  <data name="PVS_TAB_TITLE" xml:space="preserve">
+    <value>PVS</value>
```
Frezzle: The HA and WLB tabs have more descriptive titles ("High Availability" and "Workload Balancing", respectively); should the title for the PVS tab be "Provisioning Services" instead of "PVS"?

XenAdmin/TabPages/PvsPage.cs
```diff
@@ -0,0 +1,208 @@
...
+
+            return newRow;
+        }
+
+        private void UnregisterVMHandlers()
+        {
+            foreach (DataGridViewRow row in dataGridViewVms.Rows)
+            {
+                var vm = row.Tag as VM;
+                if (vm == null)
```
Frezzle: Is this null check needed (or the one on line 203)? Can the VM ever be null here? On line 122 we only add a row when the VM for a PVS Proxy is not null, so is it possible for the VM here to become null after being added as a row?

XenAdmin/TabPages/PvsPage.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            Program.Invoke(this, LoadFarms); 
+        }
+
+        private void PvsProxyBatchCollectionChanged(object sender, EventArgs e)
+        {
+            Program.Invoke(this, LoadVMs);
+        }
+
+        private void VmPropertyChanged(object sender, PropertyChangedEventArgs e)
```
Frezzle: In this function, why do we only check for changes in the name_label? Are the VM's other PVS-related properties (such as if it's currently cached, for example) not also able to change?

XenAdmin/TabPages/PvsPage.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            if (e.PropertyName != "name_label")
+                return;
+
+            foreach (DataGridViewRow row in dataGridViewVms.Rows)
+            {
+                var vm = row.Tag as VM;
+                if (vm != null && vm.Equals(sender))
+                    row.Cells["columnVM"].Value = vm.Name;
```
Frezzle: Is there ever only one or zero VMs here which are equal to sender? If so, we could break out of the loop early here once we've found the match.

XenModel/XenModel.csproj
```diff
@@ -353,6 +354,9 @@
     <Compile Include="XenAPI\pgpu_dom0_access.cs" />
     <Compile Include="XenAPI\pool_allowed_operations.cs" />
     <Compile Include="XenAPI\primary_address_type.cs" />
+    <Compile Include="XenAPI\PVS_farm.cs" />
+    <Compile Include="XenAPI\PVS_proxy.cs" />
+    <Compile Include="XenAPI\PVS_server.cs" />
```
Frezzle: I'm curious; these files were added in a previous commit/ticket, right? (The bindings update one) So why have these lines just appeared here?

Frezzle:
There are some UI-related things I think could be improved:
1. When a VM's name is too long to fit in the column, and has some spaces, instead of getting truncated it goes onto a new line and the row increases in height. Ideally it should be truncated with an ellipsis?
2. Resizing the columns doesn't happen in real-time. Also, making them smaller than their title doesn't truncate them with an ellipsis, it just removes the last word.
3. There is no margin between the two tables and the right edge of the window.

Note: The Physical Storage page doesn't have any of these issues; it uses a ListViewEx for its table, whereas the PVS page is currently using DataGridViewEx.

XenAdmin/TabPages/PvsPage.cs
MihaelaStoica: The other PVS-related properties are in the PVS_proxy objects, and changes to these objects are handled in PvsProxyBatchCollectionChanged. name_label is currently the only VM property displayed in this grid.

XenAdmin/TabPages/PvsPage.cs
MihaelaStoica: @Frezzle, you are right, the row.Tag property shoulld always be a VM object. I could probably add an assertion instead, to make sure we always set the Tag property this way.

XenAdmin/TabPages/PvsPage.cs
MihaelaStoica: Good point!

XenModel/XenModel.csproj
MihaelaStoica: Because in the previous commit I only pulled the binsings from the SDKs and therefore only added the new files to the repository, without including them in the project.

XenModel/Messages.resx
MihaelaStoica: Good point, I will change it to "PVS read cache"

MihaelaStoica:
@Frezzle: I've updated the pull request with a new commit. Please review.

Frezzle:
All good; merging.

PR #1069, opened 2016-07-18
CA-136768: Fixed bad layout of buttons at the top of the snapshots ta…

kc284:
…b page. Also fixed text alignment on the chevron button.


MihaelaStoica:
Looks good. Merging.

PR #1072, opened 2016-07-19
CA-209198: Allowed to attempt to install hotfix when it is already applied

Frezzle:
XenCenter no longer ignores checking if the master host already has the specified patch applied; it now checks before trying to apply it, which means we get no error in the Patching Wizard's Patching Page.


GaborApatiNagy:
Looks good

PR #1073, opened 2016-07-19
CA-192541: Added detail to the Import/Export OVF/OVA and Import Disk …

kc284:
…Image actions

title so we can tell the one from the other. For consistency I used a format similar
to Import/export XVA. I also removed the gerund from the latter as it cannot be
used when the action is completed.


MihaelaStoica:
Looks good. Merging

PR #1074, opened 2016-07-20
CA-216697: MTU message present on the Network Settings page when crea…

MihaelaStoica:
…ting an internal network

In the New Network wizard, when the "Single-Server Private Network" type is selected, the Network settings page should not display the MTU hint message, because it is not relevant to any controls on the page


kc284:
Looks ok to me.

PR #1075, opened 2016-07-21
CA-213493: [VMSS] Quiesce info icon and alert icon in VMSS wizard are not aligned properly

geosharath:
[VMSS] Quiesce info icon and alert icon in VMSS wizard are not aligned properly


MihaelaStoica:
Looks good. Merging

PR #1077, opened 2016-07-22
CA-213498: [VMSS] XC create new VMSS Policy wizard disables "quiesce" option when no VMs are selected in policy

geosharath:
XC create new VMSS Policy wizard disables "quiesce" option when no VMs are selected in policy

### DIFF
--- a/XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
+++ b/XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
@@ -95,7 +95,7 @@ namespace XenAdmin.Wizards.NewPolicyWizard
         {
             base.PageLoaded(direction);
             if (direction == PageLoadedDirection.Forward)
-                EnableShapshotTypes(Connection);
+                EnableShapshotTypes(Connection, false);
         }
 
         public override Image Image
@@ -166,6 +166,12 @@ namespace XenAdmin.Wizards.NewPolicyWizard
 
         private void RefreshTab(IVMPolicy policy)
         {
+            /* when a policy does not have any VMs, irrespective of
+             * the snapshot type, enable Quiesce 
+             */
+
+            quiesceCheckBox.Enabled = (policy.VMs.Count == 0);
+ 
             switch (policy.policy_type)
             {
                 case policy_backup_type.checkpoint:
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
```diff
@@ -166,6 +166,15 @@ public policy_backup_type BackupType
 
         private void RefreshTab(IVMPolicy policy)
         {
+            /* when a policy does not have any VMs, irrespective of
+             * the snapshot type, enable Quiesce 
+             */
+
+            if (policy.VMs.Count == 0)
```
kc284: You can write this in a more compact way:
quiesceCheckBox.Enabled = (policy.VMs.Count == 0);

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
```diff
@@ -166,6 +166,15 @@ public policy_backup_type BackupType
...
+                quiesceCheckBox.Enabled = true;
+            else
+                quiesceCheckBox.Enabled = false;
+
             switch (policy.policy_type)
```
kc284: I'm a bit confused with this method. The comment in line 169 is inconsistent with what's happening in this switch block.
If quiesce for 0 VMs shoud be enabled irrespective of the snapshot type, then you may want to move the check after the switch block, because quiesceCheckBox.Enabled will be overwritten in there.
On the other hand, if quiesce is only relevant for policy_backup_type.snapshot, you may want to move the check into this particular case and remove the comment.
What exactly is the intention here?

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
```diff
@@ -174,42 +183,46 @@ private void RefreshTab(IVMPolicy policy)
...
+                                this.quiesceCheckBox.Checked = false;
+                                break;
+                            }
                         }
+
                     }
+
                 }
-                else
+                else /* we entry this block only when we are editing a policy, in that case the decision has already been taken in RefreshTab function */
```
kc284: We "enter". "Entry" is a noun.

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
geosharath: Thanks for the suggestion, ill update the same.

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePageSpecific.cs
geosharath: The case inside switch handles when the policy's snapshot type is quiesce, then irrespective of whether the policy has VMs or not we should enable the quiesce option.

kc284:
Looks ok. Merging.

PR #1079, opened 2016-07-25
CA-213499: List view of snapshots not working as expected with VMSS

geosharath:
vmss snapshots are shown even when Scheduled snapshots is
unchecked in same menu.

### DIFF
--- a/XenAdmin/TabPages/SnapshotsPage.cs
+++ b/XenAdmin/TabPages/SnapshotsPage.cs
@@ -426,7 +426,7 @@ namespace XenAdmin.TabPages
             for (int i = 0; i < snapshots.Count; i++)
             {
                 VM snapshot = snapshots[i];
-                if (!(snapshot.is_snapshot_from_vmpp && !toolStripMenuItemScheduledSnapshots.Checked))
+                if (!(snapshot.is_snapshot_from_vmpp || snapshot.is_vmss_snapshot) || toolStripMenuItemScheduledSnapshots.Checked)
                 {
                     snapshot.PropertyChanged -= snapshot_PropertyChanged;
                     snapshot.PropertyChanged += snapshot_PropertyChanged;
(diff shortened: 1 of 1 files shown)


XenAdmin/TabPages/SnapshotsPage.cs
```diff
@@ -426,16 +426,7 @@ private IList<VM> RefreshDataGridView(IList<VM> snapshots)
...
-                {
-                    snapshot.PropertyChanged -= snapshot_PropertyChanged;
-                    snapshot.PropertyChanged += snapshot_PropertyChanged;
-                    //Build DataGridView
-                    SnapshotDataGridViewRow row = new SnapshotDataGridViewRow(snapshot);
-                    row.Tag = snapshot;
-                    DataGridView.Rows.Add(row);
-                }
-                else if (!(snapshot.is_vmss_snapshot && !toolStripMenuItemScheduledSnapshots.Checked))
+                if (!((snapshot.is_snapshot_from_vmpp || snapshot.is_vmss_snapshot) && !toolStripMenuItemScheduledSnapshots.Checked))
```
MihaelaStoica: I think this condition can be simplifird a bit, for better readability: 
`if (!(snapshot.is_snapshot_from_vmpp || snapshot.is_vmss_snapshot) || toolStripMenuItemScheduledSnapshots.Checked)` 
which basically reads "if not scheduled snapshot OR show_everything"

geosharath: there are 4 cases 

scheduled_snapshots - checkbox_value - display (outcome) 
0                                  0                             1
0                                  1                              0
1                                  0                             1
1                                   1                           1

if you take the second case where the snapshot is not a scheduled one, and the checkbox is true (which means display only scheduled snapshots) the suggested expression will evaluate to true and display the snapshot which is not desired. can you please share your thoughts on this ?

MihaelaStoica: I think that the toolStripMenuItemScheduledSnapshots.Checked means "include scheduled snapshots"

geosharath: oh alright, i had misunderstood the interpretation of checkbox, will update the PR. Thanks for your suggestion.

geosharath:
@MihaelaStoica  I have updated the PR.

MihaelaStoica:
Looks good. Merging

PR #1084, opened 2016-07-25
CA-213728: Going to Home Server tab can change home server

GaborApatiNagy:
Even though the correct row have been selected by the time the grid becomes visible, it is not displayed properly. This change set fixes the problem by selecting the row again.


MihaelaStoica:
Looks good. Merging.

PR #1085, opened 2016-07-25
CP-16936: Implement GUI for viewing the PVS farms and PVS servers (read-only)

Frezzle:
Added new dialog for viewing PVS farms and servers.


XenAdmin/Dialogs/PvsFarmDialog.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// Creates a dialog for viewing the PVS farms and PVS servers on a particular connection.
+        /// </summary>
+        /// <param name="_connection">May not be null.</param>
+        public PvsFarmDialog(IXenConnection _connection)
+        {
+            System.Diagnostics.Trace.Assert(_connection != null);
+            connection = _connection;
+
+            InitializeComponent();
+            Text += string.Format(" - '{0}'", this.connection.Name);
```
MihaelaStoica: We should put the string in the resources (Messages.resx). Also, shouldn't this be something like "PVS Farms - {0}"?

XenAdmin/Dialogs/PvsFarmDialog.cs
```diff
@@ -0,0 +1,127 @@
...
+        private void Rebuild()
+        {
+            Program.AssertOnEventThread();
+
+            try
+            {
+                gridView.SuspendLayout();
+                gridView.Rows.Clear();
+
+                PVS_farm[] pvsFarmsArray = connection.Cache.PVS_farms;
```
MihaelaStoica: We can use Sort() function here, which will sort the XenObjets by Name property (which we will need to define). E.g.
`var pvsFarms = new List<PVS_farm>(Connection.Cache.PVS_farms);`
`pvsFarms .Sort();`

XenAdmin/Dialogs/PvsFarmDialog.resx
```diff
@@ -0,0 +1,521 @@
...
+  <data name="$this.AutoScaleDimensions" type="System.Drawing.SizeF, System.Drawing">
+    <value>96, 96</value>
+  </data>
+  <data name="$this.ClientSize" type="System.Drawing.Size, System.Drawing">
+    <value>618, 361</value>
+  </data>
+  <data name="$this.Font" type="System.Drawing.Font, System.Drawing">
+    <value>Tahoma, 8pt</value>
+  </data>
+  <data name="$this.Icon" type="System.Drawing.Icon, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
MihaelaStoica: The dialog shouldn't have the icon embedded in the resource, it should inherit application's icon

XenAdmin/Controls/DataGridViewEx/CollapsingPvsFarmServerDataGridViewRow.cs
```diff
@@ -0,0 +1,165 @@
...
+
+                SetNoIcon();
+
+                ipAddressesCell.Value = string.Join(", ", server.addresses);
+                firstPortCell.Value = server.first_port;
+                lastPortCell.Value = server.last_port;
+            }
+        }
+
+        public int ExpansionCellIndex
```
MihaelaStoica: Are these public properties used anywhere?

Frezzle: UnderlyingFarm and IsFarmRow are used but the others aren't. I will remove the index-related ones, but should I also remove the UnderlyingServer and IsServerRow, or should I keep them for completeness/future-use (as their "counterparts" are currently in use)?

XenAdmin/Dialogs/PvsFarmDialog.cs
Frezzle: The Text property is initialized to "PVS Farms" by the designer and then here we concatenate " - '{0}'". I'll add a string resource and change this to an assignment similar to:
Text = string.Format(Messages.PVS_FARMS_DIALOG_TITLE, this.connection.Name);

stephen-turner: I'm not commenting on this specific case, but concatenation of strings generally doesn't work well when localising: other languages may have the parts in a different order, or use different separators, or different parts of speech in different contexts, or... So it's generally better to specify the whole string.

XenAdmin/Controls/DataGridViewEx/CollapsingPvsFarmServerDataGridViewRow.cs
MihaelaStoica: Let's keep UnderlyingServer and IsServerRow, for completeness

Frezzle:
@MihaelaStoica All changes made :+1:

MihaelaStoica:
Looks good. Merging

PR #1086, opened 2016-07-28
Fixed issue on the New SR wizard where the SR type details panel was …

kc284:
…not populated

when the default checked button was invisible.


Frezzle:
Looks good; merging.

PR #1087, opened 2016-07-28
CA-214622: Confusing info shown on Install Update screen

Frezzle:
Changed message to "Deleting update installation file ..." to match wording in previous page of the wizard.


GaborApatiNagy:
Looks good

PR #1088, opened 2016-08-01
CA-217449 &  CA-173661 - Control domain memory alert fixes

MihaelaStoica:


kc284:
Looks ok to me. I'm merging it.

PR #1089, opened 2016-08-02
CA-216935: Fixed a bug that can occur when multiple pools are being u…

GaborApatiNagy:
…pdated with the same update and one of the prechecks happen before the other pool is finished applying the update

The PatchPrechecksOnMultipleHostsInAPoolPlanAction did not filter the patchmappings by masterhost, so with some probability, it could choose the wrong mapping (pool_patch object for the same update, but in different pool). This bug is being fixed in this commit, other changes only renamed (for more consistent naming) a class and removed an unused field.


MihaelaStoica:
Looks good. Merging.

PR #1090, opened 2016-08-02
CA-217232: Batch updates: Install Dundee released hotfixes in Batch f…

GaborApatiNagy:
…ails with wrong message in hotfix wizard

Despite the update sequences are correct by host, when they are flattened, the original order (order in minimal patches) was not considered properly, so a missing patch was scheduled for later in the sequence than it should have been, thus the upgrade failed with a precheck-error.

This bug occurs when we are given
* a pool of more than one host,
* with (at least) patchA not installed on 2 or more hosts
* patchA requires patchR that is not installed on at least one of the above hosts (but not not on all)
* the host that is missing patchR is ahead of other hosts in pool.Hosts.Cache

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -477,11 +477,11 @@ namespace XenAdmin.Core
                 // Take all the hotfixes for this version
                 allPatches.AddRange(serverVersions[0].Patches);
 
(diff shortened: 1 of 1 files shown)


MihaelaStoica:
Looks good. Merging

PR #1092, opened 2016-08-03
CA-104472: Refactoring and removal of duplicate code. This also fixes…

kc284:
… descrepancies

in the calculation of the never-expiring licence ("greater than 3653 days" in
LicenceAlert and LicenceStatus vs "greater than or equal to 3653 days" elsewhere).


MihaelaStoica:
Looks good. Merging.

PR #1094, opened 2016-08-05
CA-210549: Make MultipleAction and ParallelAction cancellable

MihaelaStoica:
- The multiple action is cancellable if is at least one sub action is not completed
- When a multiple action is cancelled, we try to cancel all sub actions that are not completed
- Also changed the way we calculate the percent complete for the parent action, to report partial progress of the sub actions

### DIFF
--- a/XenModel/Actions/MultipleAction.cs
+++ b/XenModel/Actions/MultipleAction.cs
@@ -148,7 +148,7 @@ namespace XenAdmin.Actions
             }
         }
 
-        private void RecalculatePercentComplete()
+        protected virtual void RecalculatePercentComplete()
         {
             int total = 0;
             int n = subActions.Count;
--- a/XenModel/Actions/MultipleAction.cs
+++ b/XenModel/Actions/MultipleAction.cs
@@ -34,6 +34,7 @@ using System.Collections.Generic;
 using XenAdmin.Core;
 using XenAdmin.Network;
 using XenAPI;
+using System.Linq;
 
 
 namespace XenAdmin.Actions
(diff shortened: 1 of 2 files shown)


XenModel/Actions/MultipleAction.cs
```diff
@@ -127,15 +143,26 @@ private void SubActionChanged(ActionBase sender)
...
                 SubActionDescription = subAction.Description;
+                RecalculatePercentComplete();
                 OnChanged();
             }
         }
 
+        private void RecalculatePercentComplete()
+        {
+            int total = 0;
+            int n = subActions.Count;
```
kc284: I believe we should be doing the calculation with (actionsByConnection.Count+actionsWithNoConnection.Count) instead of subActions.Count, because actionsByConnection does not include actions with disconnected connections, so it may happen that we'll be calculating the percentage waiting for the completion of actions that never actually started.

MihaelaStoica:
Thanks @kc284. I've updated the pull request.

XenModel/Actions/MultipleAction.cs
```diff
@@ -127,15 +143,26 @@ private void SubActionChanged(ActionBase sender)
             {
                 SubActionTitle = subAction.Title;
                 SubActionDescription = subAction.Description;
+                RecalculatePercentComplete();
                 OnChanged();
             }
         }
 
+        protected virtual void RecalculatePercentComplete()
```
kc284: Now that I'm looking at it again, and apologies for not thinking of it the first time, wouldn't it be more efficient to increment the multiple action percentage by  an n-th of the percentage of the sender in the SubActionChanged handler instead of iterating though the relevant subactions or actionsWith/WithoutConnection? In which case, actionsCount could become protected in the parent class and the Recalculate method could go completely. Any hidden caveats in doing so?

MihaelaStoica: If we do that, then we keep incrementing with the percentage of the sender every time the sender changes, which is not correct, as an action's percentage should only be added once.

kc284: Yes, indeed, that would work only if we had the percentage delta.
In any case, the PR looks ok now, I'm merging it.

PR #1099, opened 2016-08-08
CA-217844: Handling Automatic update not supported

GaborApatiNagy:
If there is no <minimalpatches /> tag, the version is not supported. This check has priority over licensed state when displaying the reason why a server is not supported.

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -442,6 +442,9 @@ namespace XenAdmin.Core
                 var minimumPatches = new List<XenServerPatch>();
                 minimumPatches = serverVersions[0].MinimalPatches;
 
+                if (minimumPatches == null) //unknown
+                    return recommendedPatches;
+
                 var appliedPatches = host.AppliedPatches();
                 recommendedPatches = minimumPatches.FindAll(p => !appliedPatches.Any(ap => string.Equals(ap.uuid, p.Uuid, StringComparison.OrdinalIgnoreCase)));
 
@@ -477,12 +480,19 @@ namespace XenAdmin.Core
                 // Take all the hotfixes for this version
                 allPatches.AddRange(serverVersions[0].Patches);
 
+                if (serverVersions[0].MinimalPatches == null)
+                    return null;
+
                 uSeq.MinimalPatches = serverVersions[0].MinimalPatches;
 
                 foreach (Host h in hosts)
                 {
                     uSeq[h] = GetUpgradeSequenceForHost(h, allPatches, uSeq.MinimalPatches);
                 }
+            }
+            else
+            {
+                return null;
             }
 
             return uSeq;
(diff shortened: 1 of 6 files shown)


XenModel/Messages.resx
```diff
@@ -9006,6 +9006,9 @@ However, there is not enough space to perform the repartitioning, so the current
   <data name="PATCHINGWIZARD_SELECTPATCHPAGE_UPDATESEXT" xml:space="preserve">
     <value>[XenServer] Updates and Supplemental Packs (*.{0}, *.iso)|*.{0};*.iso</value>
   </data>
+  <data name="PATCHINGWIZARD_SELECTSERVERPAGE_AUTO_UPDATE_NOT_SUPPORTED_HOST_VERSION" xml:space="preserve">
+    <value>Automatic update is not supported on this [XenServer] version.</value>
```
kc284: You may want to remove the full stop fromt his tooltip too :)

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -254,12 +254,14 @@ private void GetXenServerVersions(XmlDocument xdoc)
                     }
 
                     List<XenServerPatch> patches = new List<XenServerPatch>();
```
kc284: Should we be doing the same with patches? What would happen if the patch node were absent?

XenAdmin/Core/Updates.cs
```diff
@@ -442,6 +442,9 @@ public static List<XenServerPatch> RecommendedPatchesForHost(Host host)
                 var minimumPatches = new List<XenServerPatch>();
```
kc284: These two lines can be compacted:
`var minimumPatches = serverVersions[0].MinimalPatches;`

GaborApatiNagy: Good catch! Let me not fix this here, I have been already working on my next PR that will fix this.

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
GaborApatiNagy: Comment I accidentally deleted by @kc284: "Should we be doing the same with patches? What would happen if the patch node were absent?"

GaborApatiNagy: Well, I could, but it wouldn't make any difference to existing functionalities. This is a must for minimalPatches, but not needed for the others.

I think -as part of my bugfix- I am better off not touching what I don't have to and live exactly it as it's been before.

As a future improvement, I'd rather suggest we completely get rid off this code and parse the XML with a DOM parser. (Actually, it has already got some issues like: handling duplicate nodes and not validating the document against a schema)

XenModel/Messages.resx
GaborApatiNagy: Will do, thanks!

GaborApatiNagy:
Thanks @kc284 for reviewing, please have a look at my most recent commit.

kc284:
Ok, merging.

PR #1101, opened 2016-08-09
CP-18407: Rename PVS farm to PVS site + CP-16935: Configure Control domain memory for a host 

MihaelaStoica:


XenAdmin/Dialogs/ControlDomainMemoryDialog.cs
```diff
@@ -0,0 +1,177 @@
...
+            Program.Invoke(this, Populate);
+        }
+
+        private void OkButton_Click(object sender, EventArgs e)
+        {
+            if (SaveChanges())
+            {
+                DialogResult = DialogResult.OK;
+            }
+            Close();
```
kc284: Here and in line 174 below: you shouldn't need to call Close, setting DialogResult (and both buttons already have a value set in the designer) should be ok to close it and return from ShowDialog.

XenAdmin/Dialogs/ControlDomainMemoryDialog.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                double min = vm.memory_static_min;
+                double max = Math.Min(host.memory_available_calc, MAXIMUM_DOM0_MEMORY_GB * Util.BINARY_GIGA);
+                double value = vm.memory_dynamic_min;
+                // Avoid setting the range to exclude the current value: CA-40041
+                if (value > max)
+                    max = value;
+                if (value < min)
+                    min = value;
+                memorySpinner.Initialize(Messages.CONTROL_DOMAIN_MEMORY_LABEL, null, value, max);
```
kc284: As discussed, the memory spinner needs to be refactored to ensure that whenever we initialise it or change the range we don't find ourselves with a value out of range due to rounding issues. Agreed that this will be fixed in a different PR.

MihaelaStoica:
Thanks @kc284 , I've updated the pull request

kc284:
Looks ok. Merging.

PR #1102, opened 2016-08-09
CA-217847: Fixed issue where servers were marked as unable to be upda…

kc284:
…ted automatically

even when automatic mode was supported for them. The problem was that the xenserver
versions were not retrieved from the DownloadUpdatesXmlAction once the latter had run.
Also, minor refactoring.


GaborApatiNagy:
Looks good, merging

PR #1103, opened 2016-08-10
CA-213500: Revert To, Actions and Delete buttons are enabled even if there is no manual snapshot.

geosharath:
CA-213500: Revert To, Actions and Delete buttons are enabled even
 if there is no manual snapshot.


MihaelaStoica:
Looks good. Merging.

PR #1105, opened 2016-08-10
CA-217862: Batch Updates: handle partially upgraded pools

GaborApatiNagy:
Partially upgraded pools and mixed version pools are not supported, XenCenter will display a proper error message.

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -439,8 +439,7 @@ namespace XenAdmin.Core
 
             if (serverVersions.Count != 0)
             {
-                var minimumPatches = new List<XenServerPatch>();
-                minimumPatches = serverVersions[0].MinimalPatches;
+                var minimumPatches = serverVersions[0].MinimalPatches;
 
                 if (minimumPatches == null) //unknown
                     return recommendedPatches;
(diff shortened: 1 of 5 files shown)


XenAdmin/Core/Updates.cs
```diff
@@ -503,7 +491,56 @@ public static UpgradeSequence GetUpgradeSequence(IXenConnection conn, List<XenSe
             }
         }
 
-        private static List<XenServerPatch> GetUpgradeSequenceForHost(Host h, List<XenServerPatch> allPatches, List<XenServerPatch> latestPatches)
+        /// <summary>
+        /// Returns a XenServerVersion if all hosts of the pool has the same version
```
kc284: "have"

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -193,6 +193,15 @@ private void EnabledRow(Host host, UpdateType type, int index)
                     return;
                 }
 
+                var pool = Helpers.GetPool(host.Connection);
```
kc284: The method Updates.GetUpgradeSequence returns null when the host versions are not the same, in which case we'll go into the first if-block and we won't reach down here to show the right reason. This block here needs to move into the first if-block.

GaborApatiNagy:
Thanks @kc284, I have made the necessary code changes.

kc284:
Looks better now. Merging.

PR #1109, opened 2016-08-12
CA-217846: Fixed two more tooltip texts with fullstops.

kc284:


GaborApatiNagy:
Thanks @kc284

PR #1110, opened 2016-08-12
Cp 18011: [XenCenter] Add live patching checkbox

mcintyre94:
Add a live patching tab with a checkbox to toggle whether live patching is allowed for a licensed pool or standalone host.

### DIFF
--- a/XenModel/XenAPI-Extensions/Host.cs
+++ b/XenModel/XenAPI-Extensions/Host.cs
@@ -521,6 +521,16 @@ namespace XenAPI
             return h._RestrictSslLegacySwitch;
         }
 
+        private bool _RestrictLivePatching
+        {
+            get { return BoolKey(license_params, "restrict_live_patching"); }
+        }
+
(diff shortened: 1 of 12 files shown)


XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -521,6 +521,16 @@ public static bool RestrictSslLegacySwitch(Host h)
             return h._RestrictSslLegacySwitch;
         }
 
+        private bool _RestrictLivePatching
+        {
+            get { return BoolKey(license_params, "restrict_live_patching"); }
```
MihaelaStoica: We need to call BoolKeyPreferTrue() here, so that the deature is consideted restricted if the license key is not present.

mcintyre94:
Those 2 commits should cover all the changes we discussed.

MihaelaStoica:
Looks good now. Merging

GaborApatiNagy:
This is in branch CAR-2197.

PR #1117, opened 2016-08-15
CA-218304: Text cut off in Properties dialog, Alerts tab

mcintyre94:
This is resolved by moving the label into a TableLayoutPanel, docked to the top and with auto-grow on.


MihaelaStoica:
With this change you've added another TableLayoutPanel, but you could've used the existing one, and add the rubric label (and the controls related to the repeat interval) to it.

MihaelaStoica:
Could you please also adjust the tab order for the Alert repeat interval control?

MihaelaStoica:
Looks good now. Merging.

PR #1118, opened 2016-08-15
CA-147142: ReadOnly user should not be allowed to delete crash dump f…

MihaelaStoica:
…iles

- Made the action PureAsyncAction, so that it automatically checks if the current user can perform all the necessary api calls


GaborApatiNagy:
Looks good

PR #1119, opened 2016-08-15
CA-217821: Specified home server in cross pool migration should behave as home server

mcintyre94:
Mostly just terminology changes from home server to target server in the CPM wizard. Changes in shared component SelectMultipleVMDestinationPage (shared with ImportWizard) to allow CPM wizard to rename the Home Server column (as it can the VM/template one - using the same pattern).


MihaelaStoica:
I don't think we should rename the CrossPoolMigrateToHomeCommand (including the MenuText), because this command is intended to be used for migrating a VM to it's home server.

MihaelaStoica:
Looks ok now. Merging

PR #1120, opened 2016-08-15
CA-180127: Unable to unset VM home server if all shared SRs are full

MihaelaStoica:
By passing null as the srHost parameter to the affinity picker, the "no home server" radio button remains enabled, even when no shared SRs are available, allowing users to unset the home server of a VM


GaborApatiNagy:
Looks fine

PR #1121, opened 2016-08-15
CA-218241: Cancel connect right click menu item does not seem to be doing anything

mcintyre94:
The CancelHostConnectionCommand executes by using a DisconnectCommand, so the fix here is to allow that command to execute on connections in progress as well as those that are connected.


MihaelaStoica:
Looks good. Merging.

PR #1122, opened 2016-08-15
CA-150198: EN: Missing hotkey for “Properties” after right clicking virtual disk

GaborApatiNagy:


MihaelaStoica:
Looks good. Merging.

PR #1123, opened 2016-08-15
CA-150816: fix: G11n:I18n: Hardcode issue "Set VM other_config" on Authentication dialog.

GaborApatiNagy:


stephen-turner:
I don't think the string "Set VM other_config" should be visible to the user, even localised. Can we find a solution which has a message that is meaningful to the user?

MihaelaStoica:
I was going to suggest something similar to Stephen's comment: extend SetVMOtherConfigAction constructor with an action title parameter which can then be specific to each usage of this action.

kc284:
I'm skeptical about extending the action to be usage specific, because the other_config may contain all sorts of key/value pairs and I don't think we would want to end up with actions for each one of them in the long term.

Do we even need the SetVMOtherConfigAction itself? There are similar cases elsewhere in the code where we use a DelegatedAsyncAction to perform simple calls. As for the message, can we get away with adding a generic "Configuring VM..." or "Saving configuration..."?

GaborApatiNagy:
I've updated this PR as we discussed following @kc284 's suggestion. The message I used is the same as what is shown when modifications on VM properties dialog are getting saved. "Saving VM Configuration"

PR #1124, opened 2016-08-15
CA-174979: Confusing right-click add to pool menu (if you accept the defaults)

GaborApatiNagy:
Changed "New Pool" to "Unnamed Pool"


stephen-turner:
I want to see this one please. I don't think I agree with the suggestion in the ticket, but I want to see it to be sure.

kc284:
Personally I don't like "unnamed pool" very much. How about offering a name like "pool (1)", like we do for the new VMs? (I've put this comment on the ticket too).

GaborApatiNagy:
I changed it to say "Pool"

stephen-turner:
I still want to see a demo of this.

stephen-turner:
I don't really like calling the pool "Pool" either. I propose leaving the name blank so the the user is forced to fill in a name before continuing.

stephen-turner:
I also think we should implement Tina's separate suggestion in the same ticket, if it's easy.

XenModel/Messages.resx
```diff
@@ -8461,7 +8461,7 @@ Once the VM has restarted click the Install [XenServer product] Tools menu item
     <value>&amp;New Policy...</value>
   </data>
   <data name="NEW_POOL" xml:space="preserve">
```
kc284: I believe this string is not used anywhere any more - can we remove it to keep the code lean?

GaborApatiNagy: Makes sense, removed it now

kc284:
Looks ok now; merging.

PR #1125, opened 2016-08-17
CP-18013: [XenCenter] Update host/pool updates pane

mcintyre94:
For Ely or greater, use the host.patches_requiring_reboot field to generate the messages for hosts needing rebooting in the host and pool updates panes. For versions prior to Ely there is no change.


GaborApatiNagy:
Looks good, merging

PR #1126, opened 2016-08-17
CA-213496: Snapshots taken part of VMSS have same symbol as manually taken snapshots

geosharath:
Snapshots taken part of VMSS have same symbol as manually taken snapshots in saved snapshot searchas manually taken snapshots in saved snapshot search


kc284:
Looks ok; I'm merging it.

PR #1127, opened 2016-08-17
Improvements made while investigating CA-218953:

kc284:
- when the cd drive or the tools iso cannot be found, the action should fail.
- if the tools SR is broken, do not show the error on a pop-up because the UX is
bad when installation is attempted on multiple VMS.
- added one more entry to the list of possible tools iso names.
- small performance improvement.

### DIFF
--- a/XenModel/Actions/VM/InstallPVToolsAction.cs
+++ b/XenModel/Actions/VM/InstallPVToolsAction.cs
@@ -30,6 +30,7 @@
  */
 
 using System;
+using System.Linq;
 using XenAPI;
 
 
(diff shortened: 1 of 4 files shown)


XenModel/Actions/VM/InstallPVToolsAction.cs
```diff
@@ -134,11 +131,8 @@ private XenAPI.VDI findWinISO(bool searchHiddenISOs)
             {
                 if (XenAPI.SR.SRTypes.iso.ToString() == sr.content_type)
                 {
-                    foreach (VDI vdi in Connection.ResolveAllShownXenModelObjects(sr.VDIs, searchHiddenISOs))
-                    {
-                        if (vdi.IsToolsIso)
-                            return vdi;
-                    }
+                    var vdis = Connection.ResolveAllShownXenModelObjects(sr.VDIs, searchHiddenISOs);
+                    return vdis.FirstOrDefault(v => v.IsToolsIso);
```
GaborApatiNagy: I think this will return null for the first iso-SR (if no VDI is IsToolsIso) and will not check further SRs

kc284: You're right, I missed this  :flushed:

GaborApatiNagy:
Thanks @kc284 , merging

PR #1128, opened 2016-08-18
CA-217232: Fixed regression caused by the fix for CA-216935

GaborApatiNagy:


kc284:
I believe the equality method of PoolPatchMapping needs correction too.
Once you fix this, and to avoid repeating code, you could use that method to find the mapping each time.

XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
```diff
@@ -53,15 +54,23 @@ public ApplyXenServerPatchPlanAction(Host host, XenServerPatch xenServerPatch, L
         protected override void RunWithSession(ref Session session)
         {
 
-            var mapping = mappings.Find(m => m.XenServerPatch.Uuid == xenServerPatch.Uuid && m.MasterHost == Helpers.GetMaster(host.Connection));
+            var master = Helpers.GetMaster(host.Connection);
+            var mapping = mappings.Find(m => m.XenServerPatch.Uuid == xenServerPatch.Uuid && m.MasterHost.uuid == master.uuid);
```
kc284: The XenServerPatch equality checks further down are inconcistent with this one. Shouldn't you check the patch uuid further down too?
Also, why not use the XenServerPatch.Equals method that does this. You could actaully expand that class with proper operator overloads and Equals overrides so we avoid this error in future.

XenAdmin/Wizards/PatchingWizard/PlanActions/PoolPatchMapping.cs
```diff
@@ -10,9 +10,25 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
 {
     public class PoolPatchMapping
```
kc284: Same in this class: you could extend it equality operator overloads etc to avoid similar issues in future.

GaborApatiNagy:
Thanks for your suggestions @kc284

I thought about operator overloading of == and != operators, but I think this will cause readability issues, since we don't normally have operator overloads for these in XenCenter and == should really stay as default, stay as a reference comparison in my opinion. (If we widely change that, I am fine, but would not invent this only here...)

Extending XenServerPatch with overloads I think should be a part of refactoring exercise and we shall not risk it at this stage of this feature (dev complete & last bugs). XenServerPatch class is widely used, I'd rather not change anything there.

Regarding existing code, I will create a(n internal) ticket to cover double checking of all equality operator usages in XenCenter, especially the ones around Hosts and other XenObjects. Having done that I think we should decide how to go on with these operator overloads.

For now, I will submit only some minor changes, hoping that is sufficient for this bugfix in this feature while I am suggesting we should get back to these questions later, covering all code in XenCenter.

XenAdmin/Wizards/PatchingWizard/PlanActions/UploadPatchToMasterPlanAction.cs
```diff
@@ -63,8 +63,10 @@ protected override void RunWithSession(ref Session session)
 
             var poolPatches = new List<Pool_patch>(session.Connection.Cache.Pool_patches);
             var conn = session.Connection;
+            var master = Helpers.GetMaster(conn);
 
-            var existingMapping = mappings.Find(m => m.MasterHost == Helpers.GetMaster(conn) && m.Pool_patch != null && m.XenServerPatch == patch);
+            var existingMapping = mappings.Find(m => m.MasterHost.uuid == master.uuid && m.Pool_patch != null && m.XenServerPatch == patch);
```
kc284: here uuids of XenServerPatch are still missing

XenAdmin/Wizards/PatchingWizard/PlanActions/UploadPatchToMasterPlanAction.cs
```diff
@@ -90,14 +92,9 @@ protected override void RunWithSession(ref Session session)
...
-                    var newMapping = new PoolPatchMapping()
-                    {
-                        MasterHost = Helpers.GetMaster(session.Connection),
-                        XenServerPatch = patch,
-                        Pool_patch = poolPatch
-                    };
+                    var newMapping = new PoolPatchMapping(patch, poolPatch, Helpers.GetMaster(session.Connection));
 
-                    if (!mappings.Any(m => m.MasterHost == newMapping.MasterHost && m.Pool_patch == newMapping.Pool_patch && m.XenServerPatch == patch))
+                    if (!mappings.Contains(newMapping))
```
kc284: Contains will use the Equals from the PoolPatchMapping class which is not comparing uuids, so you'll have the same problem

XenAdmin/Wizards/PatchingWizard/PlanActions/PoolPatchMapping.cs
```diff
@@ -10,9 +10,25 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
...
+
+            if (masterHost == null)
+                throw new ArgumentNullException("masterHost");
+
+            this.XenServerPatch = xenServerPatch;
+            this.Pool_patch = pool_patch;
+            this.MasterHost = masterHost;
+        }
 
         public override bool Equals(object obj)
```
kc284: need to compare uuids in here

GaborApatiNagy:
Thanks @kc284, I have update this PR

kc284:
Thanks for these changes @GaborApatiNagy. It's just occurred to me we can be a bit more defensive and add some extra null ref checks. I'm merging this and will do in a new commit.

GaborApatiNagy:
Thanks Tina, I appreciate your help on this

PR #1130, opened 2016-08-19
CA-217232: Added some null ref checks.

kc284:

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
@@ -55,7 +55,8 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
         {
 
             var master = Helpers.GetMaster(host.Connection);
-            var mapping = mappings.Find(m => m.XenServerPatch.Equals(xenServerPatch) && m.MasterHost.uuid == master.uuid);
+            var mapping = mappings.Find(m => m.XenServerPatch.Equals(xenServerPatch)
+                                             && m.MasterHost != null && master != null && m.MasterHost.uuid == master.uuid);
 
             if (mapping != null && mapping.Pool_patch != null)
             {
@@ -67,7 +68,8 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
             }
             else
             {
-                log.ErrorFormat("Mapping not found for patch {0} on master {1}", xenServerPatch.Uuid, master.uuid);
+                if (xenServerPatch != null && master != null)
+                    log.ErrorFormat("Mapping not found for patch {0} on master {1}", xenServerPatch.Uuid, master.uuid);
 
                 throw new Exception("Pool_patch not found.");
             }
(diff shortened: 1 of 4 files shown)


kc284:
I'm afraid this commit picked also some automatic line endings conversion...

XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
```diff
@@ -67,7 +68,8 @@ protected override void RunWithSession(ref Session session)
             }
             else
             {
-                log.ErrorFormat("Mapping not found for patch {0} on master {1}", xenServerPatch.Uuid, master.uuid);
+                if (xenServerPatch != null && master != null)
+                    log.ErrorFormat("Mapping not found for patch {0} on master {1}", xenServerPatch.Uuid, master.uuid);
```
Frezzle: What if XenServerPatch and/or master is null here... would you still want to log something?

kc284: I didn't deem it necessary because we're throwing the exception and the error will be recorded anyway, so I thought that an extra log line without any useful information such as the patch id wouldn't offer much.

XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrechecksOnMultipleHostsAction.cs
```diff
@@ -38,28 +38,28 @@
...
-        }
-
-        protected override void RunWithSession(ref Session session)
+        }
+
+        protected override void RunWithSession(ref Session session)
         {
             var master = Helpers.GetMaster(Connection);
-            var mapping = mappings.Find(m => m.XenServerPatch.Equals(patch) && m.MasterHost.uuid == master.uuid);
+            var mapping = mappings.Find(m => m.XenServerPatch.Equals(patch) && m.MasterHost != null && master != null && m.MasterHost.uuid == master.uuid);
```
Frezzle: The documentation for Find says it returns "The first element that matches the conditions defined by the specified predicate, if found; otherwise, the default value for type T". What is the default value for a PoolPatchMapping? Is it null?

kc284: PoolPatchMapping is a class, so the default is null

GaborApatiNagy: The constructor of PoolPatchMapping enforces non-null values for these
properties, but Fred's point is worth considering later on

Frezzle: That's okay then, thanks. As for the PoolPatchMapping properties, since they are checked for being assigned as null before being set, why do the checks exist for them being null (such as in PoolPatchMapping.Equals() and the RunWithSession() methods)? The setter for the properties are private and only used in the constructor, so isn't it redundant? Anyway, it's not harming anything.

Frezzle:
Looks good.

XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrechecksOnMultipleHostsAction.cs
kc284: ok, the pm.Master null check is redundant now, but not the master check.

PR #1131, opened 2016-08-19
Set checkbox state in the designer and not in the constructor to avoi…

kc284:
…d starting

the checks before the page is loaded. Corrected the conditions for enabling the
Previous and Next buttons on this page. This commit fixes broken RPU wizard test.


Frezzle:
Looks correct; merging.

PR #1134, opened 2016-08-22
CA-140181: Inconsistent name displaying when it contains symbol '&’

mcintyre94:
The ticket said the character after the & became a hotkey (like labels etc do with UseMnemonic enabled) but I couldn't reproduce that - & symbols were simply removed from the string. The fix is to replace them with &&, so the string.format prints a single & as it should.

### DIFF
--- a/XenAdmin/TabPages/PerformancePage.cs
+++ b/XenAdmin/TabPages/PerformancePage.cs
@@ -360,9 +360,10 @@ namespace XenAdmin.TabPages
 
         private void DeleteGraph()
         {
+            var formattableName = GraphList.SelectedGraph.DisplayName.Replace("&", "&&");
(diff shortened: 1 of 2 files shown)


XenAdmin/TabPages/PerformancePage.cs
```diff
@@ -360,9 +360,10 @@ private void EditGraph()
 
         private void DeleteGraph()
         {
+            var formattableName = GraphList.SelectedGraph.DisplayName.Replace("&", "&&");
```
kc284: You can use the existing string extension method EscapeAmpersands, see file XenModel/StringExtensions

kc284:
Looks good. Merging.

PR #1136, opened 2016-08-23
CP-18012: Use extended precheck information in Update Wizard

GaborApatiNagy:
This commit implements the check for LiveUpdate capability of a patch. If
the patch is completely live-updatable, XenCenter will not restart the
host (but will do further checks once it has finished installing the
update - separate ticket)


mcintyre94:
Discussed + looks good, merging.

PR #1137, opened 2016-08-24
CP-18618: If live patching fails, warn the user

mcintyre94:
If we are expecting to live patch, but at the end
hosts.patches_requring_reboot contains the patch, then warn the user that
live patching has failed for the host - and it needs to be rebooted.


GaborApatiNagy:
I have seen the demo, looks good - merging

PR #1138, opened 2016-08-24
CA-215575: Top connection bar on full screen console was not listening to server events

kc284:


MihaelaStoica:
I had tested it and the only thing I noticed is that if I have XenCenter displayed on a secondary screen, the full screen form is always displayed on the primary screen. This used to work correctly prior to these changes.

kc284:
Thanks @MihaelaStoica, I've made the correction.

MihaelaStoica:
Looks good now. Merging.

PR #1139, opened 2016-08-25
CA-107341:  Truncation issue occurred on rolling pool upgrade window.

mcintyre94:
Ellipsised pool and host names on the Ready to Upgrade page to 64 characters, which resolves the layout problem from extremely long names.


GaborApatiNagy:
Looks fine, merging

PR #1140, opened 2016-08-26
CA-112317: Sorting on SR storage tab is incorrect

mcintyre94:
Change the sorting on the Name and Description columns to use StringUtility.NaturalCompare, which correctly sorts strings with numbers at the end (eg name, name0, name1, name2, name10 will now sort correctly while alphabetic sorting is incorrect). Also added a tie break on uuid to make sorting behave predictably with duplicates in these columns.


GaborApatiNagy:
Looks good, merging

stephen-turner:
I realise this has already been merged, but I think it needs revising.

We already have a built-in compare function on VDIs, which uses Name then opaque_ref. So for the name sorting, it would be better to use this, rather than rewriting it.

For the description, it would be better to use opaque_ref rather than uuid as a tie breaker, for consistency.

mcintyre94:
Changes made in #1144

PR #1144, opened 2016-08-31
CA-112317: Sorting on SR storage tab is incorrect

mcintyre94:
Making the requested changes: Name is now sorted by the generic CompareTo method (which uses name with opaque ref as tiebreaker), and Description also uses opaque ref as tiebreaker for consistency.

Changes requested to closed #1140


MihaelaStoica:
Looks good. Merging.

PR #1145, opened 2016-09-01
CA-140338: Truncation on New Search Page with long Custom Field Filters name

mcintyre94:
The truncating of the custom field name by space (instead of ellipsising) seems to be a Windows default, but it's not ideal. This change edits the text displayed in the dropdown when an item is selected so that if a custom field is selected it is ellipsised to an appropriate length. It doesn't affect the dropdown itself, which continues to display the full custom field name.


MihaelaStoica:
Looks good. Merging.

PR #1147, opened 2016-09-01
CA-220040: Patching wizard shows reboot required where it is not expected

MihaelaStoica:
**CA-220040: Patching wizard shows reboot required where it is not expected**
- When building the list of servers that require post-install actions, ignore the ones where the patch has already been installed.
- Also fixing the infinite recursion in Build() method (causing StackOverflowException when installing a supplemental pack).

**Fix an error in ParallelAction with no subactions**
- Immediately complete the ParallelAction when there are no subactions to be executed, to avoid on infinite lock while waiting for the actions to be completed.


GaborApatiNagy:
looks great, thanks

PR #1148, opened 2016-09-01
CA-220049: Non-livepatch hotfix installation is failing

MihaelaStoica:
This fixes the NullReferenceException that occurs when trying to evacuate a host after a non-livepatch update (that requires reboot) has been installed


GaborApatiNagy:
looks great, thanks

PR #1149, opened 2016-09-02
CA-220171: Configure HA dialog incorrectly states failover tolerance …

MihaelaStoica:
…would be 0


GaborApatiNagy:
Looks great, merging

PR #1150, opened 2016-09-02
CA-149965: Snapshot name is truncated on “Revert To Snapback” dialog.

mcintyre94:
Changed the ellipsise length from 45 to 90 characters. This means that the first row of text becomes variable length, so I've reworked the dialog to support this and work like others: icon/rows of text in table cells, right aligned flow panel for buttons.


kc284:
I'm afraid these changes have caused some side effects, namely: the form icon has reverted to the designer default icon; the width of the yes/no buttons has shrunk; so has the size of the warning triangle.
Also, I think 90 is too long, because it's causing the snapshot name to overflow into a different line, thus leaving a lot of empty space on the first line; how about leaving it at 45 or 50 and just ensuring that the closing quote and question mak are not truncated, which is what the complaint was mostly about.

kc284:
Look ok; merging.

PR #1151, opened 2016-09-05
CA-218956: Expose HIMN when showing hidden objects

huazhihao:
Expose both the HIMN network and any VIFs associated with the HIMN
network using XenCenter when the "View->Show Hidden Objects" option
is displayed. However the shown HIMN should not be able to autoplug,
edit or delete.

### DIFF
--- a/XenAdmin/Wizards/ImportWizard/NetworkPickerPage.cs
+++ b/XenAdmin/Wizards/ImportWizard/NetworkPickerPage.cs
@@ -295,7 +295,8 @@ namespace XenAdmin.Wizards.ImportWizard
 
             // CA-73056: A row for the guest installer network shouldn't show up. But we still need
             // it present but invisible, otherwise the corresponding VIF doesn't get created at all.
-            if (isGuestInstallerNetwork)
+            // CA-218956 - Expose HIMN when showing hidden objects
+            if (isGuestInstallerNetwork && !XenAdmin.Properties.Settings.Default.ShowHiddenVMs)
                 row.Visible = false;
 
 			m_networkGridView.Rows.Add(row);
--- a/XenModel/XenAPI-Extensions/Network.cs
+++ b/XenModel/XenAPI-Extensions/Network.cs
@@ -147,8 +147,8 @@ namespace XenAPI
 
         public override bool Show(bool showHiddenVMs)
         {
-            
-                if (IsGuestInstallerNetwork)
+                // CA-218956 - Expose HIMN when showing hidden objects
+                if (IsGuestInstallerNetwork && !showHiddenVMs)
                     return false;
 
                 if (!ShowAllPifs(showHiddenVMs))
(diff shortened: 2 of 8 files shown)


MihaelaStoica:
@huazhihao Please note that the Properties dialog is still available from the Network's context menu if you switch the view in XenCenter to "Objects" (or "VMs by Network" saved search) and the HIMN is displayed in the left-hand side tree.

XenModel/XenAPI-Extensions/Network.cs
```diff
@@ -125,13 +125,20 @@ public bool AutoPlug
             {
                 // Note that this is not equivalent to BoolKey, because here
                 // absence of the key gives AutoPlug=true, not false.
-                return Get(other_config, "automatic") != "false";
+                // CA-218956 - Expose HIMN when showing hidden objects
```
MihaelaStoica: I would prefer to leave the Autoplug property unchanged (i.e. read/write the automatic flag regardless of the network type) and prevent this being set/used if the network is HIMN in XenCenter, for example by hiding the checkbox from the Network Properties dialog (although that wouldn't really be necessary if the Properties dialog would never be accessible for HIMN)

MihaelaStoica:
There are some other usages of network.IsGuestInstallerNetwork which haven't been changed to expose HIMN:
- in the ExportResourceReportAction class, the HIMN is ignored when the list of networks is built. I think that it should be included if is visible in XenCenter.
- in the VM import wizard, we might want to expose the HIMN in the same way it is exposed in the New VM Wizard.

huazhihao:
@MihaelaStoica Thanks a lot. You've been helpful. In the updated commit, I've:

- Reverted Autoplug property
- Dsiabled Properties context menu in Object view
- Expose HIMN in VM import/export wizard

Could you have a look if my change has addressed your comments? Thank you.

XenAdmin/Wizards/ImportWizard/NetworkPickerPage.cs
```diff
@@ -274,7 +274,9 @@ private void AddVIFRow(VIF vif)
 		{
             var row = new VifRow(vif);
             XenAPI.Network network = m_selectedConnection.Resolve(vif.network);
-            bool isGuestInstallerNetwork = network != null && network.IsGuestInstallerNetwork;
+            bool isGuestInstallerNetwork = network != null &&
+                network.IsGuestInstallerNetwork && !XenAdmin.Properties.Settings.Default.ShowHiddenVMs;
+            // CA-218956 - Expose HIMN when showing hidden objects
```
MihaelaStoica: The check for hidden objects shouldn't be done here, but on line 300 below 
```
if (isGuestInstallerNetwork)
    row.Visible = false;
```
The logic here is that we should always add a row for the guest installer network, to ensure that the corresponding VIF gets created; and we should then hide it, **if hidden objects are not shown**.

huazhihao: Thanks. Done.

MihaelaStoica:
Thanks @huazhihao for the rework. Looks good now.

PR #1164, opened 2016-09-14
Did some work on the SrStoragePage:

kc284:
- show/hide the context menu items instead of creating and adding new ones each time the context menu opens
- right click menu with rescan/add was not available for empty list or when empty space was clicked
- the condition for the edit button was slightly different from the edit toolstripmenuitem
- improved button layout (part of CA-100083)
- tidied up button event handlers and renamed one or two class members
- removed unused parameter in constructor and method of VDIRow

### DIFF
--- a/XenAdmin/TabPages/SrStoragePage.resx
+++ b/XenAdmin/TabPages/SrStoragePage.resx
@@ -547,13 +547,13 @@
     <value>205, 26</value>
   </data>
   <data name="moveVirtualDiskToolStripMenuItem.Text" xml:space="preserve">
-    <value>&amp;Move virtual disk...</value>
+    <value>Mov&amp;e virtual disk...</value>
   </data>
   <data name="deleteVirtualDiskToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
     <value>205, 26</value>
   </data>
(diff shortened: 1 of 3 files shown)


XenAdmin/TabPages/SrStoragePage.resx
```diff
@@ -655,139 +307,313 @@
...
+    <value>205, 26</value>
+  </data>
+  <data name="moveVirtualDiskToolStripMenuItem.Text" xml:space="preserve">
+    <value>&amp;Move virtual disk...</value>
+  </data>
+  <data name="deleteVirtualDiskToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
+    <value>205, 26</value>
+  </data>
+  <data name="deleteVirtualDiskToolStripMenuItem.Text" xml:space="preserve">
+    <value>Delete virtual disk</value>
```
MihaelaStoica: Hotkey is missing here.

kc284: Indeed. And I've just noticed the hotkey for the move menu item is different from the one for the move button, it would be nicer to use the same one.

MihaelaStoica:
Looks good. Merging.

PR #1165, opened 2016-09-14
Cp 18417: Modify the hotfix wizard to make it work for hosts that cannot autorestart

mcintyre94:
If a pool has OtherConfig key "XenCenter.CustomFields.hci-forbid-update-auto-restart", disable automatic updates + option to automatically restart. 


XenAdmin/Wizards/PatchingWizard/PatchingWizard_ModePage.cs
```diff
@@ -90,19 +87,42 @@ public override void PageLoaded(PageLoadedDirection direction)
...
+                    unknownType = true;
+                    break;
+            }
+
+            var anyPoolForbidsAutoRestart = AnyPoolForbidsAutoRestart();
+            var patchRequiresReboot = PatchRequiresReboot();
+
+            var automaticDisabled = unknownType || (anyPoolForbidsAutoRestart && patchRequiresReboot);
+
+            if (automaticDisabled)
```
kc284: This block can be compacted a bit:
```
AutomaticRadioButton.Checked = !automaticDisabled;
AutomaticRadioButton.Enabled = !automaticDisabled;
ManualRadioButton.Checked = automaticDisabled;

if (automaticDisabled)
    allowRadioButtonContainer.SetToolTip(Messages.POOL_FORBIDS_AUTOMATIC_RESTARTS);
```

XenAdmin/Wizards/PatchingWizard/PatchingWizard_ModePage.cs
```diff
@@ -165,6 +185,29 @@ private void ManualRadioButton_CheckedChanged(object sender, EventArgs e)
             UpdateEnablement();
         }
 
+        private bool AnyPoolForbidsAutoRestart()
+        {
+            var poolForbidRestartKey = "XenCenter.CustomFields.hci-forbid-update-auto-restart";
```
kc284: These keys are normally like "hci-forbid-update-auto-restart: 1;" without prefixes and usually just checking whether the key exists is enough, see for example the key rolling_upgrade_in_progress.
I would also suggest moving this to the pool extensions as well as avoiding the use of the Helpers.GetOtherConfig method on an object of known type, as it's not necessary.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_ModePage.cs
```diff
@@ -165,6 +185,29 @@ private void ManualRadioButton_CheckedChanged(object sender, EventArgs e)
...
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private bool PatchRequiresReboot()
+        {
+            return !textBoxLog.Text.Equals(Messages.PATCHINGWIZARD_MODEPAGE_NOACTION);
```
kc284: Is there a better way to check this than comparing strings?

An approach I find better is to modify the ModeRetailPatch and ModeSuppPack methods to return a bool signifying whether action is required and adding an out string parameter to return the guidance text that will populate the box - then you check that bool.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -222,6 +236,14 @@ private void EnabledRow(Host host, UpdateType type, int index)
                     return;
                 }
 
+                
+                if (poolOfOne != null && PoolForbidsAutomaticUpdates(poolOfOne))
```
kc284: For performance reasons I would move this check to the beginning, I don'st see why we should start calculating upgrading sequences if the pool does not support automatic updates.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_ModePage.cs
GaborApatiNagy: If you are already making changes here, I'd suggest this should be a getter property. For the string, you could invent another new property to avoid having an out parameter.

Also, it's a good practice to name such bool properties to start with "Is". "IsRebootRequired" or similar.

kc284:
The ticket description wants also to disable the autopmatic radio button on the SelectPatchPage, has this change been implemented?

mcintyre94:
@kc284 the Select Patch page was discussed with Stephen and he said we should disable hosts on the Select Server page if automatic is selected instead.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
GaborApatiNagy: I think this is good here, as we need to do the previous checks first, such as license status or partially upgraded pools, thus XC will show the restart-related tooltip only if that is the only reason we can't do the auto (batch) update.

I am happy to rethink this, but I think a reason such as having a partially upgraded pool or an unlicesed pool or the pool is up-to-date shall be shown even if restarting was forbidden for any reason.

GaborApatiNagy: Also, having it here, this feature can still be used to see if the pool is up-to-date at least.

GaborApatiNagy: (Personal opinion and preferences here, I am open to discussions)

mcintyre94: I'm happy to rethink this too, but I didn't want to change any of the current behaviour by overriding error messages.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_ModePage.cs
kc284: Having two properties means that he checks would be run twice; with the out parameter we avoid this - there is nothing wrong with out parameters :)

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
kc284: @GaborApatiNagy , thanks for this input, then it seems to me we agree that the right order of checks should be: pool partially upgraded; pool unlicensed; automatic updates unsupported; everything else that requires calculation of upgrading sequences. Right?

XenAdmin/Wizards/PatchingWizard/PatchingWizardModeGuidanceBuilder.cs
```diff
@@ -40,27 +40,33 @@ namespace XenAdmin.Wizards.PatchingWizard
 {
     public class PatchingWizardModeGuidanceBuilder
     {
-        public static string ModeRetailPatch(List<Host> servers, Pool_patch patch, Dictionary<string, LivePatchCode> LivePatchCodesByHost)
+        public static string ModeRetailPatch(List<Host> servers, Pool_patch patch, Dictionary<string, LivePatchCode> LivePatchCodesByHost, out bool someHostMayRequireRestart)
```
kc284: This method is also used in the tests, please amend them so they compile.

XenModel/Messages.resx
```diff
@@ -9442,6 +9442,12 @@ The VM protection policy for this VM does not have automatic archiving configure
   <data name="POOL_FAILURE_TOLERATE" xml:space="preserve">
     <value>{0} failure tolerated</value>
   </data>
+  <data name="POOL_FORBIDS_AUTOMATIC_RESTARTS" xml:space="preserve">
+    <value>Automatic post-update tasks are not supported for at least one server to be updated.</value>
+  </data>
+  <data name="POOL_FORBIDS_AUTOMATIC_UPDATES" xml:space="preserve">
+    <value>Automatic updates are not supported for this server.</value>
```
kc284: It's just occurred to me that we have another two messages saying "Automatic update is not supported", could you please change this to singular so they're all the same.

Also, I had a rethink about the top message, I think "Post-update tasks cannot be carried out automatically for all the servers to be updated." conforms better with existing messages in this context.

kc284:
Looks ok to me now :) Merging.

PR #1170, opened 2016-09-17
CA-100083 (part of): Set minimum width of the main window's tab contr…

kc284:
…ol to 40% to

avoid buttons on the console tab overlapping. Removed some panels from the console
tab to reduce control nesting.


MihaelaStoica:
This looks fine. The only thing I noticed was the tab order on the Console tab: in order to get to the console itself you need to tab through all the buttons on the page, including the ones at the buttom.

MihaelaStoica:
:+1:

PR #1171, opened 2016-09-19
CA-219777: Bad progress dialog message

mcintyre94:
The bad message was in the dialog created in the BugToolWizard after selecting servers. Since this method is the only one using the ActionProgressDialog(String) constructor, I've modified that ctor to hide the subaction label - which makes sense because there is no action in that ctor (to set a subaction) and no other appropriate label.


MihaelaStoica:
Looks good. Merging.

PR #1173, opened 2016-09-19
CA-220511, CA-187659, CA-220216 plus some refactoring and minor enhancements.

kc284:


kc284:
Closing this temporarily as I need to do some more work on migration.

XenAdminTests/CommandTests/MigrateVirtualDiskCommandTest.cs
```diff
@@ -50,7 +50,7 @@ public class MigrateVirtualDiskCommandTest : DatabaseTester_TestFixture
         [Test]
         public void CheckContextMenuText()
         {
-            MigrateVirtualDiskCommand cmd = new MigrateVirtualDiskCommand(mw, null as VDI);
+            var cmd = new MigrateVirtualDiskCommand(mw, null);
```
MihaelaStoica: I believe that this change causes a this test to fail with ArgumentNullException, as the second parameter in the constructor (Selection) cannot be null. Same for CommandCannotExecuteNullVdi.

The ContextMenu_VDI test is also failing, but probably for a different reason.

kc284: Now that I'm looking at it again, this test is conceptually wrong because it's checking the text is correct on a command that shouldn't have been constructed at all due to null selection...

MihaelaStoica:
Looks ok now. Merging.

PR #1175, opened 2016-09-19
CA-218242: Don't assign values outside the range in the memory spinne…

MihaelaStoica:
…r control

- The value of the spinner control is rounded to the nearest, while the minimum is rounded up and the maximum down (or, in some cases, min and max are not rounded at all).
In most cases this is fine, but if the current value (in bytes) is equal to one of these ranges, by rounding this way we might end up with a value that is outside of the control's ranges (causing ArgumentOutOfRange exception).


kc284:
Looks ok to me. Please remember to cherry pick this change into the feature branch too.

PR #1176, opened 2016-09-19
CA-218242: Fix crash in Dom0 memory dialog

MihaelaStoica:
- reset the spinner limits before changing the value


MihaelaStoica:
Please note that the fix in #1175 is also needed on this branch, to avoid setting the spinner's value outside the range due to rounding.

kc284:
Looks ok to me.

PR #1177, opened 2016-09-20
CAR-2184: XAPI changes.

mcintyre94:
I've made the minimal possible changes to Pvs Page to ensure it compiles, but it's had a lot of feature work (CP-16922) that supersedes this, these small changes should keep that merge painless.


MihaelaStoica:
@mcintyre94, could we please update only the XenAPI classes related to this CAR? All the other changes should already exist in the master branch.

Also, why the file XenAPI-Extensions\PVS_site.cs has been removed (actually renamed to XenAPI\pvs_proxy_status.cs). Is it not needed anymore?

mcintyre94:
Ah okay, I see - I'll reverse the others. I removed the sites extension because it only contained Name which is now a property of the pvs_site object (but I'll reverse that). pvs_proxy_status is a different, new XAPI file.

MihaelaStoica:
Looks good now. Merging.

PR #1180, opened 2016-09-21
CA-218260: Reverting to Snapshots causes VM to go back to parent host

mcintyre94:
Where we have a host the VM is running on (ie vm.resident_on is not null) at the time it is reverted, attempt to start/resume on that host if possible, instead of the home server of the VM.

### DIFF
--- a/XenModel/XenAPI-Extensions/VM.cs
+++ b/XenModel/XenAPI-Extensions/VM.cs
@@ -2073,6 +2073,20 @@ namespace XenAPI
                 return other_config != null && other_config.ContainsKey("hci-warn-before-shutdown");
             }
         }
+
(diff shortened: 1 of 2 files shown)


XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -2073,6 +2073,20 @@ public bool HciWarnBeforeShutdown
                 return other_config != null && other_config.ContainsKey("hci-warn-before-shutdown");
             }
         }
+
+        public bool CanBootOnHost(Host host)
```
MihaelaStoica: We don't generally execute API calls in object's properties. This function should be moved into the action class, and use the action's session.

MihaelaStoica:
Looks good. Merging.

PR #1181, opened 2016-09-23
CA-223082: Replace PATCH_PRECHECK_LIVEPATCH_MISSING

GaborApatiNagy:
Replace PATCH_PRECHECK_LIVEPATCH_MISSING to PATCH_PRECHECK_LIVEPATCH_NOT_APPLICABLE due to changes in XAPI
Note: if XAPI change gets active before this change we will see UNKNOWN instead of PATCH_PRECHECK_LIVEPATCH_NOT_APPLICABLE in XenCenter what will make no difference


MihaelaStoica:
Looks good. Merging

PR #1182, opened 2016-09-23
CP-17920: Show the PVS tab only when the supplemental pack is installed & Hide Memory SR

MihaelaStoica:
- Show the PVS tab only when the supplemental pack is installed 
- Hide memory SR in XenCenter


kc284:
Look ok to me.

PR #1184, opened 2016-09-26
CP-17920: Add dialog for PVS cache configuration

MihaelaStoica:
- this commit contains a temporary icon for PVS site


GaborApatiNagy:
It looks like the subtext for PVS sites stay "Not configured" even when we configure them

GaborApatiNagy:
I think it would be nice to have a "Delete PVS site" button for each PVS site so it could be accessed by keyboard (similar to how it works on Networking pages)

GaborApatiNagy:
It looks easy to overbook an SR when configuring PVS sites in this setup - some warnings at least would be nice to have. (As a suggestion for future improvement)

XenAdmin/Dialogs/PvsPages/PvsCacheConfigurationPage.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        public AsyncAction SaveSettings()
+        {
+            if (!ValidToSave)
+                return null;
+
+            var pvsCacheStorages = new List<PVS_cache_storage>();
+            foreach (var row in rows.Where(r => r.HasChanged))
+            {
+                var pvsCacheStorage = new PVS_cache_storage
```
GaborApatiNagy: I suggest we rename this to reflect that this contains the added/new items.

PR #1188, opened 2016-09-30
CA-220218, CA-205799, CA-205791 plus one or two other issues and refactoring

kc284:


XenAdmin/Wizards/CrossPoolMigrateWizard/Filters/CrossPoolMigrateCanMigrateFilter.cs
```diff
@@ -108,103 +100,77 @@ public override bool FailureFound
...
             }
         }
 
-        /// <summary>
-        /// If we have a pool, then we need to fail only when there are failures for 
-        /// each of the hosts. Otherwise any failure counts.
-        /// </summary>
-        /// <returns></returns>
-        private bool DetermineIfFailureGivenReasons()
+        private List<Host> CollateHosts(out Pool thePool)
```
GaborApatiNagy: I would rather avoid having an out parameter here. I can't see any reasons why this is not a separate method/property. GetTargetPool I would suggest. I think this method now tries to do two things. Out parameter is a red flag to me, sorry :)

XenModel/Messages.resx
```diff
@@ -10873,6 +10879,9 @@ The upper limit: SR size / {2}</value>
   <data name="STORAGE_WORST" xml:space="preserve">
     <value>0 (Lowest)</value>
   </data>
+  <data name="String1" xml:space="preserve">
```
GaborApatiNagy: This seems to be here by accident

XenAdmin/Commands/Controls/MigrateVMToolStripMenuItem.cs
```diff
@@ -76,60 +76,59 @@ public new Command Command
...
 
-                    return SelectionInVisiblePool(vm.Connection) && vm.Connection.Cache.HostCount > 1 && vm.allowed_operations != null && vm.allowed_operations.Contains(vm_operations.pool_migrate);
-                    
-                }
-                return false;
+                return SelectionInVisiblePool(vm.Connection) && vm.Connection.Cache.HostCount > 1 && vm.allowed_operations != null && vm.allowed_operations.Contains(vm_operations.pool_migrate);
             }
 
             protected override bool CanExecuteCore(SelectedItemCollection selection)
             {
```
GaborApatiNagy: I don't mind if it stays like this, but if this method is being refactored it could be made a lot clearer as 

if (selection.All(...)) then return selection.Any()

kc284: Yes, I was thinking about it, but then I was concerned that we'd have to iterate through the collection once for All and then once more (even if partially) for Any, so in the end I thought better leave it as is.

XenAdmin/Wizards/CrossPoolMigrateWizard/Filters/CrossPoolMigrateCanMigrateFilter.cs
GaborApatiNagy: Maybe personal preferences don't match completely here, but of course I am fine with leaving this as is.

GaborApatiNagy:
Thanks for the changes @kc284

PR #1192, opened 2016-09-30
CP-19086: Improvements to the PVS cache UI

MihaelaStoica:
PVS Tab:
- Move the "View PVS sites" button inside the "Configure PVS Cache" dialog (next to Remove button) and make it per-site (showing only the PVS servers from that site)
- Remove "Currently attached" column from the VMs section
- Rename "Status" column to "Read caching status"

Configure PVS Cache" dialog:
- Make sure there is a scrollbar to accommodate all hosts in a pool
- Add a confirmation dialog when a site is removed. The action is executed when the Remove button is pressed.
- When Memory SR is selected, set the maximum value to the dom0-extra-memory..
- Allow empty configuration (but see next point): The OK button should only be disabled if the PVS site name is empty
- Add an extra option in the "Cache storage" combobox, with text "Not configured", selected by default on a new site. When this is selected, we will remove the existing entry in pvs_cache_storage and will not create a new one.
- When a PVS site is deleted (forgotten), we first delete all the associated PVS_servers, otherwise the site cannot be deleted.


XenModel/Messages.resx
```diff
@@ -13249,4 +13249,7 @@ You will need to navigate to the Console on each of the selected VMs to complete
   <data name="CONFIRM_DELETE_PVS_SITE" xml:space="preserve">
     <value>Are you sure you want to delete '{0}'?</value>
   </data>
+  <data name="PVS_SITE_CANNOT_BE_REMOVED" xml:space="preserve">
+    <value>This PVS site cannot be removed because it contains running proxies.</value>
```
mcintyre94: We don't have proxies mentioned anywhere else in the PVS UI, is this clear enough or should we refer to the VM table to help them resolve this?

XenAdmin/Dialogs/PvsPages/PvsCacheConfigurationPage.cs
```diff
@@ -82,8 +82,9 @@ public void SetXenObjects(IXenObject orig, IXenObject clone)
                 // Generate a unique suggested name for the new site
                 textBox1.Text = Helpers.MakeUniqueName(Messages.PVS_SITE_NAME, takenNames);
             }
-            
+
             LoadServers();
+            viewPvsServersButton.Enabled = PvsSite != null;
```
mcintyre94: Should this be enabled only if there are PVS servers for the site?

MihaelaStoica:
@mcintyre94 , please see my latest commit. 
Regarding your questions:

- Should the sidebar update if new PVS sites are added externally? We treat this dialog as a "Properties" dialog, so we are not refreshing when the object changes, because we don't want to have things that we are editing changing while we are doing this

- Could we perform the delete action without blocking the UI to make deleting multiple sites quicker? 
We need to wait for the action to complete,in order to delete the site tab if it the action is successful.

mcintyre94:
We agreed on two more changes for this ticket:

- When the PVS Site collection changes, if it is an added site then we should add it to the UI with a configuration page. This improves the case where sites being created aren't shown in the dialog.
- To defend against the case where the collection changes and a site is deleted, check a site exists before attempting to modify it when the dialog is saved.

mcintyre94:
This looks good now :) Merging

PR #1196, opened 2016-10-07
[CA-114492] Stop GroupBy buttons remaining overlapped…

mcintyre94:
...after a drag operation by ensuring they snap back to a correct setup on mouseup

### DIFF
--- a/XenAdmin/Controls/XenSearch/GroupingControl.cs
+++ b/XenAdmin/Controls/XenSearch/GroupingControl.cs
@@ -403,6 +403,7 @@ namespace XenAdmin.Controls.XenSearch
         {
             draggedButton = null;
             dragging = false;
+            Setup();
         }
 
         void button_MouseDown(object sender, MouseEventArgs e)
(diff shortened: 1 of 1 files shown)


XenAdmin/Controls/XenSearch/GroupingControl.cs
```diff
@@ -403,6 +403,7 @@ void button_MouseUp(object sender, MouseEventArgs e)
         {
             draggedButton = null;
             dragging = false;
+            Setup();
```
kc284: The problem with this solution is that Setup is called and does things twice (first for MouseMove and then for MouseUp). Also, it's called when not needed (when just clicking on the button without dragging it).

kc284:
Personally, and if I had time in my hands, I would just get rid of all this code with the custom draggable buttons and use a toolstrip styled to look like panel+normal buttons. Toolstrip buttons can be dragged around etc and all is done by the framework without extra code. But this is probably low priority.

XenAdmin/Controls/XenSearch/GroupingControl.cs
kc284: Yes, the priority here is to avoid somehow Setup being called multiple times (my other comment was rhetoric :) )

PR #1197, opened 2016-10-07
CA-224619: XenCenter cannot use Switch to Remote Desktop option for V…

MihaelaStoica:
…Ms connected to internal network

- include networks without PIFs in the list of networks to test for RDP (at the end of the list)


MihaelaStoica:
I don't think we have IP addresses for disconnected PIFs, but I will change it to collect the addreses only if the PIF is connected or is null.

PR #1201, opened 2016-10-11
CA-214653: Work-around for XC reconnecting to proxy server with new wrong credentials

Frezzle:
Normally, a heartbeat is done every 15 seconds by getting the server time, and starts as soon as connection to host is established, but if an error occurs XC retries once more in another 15 seconds before actually closing the connection to the host.

Now, if a 407 Proxy Authentication Required error occurs on the initial heartbeat then XC does not retry again in 15 seconds; it will close the connection immediately. The disconnection was instant to up to 2 seconds, depending on time taken to get server time (testing with FreeProxy was ~instant and CCProxy was ~2 seconds). This is only a work-around, as the exact reason for this behaviour is still unknown.

### DIFF
--- a/XenModel/Network/Heartbeat.cs
+++ b/XenModel/Network/Heartbeat.cs
@@ -131,7 +131,7 @@ namespace XenAdmin.Network
 
                 // Now that we've successfully received a heartbeat, reset our 'second chance' for the server to timeout
                 if (retrying)
-                    log.DebugFormat("Heartbeat for {0} has come back", session == null ? "null" : session.Url);
+                    log.DebugFormat("Heartbeat for {0} has come back", session.Url);
                 retrying = false;
             }
             catch (TargetInvocationException exn)
@@ -146,16 +146,19 @@ namespace XenAdmin.Network
                     log.Error(exn);
                 }
                 HandleConnectionLoss();
-                return;
             }
             catch (WebException exn)
             {
                 log.Error(exn);
                 if (((HttpWebResponse)exn.Response).StatusCode == HttpStatusCode.ProxyAuthenticationRequired) // work-around for CA-214653
                 {
-                    log.DebugFormat("Heartbeat for {0} has failed due to {1} credentials; closing the main connection",
-                        session == null ? "null" : session.Url,
-                        session.proxy.Proxy.Credentials == null ? "missing" : "incorrect");
+                    if (session == null)
+                        log.Debug("Heartbeat has failed due to null session; closing the main connection");
+                    else if (session.proxy.Proxy.Credentials == null)
+                        log.DebugFormat("Heartbeat for {0} has failed due to missing credentials; closing the main connection", session.Url);
+                    else
+                        log.DebugFormat("Heartbeat for {0} has failed due to incorrect credentials; closing the main connection", session.Url);
+
                     connection.Interrupt();
                     DropSession();
                 }
@@ -163,13 +166,11 @@ namespace XenAdmin.Network
                 {
                     HandleConnectionLoss();
                 }
-                return;
             }
             catch (Exception exn)
             {
                 log.Error(exn);
                 HandleConnectionLoss();
-                return;
             }
         }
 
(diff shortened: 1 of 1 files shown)


XenModel/Network/Heartbeat.cs
```diff
@@ -148,6 +148,23 @@ private void DoHeartbeat()
...
                 return;
             }
+            catch (WebException exn)
+            {
+                log.Error(exn);
+                if (((HttpWebResponse)exn.Response).StatusCode == HttpStatusCode.ProxyAuthenticationRequired) // work-around for CA-214653
+                {
+                    log.DebugFormat("Heartbeat for {0} has failed due to {1} credentials; closing the main connection",
+                        session == null ? "null" : session.Url,
+                        session.proxy.Proxy.Credentials == null ? "missing" : "incorrect");
```
kc284: If session==null you'll have trouble with argument {1}.
Also, although we currently don't localise log messages, it is generally not good practice to construct strings like this. Better provide the whole sentence, e.g.
if (session ==null) log.DebugFormat(....)
else if (session.proxy.Proxy.Credentials == null) log.DebugFormat(...)
else ...

XenModel/Network/Heartbeat.cs
```diff
@@ -148,6 +148,23 @@ private void DoHeartbeat()
...
+                    connection.Interrupt();
+                    DropSession();
+                }
+                else
+                {
+                    HandleConnectionLoss();
+                }
+                return;
```
kc284: The returns at the end of each catch case are not necessary, it will return anyway since there is nothing to do after the try-catch and before the end of the method.

kc284:
Line 134: there's a check for null session - can it be so at that point? Then we can't have gone past the GetServerTime line, I would expect the latter to have thrown an exeption.

Frezzle:
@kc284 you're right about line 134; GetServerTime on line 130 should throw a NullReferenceException if the session is null, not reaching line 134. I will remove the check.

XenModel/Network/Heartbeat.cs
Frezzle: In fact, I don't think the session could be null, since we're catching a WebException?

Frezzle: So the way I understand it then is that any formatted string ideally shouldn't have any argument which can only be of one language (e.g. {1} in this case, which can only be "missing" or "incorrect")? I will change that.

But arguments which are data (such as links, numbers, UUIDs, etc) are fine to keep (as they are localized 'by nature')? If the answer to this is yes then is the string "null" considered localised or not? Is it okay to keep line 207, for example, which uses the "null" string when session is null:
`log.DebugFormat("Heartbeat for {0} has failed; retrying", session == null ? "null" : session.Url);`
..or should these instances also be changed?

XenModel/Network/Heartbeat.cs
Frezzle: Could this be considered good practice? As if explicitly saying that once we've caught an exception we WILL return, thereby warding off any future code being written after the try-catch block?

XenModel/Network/Heartbeat.cs
kc284: In this particular case it doesn't make much sense in english either to say "heartbeat for null has failed". Back to my earlier snippet I would write, for example:
```
if (session ==null) log.Debug("cannot find session") //or "session is null" etc.
else if (session.proxy.Proxy.Credentials == null) log.DebugFormat("heartbeat for {0} failed due to missing credentials", sessionurl);
else log.DebugFormat("heartbeat for {0} failed due to incorrect credentials", sessionurl);
```

XenModel/Network/Heartbeat.cs
kc284: I don't know what that future code will be doing, so how can I know I will want to return? :) Better handle it when the time comes.
I presume it's a matter of style, but in general I consider good practice leaving out code that makes other developers wonder "now why is this thing here since theoretically it's not needed, there must be a reason, but what is it?"

Frezzle: That makes sense, thanks!

Frezzle:
@kc284 New commit with changes made.

kc284:
It looks ok to me; merging.

PR #1203, opened 2016-10-12
CA-225015: XenCenter Allows VM migration to MemorySR

MihaelaStoica:
- We shouldn't include the hidden SRs in the list of available SRs in the Migration wizard
- Updated the SR.SupportsVdiCreate function to return false for the Memory SR; this fixes other places where the Memory SR might be visible (e.g. all places where SrPicker is used)
- Also fixed the same issue in the  Attach disk dialog


kc284:
Looks ok; merging.

PR #1207, opened 2016-10-12
CP-19111: Added ability to choose Basic or Digest authentication for proxy server

Frezzle:
2 new radio buttons in Connection Options page, for Basic and Digest; Digest is the default as it's the most secure.
New proxy authentication method setting, which is used to configure XenAPI's HTTP class and the .NET AuthenticationManager (which handles which authentication schemes can be used by the .NET web classes).
The new setting is also transferred and used by the Health Check service.

The bug noted in CA-214653 also occurs here, but the work-around on PR#1201 for that bug will also work for this.

### DIFF
--- a/XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
+++ b/XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
@@ -151,73 +151,19 @@ namespace XenAdmin.Dialogs.OptionsPages
             enableOK();
         }
 
-        private void ProxyAddressTextBox_TextChanged(object sender, EventArgs e)
+        private void GeneralProxySettingsChanged(object sender, EventArgs e)
         {
             if (eventsDisabled)
                 return;
-
             SelectUseThisProxyServer();
-
             enableOK();
         }
 
-        private void ProxyPortTextBox_TextChanged(object sender, EventArgs e)
+        private void ProxyAuthenticationSettingsChanged(object sender, EventArgs e)
         {
             if (eventsDisabled)
                 return;
-
-            SelectUseThisProxyServer();
-
-            enableOK();
-        }
-
-        private void ProxyUsernameTextBox_TextChanged(object sender, EventArgs e)
-        {
-            if (eventsDisabled)
-                return;
-
             SelectProvideCredentials();
-
-            enableOK();
-        }
-
-        private void ProxyPasswordTextBox_TextChanged(object sender, EventArgs e)
-        {
-            if (eventsDisabled)
-                return;
-
-            SelectProvideCredentials();
-
-            enableOK();
-        }
-
-        private void BypassForServersCheckbox_CheckedChanged(object sender, EventArgs e)
-        {
-            if (eventsDisabled)
-                return;
-
-            SelectUseThisProxyServer();
-
-            enableOK();
-        }
-
-        private void BasicRadioButton_CheckedChanged(object sender, EventArgs e)
-        {
-            if (eventsDisabled)
-                return;
-
-            SelectProvideCredentials();
-
-            enableOK();
-        }
-
-        private void DigestRadioButton_CheckedChanged(object sender, EventArgs e)
-        {
-            if (eventsDisabled)
-                return;
-
-            SelectProvideCredentials();
-
             enableOK();
         }
 
--- a/XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
+++ b/XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
@@ -30,16 +30,11 @@
  */
 
 using System;
-using System.Collections.Generic;
-using System.ComponentModel;
 using System.Drawing;
-using System.Data;
-using System.Text;
 using System.Windows.Forms;
 using XenAdmin.Properties;
 using XenAdmin.Actions;
 using XenAPI;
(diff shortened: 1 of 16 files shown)


Frezzle:
Will PR HTTP.cs separately into xen-api-sdk once(/if?) approved here.

XenAdmin/Program.cs
```diff
@@ -990,9 +998,59 @@ public static void ReconfigureConnectionSettings()
...
+            while (authModules.MoveNext())
+            {
+                var module = (IAuthenticationModule)authModules.Current;
+                modulesToUnregister.Add(module);
+            }
+
+            foreach (var module in modulesToUnregister)
+                AuthenticationManager.Unregister(module);
+
+            var authSetting = (HTTPHelper.ProxyAuthenticationMethod)Properties.Settings.Default.ProxyAuthenticationMethod;
```
MihaelaStoica: What will happen here if the ProxyAuthenticationMethod value is different than any value in the HTTPHelper.ProxyAuthenticationMethod enum (for example if someone changes it  manually to 2)?

XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
```diff
@@ -107,6 +115,8 @@ private void build()
                 ProxyUsernameTextBox.Visible = false;
                 ProxyPasswordLabel.Visible = false;
                 ProxyPasswordTextBox.Visible = false;
+                AuthenticationMethodPanel.Visible = false;
```
Frezzle: build() is called only once, in the ConnectionOptionsPage constructor. The authentication controls are created every time the options dialog is opened, are visible by default, and are made invisible if proxy authentication is disabled; so the visibility works correctly when reopening the dialog after enabling/disabling the registry key. Does the visibility need to update while the options dialog is still open?

MihaelaStoica: If build() is called only once in the constructor, then it's fine how it is.
We don't normally update the options dialog while it's open so no need to do it here.

XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
```diff
@@ -191,6 +201,26 @@ private void BypassForServersCheckbox_CheckedChanged(object sender, EventArgs e)
...
+        {
+            if (eventsDisabled)
+                return;
+
+            SelectProvideCredentials();
+
+            enableOK();
+        }
+
+        private void DigestRadioButton_CheckedChanged(object sender, EventArgs e)
```
MihaelaStoica: Yes, that'd be good.

XenAdmin/Program.cs
Frezzle: If the value is manually changed to 2 then authSetting gets set to 2 and the 'else' clause below it will set the .NET authentication manager to use Digest, so at the moment any values that are not 1 sets XenCenter to use Digest. This matches up with the radio button display behaviour, which checks the Digest radio button for invalid values. We could either:
1. Leave it like this so they both match, which I prefer.
2. Have an if/elseif here which, if the setting is manually made invalid, will make the authentication manager not able to use any authentication method (which will disallow any connection that needs authentication), but then the setting in the connection options page will still display Digest as default.

XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
```diff
@@ -92,6 +87,19 @@ private void build()
             if (Registry.ProxyAuthenticationEnabled)
             {
                 AuthenticationCheckBox.Checked = Properties.Settings.Default.ProvideProxyAuthentication;
+                
+                switch ((HTTPHelper.ProxyAuthenticationMethod)Properties.Settings.Default.ProxyAuthenticationMethod)
+                {
+                    case HTTPHelper.ProxyAuthenticationMethod.Basic:
+                        BasicRadioButton.Checked = true;
+                        break;
+                    case HTTPHelper.ProxyAuthenticationMethod.Digest:
```
MihaelaStoica: "case HTTPHelper.ProxyAuthenticationMethod.Digest" is not really needed here, as we do the same thing for default case. Actually, I think it would be enough to do just:
`BasicRadioButton.Checked = (HTTPHelper.ProxyAuthenticationMethod)Properties.Settings.Default.ProxyAuthenticationMethod == HTTPHelper.ProxyAuthenticationMethod.Basic;`
because the other radio button will be checked by default if this is unchecked.

XenAdmin/Program.cs
MihaelaStoica: I was thinking that maybe the cast to HTTPHelper.ProxyAuthenticationMethod will fail.

MihaelaStoica: I am happy to default to Digest if the saved value is invalid, but I was thinking that maybe the cast to HTTPHelper.ProxyAuthenticationMethod will fail.

XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs
Frezzle: We decided to leave this as is to make any future extension/change easier (w.r.t. authentication schemes).

Frezzle:
@MihaelaStoica changes have been made!

MihaelaStoica:
Looks good now. Merging.

PR #1208, opened 2016-10-13
CA-224567: PVS Site renaming plus some other improvements

MihaelaStoica:
- Renamed Add/Remove PVS Site buttons
- Added a note on the PVS Cache configuration page, regarding the recommended cache size and the allowed size for "memory only"
- Added a warning about the Site configuration not being complete until the PVS part is done as well (only visible if PVS_uuid is not set for a PVS_site)
-  Added PVS Read Caching status on the VM's general tab
- The absolute maximum allowed for a cache size is 2TB
- Changed the rubric on the Enable PVS read caching dialog


kc284:
Just an article missing on the text of the info on the configuration dialog; otherwise it looks ok to me.

PR #1209, opened 2016-10-13
CA-225057: XenCenter throws an Exception when trying to copy somethin…

MihaelaStoica:
…g to the clipboard on the Network page


kc284:
This change is ok.
I had a look around and there are a few places where we set the Clipboard text but not always within a try-catch block. Since the change is fairly straightforward, could we audit the rest and ensure they're all fail safe? Perhaps we could make a reusable helper method to do this, since all we presumably want is log the exception.

MihaelaStoica:
Sure, I'll put it in a helper function. I wonder if our Clip class is a suitable place for such helper function?

kc284:
Looks ok now. Merging.

PR #1212, opened 2016-10-14
CA-157786: General tab for VDIs.

kc284:


XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -1309,38 +1308,51 @@ private void generateGeneralBox()
...
 
+            VDI vdi = xenObject as VDI;
+            if (vdi != null)
+            {
+                s.AddEntry(Messages.SIZE, vdi.SizeText,
+                    new PropertiesToolStripMenuItem(new VdiEditSizeLocationCommand(Program.MainWindow, xenObject)));
+
+                SR vdiSr = vdi.Connection.Resolve(vdi.SR);
+                if (vdiSr != null && !vdiSr.IsToolsSR)
+                    s.AddEntry(Messages.DATATYPE_STORAGE, vdiSr.Name);
```
MihaelaStoica: Shall we use sr.NameWithLocation here, to include the pool/host name (usefull if there are SRs with the same name in different pools)?

kc284: Actually the SR.Name does give the host information, but since you mentioned it I'll have a closer look to see how the one property is dfferent from the other, we seem to have too many properties dealing with names and locations...

MihaelaStoica: SR.Name gives the host information for Local storage only

MihaelaStoica:
Looks good. Merging.

PR #1214, opened 2016-10-17
CA-225708: Can not increase dom0 Memory multiple times

MihaelaStoica:
- take the already assigned dom0 memory into account when calculating maximum memory available


mcintyre94:
Looks good, merging

PR #1218, opened 2016-10-18
[CA-225062] For Ely and greater hosts, don't check for IO tools when activating/deactivating (this covers detatch + delete too in UI) a VBD

mcintyre94:
Instead we use the vbd.allowed_operations list, and check for the plug/unplug operation respectively.


MihaelaStoica:
Looks good. Merging.

PR #1220, opened 2016-10-19
CA-225731: XenCenter: Find the ISO update in the zip file defined in CFU

GaborApatiNagy:

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
@@ -199,7 +199,7 @@ namespace XenAdmin.Wizards.PatchingWizard
                         {
                             if (fileName.EndsWith(UpdateExtension.ToLowerInvariant()))
                                 SelectedUpdateType = UpdateType.NewRetail;
-                            else if (fileName.EndsWith(".iso"))
+                            else if (fileName.EndsWith("." + Branding.UpdateIso))
                                 SelectedUpdateType = UpdateType.ISO;
                             else
                                 SelectedUpdateType = UpdateType.Existing;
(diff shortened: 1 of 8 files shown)


XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -123,14 +123,22 @@ private void ExtractFile()
 
                     while (iterator.HasNext())
                     {
-                        if (Path.GetExtension(iterator.CurrentFileName()) == "." + updateFileExtension)
+                        string currentExtension = Path.GetExtension(iterator.CurrentFileName());
+
+                        if (!string.IsNullOrEmpty(updateFileExtension) && currentExtension == "." + updateFileExtension
+                            || currentExtension == ".iso" || currentExtension == ".xsupdate")
```
MihaelaStoica: We cannot use hardcode "xsupdate" here, we should use Branding.Update instead.
Also, I think it would be better to put ".iso" in a constant.

XenAdmin/Wizards/PatchingWizard/PlanActions/DownloadPatchPlanAction.cs
```diff
@@ -91,7 +91,7 @@ private void DownloadFile(ref Session session)
 
             bool isIso = patchUri.ToLowerInvariant().EndsWith("iso");
```
MihaelaStoica: this variable is not used anymore

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -199,7 +199,7 @@ public override void PageLeave(PageLoadedDirection direction, ref bool cancel)
                         {
                             if (fileName.EndsWith(UpdateExtension.ToLowerInvariant()))
                                 SelectedUpdateType = UpdateType.NewRetail;
-                            else if (fileName.EndsWith(".iso"))
+                            else if (fileName.EndsWith("." + Branding.UpdateIso))
```
MihaelaStoica: We need to check fileName.EndsWith(Branding.Update) on line 200

PR #1224, opened 2016-10-19
Some enhancements on the New Graph Dialog:

kc284:
- on certain occasions retrieving datasources can be slow; show a spinner while in progress.
- on certain systems the list of datasources may be long; added search box to aid the user filter them by name.


MihaelaStoica:
:+1:

PR #1227, opened 2016-10-20
CA-226112: Disable PVS read caching causes System.NullReferenceException

MihaelaStoica:

### DIFF
--- a/XenAdmin/Commands/DisablePvsReadCachingCommand.cs
+++ b/XenAdmin/Commands/DisablePvsReadCachingCommand.cs
@@ -66,11 +66,9 @@ namespace XenAdmin.Commands
 
             foreach (var vm in vms)
             {
-                var action = GetAsyncActionForVm(vm);
-                if (action != null)
-                {
-                    actions.Add(action);
-                }
+                var pvsProxy = vm.PvsProxy;
+                if (pvsProxy != null)
+                    actions.Add(new PvsProxyDestroyAction(pvsProxy));
             }
 
             if (actions.Any())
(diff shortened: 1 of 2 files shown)


XenAdmin/Commands/DisablePvsReadCachingCommand.cs
```diff
@@ -97,8 +97,7 @@ protected virtual void Execute(IList<VM> vms)
         /// <returns></returns>
         private AsyncAction GetAsyncActionForVm(VM vm)
         {
-            var pvsProxy = vm.Connection.Cache.PVS_proxies.FirstOrDefault(p => p.VM.Equals(vm)); // null if no proxy
-
+            var pvsProxy = vm.PvsProxy;
```
kc284: Can't we just write in line 67:
```
foreach (var vm in vms)
{
    var pvsProxy = vm.PvsProxy;
    if (pvsProxy != null)
        actions.Add(new PvsProxyDestroyAction(pvsProxy));
}
```
Then we don't need the GetAsyncActionForVm at all - simpler and we gain a few picoseconds by avoiding extra calls and checks.

XenAdmin/Dialogs/EnablePvsReadCachingDialog.cs
```diff
@@ -139,9 +139,7 @@ private AsyncAction GetAsyncActionForVm(VM vm, PVS_site siteSelected)
         /// <returns></returns>
         private bool PvsProxyAlreadyEnabled(VM vm)
```
kc284: Here too I don't think we need to keep this method any longer, we can just keep it simple and write in line 121:
```
if (vm.PvsProxy != null)
    return null;
```

MihaelaStoica:
Thanks @kc284, I've done the changes.

kc284:
Looks ok; merging.

PR #1228, opened 2016-10-20
CP-19303: XenCenter: Use new livepatching codes

GaborApatiNagy:


XenModel/Actions/SupplementalPack/UploadSupplementalPackAction.cs
```diff
@@ -189,6 +189,15 @@ private string UploadSupplementalPack(SR sr)
                     var poolUpdateRef = Pool_update.introduce(Connection.Session, vdiRef);
                     poolUpdate = Connection.WaitForCache(poolUpdateRef);
                 }
+                catch (Failure ex)
+                {
+                    if (ex.ErrorDescription != null && ex.ErrorDescription.Count > 1 && "UPDATE_ALREADY_EXISTS".Equals(ex.ErrorDescription[0], StringComparison.InvariantCultureIgnoreCase))
```
MihaelaStoica: I would prefer to use string.Equals() here, similar to the one 4 lines below

MihaelaStoica:
Apart from the "Equals" comment above, I'm happy with this PR.

GaborApatiNagy:
I've fixed it and as we've discussed I'm merging this myself now.

PR #1236, opened 2016-10-21
[CA-211267] Add a lightweight polling mechanism to the appliance tasks, to provide an accurate timer

mcintyre94:
For other actions we get a ticking timer in the Events view because they use PollToCompletion, which calls their Changed event (via Action.Tick) every 900ms. These actions don't use PollToCompletion, and didn't regularly call their Changed event - so their timer didn't tick regularly. This change adds a lightweight mechanism to call their changed event every 900ms until they complete. This allows the events page timer to tick every second in the same way as it does for other actions.

### DIFF
--- a/XenAdmin/Actions/OVFActions/ApplianceAction.cs
+++ b/XenAdmin/Actions/OVFActions/ApplianceAction.cs
@@ -46,6 +46,8 @@ namespace XenAdmin.Actions.OVFActions
 		protected string m_tvmGateway;
 		protected XenOvfTransportBase m_transportAction;
 
+        private const int SLEEP_TIME = 900;
+
 		/// <summary>
 		/// RBAC dependencies needed to import appliance/export an appliance/import disk image.
 		/// </summary>
@@ -101,17 +103,15 @@ namespace XenAdmin.Actions.OVFActions
 
 	    protected void InitialiseTicker()
 	    {
-	        System.Threading.Tasks.Task.Run(() => TickUntilCompletion(this));
+	        System.Threading.Tasks.Task.Run(() => TickUntilCompletion());
 	    }
 
-	    private void TickUntilCompletion(ApplianceAction action)
+	    private void TickUntilCompletion()
 	    {
-	        int sleepTime = 900;
-
-	        while (!action.IsCompleted)
+	        while (!IsCompleted)
 	        {
-	            action.OnChanged();
-                Thread.Sleep(sleepTime);
+	            OnChanged();
+                Thread.Sleep(SLEEP_TIME);
 	        }
 	    }
 
--- a/XenAdmin/Actions/OVFActions/ApplianceAction.cs
+++ b/XenAdmin/Actions/OVFActions/ApplianceAction.cs
@@ -47,6 +47,7 @@ namespace XenAdmin.Actions.OVFActions
 		protected XenOvfTransportBase m_transportAction;
 
         private const int SLEEP_TIME = 900;
+        private const int MAX_ITERATIONS = 60 * 60 * 24; // seconds in 24 hours
 
 		/// <summary>
 		/// RBAC dependencies needed to import appliance/export an appliance/import disk image.
@@ -114,7 +115,8 @@ namespace XenAdmin.Actions.OVFActions
 
 	    private void TickUntilCompletion()
 	    {
-	        while (!IsCompleted)
+	        int i = 0;
+	        while (!IsCompleted && ++i<MAX_ITERATIONS)
 	        {
 	            OnChanged();
                 Thread.Sleep(SLEEP_TIME);
--- a/XenAdmin/Actions/OVFActions/ApplianceAction.cs
+++ b/XenAdmin/Actions/OVFActions/ApplianceAction.cs
@@ -47,7 +47,7 @@ namespace XenAdmin.Actions.OVFActions
 		protected XenOvfTransportBase m_transportAction;
 
         private const int SLEEP_TIME = 900;
-        private const int MAX_ITERATIONS = 60 * 60 * 24; // seconds in 24 hours
+        private const int MAX_ITERATIONS = 60 * 60 * 24 / SLEEP_TIME * 1000; //iterations in 24h
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


Frezzle:
should this be CA-211267?

mcintyre94:
Okay, renamed the first commit and moved the sleep timer to const int in the class.

XenAdmin/Actions/OVFActions/ApplianceAction.cs
```diff
@@ -97,5 +100,21 @@ protected override void CancelRelatedTask()
...
 		}
+
+	    protected void InitialiseTicker()
+	    {
+	        System.Threading.Tasks.Task.Run(() => TickUntilCompletion());
+	    }
+
+	    private void TickUntilCompletion()
+	    {
+	        while (!IsCompleted)
```
kc284: Could we add a null check here? I'm a bit paranoid the action might be nullified on a different thread and this one can't notice.

XenAdmin/Actions/OVFActions/ExportApplianceAction.cs
```diff
@@ -92,6 +92,9 @@ protected override void Run()
 		{
 		    SafeToExit = false;
 			var session = Connection.Session;
+
+            InitialiseTicker();
```
kc284: Personally I'd rather the child classes didn't have to call the ticker, e.g. I would write in the parent class:
```
protected override void Run()
{
    SafeToExit = false;
    InitialiseTicker();
}
```
and in the child class
```
protected override void Run()
{
    base.Run();
    everythingelse();
}
```

stephen-turner:
Sorry, I'm confused why that last change is needed. "this" can never be null, can it?

kc284:
Apologies for the delay, I had to find the relevant article http://stackoverflow.com/questions/10625326/this-null-inside-net-instance-method-why-is-that-possible

mcintyre94:
I think something like:

var action = new ExportApplianceAction(...);
action.Run(); // starts timer thread
action = null;

Would cause a null dereference. I don't think there's anything we can do to stop the creator of an action suddenly making it null, and I don't think the spawned thread would be destroyed by that, so it'd keep trying to use the null action.

stephen-turner:
No, I don't think that's the case. My reading of that StackOverflow thread and others is that "this" can only be null if you use reflection, or fiddle with the IL. An ordinary call to a member function cannot have "this" as null, and the function will continue running with the same "this" until it exits: another thread changing its "action" variable to null doesn't suddenly change "this" in the function that's running. Otherwise, you'd have to keep testing for (this != null) before any (explicit or implicit) dereferencing of "this", including any read of a data member.

If you're not convinced, your example is easy to test.

GaborApatiNagy:
"action = null;" this will not make the instance null or will not remove it at all. It will only make 'action' to point to nowhere, the instance will still be in the memory until the garbage collector decides to clear it. I haven't tried, but I assume it won't, as the events are still subscribed in it.

stephen-turner:
What I'm more worried about in this code is, is the action guaranteed to reach the "IsCompleted" state eventually? If not, the timer will keep ticking, the action cannot be GCed, and we have a memory leak.

GaborApatiNagy:
I think if we had to start testing (this == null) everywhere, that would be a chaos :) and broken design somewhere...

stephen-turner:
See Eric Lippert's answer in http://stackoverflow.com/questions/19300245/is-it-possible-for-this-keyword-to-equal-null:

"So there are several ways in which this could be null in C#, but they are all very much out of the mainstream. It is therefore very rare for people to write code as your professor has. C# programmers idiomatically suppose that this is not null and never check for it."

GaborApatiNagy:
Just looking into this a bit more, I may just take a step back and ask why do we need to pass 'this' around at all?
Could we just have the following?

 private void InitialiseTicker()
+	    {
+	        System.Threading.Tasks.Task.Run(() => TickUntilCompletion());
+	    }
+
+	    private static  void TickUntilCompletion()
+	    {
+              #use this here
+         }

mcintyre94:
That's how it was one commit ago (https://github.com/xenserver/xenadmin/pull/1236/commits/15b47059428f61e8c29a724e5363cfb3572d92f3), I'm happy to go back to that if the null issue isn't a concern.

stephen-turner:
Yes, @GaborApatiNagy, that's exactly what the code was until the last changeset (7d7a040). I think the previous version was correct.

stephen-turner:
@mcintyre94 Yes, you can revert 7d7a040, but I would like to be sure there isn't a memory leak before merging it.

mcintyre94:
I didn't revert the whole commit, the rest of it is still worth having.

Some investigation on the reliabiltiy of Action.IsCompleted: I think it's okay if the events always run using the RunAsync() or RunExternal() methods. Those both use AsyncAction.RunWorkerThread(), which, by calling Action.MarkCompleted(), guarantees completion:

- At the end of Run()
- On the action throwing a CancelledException
- On any other caught Exception

The only case I can see where RunWorkerThread() won't mark an event as completed is where it has already been cancelled (RunWorkedThread just returns), but in this case the CancellingAction.Cancel() event will mark the action as completed instead. 

If the events are run using just their Run() method (which we don't currently for any of these events), then I don't think they will complete. Notably setting PercentComplete to 100 doesn't seem to trigger AsyncAction.IsCompleted, and they don't set IsCompleted themselves on exceptions.  I'm not sure if this is a concern though, because we don't and shouldn't run these synchronously.

stephen-turner:
I'm wondering if we should have a 24-hour timeout, just to guard against some code path we don't understand or might accidentally add later.

Admittedly, a VM import can take more than 24 hours, but even if it does, it's not too bad a bug to stop the counter then.

GaborApatiNagy:
This would do the same, but more meaningful

GaborApatiNagy:
The reason I've suggested this is just to make sure 24h (approx, I know) stays 24h even if someone changes SLEEP_TIME_MS in the future.

GaborApatiNagy:
:+1: Thanks, merging

PR #1240, opened 2016-10-26
CA-220475: [PAR-173] XenCenter fails when updating network MTU with V…

MihaelaStoica:
…LAN tag already in use error

- use the PIF of the master when trying to find out if an external network has changed


stephen-turner:
OK, that's a smaller change than I was expecting! Good!

PR #1241, opened 2016-10-26
[CA-220457] Add friendly names for new graphs, and filter duplicates

mcintyre94:
We have some new RRDs - read, write, read_latency and write_latency. This adds support for them by adding a friendly name and using it to display them. It also adds some filtering (see GraphHelpers.cs) because for backward compatibility io_throughput_read and io_throughput_write (which differ from read and write only in scale) are still available, but shouldn't be displayed in XC. If the new read or write are available, we filter these old datasources out.

### DIFF
--- a/XenAdmin/Controls/CustomDataGraph/GraphHelpers.cs
+++ b/XenAdmin/Controls/CustomDataGraph/GraphHelpers.cs
@@ -32,6 +32,7 @@
 using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Linq;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
 using XenAPI;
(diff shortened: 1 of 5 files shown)


XenAdmin/Controls/CustomDataGraph/GraphHelpers.cs
```diff
@@ -192,6 +193,16 @@ public static List<DataSourceItem> BuildList(IXenObject xenObject, List<Data_sou
                 dataSourceItems.Add(new DataSourceItem(dataSource, friendlyName, Palette.GetColour(itemUuid), itemUuid, xenObject));
             }
 
+            // Filter old datasources only if we have their replacement ones
+            Regex io_throughput_rw_regex = new Regex("^io_throughput_(read|write)_([a-f0-9]{8})$"); // old datasources
```
GaborApatiNagy: I think these Regex objects could be moved outside this function so it won't get created each time this is called.

GaborApatiNagy: Actually, it looks like they could be moved to XenModel/Utils/Helpers.cs

MihaelaStoica:
Looks good. Merging.

PR #1242, opened 2016-10-26
[CA-223486] Storage Link Status validation is wrongly included in Precheck During RPU

mcintyre94:
We now only include the StorageLink check if we have any hosts earlier than Creedence in the RPU.

### DIFF
--- a/XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
+++ b/XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
@@ -138,6 +138,12 @@ namespace XenAdmin.Wizards.RollingUpgradeWizard
             get { return "Upgradeprechecks"; }
         }
 
+        private static bool HostNeedsLicenseCheck(Host host)
+        {
+            var edition = Host.GetEdition(host.edition);
+            return edition != Host.Edition.EnterpriseXD && edition != Host.Edition.XenDesktop;
+        }
+
         protected override List<KeyValuePair<string, List<Check>>> GenerateChecks(Pool_patch patch)
         {
             List<KeyValuePair<string, List<Check>>> checks = new List<KeyValuePair<string, List<Check>>>();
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -197,12 +197,15 @@ public override string HelpID
...
-            foreach (Host host in SelectedServers)
+            //iSL (StorageLink) check - CA-223486: only for pre-Creedence
+            if (SelectedServers.Any(h => !Helpers.CreedenceOrGreater(h)))
             {
-                checkGroup.Add(new HostHasUnsupportedStorageLinkSRCheck(host));
+                checks.Add(new KeyValuePair<string, List<Check>>(Messages.CHECKING_STORAGELINK_STATUS, new List<Check>()));
+                checkGroup = checks[checks.Count - 1].Value;
+                foreach (Host host in SelectedServers)
+                {
+                    checkGroup.Add(new HostHasUnsupportedStorageLinkSRCheck(host));
```
kc284: I believe the check needs to be added on the individual servers too, because in case we are upgrading more than one pools the check will still run on all of them including the post-creedence ones.
I think the next check (clearwater) is similarly afflicted.
One can check the above by deselecting the hide successful checkbox at the bottom of the page so the checks on the individual servers are shown in detail.

mcintyre94: Agreed, that's a good idea - I'll update both.

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -216,7 +220,8 @@ public override string HelpID
                     checkGroup = checks[checks.Count - 1].Value;
                     foreach (Host host in SelectedServers)
                     {
-                        checkGroup.Add(new UpgradingFromTampaAndOlderCheck(host));
+                        if(!Helpers.ClearwaterOrGreater(host))
```
kc284: Missing the licence check here; the version check applies also in lines 231 and 239

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -197,45 +203,47 @@ public override string HelpID
...
-                if (SelectedServers.Any(h => Host.GetEdition(h.edition) != Host.Edition.EnterpriseXD && 
-                                             Host.GetEdition(h.edition) != Host.Edition.XenDesktop))
                 {
                     checks.Add(new KeyValuePair<string, List<Check>>(Messages.CHECKING_LICENSING_STATUS, new List<Check>()));
                     checkGroup = checks[checks.Count - 1].Value;
-                    foreach (Host host in SelectedServers)
+                    foreach (Host host in preClearwaterServers)
                     {
-                        checkGroup.Add(new UpgradingFromTampaAndOlderCheck(host));
+                        if(HostNeedsLicenseCheck(host))
```
MihaelaStoica: Here we will end up with an empty group if none of the preClearwaterServers needs the license check.
I think it is OK to have a check for every server in preClearwaterServers, even if it would say OK for some of them (i.e. EnterpriseXD  and XenDesktop)

kc284: The HostNeedsLicenseCheck needs to be left in line 222 too.

kc284:
Look better now. Unless @MihaelaStoica has any remarks I'll merge it.

MihaelaStoica:
Looks OK to me.

PR #1245, opened 2016-10-27
CA-226817: Do not allow OVF/OVA import on a host incompatible with th…

kc284:
…e VM's hardware

platform version. Also, preserve between export and import the VM's has_vendor_device settings.


mcintyre94:
This looks good, discussed with Tina and tested. Merging.

PR #1247, opened 2016-10-27
[CA-207601] License expiry time in general tab is incorrect

mcintyre94:
The problem here was that TimeSpan.Minutes is always below 60, eg. for 67 minutes it is 7 (with TimeSpan.Hours being 1 then). So our TimeSpan was correct, but if it was between 1 and 2 hours then we would write the minutes in the second hour (which is below 60), ignoring the hour. The fix is to use s.TotalMinutes which for 67 minutes is 67. Hours also had a similar issue, except it would be off by 24 hours for spans between 1 and 2 days.


MihaelaStoica:
Looks good. Merging.

PR #1248, opened 2016-10-27
[CA-203752] New VM from Snapshot doesn't use default host

mcintyre94:
The reported bug was that the new VM would crash because when clicking through the wizard the pool master is selected by default, not the slave the VM snapshot is on. We already select a default host when a new VM is created from a selected host using HostAncestor, but this property is null for a template (which isn't in the server/VM tree). The fix is that if the selection is a template (which it is when we use new VM from Snapshot) then we instead use its host (using Home()) as the default host. Now the correct host is selected in the VM wizard and the VM creation succeeds.

### DIFF
--- a/XenAdmin/Commands/NewVMCommand.cs
+++ b/XenAdmin/Commands/NewVMCommand.cs
@@ -83,7 +83,19 @@ namespace XenAdmin.Commands
                 }
             }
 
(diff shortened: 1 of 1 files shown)


MihaelaStoica:
Looks good. Merging.

PR #1249, opened 2016-10-27
Avoid KeyNotFoundException when moving VMs with unresolved VDIs

kc284:
A VM can find itself with a VBD with an unresolved VDI (e.g. ejected CD). The VDI will not have been included in the storage mapping and the user will be presented with a KeyNotFoundException message even if the move of the other VDIs has succeeded. To fix this ignore unresolved VDIs first, then check 
if the storage mapping contains the resolved ones and then try getting the target SR from the mapping.


MihaelaStoica:
Looks good. Merging.

PR #1251, opened 2016-10-28
[CA-226116] Add Windows update as a hidden feature

mcintyre94:
If windows_update is in the hidden features registry key, then we hide the virtualization state line about Windows update.


MihaelaStoica:
Looks good. Merging.

PR #1254, opened 2016-11-01
[CA-227602] Truncation on the Install Mode page in the Install Upgrade wizard

mcintyre94:
I've looked at how this was implemented in Dundee, and there we used an absolute height on the row containing the manual checkbox. I've done the same here, changing row 3 from autosize to absoliute 47px. I've also removed the empty row that was between the two radio buttons. This is a Winforms problem, without workarounds an autosize radio button in an autosize container (say the table row) will report its height as one row of text, no matter whether it needs more. This seems to be the only place we use a long radio label (as opposed to a separate label underneath the radio), so I doubt it's worth doing anything more complex than this. I've tested it with the CN string used in the ticket, and it uses less space than the English translation. The string also hasn't changed from Dundee.


MihaelaStoica:
This change looks fine, but trying this out I've noticed a some extra padding at the bottom of the page, just below "Allow XenCenter to delete..." checkbox

PR #1257, opened 2016-11-02
CA-213563: VM policy cannot be changed to snapshot with quiesce after installing vss tools

geosharath:
Enabling snapshot policy type to be changed to quiesce from other snapshot policy types if the selected VMs are quiesce snapshot capable.

Signed-off-by: Sharath Babu <sharath.babu@citrix.com>

### DIFF
--- a/XenAdmin/Dialogs/PropertiesDialog.cs
+++ b/XenAdmin/Dialogs/PropertiesDialog.cs
@@ -465,6 +465,14 @@ namespace XenAdmin.Dialogs
 
         private void verticalTabs_SelectedIndexChanged(object sender, EventArgs e)
         {
+            
+            var snapshotTypePageSpecific = verticalTabs.SelectedItem as NewPolicySnapshotTypePageSpecific<VMSS>;
+            if (snapshotTypePageSpecific != null)
+            {
+                newPolicyVMSSTypePage1.ToggleQuiesceCheckBox(newVMSSVMsPage1.SelectedVMs);
+                return;
+            }
+
             var selectedPage = verticalTabs.SelectedItem as NewPolicyArchivePage;
             if (selectedPage != null)
             {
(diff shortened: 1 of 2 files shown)


XenAdmin/Dialogs/PropertiesDialog.cs
```diff
@@ -465,6 +465,41 @@ protected virtual void OnFormClosing(PropertiesDialogClosingEventArgs e)
 
         private void verticalTabs_SelectedIndexChanged(object sender, EventArgs e)
         {
+            
+            var snapshotTypePageSpecific = verticalTabs.SelectedItem as NewPolicySnapshotTypePageSpecific<VMSS>;
+            if (snapshotTypePageSpecific != null)
+            {
+              
+                switch (newPolicyVMSSTypePage1.BackupType)
```
MihaelaStoica: I suggest to move all this logic into the NewPolicySnapshotTypePageSpecific class to keep the specific implementation outside of the PropertiesDialog. For example:
`snapshotTypePageSpecific.ToggleQuiesceCheckBox(newVMSSVMsPage1.SelectedVMs);`

XenAdmin/Dialogs/PropertiesDialog.cs
```diff
@@ -465,6 +465,41 @@ protected virtual void OnFormClosing(PropertiesDialogClosingEventArgs e)
...
+                case policy_backup_type.snapshot_with_quiesce:
+                       newPolicyVMSSTypePage1.ToggleQuiesceCheckBox(true, true);
+                       break;
+                
+                case policy_backup_type.checkpoint:
+                       newPolicyVMSSTypePage1.ToggleQuiesceCheckBox(true, false);
+                       break;
+                }
+
+                if (newVMSSVMsPage1.SelectedVMs.Count > 0)
```
MihaelaStoica: This check is not necessary here, as the foreach can cope with Count=0

geosharath:
@MihaelaStoica  I have updated the PR.

PR #1258, opened 2016-11-02
CA-227274: Fix CN and JP version process lable size

cheng-z:
This change is to change CN and JP process_lable height same as EN version


cheng-z:
Hi @MihaelaStoica 
Could you please help to review the code and let me know if any concern?
Thanks.

PR #1259, opened 2016-11-02
Changes for CP-18097 and some removal of obsolete code.

kc284:


stephen-turner:
The spell checker is required, please replace it. It has caught many errors in the past.

GaborApatiNagy:
I agree, the spellchecker has been useful quite a few times.
Other code looks good

kc284:
Hmmmm, there are many more errors the spellchecker hasn't caught because it's not possible to fill the list with the infinite misspellings of the english language, which is why I don't think it's worth having it. This commit was not a sudden decision but rather something we have discussed before (long time ago); in any case I've replaced it, but it may be be worth having a look for a proper spell checker at some point - if it could be implemented as a PR hook it would be ideal.

stephen-turner:
The spell checker checks all the text, including identifiers, which obviously should be allowed to be non-words: but I don't know how to do anything better without completely parsing the code.

Furthermore, it also needs to catch British spellings and antiquated language that are linguistically correct but not suitable for XenCenter.

I would be happy to have a proper spell checker on all text, but it needs to solve these two problems.

stephen-turner:
And if you find misspellings it hasn't caught, please add them to the "dictionary"! It seems that most errors come up repeatedly.

kc284:
The second problem is not a problem because spellcheckers normally work against a dictionary, in our case we would point to an en-US dictionary, so british spelling would be caught (and modern computer dictionaries are normally stripped of old fashioned words too).

One should be able to specify filetypes to check - we're mainly interested in resx rather than other code, so not all of the code needs to be parsed. Some spellcheckers (e.g. some used by IDEs, won't mention names here) are pretty clever with identifiers too, which can be a good thing, e.g. they like myVariable, but not myvariable or myVaraible.

Ok, this can be investigated. For the time being, could we merge this PR if there are no other corrections needed?

PR #1260, opened 2016-11-02
[CA-225715] Performance Graph reporting for CPU for a VM based on max VCPUs

mcintyre94:
(not current vcpus)

Instead of using VCPUs_max, use VCPUs_at_startup, which is the value set when we change the VCPUs count on the VM. Note that despite its name we don't need to restart to see this value change.


stephen-turner:
What happens if the current number of CPUs is changed?

mcintyre94:
Replying here so we have it on the commit: VCPUs_at_startup changes immediately when the VCPU count is changed, and will be correct on the next load of the graph page.

MihaelaStoica:
Looks good. Merging.

PR #1262, opened 2016-11-03
CA-226103: Hotkey conflict in the “Choose the type of new storage” UI…

MihaelaStoica:
… while creating new storage

Also changed the page to use different resource strings for the Storage type in the right hand panel (no hotkeys should be shown there)


mcintyre94:
Looks good, merging

PR #1263, opened 2016-11-03
CA-147722: Truncation issue occurred on CPU & Memory window when VM m…

MihaelaStoica:
…emory more than physical memory.

Also aligning the warning icon with the label.


mcintyre94:
Is the resx change here intentional? It looks fine either way, but just to check.

mcintyre94:
This looks good, merging

PR #1264, opened 2016-11-03
[CA-226897] Resume on Server option won't automatically resume the VM

mcintyre94:
The resume on server uses the CrossPoolMigrateCommand to implement the migration of the VM to the desired host. Before this change that command had no support for resuming a VM after migrating it, so the VM was not resumed. With this change, the CrossPoolMigrateWizard can take an optional (default false) parameter to restart the VM after migration. When this is true, the migrate action becomes a MultipleAction, first migrating and then restarting the VM.

### DIFF
--- a/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateWizard.cs
+++ b/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateWizard.cs
@@ -247,6 +247,8 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard
                 else
                 {
                     var isCopy = wizardMode == WizardMode.Copy;
+                    var migrateAction = 
+                        new VMCrossPoolMigrateAction(vm, target, SelectedTransferNetwork, pair.Value, isCopy);
 
                     if (_resumeAfterMigrate)
                     {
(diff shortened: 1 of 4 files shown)


XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateWizard.cs
```diff
@@ -239,8 +244,30 @@ protected override void FinishWizard()
...
+
+                    if (_resumeAfterMigrate)
+                    {
+                        var title = VMCrossPoolMigrateAction.GetTitle(vm, target, isCopy);
+                        var startDescription = isCopy ? Messages.ACTION_VM_COPYING: Messages.ACTION_VM_MIGRATING;
+                        var endDescription = isCopy ? Messages.ACTION_VM_COPIED: Messages.ACTION_VM_MIGRATED;
+
+                        var actions = new List<AsyncAction>()
+                        {
+                            new VMCrossPoolMigrateAction(vm, target, SelectedTransferNetwork, pair.Value, isCopy),
```
MihaelaStoica: I suggest to put the VMCrossPoolMigrateAction into a variable outside the `if (_resumeAfterMigrate)` block and use this on both cases (as it is essentially the same action).

XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateWizard.cs
```diff
@@ -239,8 +244,30 @@ protected override void FinishWizard()
...
+                            new ResumeAndStartVMsAction(vm.Connection, target, new List<VM>{vm}, new List<VM>(), null, null)
+                        };
+
+                        new MultipleAction(vm.Connection, title, startDescription, endDescription,
+                            actions, false, false, true).RunAsync();
```
MihaelaStoica: We should probably use the MultipleAction with suppressHistory parameter set to true, so that the multiple action does not appear on the Events page, only the subactions. In this case the title, startDescription and stopDescription are not that important, but it is still good to have them properly set.

PR #1266, opened 2016-11-04
[CA-227845] Fixing problems caused by tooltip container on hotfix wizard mode page

mcintyre94:
Due to the tooltip container introduced for the automatic radio button, winforms didn't treat these radio buttons properly - we needed code to make them mutually exclusive, tab didn't behave properly and keyboard arrow navigation between them didn't work. Therefore I have removed the tooltip container (so our radio buttons behave properly without any extra work) and we instead use a winforms tooltip shown when the mouse is over the automatic radio button. Note that since the control is disabled when we want a tooltip, neither hover events nor attaching a winforms tooltip in the usual way will work. Therefore I use a mousemove event on the table, which displays the tooltip if and only if we are over the disabled control. 

Also adjusted the tab indexes to make sure tab works correctly across the whole page.


MihaelaStoica:
Looks good. Merging.

PR #1267, opened 2016-11-07
CA-227601: check SR supports db replication on a background thread.

kc284:


mcintyre94:
Looks good :) Merging

PR #1269, opened 2016-11-07
[CA-228051] XenCenter allows user to create unusable PVS Proxy cache config

mcintyre94:
We now disable sites that aren't configured in the enable read caching dialog, using an EnableableComboBox. I've also changed the DrawMode of that ComboBox component, OwnerDrawFixed more appropriately sizes the dropdown container when the number of items is small.

### DIFF
--- a/XenModel/XenAPI-Extensions/PVS_site.cs
+++ b/XenModel/XenAPI-Extensions/PVS_site.cs
@@ -30,6 +30,7 @@
 
 using System;
 using System.Linq;
+using XenAdmin;
 
 namespace XenAPI
 {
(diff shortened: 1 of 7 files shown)


MihaelaStoica:
As discussed on the ticket, we might need to do more changes, so I am marking this PR as "needs updating" for now

XenModel/XenAPI-Extensions/PVS_site.cs
```diff
@@ -54,6 +55,39 @@ public PVS_cache_storage PvsCacheStorage(Host host)
                 pvsCacheStorage.site.opaque_ref == opaque_ref && pvsCacheStorage.host.opaque_ref == host.opaque_ref);
         }
 
+        public string NameWithWarning
+        {
+            get
+            {
+                if (!IsCacheConfigured())
+                {
+                    return string.Format(Messages.NEWVM_DEFAULTNAME, Name, Messages.PVS_CACHE_NOT_CONFIGURED);
```
MihaelaStoica: I think it would be better to not use NEWVM_DEFAULTNAME, as this is generally used to default a new object name and may change in the feature.

MihaelaStoica:
Looks good now. Merging

PR #1270, opened 2016-11-08
CA-227034: Slipstream packaged Live patching hotfix is rebooting host

MihaelaStoica:
This happened because the LivePatchCodesByHost was null when the prechecks were generated for an update, so the live patch codes returned by the precheck were ignored.


GaborApatiNagy:
Cool

PR #1271, opened 2016-11-08
[CA-228723]  Cancelling update installation leaves the update in xapi db

mcintyre94:
Modify the cleanup pool update action to also call update destroy (removing the update from xapi db) when the update has not been applied to any hosts (ie we cancelled the wizard before completion).

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard.cs
@@ -395,7 +395,7 @@ namespace XenAdmin.Wizards.PatchingWizard
                     {
                         Pool_update.pool_clean(session, poolUpdate.opaque_ref);
                         if(!poolUpdate.AppliedOnHosts.Any())
-                            Pool_update.async_destroy(session, poolUpdate.opaque_ref);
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard.cs
```diff
@@ -380,16 +380,28 @@ private List<AsyncAction> GetCleanUpPoolUpdateActions()
...
         {
             return
-                new DelegatedAsyncAction(poolUpdate.Connection, Messages.REMOVE_PATCH, "", "", session => Pool_update.pool_clean(session, poolUpdate.opaque_ref));
+                new DelegatedAsyncAction(poolUpdate.Connection, Messages.REMOVE_PATCH, "", "", session =>
+                {
+                    try
+                    {
+                        Pool_update.pool_clean(session, poolUpdate.opaque_ref);
+                        if(!poolUpdate.AppliedOnHosts.Any())
+                            Pool_update.async_destroy(session, poolUpdate.opaque_ref);
```
MihaelaStoica: If we use async api calls, then we need to call PollToCompletion() (as well as assigning RelatedTask) so that the action does not get marked as completed before the api task is actually finished.

PR #1272, opened 2016-11-08
[CA-228729] Improve the message shown on finishing the update installation

mcintyre94:
Previous message was Remove/Removing/Removed Updates, which is confusing. Changed to Clean/Cleaning/Cleaned Update Files.


MihaelaStoica:
Looks good. Merging.

PR #1276, opened 2016-11-10
CA-192747: Default name for the performance graphs

MihaelaStoica:


mcintyre94:
Looks good, merging :)

PR #1280, opened 2016-11-10
[CA-149812] EN: Duplicate hotkeys about “Dismiss All” and “Dismiss Selected" tabs.

mcintyre94:
Moved to A for Dismiss All, and E for Dismiss Selected. Since this frees D (the toolstrip menu split button uses one of these, so was previously always D), moved Filter by Date from A to D.


MihaelaStoica:
This looks good now. The only thing that I had noticed was that on the Updates page, the "Dismiss" item is selected by default when you first go to that page and it does not actually do anything (plus it duplicates the D hotkey). On the other pages "Dismiss All" is the default item.

MihaelaStoica:
Merging.

PR #1281, opened 2016-11-10
CA-229019 and CA-229009

GaborApatiNagy:


XenModel/Actions/SupplementalPack/UploadSupplementalPackAction.cs
```diff
@@ -188,6 +188,8 @@ private string UploadSupplementalPack(SR sr)
                 {
                     var poolUpdateRef = Pool_update.introduce(Connection.Session, vdiRef);
                     poolUpdate = Connection.WaitForCache(poolUpdateRef);
+
```
MihaelaStoica: I think the action should throw an exception if the poolUpdate is null (an update hasn't been retrieved from the cache)

MihaelaStoica:
Looks good now.

PR #1286, opened 2016-11-11
[CA-227813] Fixing some messages

mcintyre94:

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -12784,7 +12784,7 @@ A {1} user cannot alter the Workload Balancing settings.</value>
     <value>One or more hosts in the pool do not have an appropriate license.</value>
   </data>
   <data name="WLB_ERROR_4021" xml:space="preserve">
-    <value>WLB could not connect to [XenServer] because the supplied credentials were invalid.</value>
+    <value>WLB could not connect to the server because the supplied credentials were invalid.</value>
   </data>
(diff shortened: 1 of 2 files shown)


XenModel/Messages.resx
```diff
@@ -12784,7 +12784,7 @@ A {1} user cannot alter the Workload Balancing settings.</value>
     <value>One or more hosts in the pool do not have an appropriate license.</value>
   </data>
   <data name="WLB_ERROR_4021" xml:space="preserve">
-    <value>WLB could not start a [XenServer] session with supplied invalid credentials.</value>
+    <value>WLB could not connect to [XenServer] because the supplied credentials were invalid.</value>
```
kc284: The original implies "could not open a [XenServer] session to the server", so I believe this should be "could not connect to the server".

kc284: Hmm, ok, maybe, but then it should probably be "to the [XenServer] host", just "connect to [XenServer]" sounds a bit weird to me.

MihaelaStoica: We usually try not to say "host" in the UI, we prefer "server". Although in this case "XenServer server" is not ideal :). If it makes the message clearer, then I'm OK with "XenServer host"

stephen-turner: I'm equally happy with either "connect to XenServer" or "connect to the XenServer host".

MihaelaStoica:
Looks good. Merging.

PR #1287, opened 2016-11-13
CA-145890: added new Page to the options dialog where the user can co…

kc284:
…nfigure how

confirmations are displayed. Added checkboxes to the dialogs requesting confirmation
of dismissal of notifications so the user can opt to skip this step.


kc284:
Apart from checking the code, please also test whether the behaviour is the desired and also easy to grasp.

kc284:
Thanks @MihaelaStoica , I've updated the test.

PR #1288, opened 2016-11-14
[CA-229350] PVS allows Memory Only even when no control domain memory available

mcintyre94:
We now disable the Memory Only option if the available dom0 memory is less than the minimum the spinner supports (1Gb)

### DIFF
--- a/XenAdmin/Controls/PvsCacheStorageRow.cs
+++ b/XenAdmin/Controls/PvsCacheStorageRow.cs
@@ -98,7 +98,7 @@ namespace XenAdmin.Controls
                 };
             }
 
-            var enabled = Host.dom0_memory_extra >= numericUpDownCacheSize.Minimum;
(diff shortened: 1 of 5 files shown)


stephen-turner:
I think you mean "no memory available" rather than "no control domain available". Hopefully you have a control domain!

XenAdmin/Controls/PvsCacheStorageRow.cs
```diff
@@ -95,7 +97,10 @@ private void PopulateCacheSrCombobox()
                     opaque_ref = Helper.NullOpaqueRef
                 };
             }
-            var memorySrItem = new ToStringWrapper<SR>(memorySr, Messages.PVS_CACHE_MEMORY_ONLY);
+
+            var enabled = Host.dom0_memory_extra >= numericUpDownCacheSize.Minimum;
```
MihaelaStoica: Also, we could use the MIN_CACHE_SIZE_GB constant instead of numericUpDownCacheSize.Minimum

MihaelaStoica:
Looks good now.

PR #1293, opened 2016-11-16
CA-230461: NullReferenceException during batch update when no check-for-updates has been performed

GaborApatiNagy:
Much better way to update/download the CFU

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
@@ -193,7 +193,7 @@ namespace XenAdmin.Wizards.PatchingWizard
                 }
 
                 //check updgrade sequences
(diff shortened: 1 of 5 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -195,7 +193,7 @@ private void EnabledRow(Host host, UpdateType type, int index)
                 }
 
                 //check updgrade sequences
-                Updates.UpgradeSequence us = Updates.GetUpgradeSequence(host.Connection, AutoDownloadedXenServerVersions);
+                Updates.UpgradeSequence us = Updates.GetUpgradeSequence(host.Connection, Updates.XenServerVersions);
```
MihaelaStoica: I think we should use Updates.GetUpgradeSequence(conn) here, as we do in the other 2 places in the wizard. And we can even get rid of the GetUpgradeSequence(conn, xsVersions) function as it won't be used anymore.

MihaelaStoica:
Looks good. Merging.

PR #1300, opened 2016-11-18
Get the git branch from the environment variable and do not fall back…

kc284:
… to trunk

if it can't be found.Use GIT_LOCAL_BRANCH instead of GIT_BRANCH to retrieve the
branch name without the remote.


kc284:
@GaborApatiNagy another PR to review in arrears.

PR #1301, opened 2016-11-18
Restored falling back to trunk for xenadmin until we switch the brand…

kc284:
…ing repos to a different internal location.


MihaelaStoica:
Looks good. Merging.

PR #1303, opened 2016-11-21
CA-104329: Removed calculation which might cause lag from the page lo…

kc284:
…ading step;

have the wizard do it before we get there. Also, replaced the labels in such way
that, if there should be lag, the user will not see the draft message used only
on the designer.

### DIFF
--- a/XenAdmin/Wizards/HAWizard.cs
+++ b/XenAdmin/Wizards/HAWizard.cs
@@ -31,13 +31,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.Windows.Forms;
-
 using XenAdmin.Actions;
 using XenAdmin.Controls;
-using XenAdmin.Core;
 using XenAdmin.Dialogs;
-using XenAdmin.Network;
 using XenAdmin.Wizards.HAWizard_Pages;
 using XenAPI;
 using System.Drawing;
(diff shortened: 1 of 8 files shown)


GaborApatiNagy:
Hi Tina, it looks good - could I see a very quick demo later today?

XenAdmin/Wizards/HAWizard.cs
```diff
@@ -108,7 +104,32 @@ protected override void UpdateWizardContent(XenTabPage senderPage)
             {
                 xenTabPageHaFinish.HeartbeatSrName = xenTabPageChooseSR.SelectedHeartbeatSR.Name;
                 xenTabPageHaFinish.Ntol = xenTabPageAssignPriorities.Ntol;
-                xenTabPageHaFinish.RestartPriorities = xenTabPageAssignPriorities.getCurrentSettings().Values;
+
+                int alwaysRestartHighPriority = 0, alwaysRestart = 0, bestEffort = 0, doNotRestart = 0;
+                foreach (VM.HA_Restart_Priority priority in xenTabPageAssignPriorities.getCurrentSettings().Values)
```
GaborApatiNagy: Thanks for demoing @kc284.

I think "xenTabPageAssignPriorities.getCurrentSettings()" could take long, because that method iterates through the datagridview from the previous page (and casts each row).

I would suggest we change this to have the previous page to calculate this on page leave and then from this page we could just access the result Dictionary what could be just a field in that class. That way the previous page will wait a bit before leaving as it can have a large grid with a lot of data - then this page will load immediately.

kc284: Had another look at the wizard workflow: UpdateWizardContent is indeed called in PageLeave, so it seems to me the alternative way to do this is calculate and set the field each time the settings change on the page, so when we leave the page we'll only have to get that value, without any calculation at all.

GaborApatiNagy:
We had a quick conversation with @kc284, so merging it as it does work nicely. Thanks

PR #1306, opened 2016-11-23
CA-229009: VDI is not deleted for the updates

GaborApatiNagy:
If batch updating fails in a pool, this code will do the clean up by running the appropriate (next in the queue) RemoveUpdateFilesFromMaster action. In addition to Pool_update.clean(), Pool_update.destroy() will be called when needed.


MihaelaStoica:
Looks good. Merging

PR #1309, opened 2016-11-24
CP-17276: Out of Memory exception when XenCenter exceeds the maximum …

MihaelaStoica:
…number of GDI handles

Changed the tree builder so that the tree nodes rely on the TreeView's font. If the nodes have their NodeFont != null, then a tree update causes new fonts to be generated for each node and when fonts are being created faster than they could be cleaned up, we reach the GLI handles limit of 10000 (which causes the Out of Memory exception).


stephen-turner:
Such a simple change!!!

PR #1310, opened 2016-11-24
[CA-232963] XC hangs after restarting toolstack while viewing PVS tab

mcintyre94:
Changed the collection handlers to use ProgramInvokeHandler in the same way as other tabs, instead of Program.Invoke which caused the hang.


MihaelaStoica:
Looks good. Merging.

PR #1311, opened 2016-11-24
CA-232294: Can't connect if an Add Server dialog has been cancelled before

GaborApatiNagy:
This commit is to fix the following: "when right clicking a disconnected pool the context menu gives you the option to connect, this opens a window to enter credentials. if you close this window (either by cancel or the X), next time when you right click and click connect, nothing happens. you have to restart XenCenter to connect to the pool"


MihaelaStoica:
Looks ok to me. Merging.

PR #1314, opened 2016-11-25
[CA-233454] PVS tab doesn't show a new VM when it's created

mcintyre94:
Refined the rules for not adding a VM to the table, if it is a template (thus not_a_real_vm), and it has the __gui__ prefix (thus hidden), we still add it, but hide it.
When its name is changed (to remove the __gui__ prefix), we update its name and re-calculate whether it should be visible (in the case of a new VM this will be true once the __gui__ prefix is gone). Also resort the table if a node changes from hidden to visible, because it appears as an addition to the table.

### DIFF
--- a/XenModel/Actions/VM/CreateVMAction.cs
+++ b/XenModel/Actions/VM/CreateVMAction.cs
@@ -690,7 +690,7 @@ namespace XenAdmin.Actions.VMActions
         {
             get
             {
-                return string.Format("{0}{1}", Helpers.GuiTempObjectPrefix, NameLabel);
+                return Helpers.MakeHiddenName(NameLabel);
             }
         }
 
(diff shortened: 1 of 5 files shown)


XenModel/Actions/VM/CreateVMAction.cs
```diff
@@ -210,6 +210,7 @@ protected override void Run()
             AddDisks();
             AddNetworks();
             XenAdminConfigManager.Provider.ShowObject(VM.opaque_ref);
+            VM.IsBeingCreated = false;
```
MihaelaStoica: There is another action, CreateVMFastAction, that is using the same mechanism of hiding the VM while it's being created. We should set the IsBeingCreated property there as well.

PR #1321, opened 2016-12-01
CA-234628: Errors with no title in XenCenter

MihaelaStoica:
Set the suppressHistory flag to true for the parallel action that we run periodically to check the Health Check analysis result


GaborApatiNagy:
Looks fine

PR #1328, opened 2016-12-06
CA-207011: Don't allow XC to create more than one SMB ISO SR on the same share

GaborApatiNagy:
This commit implements the same logic that is used for NFS ISOs.


XenAdmin/Wizards/NewSRWizard_Pages/Frontends/CIFS_ISO.cs
```diff
@@ -165,9 +166,22 @@ public string SrDescription
 
         private void textBoxCifsSharename_TextChanged(object sender, EventArgs e)
         {
+            passwordFailure1.PerformCheck(IsSRNameUnique);
+
             OnPageUpdated();
         }
 
+        private bool IsSRNameUnique(out string error)
```
MihaelaStoica: Could we please change this method name to a more relevant name? Because this method actually checks if the iso is already attached , not if the name is unique.

GaborApatiNagy:
Hi @MihaelaStoica - I've fixed the naming problem you noticed. Thanks

PR #1329, opened 2016-12-06
[CP-19888] Add support for code coverage in build jobs

mcintyre94:
Add support for a new environment variable REPORT_COVERAGE, which causes the tests phase to use the new script tests-checks-cover instead of tests-checks.This new script wraps the tests-checks script, by instrumenting NCover to monitor before it runs, and writing a summary log and generating a HTML report of coverage in the workspace output directory.


mcintyre94:
Just noticed there's a conflict here since I first made this PR, will leave open and get back to it.

mk/build.sh
```diff
@@ -72,14 +73,57 @@ XENADMIN_DIR="$( cd -P "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 source ${XENADMIN_DIR}/mk/declarations.sh
 
+if test -z "${XC_BRANDING}"; then XC_BRANDING=citrix; fi
```
kc284: This PR restores here code that was recently removed or moved to different files. Please compare with the current state of this file on master and remove what's not there.

mcintyre94: Oops. I badly misread that merge conflict - I'll revert that commit and do a more careful merge.

mk/tests-checks-cover.sh
```diff
@@ -0,0 +1,56 @@
...
+
+set -u
+
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/declarations.sh"
+
+net start ncover || true # May fail if ncover already running, so ignore that error
+
+echo "Running + monitoring coverage of tests..."
+
+export NCOVER_USING_PROJECT="${NCOVER_PROJECT_ID}"
```
kc284: Where are these two variables used?

mk/tests-checks-cover.sh
```diff
@@ -0,0 +1,56 @@
...
+# WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+# NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+# SUCH DAMAGE.
+
+set -u
+
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/declarations.sh"
+
+net start ncover || true # May fail if ncover already running, so ignore that error
```
kc284: Do we want to ignore it if already running or kill it?

mk/tests-checks-cover.sh
```diff
@@ -0,0 +1,56 @@
...
+
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/tests-checks.sh"
+
+ncover summarize --project="XC Tests" --wait
+
+echo "Generating coverage report..."
+ncover report --project="XC Tests" --execution="${BUILD_ID}" --file="$(cygpath -d ${TEST_DIR})\coverage.html" --detail=method
+cp ${TEST_DIR}/coverage.html ${OUTPUT_DIR}
+
+net stop ncover
```
kc284: Same question as above: should we check if it hasn't stopped and kill it if not? Or is is not an issue?

mk/tests-checks-cover.sh
mcintyre94: I don't think it really matters, the most likely cause is that someone started the service manually and forgot to stop it. At a guess it might be safer to try to stop it (ignoring the error if that fails because of already being stopped) so we have a freshly started service though.

mk/tests-checks-cover.sh
mcintyre94: NCover uses them to know where to store the code coverage information. It doesn't have an explicit "start recording", it logs all .dlls that match the project while it's running. So these are set, then tests are run, and when they do they're recorded against that project ID and build ID. See https://www.ncover.com/support/docs/extras/command-line/ncover-use

mk/tests-checks-cover.sh
mcintyre94: I think if the service had stopped this would be a build error and we should ignore it. ncover summarize and ncover report will require the service, so if those didn't error then it must've been stopped somewhere that doesn't matter. I'll add a || true here to ignore this one.

PR #1333, opened 2016-12-07
CA-235381: XenCenter should not allow users to edit the PVS cache sto…

MihaelaStoica:
…rage if it is in use

- On the PVS Cache Configuration dialog, disable the cache storage fields with an information tooltip stating that this PVS cache storage cannot be changed because it is in use.
- Also set the minimum width for the splitter panels, to avoid display issues when the splitter is moved and either of the panels becomes unusable.

### DIFF
--- a/XenAdmin/Controls/PvsCacheStorageRow.cs
+++ b/XenAdmin/Controls/PvsCacheStorageRow.cs
@@ -70,6 +70,9 @@ namespace XenAdmin.Controls
             origCacheSizeGb = numericUpDownCacheSize.Value;
 
             PopulateCacheSrCombobox();
+            var isInUse = OrigPvsCacheStorage != null && OrigPvsCacheStorage.IsInUse;
+            comboBoxCacheSr.Enabled = numericUpDownCacheSize.Enabled = !isInUse;
+            pictureBoxInfo.Visible = isInUse;
         }
 
         private void PopulateCacheSrCombobox()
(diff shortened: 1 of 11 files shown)


XenModel/XenAPI-Extensions/PVS_cache_storage.cs
```diff
@@ -0,0 +1,52 @@
...
+    public partial class PVS_cache_storage
+    {
+        public bool IsInUse
+        {
+            get
+            {
+                foreach (var pvsProxy in Connection.Cache.PVS_proxies)
+                {
+                    var vm = pvsProxy.VM;
+                    if (vm != null && vm.resident_on.opaque_ref == host.opaque_ref &&
```
GaborApatiNagy: I am not sure if we ever get a null here, but it would make it safer to null-check. (There are other places where it's checked in XC)

XenModel/XenAPI-Extensions/PVS_cache_storage.cs
```diff
@@ -0,0 +1,52 @@
...
+                        pvsProxy.site.opaque_ref == site.opaque_ref && pvsProxy.currently_attached)
```
GaborApatiNagy: Can the 'site' be null in any case?

XenAdmin/Controls/PvsCacheStorageRow.cs
```diff
@@ -199,6 +202,16 @@ private void comboBoxCacheSr_SelectedIndexChanged(object sender, EventArgs e)
             }
             SomethingChanged(this, e);
         }
+
+        private void pictureBoxInfo_Click(object sender, EventArgs e)
```
GaborApatiNagy: I think the tooltip would look better if it was displayed on the disabled controls, not just once the (i) icon is clicked.

PR #1334, opened 2016-12-08
CA-216752: Not all wizards progress through to the next page by press…

MihaelaStoica:
…ing the Enter key

- Added a new method in XenTabPage which the derived classes can implement to select a default control after the page is loaded.
- Added a new wizard test that runs through the wizard by pressing the Enter key and checks if it has landed on the right page.

### DIFF
--- a/XenAdmin/Controls/XenTabPage.cs
+++ b/XenAdmin/Controls/XenTabPage.cs
@@ -191,6 +191,11 @@ namespace XenAdmin.Controls
         /// Check whether this step needs to be disabled. Not always overriden in derived classes
         /// </summary>
         public virtual void CheckPageDisabled() { }
+        
+        /// <summary>
+        /// Select a control on the page. Not always overriden in derived classes
+        /// </summary>
+        public virtual void SelectDefaultControl() { }
 
         protected void OnPageUpdated()
         {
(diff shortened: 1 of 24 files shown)


kc284:
This looks ok. I think that, in addition to the above pages, the SelectDefaultControl() should be implemented on the following:

- Finish page of the BugTool wizard
- Finish page of the NewSr wizard
- Ready to upgrade page of the RPU wizard
- AssignPriorities page of the HA wizard.

In the above pages the cancel button gets the focus by default, but it would be better if one of the textboxes were selected on the first 3, as the user can't go further without providing input, and the next button on the last.

MihaelaStoica:
@kc284 I've added SelectDefaultControl() implementations to more wizard pages. Regarding your comments:
- Finish page of the BugTool wizard: the location textbox will get the focus only if it's empty. Otherwise (i.e it is defaulted to something) no user input is needed ant the user can click the Finish button. In general, the wizards will focus the Finish button on the last page of the wizard (if it's not disabled, of course).
- Ready to upgrade page of the RPU wizard does not need the focus on another control, because there is no user input required on this page and the user needs to click the "Start Upgrade" button anyway, so it's expected that this button has the focus.
- AssignPriorities already had the function

kc284:
Hi @MihaelaStoica, 
The RPU wizard does require input for automatic mode (location of installation files); in any case this is not a blocker for this PR, so I'm going to merge it because it has stayed open for quite a while and we can revisit this page at a later occasion.

PR #1339, opened 2016-12-13
CA-211363: Replace 'User name' labels with 'Username'

GaborApatiNagy:
According to the Microsoft style guide XC to use 'username' for UI fields and use 'user name' when it's being referred to.


kc284:
I think this ticket goes the other way round too, i.e. usages of "username" in text should be split. I have compiled a list of places where I think a change is needed and I've attached it to the ticket. Some of them are a bit ambiguous, so I'm happy to discuss them.

GaborApatiNagy:
Thanks for reviewing - good findings, I will get back to this soon.

kc284:
Looks ok now, I'm merging it.

PR #1344, opened 2016-12-15
CP-20149: Some more script adjustments.

kc284:


GaborApatiNagy:
It looks good to me, but may I suggest we use ticket numbers in (most) commits to make the history look better?

For this PR, I think it's fine if we just change the title.

stephen-turner:
Yes, ticket numbers in all commits please (except for trivial typo corrections and the like).

kc284:
Ok, I'll keep this in mind for the future. For the time being, can we merge this because I wouldn't like to delay it further.

GaborApatiNagy:
I added CP-20149 to the title of this PR, so the merge commit will have it and that's more than enough imho.

PR #1346, opened 2016-12-15
CA-236705: [XSO-662]: Applying Updates as Non-root Pool_admin user causes exception and crash of XenCenter

GaborApatiNagy:
Fixed a typo that caused XenCenter to show a Trace error dialog (and to fail to do the clean-up) after applying an update in manual mode. (only if RBAC is involved)


MihaelaStoica:
👍

PR #1348, opened 2016-12-19
CA-236547: Unexpected error while applying hotfix on multiple Xenservers of different version

GaborApatiNagy:
Fixing the error, so the hotfix installation will fail as and where it is expected


MihaelaStoica:
Looks ok to me. Merging.

PR #1349, opened 2016-12-19
CA-236566: test-hotfix-basic-7 throws error 'VDI is in use by some other operation'

GaborApatiNagy:
Remove the VDI only on a best effort basis if the upload failed in the UploadSupplementalPackAction. This will make XenCenter to display the message from the original exception and not from the removal action if the latter had also failed.


MihaelaStoica:
Looks good. Merging.

PR #1352, opened 2016-12-20
CA-234494: System template cross-pool copy fails with Internal error:assert_can_migrate: inter_pool_metadata_transfer returned nonempty list

GaborApatiNagy:
Added a check for the vm's default_template as a workaround for an issue where the allowed_operations of a default template _initially_ indicated that cross pool migration was allowed.


MihaelaStoica:
Looks good. Merging.

PR #1354, opened 2016-12-21
CA-214628: .NET Image limit causes exception

GaborApatiNagy:
Changed code at a a few places where it made sense to limit how many times XenCenter asks the Resource Manager for a new instance(*1) of the same(*2) image.
*1: RM returns a new object each time
*2: they are the same as we don't change the locale of a thread and also we don't localise these images - by design

### DIFF
--- a/XenAdmin/Commands/AddHostCommand.cs
+++ b/XenAdmin/Commands/AddHostCommand.cs
@@ -95,7 +95,7 @@ namespace XenAdmin.Commands
         {
             get
             {
-                return Resources._000_AddApplicationServer_h32bit_16;
+                return Images.StaticImages._000_AddApplicationServer_h32bit_16;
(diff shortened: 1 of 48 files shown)


GaborApatiNagy:
We reviewed it with @kc284 and we agreed that additionally, the followings are need to be done:
* implement similar logic for the Commands too as they are keep on getting updates so affected
* move these static fields to a class (at Images)
* use Camel casing once they are moved

kc284:
This looks ok to me. It might also be worth to have a test to ensure that new images get a static counterpart and any commands don't call directly the ResourceManager, e.g. a check under xenadmin/devtools?

kc284:
As discussed, I'll merge this and we can create a new ticket for adding the check.

PR #1356, opened 2016-12-21
Removed flock as it's not working correctly and is causing the build …

kc284:
…to hang.


kc284:
I don't think this PR can be merged before the holiday, so I'm going to close it and include this commit in my series of changes for the build scripts.

PR #1367, opened 2017-01-09
CA-213497: VMs not supporting checkpoint is allowed to be added in a checkpoint policy

geosharath:
[VMSS] XC create new VMSS Policy wizard allows to select "Disk and memory snapshot"
even when VMs not supporting it are selected


geosharath:
superseded by other PRs

PR #1369, opened 2017-01-09
CA-237963: On a partially upgraded pool (from Dundee to Ely) the host…

MihaelaStoica:
…s lose their consoles

When getting the control domain VM, we should use the control_domain only if it is not null. there is no need to check the version (which was, however, looking at the API version of the master)


stephen-turner:
Looks straightforward, merging.

PR #1372, opened 2017-01-10
CA-226900: Do not allow selection of individual pool members when ins…

MihaelaStoica:
…talling a hotfix

This change does not apply to supplemental packs, as we should be able to install supp packs to individual hosts in a pool.
Also, if the wizard is installing an update from disk and we cannot match it in updates.xml, then we will continue to allow host selection, because we can't tell whether it is a supp pack or a hotfix.


GaborApatiNagy:
I'm wondering about a case where we have a pool with a dead host. In this case, with the current code, we can't install updates the normal way (only by browsing for the update locally).

MihaelaStoica:
@GaborApatiNagy, I've added another commit with the rubric changes.
As we discussed, we'll address the "dead host" case in a separate ticket (CA-239951)

GaborApatiNagy:
Thanks Mihaela, I'll merge it

PR #1382, opened 2017-01-16
CA-176169: Copyright statement checking

GaborApatiNagy:
* new check script
* added and fixed statements

### DIFF
--- a/Branding/branding.sh
+++ b/Branding/branding.sh
@@ -1,5 +1,35 @@
 #!/bin/sh
 
+# Copyright (c) Citrix Systems, Inc. 
+# All rights reserved.
+# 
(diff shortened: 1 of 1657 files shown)


kc284:
This looks ok. Also a few more:
- Branding/branding.sh
- XenServerHealthCheckTests/RequestUploadTaskTests.cs
- some of the scripts in devtools

Now, there are also some non .cs file in these folders that don't have copyright:
- dotNetInstaller
- splash
- WixInstaller
- xe
- xva_verify
but I have to admit I can't remember whether there was a reason for this or whether it's an omission

kc284:
It looks better now, I'm merging it.

PR #1383, opened 2017-01-17
CA-227792: unable to change policy type and modify VMs list at the same me time

geosharath:

### DIFF
--- a/XenAdmin/Dialogs/PropertiesDialog.cs
+++ b/XenAdmin/Dialogs/PropertiesDialog.cs
@@ -338,8 +338,24 @@ namespace XenAdmin.Dialogs
 
             // Add a save changes on the beginning of the actions to enact the alterations that were just changes to the xenObjectCopy.
             // Must come first because some pages' SaveChanges() rely on modifying the object via the xenObjectCopy before their actions are run.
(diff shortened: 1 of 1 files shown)


XenAdmin/Dialogs/PropertiesDialog.cs
```diff
@@ -338,8 +338,24 @@ private void btnOK_Click(object sender, EventArgs e)
...
+                if (VMSSObj.policy_type == policy_backup_type.snapshot_with_quiesce)
+                {
+                    actions.Insert(0, new SaveChangesAction(xenObjectCopy, xenObjectBefore, true));
+                }
+                else
+                {
+                    actions.Insert(actions.Count, new SaveChangesAction(xenObjectCopy, xenObjectBefore, true));
+                }
+            }
+            else /* VMPP object*/
```
MihaelaStoica: I suggest to remove the comment /* VMPP object*/ from here, because on this path xenObjectBefore can be any XenObject type other than VMSS.

geosharath: @MihaelaStoica I have updated the same

PR #1386, opened 2017-01-18
Build overhaul: build XenCenter using a Jenkinsfile.

kc284:


mk/package-and-sign.sh
```diff
@@ -53,13 +52,7 @@ mkdir_clean()
 {
   rm -rf $1 && mkdir -p $1
 }
-
-#overwrite sign file
```
MihaelaStoica: We still need to overwrite the sign file, because someone might want to use their own sign file, not the one packed by us in XenAdminUnsigned.zip

MihaelaStoica:
Looks ok to me now. Merging.

PR #1387, opened 2017-01-19
CA-213512: [VMSS] XC mentions VMPR instead of VMSS on failure of snapshot

geosharath:


XenModel/Alerts/PolicyAlert.cs
```diff
@@ -30,29 +30,67 @@
...
+        public PolicyAlert(long priority, string name, DateTime _time, string body)
         {
             Type = (priority == 4 ? "info": "error");
             Time = _time;
-            Text = Message.FriendlyBody(name);
+            PolicyType = "VMSS";
+
+            if(Type == "info")
+            {
+                Text = Message.FriendlyBody(body);
```
MihaelaStoica: This should be `Message.FriendlyBody(name);`

XenModel/Alerts/PolicyAlert.cs
```diff
@@ -30,29 +30,67 @@
...
+                return;
+            }
+
+            /* We reach here when message type is an error hence we need 
+             * to parse the message body accordingly */
+
+            numberOfVMsFailed = Regex.Matches(body, "VM:").Count;
+            string[] messageHeader = Regex.Split(body.Replace("\n", " "), "Details:");
+            Text = messageHeader[0] + "\n\nDetails:\n";
```
MihaelaStoica: Here the Text won't be localizable, because messageHeader[0] is the taken straight from Message.body and "Details" is hardcoded.
I suggest to use Message.FriendlyBody(name) (which will be different for each VMSS message) in a string.Format() call.

Also, I would like to include the policy name in here, which I think we can obtain from the Message object.

So, for example if `Message.body-vmss_snapshot_failed` is something like: _"The snapshot schedule '{0}' failed because a snapshot could not be created on {1} VMs. \n {2}"_, then it will be used like:
`string.Format(Message.FriendlyBody(name), policyName, nmberOfVMsFailed, details);`, where details will contain an entry for each VM that encountered an error.

I think this will work with the other VMSS messages (VMSS_SNAPSHOT_MISSED_EVENT, VMSS_XAPI_LOGON_FAILURE, etc. - if any of them is considered "error") even if the message body does not contain detailed error information.

XenModel/FriendlyNames.resx
```diff
@@ -1883,4 +1883,7 @@
   <data name="Label-host.edition-basic" xml:space="preserve">
     <value>[Citrix] [XenServer product] Basic Edition</value>
   </data>
-</root>
+  <data name="Message.body-vmss_details" xml:space="preserve">
```
MihaelaStoica: This belongs better to Messages.rex

XenModel/Alerts/PolicyAlert.cs
```diff
@@ -43,16 +45,51 @@ public class PolicyAlert : IEquatable<PolicyAlert>
...
+                Text = Message.FriendlyBody(name);
+                return;
+            }
+
+            /* We reach here when message type is an error hence we need 
+             * to parse the message body accordingly */
+
+            numberOfVMsFailed = Regex.Matches(body, "VM:").Count;
+            string[] messageHeader = Regex.Split(body.Replace("\n", " "), "Details:");
+            Text = string.Format(Message.FriendlyBody(name), policyName, numberOfVMsFailed, "\n\n" + Message.FriendlyBody("vmss_details") + "\n");
```
MihaelaStoica: I think we should only add the "Details" part if we find detailed VM errors. 
So, we can remove the {2} parameter from Message.body-vmss_snapshot_failed, and add "Details:" further down and only if `vmList.Length > 1`. (Please note that I had also suggested to move "Details:" to Messages.resx)
Also, we can build the Text better with the help of a StringBuilder. Then there's no need for string concatenations and explicit addition of  the newline character, because we can do StringBuilder's methods (AppendLine, Append, AppendFormat..).

XenModel/Alerts/PolicyAlert.cs
```diff
@@ -43,16 +45,51 @@ public class PolicyAlert : IEquatable<PolicyAlert>
...
+                vmList[vmIterator].Trim();
+                string[] tmp = Regex.Split(vmList[vmIterator], "Error:");
+                string vmName = Regex.Split(tmp[0], "UUID:")[0];
+              
+                string[] errorCode = Regex.Split(tmp[1].Replace("[", "").Replace("],", "").Replace("]", ""), "\',");
+                for (int errorCodeIterator = 0; errorCodeIterator < errorCode.Length; errorCodeIterator++)
+                    errorCode[errorCodeIterator] = errorCode[errorCodeIterator].Replace("\'", "").Trim();
+
+                string errorMessage = new Failure(errorCode).ShortMessage;
+                Text += vmIterator +") " + vmName.Trim() + ": " + errorMessage + '\n';
```
MihaelaStoica: This should be done with a string format method, for example (using a StringBuilder, as suggested above):
`stringBuilder.AppendFormat(Messages.VMSS_ALERT_VM_ERROR_FORMAT, vmIterator, vmName.Trim(), errorMessage);` where Messages.VMSS_ALERT_VM_ERROR_FORMAT is `"{0}. {1}: {2}"` Here I also suggested "." instead of ")" for the numbered list.

XenModel/Alerts/PolicyAlert.cs
```diff
@@ -43,16 +45,51 @@ public class PolicyAlert : IEquatable<PolicyAlert>
...
+            {
+                vmList[vmIterator].Trim();
+                string[] tmp = Regex.Split(vmList[vmIterator], "Error:");
+                string vmName = Regex.Split(tmp[0], "UUID:")[0];
+              
+                string[] errorCode = Regex.Split(tmp[1].Replace("[", "").Replace("],", "").Replace("]", ""), "\',");
+                for (int errorCodeIterator = 0; errorCodeIterator < errorCode.Length; errorCodeIterator++)
+                    errorCode[errorCodeIterator] = errorCode[errorCodeIterator].Replace("\'", "").Trim();
+
+                string errorMessage = new Failure(errorCode).ShortMessage;
```
MihaelaStoica: We should use Failure.Message, not ShortMessage

geosharath:
@MihaelaStoica can you please share your thoughts on this.

MihaelaStoica:
This looks good now, @geosharath. I am not too sure about the wording changes of "snapshot schedule" vs "scheduled snapshot", but I will merge this PR now, and could you please create a ticket to review the terminology later?

PR #1394, opened 2017-01-20
Moved the ProxyAuthenticationMethod enum to the HTTP class so the VNC…

kc284:
…Control can compile.


GaborApatiNagy:
Looks good to me (as well)

Frezzle:
Looks good. Merging.

PR #1397, opened 2017-01-23
CA-236114 - Wrong update is installed when two different updates are …

letsboogey:
…uploaded in the same instance of the wizard


MihaelaStoica:
Looks good. Merging.

PR #1408, opened 2017-01-25
Merging CAR-2205 into master

kc284:


kc284:
Merging as the commits contained in this PR have been reviewed separately and tested on different branches on Jenkins.

PR #1413, opened 2017-01-30
Jenkinsfile: add authentication and fix GString readFile issue.

rasebo:
Added authentication and fixed the GString readFile issue.


kc284:
Looks ok.

PR #1418, opened 2017-02-01
CP-20712: Add support for FalconOrGreater function in XC

geosharath:

### DIFF
--- a/XenModel/Utils/Helpers.cs
+++ b/XenModel/Utils/Helpers.cs
@@ -421,6 +421,25 @@ namespace XenAdmin.Core
                 HostBuildNumber(host) == CUSTOM_BUILD_NUMBER;
         }
 
+        /// <param name="conn">May be null, in which case true is returned.</param>
+        public static bool FalconOrGreater(IXenConnection conn)
+        {
+            return conn == null ? true : FalconOrGreater(Helpers.GetMaster(conn));
+        }
+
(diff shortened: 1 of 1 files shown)


stephen-turner:
Sharath, do we have the corresponding change on the server side yet? I'm still seeing 2.1.5 in branding.hg. The reason I ask is that I'm not sure whether we want to bump it to 2.2.1 or (say) 2.2.5.

geosharath:
@stephen-turner ill check with the Ely team and get back on the exact platform version for Ely. As far as I know it is 2.2.0 so anything beyond that would be Falcon.

stephen-turner:
Yes, Ely is 2.2.0, but trunk is still at 2.1.5. I would like to use whichever version we first put on trunk.

geosharath:
@stephen-turner looks like in transformer world we would not be looking at branding.hg for platform version and that wont be changed also going forward

stephen-turner:
So where does the platform version come from?

geosharath:
@stephen-turner the value comes from branding.git and currently its set to 2.2.50

stephen-turner:
OK, thanks. So let's use 2.2.50 here.

geosharath:
@stephen-turner I have updated the same

stephen-turner:
Thanks, taking the change now.

PR #1419, opened 2017-02-01
CA-236844: XenCenter throws "Assertion failed" error when toggling be…

MihaelaStoica:
…tween the RDP and VNC consoles

Couldn't reproduce it, but it looks like the assertion is hit because the console view still holds a subscription to an event from an old vnc stream (maybe not garbage collected yet).
Added code to deregister event handlers when the vcnStream is disconnected and also replaced the the assertion with an if statement, ignoring the case that might have caused the assertion failure.

### DIFF
--- a/XenAdmin/ConsoleView/VNCGraphicsClient.cs
+++ b/XenAdmin/ConsoleView/VNCGraphicsClient.cs
@@ -248,8 +248,9 @@ namespace XenAdmin.ConsoleView
         private void OnError(object sender, Exception e)
         {
             Program.AssertOffEventThread();
-            System.Diagnostics.Trace.Assert(sender == this.vncStream);
-
+            if (sender != vncStream)
+                return;
+            
             this.connected = false;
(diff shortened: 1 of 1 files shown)


XenAdmin/ConsoleView/VNCGraphicsClient.cs
```diff
@@ -248,8 +248,9 @@ private void vncStream_ConnectionSuccess(object sender, EventArgs e)
         private void OnError(object sender, Exception e)
         {
             Program.AssertOffEventThread();
-            System.Diagnostics.Trace.Assert(sender == this.vncStream);
-
+            if (sender != vncStream)
```
stephen-turner: I suggest retaining the Assert (perhaps making it into a Debug assert) but keeping the new "return" too.

MihaelaStoica:
Thanks @stephen-turner. I added a new commit.

stephen-turner:
Thanks, merging it now.

PR #1421, opened 2017-02-01
CA-241844: [VMSS] Deleting a snapshot schedule raises XenCenter error

geosharath:
After deleting the schedule vmss will be null for that message, hence
we need to handle this case.

Signed-Of-By: Sharath Babu <sharath.babu@citrix.com>


MihaelaStoica:
Looks good. Merging.

PR #1422, opened 2017-02-01
CA-241847: [VMSS] VM Protection policies pop-up from Pool -> VM Protection policies raises XenCenter error on XS 6.1

geosharath:
When Text is null error is raised by split() function, we
need to handle that case.

Signed-Off-By: Sharath Babu <sharath.babu@citrix.com>

### DIFF
--- a/XenModel/Alerts/PolicyAlert.cs
+++ b/XenModel/Alerts/PolicyAlert.cs
@@ -104,7 +104,6 @@ namespace XenAdmin.Alerts
         public PolicyAlert(IXenConnection connection, string body)
         {
             PolicyType = "VMPP";
-            numberOfVMsFailed = Text.Split('\n').Length;
             var sb = new StringBuilder();
             try
             {
@@ -116,6 +115,7 @@ namespace XenAdmin.Alerts
                 if (_fnames[0].InnerText == "error")
                 {
                     Type = "error";
+                    numberOfVMsFailed = 0;
                     _fnames = xmlDocument.GetElementsByTagName("error");
                     for (int i = 0; i < _fnames.Count; i++)
                     {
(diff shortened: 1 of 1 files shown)


XenModel/Alerts/PolicyAlert.cs
```diff
@@ -104,7 +104,7 @@ public PolicyAlert(long priority, string name, DateTime _time, string body, stri
         public PolicyAlert(IXenConnection connection, string body)
         {
             PolicyType = "VMPP";
-            numberOfVMsFailed = Text.Split('\n').Length;
+            numberOfVMsFailed = (Text == null) ? 0 : Text.Split('\n').Length;
```
MihaelaStoica: numberOfVMsFailed doesn't make sense for info or warn alerts, and it needs to be moved inside the block where we identified this alert as an error and only after the Text is assigned, in line 143 `Text = sb.ToString();`

geosharath: ah, yes your right, I shall make this change. Thanks for your suggestion.

MihaelaStoica:
StringBuilder.ToString() will never return null, so the check is not needed

XenModel/Alerts/PolicyAlert.cs
```diff
@@ -141,6 +140,7 @@ public PolicyAlert(IXenConnection connection, string body)
                             sb.AppendLine();
                     }
                     Text = sb.ToString();
+                    numberOfVMsFailed = (Text == null) ? 0 : Text.Split('\n').Length;
```
GaborApatiNagy: I think this could be done better here, I suggest we just count the number of VMs above (line 129)

geosharath:
@MihaelaStoica @GaborApatiNagy updated the PR.

MihaelaStoica:
Looks good now. Merging.

PR #1424, opened 2017-02-02
CA-241431: Using standard .net WebClient to download updates.xml

GaborApatiNagy:
Switched to use WebClient when downloading updates.xml
Fixed a bug where proxy settings were ignored when downloading updates

### DIFF
--- a/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
+++ b/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
@@ -87,6 +87,9 @@ namespace XenAdmin.Actions
                 {
                     try
                     {
+                        var proxy = XenAdminConfigManager.Provider.GetProxyFromSettings(Connection);
+                        client.Proxy = proxy;
+
                         //register download events
                         client.DownloadProgressChanged += client_DownloadProgressChanged;
                         client.DownloadFileCompleted += client_DownloadFileCompleted;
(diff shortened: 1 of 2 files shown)


stephen-turner:
Are there any other places where we use HTTPHelper.GET that will cause problems?
Conversely, is there anything our HTTPHelper.GET does that WebClient doesn't do?

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -87,6 +87,9 @@ private void DownloadFile()
                 {
                     try
                     {
+                        var proxy = XenAdminConfigManager.Provider.GetProxyFromSettings(Connection);
```
MihaelaStoica: I believe this has been fixed by @Frezzle on the CAR-2214 branch

GaborApatiNagy:
re @stephen-turner 's comment: There will be no other places after this PR, because we'll use HTTPHelper.GET only at places where we communicate with the API. 

Following this PR
* for XS hosts (to the API) we'll use the SDK's implementation (this is already the case), 
* for the web, we'll use only the WebClient class (this is the case after this PR).

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
GaborApatiNagy: As we agreed, we'll solve this if it becomes a merge conflict at all.

MihaelaStoica:
Looks good. Merging.

PR #1429, opened 2017-02-08
CP-19605: Improve the performance of the CreateNewRootNode function

MihaelaStoica:
Make GroupKey a class and override GetHashCode and Equals

### DIFF
--- a/XenModel/XenSearch/GroupAlg.cs
+++ b/XenModel/XenSearch/GroupAlg.cs
@@ -257,7 +257,7 @@ namespace XenAdmin.XenSearch
         public abstract void GetNextLevel(List<GroupKey> nextLevel);
     }
 
-    public struct GroupKey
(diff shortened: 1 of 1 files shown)


XenModel/XenSearch/GroupAlg.cs
```diff
@@ -316,6 +332,8 @@ public override bool Populate(IAcceptGroups adapter, int indent, bool defaultExp
 
         public override void PopulateFor(IAcceptGroups adapter, GroupKey group, int indent, bool defaultExpand)
         {
+            //System.Console.WriteLine(group.GetHashCode().ToString() + "   " + group.key.ToString() );
```
stephen-turner: The new code is good, but I suggest we lose the (already commented-out) WriteLine.

MihaelaStoica:
@stephen-turner: Done

PR #1431, opened 2017-02-08
CP-20545: [XC] Validate the usage of "scheduled snapshots" and "snapshot schedules"

geosharath:
Signed-Off-By: Sharath Babu <sharath.babu@citrix.com>


MihaelaStoica:
Looks good. Merging.

PR #1436, opened 2017-02-10
CA-242397: Fix format exception caused by incorrect chunk reading 

MihaelaStoica:
Improved the ReadHttpHeaders function, by making sure that a chunk is read fully from the stream (the number of bytes read by the Stream.Read function can be less than the number of bytes requested, so we need to continue reading until the required number of bytes is read)


MihaelaStoica:
Merging this myself as it contains the changes approved in xen-api-sdk (xapi-project/xen-api-sdk/pull/88)

PR #1437, opened 2017-02-10
CA-227631: License Manager doesn't respond immediately to disconnected host

GaborApatiNagy:
Fixed a bug that caused a disconnected host to appear as if it was still connected in the License Manager. After this commit, if a host get disconnected in the background its row will be disabled and will be unselectable.

(Note that when the connection resumes, the user will still have to click on a different row to have the disabled state (and the buttons) fixed, but that is a Won't fix bug (CA-115261) that I couldn't fix right now.)


MihaelaStoica:
Looks good. Merging.

PR #1439, opened 2017-02-13
CA-236150 - Funny label on status report wizard

letsboogey:
Changed label to a LinkLabel instead of RichTextBox. Also fixed similar bug on the healthcheck enrollment dialog.


XenAdmin/Wizards/BugToolWizardFiles/BugToolPageDestination.cs
```diff
@@ -273,9 +277,10 @@ private void caseNumberLabelTextBox_TextChanged(object sender, EventArgs e)
 
         #endregion
 
-        private void richTextBox1_LinkClicked(object sender, LinkClickedEventArgs e)
+        private void enterCredentialsLinkLabel_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
         {
-            Program.OpenURL(e.LinkText);
+            this.enterCredentialsLinkLabel.LinkVisited = true;
+            System.Diagnostics.Process.Start(Messages.MY_CITRIX_CREDENTIALS_URL);
```
GaborApatiNagy: I think we should still use  Program.OpenURL() here as before.

GaborApatiNagy:
Looks good, merging

PR #1441, opened 2017-02-15
CA-242451: Use a different location for binaries built with CTXSIGN defined (the…

kc284:
… actual location will be updated when the build is available). Use latest dotnet-packages.


packages/DOTNET_BUILD_LOCATION_CTXSIGN
```diff
@@ -0,0 +1 @@
+xc-local-build/dotnet-packages/master/3
```
kc284: Please accept this for now, it will change once DO-2610 is done.

GaborApatiNagy:
Looks good, merging

PR #1442, opened 2017-02-17
CA-164371 - Usability issue with Popup windows from XenCenter Tools Menu

letsboogey:


stephen-turner:
Accepting this change.

PR #1443, opened 2017-02-17
CA-243964: Don't allow upload of .xsupdate files to >= 7.1 servers

stephen-turner:


GaborApatiNagy:
Good catch - I was wondering how could this restriction been left out... 
Looks good, merging.

PR #1446, opened 2017-02-22
CP-17253: [XC] Ability to select VLAN PIF as a management interface of a host/Pool

geosharath:
This is allowed for both host and pool (on a host by host basis).

Signed-Off-By: Sharath Babu <sharath.babu@xitrix.com>


MihaelaStoica:
Looks good. Merging.

PR #1447, opened 2017-02-23
CA-244656: The popup dialog for Check for Updates (on first XC start)…

MihaelaStoica:
… should enable all CFU options


stephen-turner:
I can't see anything wrong with this! Merging.

PR #1448, opened 2017-02-24
CA-244655 - XenCenter shows the wrong message on the Updates page

letsboogey:
XenCenter shows the wrong message on the Updates page when no updates are found


MihaelaStoica:
Looks good. Merging.

PR #1456, opened 2017-02-27
CA-236104: Do not allow pre and post Ely servers to be selected at th…

MihaelaStoica:
…e same time when the user browsed for an iso update or supplemental pack.


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -260,14 +260,39 @@ private void EnabledRow(Host host, UpdateType type, int index)
...
                     break;
             }
         }
 
+        private Host GetFirstCheckedHost()
+        {
+            var firstCheckedRow = dataGridViewHosts.Rows.Cast<PatchingHostsDataGridViewRow>().FirstOrDefault(row => row.CheckValue > UNCHECKED);
+            if (firstCheckedRow == null)
+                return null;
+            return firstCheckedRow.Tag as Host ?? Helpers.GetMaster(((IXenObject)firstCheckedRow.Tag).Connection);
```
GaborApatiNagy: Shall we use GetMaster(Pool) instead?

PR #1457, opened 2017-02-27
CP-20881 - Implement RDP console resize when toggling the full screen…

letsboogey:
… mode

In addition to ticket requirements, the console size is updated when the console is:
- resized in undocked mode
- resized manually by dragging window edges
- resized by dragging the panel splitter


XenAdmin/ConsoleView/XSVNCScreen.cs
```diff
@@ -1386,5 +1385,31 @@ private bool Keysym(bool pressed, Keys key, Keys extendedKey)
 
             return false;
         }
+
+        private void InitializeComponent()
```
MihaelaStoica: I think this was added while experimenting with events on this control, but is not needed anymore.

PR #1459, opened 2017-02-28
CA-164371 - Usability issue with Popup windows from XenCenter Tools Menu

letsboogey:
Added scrollbars for case when panel is resized using splitter.


MihaelaStoica:
This looks good.
The only thing I've noticed by testing it was that on the Console page, the bottom border of the "Windows Remote Desktop console options" group box is cut off when the scrollbar is visible.

PR #1460, opened 2017-02-28
CA-244618: Corrected the xml-rpc licence.

kc284:


MihaelaStoica:
Looks good. Merging

PR #1461, opened 2017-02-28
CA-244687: Install Updates Wizard: Automated mode: bugfix for omitted after-apply-guidances

GaborApatiNagy:
This commit fixes a bug in the Update Wizard that causes XenCenter to not restart Toolstack when it should in the following case: in Automated Mode, when the update sequence for a host has an update --that contains live-patch with restartHost guidance (and guidance-mandatory=false)-- and which is followed by an update with Toolstack restart guidance.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
@@ -167,7 +167,7 @@ namespace XenAdmin.Wizards.PatchingWizard
                         foreach (var host in hostsToApply)
                         {
                             planActions.Add(new ApplyXenServerPatchPlanAction(host, patch, patchMappings));
-                            planActions.AddRange(GetMandatoryActionListForPatch(delayedActionsByHost[host], host, patch));
+                            planActions.AddRange(GetMandatoryActionListForPatch(host, patch));
                             UpdateDelayedAfterPatchGuidanceActionListForHost(delayedActionsByHost[host], host, patch);
                         }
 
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -290,68 +276,125 @@ private void UpdateStatusTextBox()
...
+                            doWorkEventArgs.Cancel = true;
+                            return;
+                        }
 
-                    bgw.ReportProgress(0, action);
-                    action.Run();
+                        action = a;
 
-                    Thread.Sleep(1000);
+                        // always run the IAvoidRestartHostsAware ones (these are assumed to be preceding the rest, be careful if you touch that!)
```
MihaelaStoica: Instead of relying on the actions to be in a specific order in the list, we could iterate through the list twice:
- on first iteration, run only the IAvoidRestartHostsAware actions
- on second iteration run the other actions, if needed

GaborApatiNagy: Good idea. I can't really recall how I ended up doing it and despite the warnings (too many inline comments) I decided to submit it. I must agree, there is a better way, so updated the PR. Thanks

XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -290,68 +276,133 @@ private void UpdateStatusTextBox()
...
+                            RunPlanAction(bgw, action);
+                        }
+                        else
+                        {
+                            action.Visible = false;
+                            bgw.ReportProgress((int)((1.0 / (double)bgw.ActionsCount) * 100), action); //still need to report progress, mainly for the progress bar
 
-                    bgw.doneActions.Add(action);
-                    bgw.ReportProgress((int)((1.0 / (double)bgw.AllActions.Count) * 100), action);
+                            continue; //yes, skipping this action
```
MihaelaStoica: continue is not needed here

XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -290,68 +276,133 @@ private void UpdateStatusTextBox()
...
+                            doWorkEventArgs.Cancel = true;
+                            return;
+                        }
 
-                    Thread.Sleep(1000);
+                        // any non-restart-alike delayed action needs to be run if:
+                        // - this host is pre-Ely and there isn't any restart plan action among the delayed actions, or
+                        // - this host is Ely or above and bgw.AvoidRestartHosts contains the host's uuid (shows that live patching must have succeeded)
+                        if (!Helpers.ElyOrGreater(h) && !actions.Any(pa => pa.IsRestartRelatedPlanAction())
+                            || Helpers.ElyOrGreater(h) && bgw.AvoidRestartHosts != null && bgw.AvoidRestartHosts.Contains(h.uuid))
```
MihaelaStoica: Here we also need to include the case where there wasn't any restart related action in the list (!actions.Any(pa => pa.IsRestartRelatedPlanAction())

GaborApatiNagy:
good catch @MihaelaStoica, I've just submitted my fixes

PR #1462, opened 2017-03-01
Replaced bash script with powershell script, as not everyone has cygw…

kc284:
…in installed on their Windows machine.


kc284:
It would be nice if the reviewer tried out the script on their machine as part of the review.

MihaelaStoica:
Reviewed and tested.

PR #1463, opened 2017-03-02
CP-17250: [XC] Allow host to join an existing pool with management on a VLAN

geosharath:
Commits:

- 	Adding new APIs for pool reconfigure managment interface
- 	Enable Pool join over VLAN
-       Use pool level api for changing Pool's management interface 

Signed-Off-By: Sharath Babu <Sharath.Babu@citrix.com>

### DIFF
--- a/XenModel/Actions/Network/ChangeNetworkingAction.cs
+++ b/XenModel/Actions/Network/ChangeNetworkingAction.cs
@@ -135,6 +135,12 @@ namespace XenAdmin.Actions
                     if (Pool != null)
                     {
                         progress += inc;
+                        if (!Helpers.FeatureForbidden(Connection, Host.RestrictManagementOnVLAN))
+                       {
+                            PoolReconfigureManagement(progress);
+                            return;
+                       }
+
                         ReconfigureManagement(false, progress);
                     }
 
@@ -193,6 +199,13 @@ namespace XenAdmin.Actions
 
             NetworkingActionHelpers.ReconfigureManagement(this, downManagement, newManagement, this_host, true, hi,
                                                           clearDownManagementIP);
+        }
+
+        private void PoolReconfigureManagement(int hi)
+        {
+            System.Diagnostics.Trace.Assert(newManagement != null);
+
+            NetworkingActionHelpers.PoolReconfigureManagement(this, newManagement, downManagement, hi);
         }
 
         private void BringUp(PIF new_pif, PIF existing_pif, int hi)
--- a/XenModel/Actions/Network/ChangeNetworkingAction.cs
+++ b/XenModel/Actions/Network/ChangeNetworkingAction.cs
@@ -95,6 +95,8 @@ namespace XenAdmin.Actions
             ApiMethodsToRoleCheck.Add("pif.plug");
             ApiMethodsToRoleCheck.AddRange(XenAPI.Role.CommonSessionApiList);
             ApiMethodsToRoleCheck.AddRange(XenAPI.Role.CommonTaskApiList);
+            if(!Helpers.FeatureForbidden(Connection, Host.RestrictManagementOnVLAN))
+                ApiMethodsToRoleCheck.Add("pool.management_reconfigure");
             #endregion
 
         }
@@ -203,9 +205,12 @@ namespace XenAdmin.Actions
 
         private void PoolReconfigureManagement(int hi)
         {
-            System.Diagnostics.Trace.Assert(newManagement != null);
+            System.Diagnostics.Trace.Assert(downManagement != null);
 
-            NetworkingActionHelpers.PoolReconfigureManagement(this, newManagement, downManagement, hi);
+            if (newManagement == null)
+                return;
+
+            NetworkingActionHelpers.PoolReconfigureManagement(this, Pool, newManagement, downManagement, hi);
         }
 
         private void BringUp(PIF new_pif, PIF existing_pif, int hi)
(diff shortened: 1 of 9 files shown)


XenModel/Actions/Network/NetworkingActionHelpers.cs
```diff
@@ -269,6 +269,45 @@ private static void ReconfigureManagement_(AsyncAction action, PIF pif, int hi)
...
+           int lo = action.PercentComplete;
+           int inc = (hi - lo) / 3;
+           lo += inc;
+           PoolManagementReconfigure_( action, up_pif, lo);
+
+           /* Pool recover slaves spwans two tasks dbsync (update_env) and server_init on each slaves. 
+            * Only after their completion master will be able to execute reconfigure_IPs. 
+            * Hence, we allow some setteling time for them to complete.
+            */
+           Thread.Sleep(10000);
```
MihaelaStoica: My understanding here is that the pool_management_reconfigure task is marked as completed before its subtasks are completed. Is this correct? 
If that's the case, then isn't this a bug in xapi? 
Assuming that we have to work around this, then is there any pool property that we can check to indicate that the pool_management_reconfigure (and the associated tasks) have truly finished, instead of waiting an arbitrary 10 seconds?

XenModel/Actions/Network/ChangeNetworkingAction.cs
```diff
@@ -195,6 +201,13 @@ private void ReconfigureManagement(bool this_host, int hi)
                                                           clearDownManagementIP);
         }
 
+        private void PoolReconfigureManagement(int hi)
+        {
+            System.Diagnostics.Trace.Assert(newManagement != null);
```
geosharath: we would not hit this point in code if downmanagemnt was null

XenModel/Actions/Network/NetworkingActionHelpers.cs
geosharath: I checked the tasks for pool_management_reconfigure, it completes after pool_recover_slaves gets completed. Only after pool_recover_slaves is completed these two tasks are spawned so its difficult to track them since these are not sub tasks of pool_recover_slaves. Even I was not keen in giving a timeout value, however in case if we have to check for some pool property we would require to poll (busy loop) for some fixed amount of time until the property is changed, so ultimately the timeout factor comes into play.

stephen-turner: Can this not be fixed up on the server side somehow?

XenModel/Actions/Network/ChangeNetworkingAction.cs
MihaelaStoica: I agree that we shouldn't hit this code with null `downManagement`. But that's why assertion is suitable here. In general we use assertion to make sure that we don't hit that code with a certain condition. 
And we shouldn't throw an assertion failure if `newManagement` is null because this can happen, according to the comments on the constructor. I think that for `newManagement` we need a null check, but in an `if` statement, not an assertion.

geosharath: Thanks for the suggestion @MihaelaStoica , I will incorporate the same. Your suggesting having an assertion inside an if statement for newManagement like if (newManagement == null) then assert ?

XenModel/Actions/Network/NetworkingActionHelpers.cs
geosharath: @stephen-turner , @sharad analyzed it and these are his thoughts  `Pool.management_reconfigure` calls `pool.recover_slaves` this call purposely creates a new task `emergency_reset_master` check here
https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/xapi_pool.ml#L816
Since xapi gets restarted on slaves as part of `emergency_reset_master` we can't make sure task `pool.management_reconfigure` will get finished only when slaves are able to contact with master.

XenModel/Actions/Network/ChangeNetworkingAction.cs
MihaelaStoica: I suggest to have:
- an Assert for downManagement and
- an if for newManagement
E.g:
```
System.Diagnostics.Trace.Assert(downManagement != null);
if (newManagement == null) 
    return;
// the rest of the code here
```

XenModel/Actions/Network/NetworkingActionHelpers.cs
geosharath: @stephen-turner We have tried a range of solutions like polling of tasks that gets created for dbsync and waited for a predefined time until they complete, but not always they were able to capture them. There is no other property as far as we know that we can check to ensure that emergency reset master has indeed completed in all slaves.

XenModel/Actions/Network/NetworkingActionHelpers.cs
```diff
@@ -269,6 +269,45 @@ private static void ReconfigureManagement_(AsyncAction action, PIF pif, int hi)
...
+       /// <summary>
+       /// Switch the Pool's management interface from its current setting over to the given PIF.
+       /// </summary>
+       private static void PoolManagementReconfigure_(AsyncAction action, PIF pif, int hi)
+       {
+
+           log.DebugFormat("Switching to PIF {0} {1} for management...", pif.Name, pif.uuid);
+           action.Description = string.Format(Messages.ACTION_CHANGE_NETWORKING_MANAGEMENT_RECONFIGURING, pif.Name);
+
+           action.RelatedTask = Pool.async_management_reconfigure(action.Session, pif.network.opaque_ref);
```
geosharath: @MihaelaStoica sure will add the same

XenModel/Actions/Network/NetworkingActionHelpers.cs
MihaelaStoica: @geosharath Could you please clarify why do we need to call Pool.management_reconfigure and why the old way of calling host.management_reconfigure on all hosts in not sufficient anymore?

PR #1469, opened 2017-03-03
CA-245546 - Only the active RDP console is resized when the main XenC…

letsboogey:
…enter window is resized


MihaelaStoica:
Looks good. Merging.

PR #1470, opened 2017-03-03
CA-240234: PVS proxy setup is not preserved accross OVF export/import

GaborApatiNagy:
Implemented export/import of PVS cache/site setting of a vm for OVF

### DIFF
--- a/XenOvfTransport/Export.cs
+++ b/XenOvfTransport/Export.cs
@@ -40,6 +40,7 @@ using XenOvf;
 using XenOvf.Definitions;
 using XenOvf.Utilities;
 using XenAPI;
+using System.Linq;
 
 
 namespace XenOvfTransport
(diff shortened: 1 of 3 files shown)


XenOvfTransport/Export.cs
```diff
@@ -354,6 +355,37 @@ private EnvelopeType _export(Session xenSession, string targetPath, string ovfna
                         OVF.AddOtherSystemSettingData(ovfEnv, vsId, "vgpu", sb.ToString(), OVF.GetContentMessage("OTHER_SYSTEM_SETTING_DESCRIPTION_4"));
                     }
                 }
+
+                var allProxies = PVS_proxy.get_all_records(xenSession).Select(kvp => kvp.Value).ToList();
```
MihaelaStoica: Why do you use a xapi call (get_all_records) for PVS proxies, but we use Resolve for the VIF and VM? Can't we also use ResolveAll for proxies?

XenOvfTransport/Export.cs
```diff
@@ -354,6 +355,37 @@ private EnvelopeType _export(Session xenSession, string targetPath, string ovfna
...
+                    {
+                        var vmFromVif = xenSession.Connection.Resolve(vif.VM);
+                        if (vmFromVif != null && vmFromVif.uuid == vm.uuid)
+                        {
+                            var pvsSite = xenSession.Connection.Resolve(p.site);
+                            if (pvsSite != null)
+                            {
+                                pvsSiteUuid = pvsSite.uuid;
+                                
+                                break;
```
MihaelaStoica: I think that the `break `should be outside this `if `block, when we found the VM that this proxy is configured for. The site check is just to get site's uuid, not a condition for exiting the loop.

XenOvfTransport/Export.cs
GaborApatiNagy: true, will change it

XenOvfTransport/Export.cs
GaborApatiNagy: Currently it exits the loop once it's found a pvs proxy object that refers the current vm and has a valid pvs site assigned to it. However, I see that doing so makes the loop to continue in case it finds a pvs proxy object for the vm, but can't resolve the referred pvs site. I can't see significant difference, but as we discussed it makes sense to assume the first object is fine and not to search for a next one in the not-expected case should the pvs site be not resolvable. So moving the break statement outside.

GaborApatiNagy:
@MihaelaStoica thanks for reviewing, I've updated the PR with the suggested changes

MihaelaStoica:
Looks good to me now. Merging.

PR #1474, opened 2017-03-07
CA-137253: Make the error message more generic, because it can have v…

stephen-turner:
…arious causes


stephen-turner:
NB Not tested, because I don't have an iLO-enabled host to test it on. But it should be straightforward.

GaborApatiNagy:
looks good, merging

PR #1476, opened 2017-03-09
CA-243971: PoolJoinRules should check for equivalence of patches

MihaelaStoica:

### DIFF
--- a/XenModel/PoolJoinRules.cs
+++ b/XenModel/PoolJoinRules.cs
@@ -394,19 +394,16 @@ namespace XenAdmin.Core
             {
                 var masterUpdates = master.AppliedUpdates();
                 var slaveUpdates = slave.AppliedUpdates();
-                if (masterUpdates.Except(slaveUpdates, new UpdateUuidEqualityComparer()).Any() ||
-                    slaveUpdates.Except(masterUpdates, new UpdateUuidEqualityComparer()).Any())
-                    return true; 
+                return masterUpdates.Count != slaveUpdates.Count ||
+                       masterUpdates.Except(slaveUpdates, new UpdateUuidEqualityComparer()).Any();
             }
             else
             {
                 var masterPatches = master.AppliedPatches();
                 var slavePatches = slave.AppliedPatches();
-                if (masterPatches.Except(slavePatches, new PatchUuidEqualityComparer()).Any() ||
-                    slavePatches.Except(masterPatches, new PatchUuidEqualityComparer()).Any())
-                    return true; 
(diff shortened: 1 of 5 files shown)


XenModel/PoolJoinRules.cs
```diff
@@ -382,6 +388,27 @@ private static bool DifferentServerVersion(Host slave, Host master)
                 slave.ProductBrand != master.ProductBrand;
         }
 
+        private static bool DifferentPatches(Host slave, Host master)
+        {
+            if (Helpers.ElyOrGreater(master))
+            {
+                var masterUpdates = master.AppliedUpdates();
+                var slaveUpdates = slave.AppliedUpdates();
+                if (masterUpdates.Except(slaveUpdates, new UpdateUuidEqualityComparer()).Any() ||
```
stephen-turner: It would be more efficient to check that the length of the two Lists is the same, and that the set difference in one direction is empty. That way, you only have to calculate at most one set difference, rather than two; and in most cases of inequality, zero.

stephen-turner: I'm not sure why we're defining new classes UpdateUuidEqualityComparer and PatchUuidEqualityComparer. Can't we just make the default Equals do the right thing on updates and patches. (Or maybe it does the right thing already?)

MihaelaStoica: The default comparer in all our XenObjects is by opaque_ref. For updates and patched we will have objects with the same uuid in different pools, but the opaque_refs would be different. I didn't want to change the default comparer because there could be situations where we rely on the opaque_ref uniqueness across pools, for example in the batch hotfixing.

MihaelaStoica: Thanks for the suggestion.  I've added a new commit with these changes.

GaborApatiNagy:
It may not cause a problem, but now (since Ely) AppliedUpdates contain supplemental packs (and driver disks) as well. I assume this code will catch them too. 
Just wondering whether this is what we want here - I am not familiar to restrictions to this extent, but this may cause regression if XAPI would ignore supp packs.

XenModel/PoolJoinRules.cs
stephen-turner: OK. I suspect it would be OK with opaque_ref, but I'm happy to be cautious.

stephen-turner:
DO NOT MERGE until we've decided what the correct behaviour on the server side is

kc284:
I'm closing this PR as replaced by https://github.com/xenserver/xenadmin/pull/1534

PR #1477, opened 2017-03-09
CA-246338: Changing Network Bond Mode on a single host in a pool changes bond mode on all hosts

GaborApatiNagy:
Added "Changes will apply to all hosts in a pool." to clarify the scope of these settings.


stephen-turner:
Merging

PR #1478, opened 2017-03-09
CA-230621: Disk space calculation is logged incorrectly in the XenCen…

MihaelaStoica:
…ter log

Only log the cleanup message if the available space is less than the required space


stephen-turner:
Merging

PR #1481, opened 2017-03-10
CP-21288: Update the libraries location

MihaelaStoica:

### DIFF
--- a/mk/deps-map.json
+++ b/mk/deps-map.json
@@ -11,42 +11,42 @@
       "flat": "true"
     },
     {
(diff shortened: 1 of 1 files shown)


kc284:
You can use a wildcard for the dlls:
"pattern": "@REMOTE_DOTNET@/dotnet46/*.dll"

MihaelaStoica:
But they still need to be copied in different folders in xenadmin.git

kc284:
Aaaa, we haven't ported the master changes here, have we? Ok fair enough!

PR #1482, opened 2017-03-10
CA-236692: XenCenter does not allow host with suspended Vms to join pool

stephen-turner:
Fixed and tested


MihaelaStoica:
Looks good. Merging.

PR #1483, opened 2017-03-10
Fixed typo/regression that was introduced in a previous commit.

kc284:


MihaelaStoica:
Merging

PR #1487, opened 2017-03-14
CP-21172: Support alerts for LTSR and CR XS versions at the same time

GaborApatiNagy:


MihaelaStoica:
Looks good. Merging.

PR #1488, opened 2017-03-14
CP-21174: Multiple dismissed XenCenterVersion alerts

MihaelaStoica:
- save multiple dismissed alerts as a comma-separated list in Settings.Default.LatestXenCenterSeen

### DIFF
--- a/XenAdmin/Alerts/Types/XenCenterUpdateAlert.cs
+++ b/XenAdmin/Alerts/Types/XenCenterUpdateAlert.cs
@@ -101,11 +101,15 @@ namespace XenAdmin.Alerts
             }
         }
 
+        static int DISMISSED_XC_VERSIONS_LIMIT = 5;
+
         public override void Dismiss()
         {
             List<string> current = new List<string>(Properties.Settings.Default.LatestXenCenterSeen.Split(','));
             if (current.Contains(NewVersion.VersionAndLang))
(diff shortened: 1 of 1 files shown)


XenAdmin/Alerts/Types/XenCenterUpdateAlert.cs
```diff
@@ -102,15 +102,20 @@ public override string HelpID
         }
 
         public override void Dismiss()
-        {            
-            Properties.Settings.Default.LatestXenCenterSeen = NewVersion.VersionAndLang;
+        {
+            List<string> current = new List<string>(Properties.Settings.Default.LatestXenCenterSeen.Split(','));
+            if (current.Contains(NewVersion.VersionAndLang))
+                return;
+            current.Add(NewVersion.VersionAndLang);
```
GaborApatiNagy: This list is now going to grow forever, I suggest we limit the max length. (As the previously dismissed ones are not really needed to be kept.)

MihaelaStoica: Thanks @GaborApatiNagy for the suggestion, I've added a new commit, where the size of this list is limited to 5, cleaning up from the begining of the list.

GaborApatiNagy:
Looks good, merging

PR #1495, opened 2017-03-21
CA-248087: Updated API bindings.

kc284:


kc284:
Please note there one or two small problems with this PR, namely some missing descriptions. These are to be fixed on the sdk side.

stephen-turner:
Why have vmss_type and vmss_frequency changed? Are these just line-ending changes? If so, were they incorrect before and are correct now, or the other way round?

kc284:
Yes, these files were originally committed with unix-style eol, now they are windows-style.
The SDK does normalise the lines, so it's likely the regression to unix eol was due to the committer's local git settings - it does happen quite often.

stephen-turner:
OK, thanks, merging.

PR #1503, opened 2017-03-22
CP-21558 - On the vGPU types dialog, hide the column headers where al…

letsboogey:
…l the values are 0

### DIFF
--- a/XenAdmin/Controls/GPU/GpuConfiguration.cs
+++ b/XenAdmin/Controls/GPU/GpuConfiguration.cs
@@ -102,6 +102,8 @@ namespace XenAdmin.Controls.GPU
                                                    .ToArray());
                 }
                 SetCheckedValues();
+                HideColumnIfEmpty(MaxResolutionColumn);
+                HideColumnIfEmpty(MaxDisplaysColumn);
             }
             finally
             {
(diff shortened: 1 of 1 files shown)


XenAdmin/Controls/GPU/GpuConfiguration.cs
```diff
@@ -153,6 +155,22 @@ private void SetWindowTitle()
...
                                    : Messages.GPU_RUBRIC_PLEASE_SELECT_WHICH_GPU_MULTIPLE;
         }
+
+        private void HideColumnIfEmpty(string columnName)
+        {
+            bool columnEmpty = true;
+            foreach (DataGridViewRow row in dataGridViewEx1.Rows)
+            {
+                var value = row.Cells[columnName].Value;
+                if (value != null && !String.IsNullOrEmpty((string)value))
```
kc284: The first null check is not needed.

XenAdmin/Controls/GPU/GpuConfiguration.cs
```diff
@@ -153,6 +155,22 @@ private void SetWindowTitle()
                                    ? Messages.GPU_RUBRIC_PLEASE_SELECT_WHICH_GPU_ONE
                                    : Messages.GPU_RUBRIC_PLEASE_SELECT_WHICH_GPU_MULTIPLE;
         }
+
+        private void HideColumnIfEmpty(string columnName)
```
kc284: I would change this method to accept as a parameter 'DataGridViewColumn column' and then in line 164 you can do row.Cells[column.Name] to find the cell. In general using hardcoded strings should be avoided, because if someone renames the columns, lines 105-106 will break.

letsboogey:
Good shout, I've made the changes and refreshed the pull request

GaborApatiNagy:
Cool, merging

PR #1506, opened 2017-03-23
CA-237939: Cannot create VM from Checkpoint other than the resident h…

MihaelaStoica:
…ost in pool

Use VM.clone (and not VM.copy) when creating a VM from a snapshot.


GaborApatiNagy:
cool, merging

PR #1509, opened 2017-03-23
CP-21500: Updates wizard: Show new versions that are available as updates

GaborApatiNagy:
In the Updates wizard:
New versions that are available as updates will be listed in the "Download Updates from Citrix" section, before any other updates.
If XenCenter is not the latest version, these rows will be greyed out, with a tooltip.

### DIFF
--- a/XenAdmin/Alerts/NewVersionPriorityAlertComparer.cs
+++ b/XenAdmin/Alerts/NewVersionPriorityAlertComparer.cs
@@ -1,4 +1,35 @@
-﻿using System;
+﻿/* Copyright (c) Citrix Systems, Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer. 
+ * *   Redistributions in binary form must reproduce the above 
+ *     copyright notice, this list of conditions and the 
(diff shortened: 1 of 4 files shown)


XenAdmin/Alerts/NewVersionPriorityAlertComparer.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
```
MihaelaStoica: This new file is missing the Copyright notice

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -255,7 +253,15 @@ private void PageLeaveCancelled(string message)
         private void PopulatePatchesBox()
         {
             dataGridViewPatches.Rows.Clear();
-            var updates = new List<Alert>(Updates.UpdateAlerts);
+
+            var alerts = Updates.UpdateAlerts.ToList();
+            if (dataGridViewPatches.SortedColumn == null)
+            {
+                alerts.Sort(new NewVersionPriorityAlertComparer());
```
MihaelaStoica: Can this not be done in the `else` block of the `if (dataGridViewPatches.SortedColumn != null)` below?

XenAdmin/Alerts/NewVersionPriorityAlertComparer.cs
GaborApatiNagy: oh I'll need to fix it

GaborApatiNagy:
Updated this PR

PR #1511, opened 2017-03-24
CA-232996: Storage menu doesn't refresh in XenCenter for destroy and …

MihaelaStoica:
…forget

Update the toolbar when an SR action is completed


stephen-turner:
This was a better solution than I had in mind, because it also updates the menu items if the menus are already open. Merging.

PR #1523, opened 2017-03-30
CA-249052: [SCTX-2514]  xe.exe unable to upload patches with XS 7.1.

MihaelaStoica:
Added Content-Length header field to the HTTP PUT method

### DIFF
--- a/CommandLib/thinCLIProtocol.cs
+++ b/CommandLib/thinCLIProtocol.cs
@@ -185,8 +185,8 @@ namespace CommandLib
         public static Stream doRPC(String method, Uri uri, thinCLIProtocol tCLIprotocol, params string[] headers)
         {
             Stream http = Transport.connect(tCLIprotocol, uri.Host, uri.Port);
-            String header = string.Format("{0} {1} HTTP/1.0", method, uri.PathAndQuery);
-            writeLine(http, header);
+            String startLine = string.Format("{0} {1} HTTP/1.0", method, uri.PathAndQuery);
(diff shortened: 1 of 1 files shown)


CommandLib/thinCLIProtocol.cs
```diff
@@ -182,12 +182,15 @@ public class HTTP{
 		    return Int32.Parse(bits[1]);
 	    }
 
-        /// <param name="addr">The target URI, including scheme, hostname and path.</param>
-        public static Stream doRPC(String method, Uri uri, thinCLIProtocol tCLIprotocol)
+        public static Stream doRPC(String method, Uri uri, thinCLIProtocol tCLIprotocol, params string[] headers)
         {
             Stream http = Transport.connect(tCLIprotocol, uri.Host, uri.Port);
-            String header = method + " " + uri.PathAndQuery + " HTTP/1.0\r\n\r\n";
+            String header = string.Format("{0} {1} HTTP/1.0", method, uri.PathAndQuery);
```
GaborApatiNagy: It works as expected, but I find confusing to have a "header" and a "headers" as well. The "header" here is the start-line what is not part of the header.

I know it's existing code, but if we start using a new variable "headers" here, I think we should rename "header" to something else as that is not part of the header per definition https://tools.ietf.org/html/rfc2616#section-4

MihaelaStoica:
Thanks @GaborApatiNagy . I've renamed the variable.

GaborApatiNagy:
Great, merging

PR #1525, opened 2017-03-31
Feature REQ-378 XenCenter part

makunterry:


XenModel/Alerts/PerfmonOptionsDefinition.cs
```diff
@@ -117,11 +197,12 @@ public static PerfmonOptionsDefinition GetPerfmonOptionsDefinitions(IXenObject x
 
             string mailDestination = GetMailDestination(connection);
             string mailHub = GetSmtpMailHub(connection);
+            string mailLanguageCode = GetMailLanguageCode(connection);
 
             if (mailDestination == null || mailDestination == null)
```
MihaelaStoica: Unrelated to this commit, I think the second condition in this `if` was supposed to be `mailHub == null`

XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
```diff
@@ -132,8 +138,14 @@ public void Repopulate()
                     EmailAddressTextBox.Text = _PerfmonOptions.MailDestination;
                     SmtpServerAddrTextBox.Text = PerfmonOptionsDefinition.GetSmtpServerAddress(_PerfmonOptions.MailHub);
                     SmtpServerPortTextBox.Text = PerfmonOptionsDefinition.GetSmtpPort(_PerfmonOptions.MailHub);
+                    if (null != _PerfmonOptions.MailLanguageCode)
```
MihaelaStoica: MailLanguageCode always returns null because the language selection is not saved if changed from XenCenter. It needs to be added to PerfmonOptionsDefinitionAction in order to be saved to the pool's other-config.

XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
```diff
@@ -145,7 +157,8 @@ public bool HasChanged
                 return ((_OrigEmailNotificationCheckBox != EmailNotificationCheckBox.Checked) ||
                         (_OrigEmailAddressTextBox != EmailAddressTextBox.Text) ||
                         (_OrigSmtpServerAddrTextBox != SmtpServerAddrTextBox.Text) ||
-                        (_OrigSmtpServerPortTextBox != SmtpServerPortTextBox.Text));
+                        (_OrigSmtpServerPortTextBox != SmtpServerPortTextBox.Text) ||
+                        (_OrigMailLanguageCodeComboBox != MailLanguageComboBox.SelectedValue.ToString()));
```
MihaelaStoica: `_OrigMailLanguageCodeComboBox != MailLanguageComboBox.SelectedValue.ToString()` always returns true, because _OrigMailLanguageCodeComboBox contains the KeyValuePair (e.g. "[en-US, English]") and SelectedValue contains just the language code (e.g. "en-US")

XenModel/Messages.resx
```diff
@@ -7161,6 +7161,24 @@ Size: {3}</value>
...
+    <value>en-US</value>
+  </data>
+  <data name="MAIL_LANGUAGE_ENGLISH_NAME" xml:space="preserve">
+    <value>English</value>
+  </data>
+  <data name="MAIL_LANGUAGE_JAPANESE_CODE" xml:space="preserve">
+    <value>ja-JP</value>
+  </data>
+  <data name="MAIL_LANGUAGE_JAPANESE_NAME" xml:space="preserve">
+    <value>Janpanese</value>
```
MihaelaStoica: There's a typo here. It should say "Japanese"

XenModel/Messages.zh-CN.resx
```diff
@@ -59,46 +59,46 @@
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
-  <xsd:schema id="root" xmlns="" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
-    <xsd:import namespace="http://www.w3.org/XML/1998/namespace"/>
-    <xsd:element msdata:IsDataSet="true" name="root">
+  <xsd:schema id="root" xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
```
MihaelaStoica: There is no need to add the new strings to the Chinese and Japanese resource files. the localization team will do this as part of the translation process.

XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
makunterry: Yes. Those logic part will be covered by CP-19453. So for this ticket it's OK.

XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
makunterry: Yes you are right. I made mistake for the "_OrigMailLanguageCodeComboBox" assignment, I should use SelectedValue.ToString()

makunterry:
Added one commit for covering CP-19453.

MihaelaStoica:
We should only expose the Mail language combobox for servers that have this feature implemented on the server-side.

XenModel/Alerts/PerfmonOptionsDefinition.cs
```diff
@@ -46,15 +46,79 @@ public class PerfmonOptionsDefinition
...
+            }
+        }
 
         private static readonly char[] mailHubDelim = new char[] { ':' };
         private readonly String mailHub;
         private readonly String mailDestination;
+        private readonly String mailLanguageCode;
 
-        public PerfmonOptionsDefinition(String mailHub, String mailDestination)
+        private static MailLanguageList ml_list = new MailLanguageList(new List<KeyValuePair<String, String>>() {
```
MihaelaStoica: I think this can be simplified by using a Disctionary<string, string> directly here, instead of the MailLanguageList class.

XenModel/Alerts/PerfmonOptionsDefinition.cs
```diff
@@ -117,11 +197,12 @@ public static PerfmonOptionsDefinition GetPerfmonOptionsDefinitions(IXenObject x
 
             string mailDestination = GetMailDestination(connection);
             string mailHub = GetSmtpMailHub(connection);
+            string mailLanguageCode = GetMailLanguageCode(connection);
 
-            if (mailDestination == null || mailDestination == null)
+            if (mailDestination == null || mailHub == null || mailLanguageCode == null)
```
MihaelaStoica: We can only force not null mailLanguageCode on the pools where this featue is implemented. Since we are going to do the version/feature implementation check later, we should probably remove the `mailLanguageCode == null` check for now.

XenModel/Messages.resx
```diff
@@ -7161,6 +7161,24 @@ Size: {3}</value>
...
+  <data name="MAIL_LANGUAGE_CHINESE_CODE" xml:space="preserve">
+    <value>zh-CN</value>
+  </data>
+  <data name="MAIL_LANGUAGE_CHINESE_NAME" xml:space="preserve">
+    <value>Chinese(S)</value>
+  </data>
+  <data name="MAIL_LANGUAGE_ENGLISH_CODE" xml:space="preserve">
+    <value>en-US</value>
+  </data>
+  <data name="MAIL_LANGUAGE_ENGLISH_NAME" xml:space="preserve">
```
MihaelaStoica: We should add a space character before the brackers, e.g. _English (Unites States)_.
Also, for Chinese, wouldn't be better to say _Chinese (Simplified)_ instead of _Chinese (S)_?

makunterry:
@MihaelaStoica -- I will do the "expose the Mail language combobox for servers that have this feature implemented on the server-side" job on another pull request. So for this one if no concern, please help to merge. Thanks a lot!

makunterry:
hi @MihaelaStoica -- I updated this PR just now and include the ticket CP-21658. Please have a look. Thanks!

XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
```diff
@@ -107,33 +113,74 @@ private bool IsValidSmtpAddress()
...
             Repopulate();
 
             // Save original settings for change detection
             _OrigEmailNotificationCheckBox = EmailNotificationCheckBox.Checked;
             _OrigEmailAddressTextBox = EmailAddressTextBox.Text;
             _OrigSmtpServerAddrTextBox = SmtpServerAddrTextBox.Text;
             _OrigSmtpServerPortTextBox = SmtpServerPortTextBox.Text;
+
+            if (_bSupportMailLanguage)
+                _OrigMailLanguageCodeComboBox = null == MailLanguageComboBox.SelectedValue ? null : MailLanguageComboBox.SelectedValue.ToString();
```
MihaelaStoica: I think that the _OrigMailLanguageCodeComboBox should contain the _PerfmonOptions.MailLanguageCode as it is saved on the pool, not combobox's selected value, which in many cases is defaulted to something. because we nned to save it even in the case when the user has not changed the default value.

makunterry:
@MihaelaStoica -- Updated PR. Please help to have a look. Thanks!

XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
```diff
@@ -150,8 +141,8 @@ public void SetXenObjects(IXenObject orig, IXenObject clone)
             _OrigSmtpServerAddrTextBox = SmtpServerAddrTextBox.Text;
             _OrigSmtpServerPortTextBox = SmtpServerPortTextBox.Text;
 
-            if (_bSupportMailLanguage)
-                _OrigMailLanguageCodeComboBox = null == MailLanguageComboBox.SelectedValue ? null : MailLanguageComboBox.SelectedValue.ToString();
+            if (_bSupportMailLanguage && MailLanguageComboBox.SelectedValue != null)
+                _OrigMailLanguageCodeComboBox = MailLanguageComboBox.SelectedValue.ToString();
```
MihaelaStoica: I propose to rename `_OrigMailLanguageCodeComboBox` to `_OrigMailLanguageCode` and save the original code into it (not the combobox's selected value).
So this code becomes:

```
if (_bSupportMailLanguage && _PerfmonOptions != null)
    _OrigMailLanguageCode = _PerfmonOptions.MailLanguageCode;
else
    _OrigMailLanguageCode = null;
```
Then, on the `HasChanged` function below, we will have:
 `(_bSupportMailLanguage && _OrigMailLanguageCode != MailLanguageComboBox.SelectedValue.ToString())`

PR #1532, opened 2017-04-07
CA-249824: Unable to create vlan, bond using XenCenter

MihaelaStoica:
- set managed=true, otherwise the network creation fails with the error "The network is not managed by xapi"


stephen-turner:
Pulling. I have verified that this covers all calls to Network.create() and Network.async_create().

PR #1535, opened 2017-04-10
CA-223406 - sr-scan with NFS4 broken

letsboogey:


GaborApatiNagy:
I'll look into this tomorrow morning as agreed with @letsboogey

GaborApatiNagy:
I think some more error handling is needed on the Location page. 

If I enter a Server name that is invalid/not accessible, there is no error message shown at all. Before this change, there used to be a popup window saying:
image

This version does not seem to display any error message: (It shows "Scanning..." for a few seconds only)
image

XenAdmin/Wizards/NewSRWizard_Pages/SrWizardHelpers.cs
```diff
@@ -40,9 +40,15 @@ namespace XenAdmin.Wizards.NewSRWizard_Pages
 {
     public class SrWizardHelpers
     {
+        public static bool ValidateServerAddress(String address)
+        {
+            Regex re = new Regex(@"^[^/:*?<>|\\]+$");
```
GaborApatiNagy: I think we have a better way to validate it, see: 
https://github.com/xenserver/xenadmin/blob/master/XenAdmin/Dialogs/OptionsPages/ConnectionOptionsPage.cs#L191

Uri.CheckHostName(string)

GaborApatiNagy:
Thanks for the fixes @letsboogey 

I like that we keep the old screen for the previous versions and have the new layout for the new ones only.

I'm still at the initial UI testing - found these two issues, both with the new layout one:
* If the scan hasn't been successful, the subsequent controls shouldn't get enabled. (I see the error message is shown now, that's good.)
* I noticed that the label with the example gets updated if the text changes in the Share Name field. It should always match the version selection. I suggest you subscribe to the appropriate event of the the radio controls.

I'm still planning to look into the code in the meantime.

XenAdmin/Wizards/NewSRWizard.cs
```diff
@@ -374,7 +384,7 @@ protected override void UpdateWizardContent(XenTabPage senderPage)
                 m_srWizardType.DeviceConfig = xenTabPageNfsIso.DeviceConfig;
                 SetCustomDescription(m_srWizardType, xenTabPageNfsIso.SrDescription);
             }
-            else if (senderPagetype == typeof(VHDoNFS))
+            else if (senderPagetype == typeof(VHDoNFS) || senderPagetype == typeof(VHDoNFSv2))
```
GaborApatiNagy: Calling SetCustomDescription() for the new NFSv2 page is missing at 3 lines below

XenModel/Messages.resx
```diff
@@ -5126,6 +5126,12 @@ Would you like to eject these ISOs before continuing?</value>
   <data name="EXAMPLE" xml:space="preserve">
     <value>Example: {0}</value>
   </data>
+  <data name="EXAMPLE_PATH_NFS3" xml:space="preserve">
+    <value>Example: /export/nfs</value>
+  </data>
+  <data name="EXAMPLE_PATH_NFS4" xml:space="preserve">
+    <value>Example: /nfs</value>
```
GaborApatiNagy: To keep this similar to the previous versions, I suggest we have "/path" here.

XenAdmin/Wizards/NewSRWizard_Pages/SrWizardHelpers.cs
```diff
@@ -40,6 +40,21 @@ namespace XenAdmin.Wizards.NewSRWizard_Pages
 {
     public class SrWizardHelpers
     {
+        public static bool ValidateServerAddress(String address)
+        {
+            var serverNameType = Uri.CheckHostName(address);
+            if (serverNameType == UriHostNameType.Unknown || serverNameType == UriHostNameType.IPv6)
```
GaborApatiNagy: When the then- and the else-statement is just a return true/false, it's usually better to just return the condition itself. I'm fine with this now, but it'd be nicer to have a one-liner here.

XenAdmin/Wizards/NewSRWizard_Pages/Frontends/VHDoNFSNew.cs
```diff
@@ -0,0 +1,372 @@
...
+                {
+                    DisableControls();
+                    return;
+                }
+                                
+                NfsSharePathTextBox.Enabled = serverOptionsTextBox.Enabled = serverScanButton.Enabled = sharePathExampleLabel.Visible = true;
+                GetSupportedNfsVersionsAndSetUI(SrProbeAction.XMLFromErrorDescription);
+
+                //clear xml string otherwise it will persist across different probes and be reused if new one is not created, bypassing above if statement
+                SrProbeAction.XMLFromErrorDescription = null;
```
GaborApatiNagy: I can't see why we have a try-finally block here, but I'm fine with it.
However, if we have it around, this line is a typical example that should be in the finally block - to make sure it is always executed.
Therefore, I suggest we move this into the finally.

letsboogey:
Thanks for the review and comments @GaborApatiNagy . I have made the requested changes

XenAdmin/Wizards/NewSRWizard.cs
```diff
@@ -278,7 +280,12 @@ protected override void UpdateWizardContent(XenTabPage senderPage)
                 m_srWizardType = xenTabPageChooseSrType.SrWizardType;
 
                 if (m_srWizardType is SrWizardType_VhdoNfs)
-                    AddPage(xenTabPageVhdoNFS);
+                {
+                    if (Helpers.FalconOrGreater(xenConnection))
```
GaborApatiNagy: I think this check should be IvernessOrGreater

GaborApatiNagy:
Thanks @letsboogey , looks good now, merging

PR #1536, opened 2017-04-11
CA-249084

GaborApatiNagy:


XenAdmin/Wizards/DRWizards/DRFailoverWizardStoragePage.cs
```diff
@@ -332,8 +332,7 @@ private List<FibreChannelDevice> FiberChannelScan()
 
             Host master = Helpers.GetMaster(Connection);
 
-            if (master != null && (Helpers.HostBuildNumber(master) >= 9633
-                                || Helpers.HostBuildNumber(master) == Helpers.CUSTOM_BUILD_NUMBER))
+            if (master != null)
```
MihaelaStoica: We don't need the master check anymore and we should remove the `dconf[SrProbeAction.DEVICE] = device.Path;` completelly. That was only applicable for old versions.

XenAdmin/Wizards/NewSRWizard_Pages/SrWizardType.cs
```diff
@@ -62,8 +62,7 @@ public LvmOhbaSrDescriptor(FibreChannelDevice device, IXenConnection connection)
 
             // CA-19025: Change XenCenter SR.create for LVMoHBA to use the
             // updated SCSIid parameter rather than the device path
-            if (master != null && (Helpers.HostBuildNumber(master) >= 9633
-                                   || Helpers.HostBuildNumber(master) == Helpers.CUSTOM_BUILD_NUMBER))
+            if (master != null)
```
MihaelaStoica: As above, we don't need the old code with device path.

MihaelaStoica:
We should also fix the use of HostBuildNumber on the MainWindow.cs. Although it is just a log entry, we should use the build number as it is, without trying to convert it to integer first.
Actually, we should remove all usages of that function, and use string matching instead, where needed.

XenAdmin/Wizards/DRWizards/DRFailoverWizardStoragePage.cs
```diff
@@ -332,15 +332,10 @@ private List<FibreChannelDevice> FiberChannelScan()
...
```
MihaelaStoica: I think we don't need to resolve the master here. If the master is null, we have greater problems than the SCSIId being added to the DeviceConfig.

MihaelaStoica:
@GaborApatiNagy, could you please also resolve the merge conflicts with master?

XenAdmin/Wizards/DRWizards/DRFailoverWizardStoragePage.cs
GaborApatiNagy: 🤣 true... I'll remove that

XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -713,11 +713,15 @@ public virtual string PlatformVersion
...
+        /// From Falcon, this property is not used.
+        /// </summary>
+        /// <remarks>
         /// Return the build number of this host, or -1 if none can be found.  This will often be
         /// 0 or -1 for developer builds, so comparisons should generally treat those numbers as if
         /// they were brand new.
-        /// </summary>
-        public int BuildNumber
+        /// </remarks>
+        internal int BuildNumber
```
GaborApatiNagy: good idea

XenAdmin/Wizards/NewSRWizard_Pages/SrWizardType.cs
```diff
@@ -62,15 +62,10 @@ public LvmOhbaSrDescriptor(FibreChannelDevice device, IXenConnection connection)
...
```
MihaelaStoica: We should remove the master check from here as well

MihaelaStoica:
Looks good now. Merging.

PR #1537, opened 2017-04-18
CA-249056: Add a unique ID to XenCenter Check For Update

GaborApatiNagy:


stephen-turner:
What do typical user agents look like in your scheme. Do they correspond to the usual convention from other vendors?

GaborApatiNagy:
@stephen-turner Good point. At the moment it would be: "XenCenter (v.1.2.3)", but I think I should change it to "XenCenter/1.2.3" to make it standard and to comply with the format defined in https://tools.ietf.org/html/rfc7231#section-5.5.3

I'll make this change and update this PR now.

XenAdmin/Actions/GUIActions/DownloadUpdatesXmlAction.cs
```diff
@@ -293,6 +294,12 @@ protected virtual XmlDocument FetchCheckForUpdatesXml(string location)
             var xdoc = new XmlDocument();
             var uri = new Uri(location);
             var proxy = XenAdminConfigManager.Provider.GetProxyFromSettings(Connection, false);
+            
+            var uniqueIdHash = GetUniqueIdHash();
+
+            string buildText = Helpers.CommonCriteriaCertificationRelease
+                        ? string.Format("{0}: {1}", Program.Version.Revision, Messages.COMMON_CRITERIA_TEXT)
```
stephen-turner: We don't have any CC versions any more; but even if we did, I don't think we want the COMMON_CRITERIA_TEXT in the user-agent. Let's just lose it.

XenAdmin/Actions/GUIActions/DownloadUpdatesXmlAction.cs
```diff
@@ -293,6 +294,12 @@ protected virtual XmlDocument FetchCheckForUpdatesXml(string location)
             var xdoc = new XmlDocument();
             var uri = new Uri(location);
             var proxy = XenAdminConfigManager.Provider.GetProxyFromSettings(Connection, false);
+            
+            var uniqueIdHash = GetUniqueIdHash();
+
+            string buildText = Helpers.CommonCriteriaCertificationRelease
+                        ? string.Format("{0}: {1}", Program.Version.Revision, Messages.COMMON_CRITERIA_TEXT)
+                        : Program.Version.Revision.ToString();
```
stephen-turner: This just takes the build number. I think you want the three-digit version number instead.

XenAdmin/Actions/GUIActions/DownloadUpdatesXmlAction.cs
```diff
@@ -313,5 +322,26 @@ protected virtual XmlDocument FetchCheckForUpdatesXml(string location)
 
             return xdoc;
         }
+
+        private string GetUniqueIdHash()
+        {
+            string uniqueIdHash = "nil";
+
+            var managementObj = new System.Management.ManagementObject("Win32_OperatingSystem=@");
```
stephen-turner: I suppose all these system calls are OK, but I can't be sure that none of them can throw exceptions. Maybe we should wrap the whole of lines 330-342 in a try-catch block.

XenAdmin/Actions/GUIActions/DownloadUpdatesXmlAction.cs
```diff
@@ -313,5 +322,26 @@ protected virtual XmlDocument FetchCheckForUpdatesXml(string location)
...
+        private string GetUniqueIdHash()
+        {
+            string uniqueIdHash = "nil";
+
+            var managementObj = new System.Management.ManagementObject("Win32_OperatingSystem=@");
+            string serialNumber = (string)managementObj["SerialNumber"];
+
+            if (!string.IsNullOrWhiteSpace(serialNumber))
+            {
+                var serialBytes = Encoding.ASCII.GetBytes("JustSomeSaltForCheckForUpdates" + serialNumber);
```
stephen-turner: I'm not sure what the salt is doing here. I don't see any advantage of a publicly known salt, and it just makes the md5sum longer to compute.

XenAdmin/Actions/GUIActions/DownloadUpdatesXmlAction.cs
```diff
@@ -38,6 +38,7 @@
 using XenAdmin.Core;
 using System.Diagnostics;
 using System.Net;
+using System.Text;
```
stephen-turner: I think we should keep DownloadUpdatesXmlAction in XenModel, and pass in the user agent and user agent id that we want to use, rather than move it all to XenAdmin.

stephen-turner:
Please also step through the code and make sure that everything is behaving as expected.

stephen-turner:
http://stackoverflow.com/questions/3138893/detail-about-win32-operatingsystem-serialnumber and http://stackoverflow.com/questions/14213020/getting-windows-serial-number-was-getting-machineguid-from-registry raise doubts about the uniqueness of SerialNumber. We should consider carefully what level of uniqueness we need, and what call will best provide this.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -35,6 +35,7 @@
 using System.Threading;
 using System.IO;
 using XenCenterLib.Archive;
+using System.Management;
```
stephen-turner: Please remove unnecessary change

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -62,39 +63,44 @@ public List<XenServerVersion> XenServerVersionsForAutoCheck
...
                 }
                 return new List<XenServerVersion>();
             }
         }
 
-        private readonly bool _checkForXenCenter;
-        private readonly bool _checkForServerVersion;
-        private readonly bool _checkForPatches;
-        private readonly string _checkForUpdatesUrl;
+        private readonly bool checkForXenCenter;
```
stephen-turner: Please don't make any changes beyond what is necessary here

XenModel/XenModel.csproj
```diff
@@ -52,6 +52,7 @@
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
     <Reference Include="System.Drawing" />
+    <Reference Include="System.Management" />
```
stephen-turner: Please remove unnecessary change

stephen-turner:
Everything looks good on the server side, so merging.

GaborApatiNagy:
Great news, thanks for merging

On 26 Apr 2017 12:46, "Stephen Turner" <notifications@github.com> wrote:

Merged #1537 <https://github.com/xenserver/xenadmin/pull/1537>.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/xenserver/xenadmin/pull/1537#event-1058432643>, or mute
the thread

PR #1541, opened 2017-04-20
CA-248920: The radiobutton selection on Select Update page is not preserved while navigating to the previous page in the Install Update wizard

GaborApatiNagy:

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
@@ -54,6 +54,7 @@ namespace XenAdmin.Wizards.PatchingWizard
         private bool CheckForUpdatesInProgress;
         public XenServerPatchAlert SelectedUpdateAlert;
         public XenServerPatchAlert FileFromDiskAlert;
+        private bool firstLoad = true;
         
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -150,12 +151,14 @@ public override void PageLoaded(PageLoadedDirection direction)
                 labelWithAutomatedUpdates.Visible = automatedUpdatesOptionLabel.Visible = AutomatedUpdatesRadioButton.Visible = automatedUpdatesPossible;
                 labelWithoutAutomatedUpdates.Visible = !automatedUpdatesPossible;
 
-                AutomatedUpdatesRadioButton.Checked = automatedUpdatesPossible;
-                downloadUpdateRadioButton.Checked = !automatedUpdatesPossible;
+                AutomatedUpdatesRadioButton.Checked = firstLoad && automatedUpdatesPossible;
```
MihaelaStoica: This wil set Checked to false if `firstLoad` is false, which should not happen. We should only set Checked programatically on the firstLoad (the default values) and maybe when the AutomatedUpdatesRadioButton gets hidden while navigating through the pages (yhis might happen if some hosts get disconnected)

PR #1542, opened 2017-04-21
CA-249845: Compiling the server status reports sometime fails with Sy…

MihaelaStoica:
…stem.IO.IOException: The process cannot access the file because it is being used by another process

Add a retry mechanism around the File.Move, because the operation sometimes fails when there is an anti-virus running on the XenCenter machine (we are doing the move immediately after we finished writing the file and the anti-virus software might be checking the file at that point)

### DIFF
--- a/XenModel/XenAPI/HTTP.cs
+++ b/XenModel/XenAPI/HTTP.cs
@@ -705,14 +705,11 @@ namespace XenAPI
                 CopyStream(fileStream, requestStream, dataCopiedDelegate, cancellingDelegate);
             }
         }
-        
-        private const int FILE_MOVE_MAX_RETRIES = 5;
-        private const int FILE_MOVE_SLEEP_BETWEEN_RETRIES = 100;
 
         /// <summary>
         /// A general HTTP GET method, with delegates for progress and cancelling. May throw various exceptions.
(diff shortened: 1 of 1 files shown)


MihaelaStoica:
Please note that these changes need to go in the xen-api-sdk repo as well. I will add the commit there after it gets reviewed and approved here.

stephen-turner:
I would like to understand why we download to a temporary file and then Move at all, rather than just downloading directly to the correct place. There's surely a reason, but I don't know what it is.

stephen-turner:
We have some other instances of the same pattern in OVF export and XVA export. Are they subject to the same bug?

stephen-turner:
The code looks fine, but I would prefer to extract the "move with retries" functionality into a separate function in order to keep GET clean, and in case anything else wants to use it later.

MihaelaStoica:
The use of a temporary file has been around since the Http.Get function had been added some 10 years ago. I am not sure about the reason, but we decided to keep it as it is.
Regarding the other usages of File.Move, I will check the VM export functions and create a separate ticket if they are affected.

stephen-turner:
OK, thanks, merging now.

PR #1559, opened 2017-04-26
CP-21782: Backport the fix for SCTX-2517 (CA-249624) to ely/bugfix

MihaelaStoica:
CA-249624: [XSO-707] Convert VM to Template from XenCenter is not working if connected as an AD user

- Reset the action's Session after the PureAsyncAction is run to collect the RBAC method list.

Also fixes the bug where AD users cannot do multiple operations on the VMs [SCTX-2517]


MihaelaStoica:
Merged this myself as it has been reviewed on #1529

PR #1562, opened 2017-04-27
CA-250145: XenCenter cannot shutdown VM as VmPowerAdmin user if HA is…

MihaelaStoica:
… enabled

The error was happening because of the Pool.sync_database call (when HA is enabled and the VM has a saved restart priority other than DoNotRestart) which requires pool admin or pool operator roles.
This call is not needed, so removing it.


stephen-turner:
Merging.

PR #1563, opened 2017-04-28
CA-251616: Remove the use of the additional branding repository

MihaelaStoica:


kc284:
Since we now clone branding.git at line 139 for all cases, the SERVER_BRANDING_TIP at line 212 will always require calculation.

MihaelaStoica:
Thanks @kc284, I've added a new commit with these changes.

PR #1565, opened 2017-04-28
CA-250235 - XC is allowing user to configure Health check even with W…

letsboogey:
…rong XenServer Credentials

### DIFF
--- a/XenAdmin/Dialogs/HealthCheck/HealthCheckSettingsDialog.cs
+++ b/XenAdmin/Dialogs/HealthCheck/HealthCheckSettingsDialog.cs
@@ -168,7 +168,7 @@ namespace XenAdmin.Dialogs.HealthCheck
 
         private void UpdateButtons()
         {
-            okButton.Enabled = m_ctrlError.PerformCheck(CheckCredentialsEntered);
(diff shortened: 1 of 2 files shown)


XenAdmin/Dialogs/HealthCheck/HealthCheckSettingsDialog.cs
```diff
@@ -316,7 +319,6 @@ private void credentials_TextChanged(object sender, EventArgs e)
 
         private void xsCredentials_TextChanged(object sender, EventArgs e)
         {
-            UpdateButtons();
```
MihaelaStoica: By removing this line, the OK button stays disabled even after the XS credentials have been entered (but not checked with the "Test Ctedentials" button) 
Was this change intentional?

letsboogey: Yes this was intentional, to only allow the user to click ok after the credentials have been checked..since we couldn't do the check when the ok button is pressed

MihaelaStoica: But would the user know that they need to press "Test credentials" in order for the OK button to become enabled?

PR #1569, opened 2017-05-02
Add support for EXT_KEY_EVENT encoding in XenCenter VNC

anoobs:


stephen-turner:
Thanks, @anoobs. I haven't studied the changeset, but does it continue to work properly for older hosts too?

anoobs:
@stephen-turner : yes, it works without issues for older hosts.

PR #1573, opened 2017-05-03
download_packages: Check version of powershell before executing

anoobs:
Windows 7 (without sp1) comes with powershell v2.x and
download_packages.ps1 doesn't work with powershell v2.x.

Check for powershell version and bail out if Major version is less than
3.

Signed-off-by: Anoob Soman <anoob.soman@citrix.com>


MihaelaStoica:
Thanks @anoobs for this.

PR #1575, opened 2017-05-04
CA-251526 - Change template order in New VM wizard

letsboogey:

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/Page_Template.cs
+++ b/XenAdmin/Wizards/NewVMWizard/Page_Template.cs
@@ -301,6 +301,7 @@ namespace XenAdmin.Wizards.NewVMWizard
             switch (templateType)
             {
                 case VM.VmTemplateType.Windows:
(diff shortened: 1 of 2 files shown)


XenAdmin/Wizards/NewVMWizard/Page_Template.cs
```diff
@@ -302,6 +302,8 @@ public static string ToDisplayString(this VM.VmTemplateType templateType)
             {
                 case VM.VmTemplateType.Windows:
                     return Messages.NEWVMWIZARD_TEMPLATEPAGE_WINDOWS;
+                case VM.VmTemplateType.LegacyWindows:
```
MihaelaStoica: This (and similarly on line 345) can also be written as:
```
case VM.VmTemplateType.Windows:
case VM.VmTemplateType.LegacyWindows:
    return Messages.NEWVMWIZARD_TEMPLATEPAGE_WINDOWS;
```

PR #1576, opened 2017-05-04
CA-251630: XenCenter update wizard incorrectly states that some hosts…

MihaelaStoica:
… need rebooting and requires suspending VMs

Perform all the prechecks on applicable hosts only (i.e. the hosts that don't have the update applied already), with the following exceptions:
- The HA check, which is performed on the pool master
- The server-side precheck (PatchPrecheckCheck) which will still be performed on all servers in the pool and will show the warning if the update has already been applied on some servers

This commit also adds an override to the HAOffCheck.SuccessfulCheckDescription to show the pool name instead of the master's if the check is successful.
Also, in the PatchPrecheckCheck, I had moved the code that checks if the patch is already applied to the very beginning (before the host liveness check), so it will return the warning that the server will be skipped even if the server is not reachable

### DIFF
--- a/XenAdmin/Diagnostics/Checks/HAOffCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/HAOffCheck.cs
@@ -69,5 +69,14 @@ namespace XenAdmin.Diagnostics.Checks
                 return Messages.HA_CHECK_DESCRIPTION;
             }
         }
+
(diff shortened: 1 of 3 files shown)


GaborApatiNagy:
merging now (reviewed and accepted it before the retrospective)

PR #1580, opened 2017-05-04
CP-21997: Remove CFUValidator from the build script

MihaelaStoica:


stephen-turner:
The code has been deleted, so it seems reasonable not to try and build it...

PR #1585, opened 2017-05-08
CA-242134: Remove the code targetting <= 6.1 XenServers

MihaelaStoica:


XenAdmin/Dialogs/LicenseManager/LicenseStatus.cs
```diff
@@ -368,8 +365,7 @@ public string LicenseEntitlements
                     return Messages.LICENSE_NOT_ELIGIBLE_FOR_SUPPORT;
                 }
 
-                if ((PoolLicensingModel == LicensingModel.Creedence || PoolLicensingModel == LicensingModel.Clearwater)
-                    && CurrentState == HostState.Free)
+                if (CurrentState == HostState.Free)
```
kc284: If the enum LicensingModel acquires one more member, will this line still be correct? I presume the author meant `if (PoolLicensingModel != LicensingModel.PreClearwater)`, right?

kc284:
What has been changed above looks ok to me.
I also think there is some more code that needs looking at:
- WLB cheks starting from the definitions in XenAdmin/Wlb/WlbPoolConfiguration.cs.
- Host.IsXCP extension?
- What about the old VM restart priorities?

XenAdmin/Dialogs/LicenseManager/LicenseStatus.cs
MihaelaStoica: If the enum acquires one more member, then we'll have to revisit all its usages, but it should still be correct, because in general, if a pool is not licensed, then it is not eligible for support. And yes, I think the author (probably me 😄) meant `if (PoolLicensingModel != LicensingModel.PreClearwater)`

MihaelaStoica:
Thanks @kc284, I will add your suggestions of things to remove to the list on the ticket.

kc284:
Now that the conflict is resolved I'll merge this one.

PR #1586, opened 2017-05-10
CA-252877: Precheck actions not reverted on upgrading to new XenServer version using an update

GaborApatiNagy:
Automated Updates now undo pre-check actions once it's finished wit the updates


XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -177,9 +178,13 @@ public override void PageLoaded(PageLoadedDirection direction)
                     }//patch
                 }
 
+                //add a revert pre-check action for this pool
```
MihaelaStoica: I think we should add a final action for a pool only if there are any prechecks to revert for that pool

GaborApatiNagy:
Thanks for the comments @MihaelaStoica, I've made the requested code changes.

PR #1587, opened 2017-05-10
CA-250234 - XenCenter hit with NullReferenceException at HAPage

letsboogey:

### DIFF
--- a/XenAdmin/TabPages/HAPage.cs
+++ b/XenAdmin/TabPages/HAPage.cs
@@ -198,7 +198,7 @@ namespace XenAdmin.TabPages
                 else
                 {
                     //CA-250234 no need to rebuild HA page if we lost connection to the pool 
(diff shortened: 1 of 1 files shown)


XenAdmin/TabPages/HAPage.cs
```diff
@@ -197,6 +197,10 @@ private void Rebuild()
                 }
                 else
                 {
+                    //CA-250234 no need to rebuild HA page if we lost connection to the pool 
+                    if (pool.Connection == null)
```
MihaelaStoica: This is not enough here, as Connection can be non-null but acually not connected. I suggest:
'if (connection == null || connection.Session == null || !connection.IsConnected)'

letsboogey: Thanks for the review, I have made the check more robust

PR #1588, opened 2017-05-11
CA-249849 - Using arrows to select license in license manager are the…

letsboogey:
… wrong way

### DIFF
--- a/XenAdmin/Dialogs/AssignLicenseDialog.resx
+++ b/XenAdmin/Dialogs/AssignLicenseDialog.resx
@@ -247,7 +247,7 @@
     <value>304, 23</value>
   </data>
   <data name="licenseServerNameTextBox.TabIndex" type="System.Int32, mscorlib">
-    <value>2</value>
+    <value>1</value>
   </data>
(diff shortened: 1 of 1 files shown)


kc284:
This change is ok and does work; a small improvement would be to start the radio button indices from 0. The tab indices do not need to be unique in the main control they appear, but rather in the container where they sit. 
One way to check this is open the designer and click View>Tab Order on the menu item and then click on the radio button: Visual Studio will assign automatically the tab index.

letsboogey:
Thanks for the review and Visual Studio tip @kc284 , i'll make that improvement and update the PR

XenAdmin/Dialogs/AssignLicenseDialog.resx
```diff
@@ -685,7 +685,7 @@
     <value>525, 201</value>
   </data>
   <data name="editionsGroupBox.TabIndex" type="System.Int32, mscorlib">
-    <value>2</value>
+    <value>0</value>
```
kc284: This should stay 2 because the Groupbox sits inside the main container and there is the top label and the server planel before it.

letsboogey: Done!

PR #1591, opened 2017-05-11
CA-205187 - L10n: JA/SC: The websites are not shown as hyperlink on t…

letsboogey:
…he “Health Check Privacy Statement” window


kc284:
Not stated on the ticket, but when I tested this and attempted to launch the privcay statement from the HealthCheckSettingsDialog I noticed that the tab order of the contols at the top, namely LinkLabel, Cehckbox and schedule panel, is wrong.

kc284:
I've tested the changes; they are ok with me. @letsboogey will you add a new commit about the above comment or would you like a different ticket?

letsboogey:
@kc284 I will fix the tab order and update this pr thanks

PR #1592, opened 2017-05-12
CA-252923: Update apply on slave host fails with pool_patch or pool_update not found

GaborApatiNagy:
In rare cases, the GetMaster could return null, causing the update progress to fail with "Pool_patch or Pool_update not found". (when the Cache hasn't been populated by the execution gets here)
This commit drops GetMaster and uses a pre-saved masterUuid instead of waiting for the Cache what would be the other way to solve this issue.


MihaelaStoica:
Closing this as we decided that this will be included in a merge from master into REQ-411

PR #1596, opened 2017-05-17
CP-22336: Backport CA-249056 (CFU unique ID) from master

GaborApatiNagy:
Backport CA-249056 #1537 (CFU unique ID) from master to Honolulu

Minor conflicts were caused by the lack of the proxy-improvement code that is present in master (and not in Honolulu).

As part of the conflict resolution, I had to switch to use WebClient instead of HTTPHelper.GET in DownloadUpdatesXmlAction as this feature heavily relies on that class. On the source branch, this change had been made by the proxy code before the changes (that are being backported) were implemented. 
Note that if the proxy code ever gets merged into this (or a future branch based on this) branch, this is not expected to cause any difficulty.

### DIFF
--- a/CFUValidator/Updates/AlternativeUrlDownloadUpdatesXmlSourceAction.cs
+++ b/CFUValidator/Updates/AlternativeUrlDownloadUpdatesXmlSourceAction.cs
@@ -45,7 +45,7 @@ namespace CFUValidator.Updates
         private readonly string newLocation;
 
         public AlternativeUrlDownloadUpdatesXmlSourceAction(string url)
-            : base(true, true, true)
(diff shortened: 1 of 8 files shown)


stephen-turner:
@GaborApatiNagy Given that you've had to implement it slightly differently, are you confident in the code on Honolulu? Have you dev tested it all again?

MihaelaStoica:
@GaborApatiNagy Please note that the ticket where we changed DownloadUpdatesXmlAction to use a WebClient (CA-241431) will also be included in Honolulu, as part of the REQ-411 backport.

GaborApatiNagy:
We didn't like this as is, so closing and merging REQ-411-Ely first and then the backporting will be done.

PR #1602, opened 2017-05-22
CA-220047: Show proper precheck error when a VM cannot be migrated

GaborApatiNagy:
Detect both VM_LACKS_FEATURE and VM_LACKS_FEATURE_SUSPEND, show proper error message on the pre-check page in the Install Updates Wizard

### DIFF
--- a/XenAdmin/Diagnostics/Checks/AssertCanEvacuateCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/AssertCanEvacuateCheck.cs
@@ -180,10 +180,10 @@ namespace XenAdmin.Diagnostics.Checks
                         return new NoPVDrivers(this, vm);
 
                     case Failure.VM_LACKS_FEATURE:
-                        return new CannotMigrateVM(this, vm, CannotMigrateVM.CannotMigrateVMReason.LacksFeature);
+                        return new CannotMigrateVM(this, vm, GetMoreSpecificReasonForCannotMigrateVm(vm, CannotMigrateVM.CannotMigrateVMReason.CannotMigrateVm));
 
                     case Failure.VM_LACKS_FEATURE_SUSPEND:
-                        return new CannotMigrateVM(this, vm, CannotMigrateVM.CannotMigrateVMReason.LacksFeatureSuspend);
+                        return new CannotMigrateVM(this, vm, GetMoreSpecificReasonForCannotMigrateVm(vm, CannotMigrateVM.CannotMigrateVMReason.LacksFeatureSuspend));
 
                     case "VM_OLD_PV_DRIVERS":
                         return new PVDriversOutOfDate(this, vm);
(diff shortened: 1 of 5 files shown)


XenAdmin/Diagnostics/Checks/AssertCanEvacuateCheck.cs
```diff
@@ -184,6 +184,22 @@ private Problem GetProblem(IXenConnection connection, XenRef<VM> vmRef, string[]
 
                         return new NoPVDrivers(this, vm);
 
+                    case Failure.VM_LACKS_FEATURE:
+                        vm = connection.Resolve<VM>(vmRef);
```
kc284: These lines that resolve the vm and throw exception if null are repeated through each case. Can we move them before the switch block, e.g.
```
VM vm = connection.Resolve<VM>(vmRef);
if (vm == null)
    throw new NullReferenceException(exception[0]);
```

XenModel/Messages.resx
```diff
@@ -11627,6 +11627,9 @@ Check your settings and try again.</value>
   <data name="UPDATES_WIZARD_CANNOT_MIGRATE_VM_LICENSE_REASON" xml:space="preserve">
     <value>Cannot migrate VM '{0}' due to license restrictions.</value>
   </data>
+  <data name="UPDATES_WIZARD_CANNOT_MIGRATE_VM_SUSPEND_REASON" xml:space="preserve">
+    <value>Cannot migrate VM '{0}', because it lacks the ability to suspend. Please ensure it has up-to-date drivers installed.</value>
```
kc284: Although intransitive usage of "suspend" does exist, I think transitive usage is more correct in this context. I would write simply "Cannot migrate VM '{0}', because it cannot be suspended. Please etc."

MihaelaStoica: Also, the second part of the message ("Please ensure it has up-to-date drivers installed") suggests that installing the drivers will fix the problem, but at the same time we offer "Shut down" as the solution, which is confusing.

XenAdmin/Diagnostics/Checks/AssertCanEvacuateCheck.cs
GaborApatiNagy: I see your point, I agree that would be better, but I think that shouldn't be part of a bugfix as it wouldn't be the minimal code change anymore - I only followed the existing pattern to avoid refactoring as much as possible.  Also, please note that the VM is not being resolved in the default case whereas it would be otherwise. 

I'd suggest we leave this as is (as part of this bugfix).

kc284: I see what you're saying regarding the default case, but I think resolving should be the first check because is it meaningful to have a VM problem without a resolved VM?
Beyond this, I think this method is somewhat inefficient because it throws exceptions only to catch, log, and rethrow them, and it could be written in a better way. Is this is a general trunk quality bug? If so, we could do some refactoring.

XenModel/Messages.resx
stephen-turner: Suggestion. If it doesn't have tools, say "Cannot migrate VM '{0}', because it doesn't have up-to-date drivers installed" (but continue to offer shut down as the resolution). If it does have tools, but still can't migrate, say just "Cannot migrate VM '{0}'".

XenAdmin/Diagnostics/Checks/AssertCanEvacuateCheck.cs
```diff
@@ -253,6 +229,25 @@ private Problem GetProblem(IXenConnection connection, XenRef<VM> vmRef, string[]
             }
         }
 
+        private static CannotMigrateVM.CannotMigrateVMReason GetMoreSpecificReasonForCannotMigrateVm(VM vm, CannotMigrateVM.CannotMigrateVMReason reason)
+        {
+            var gm = vm.Connection.Resolve(vm.guest_metrics);
```
kc284: This could go into the first if block to avoid resolving unecessarily, even if it's not an expensive operation, but it's ok, you don't have to change it.

kc284:
Looks ok to me.

PR #1605, opened 2017-05-24
Increase XenCenter micro version for honolulu

MihaelaStoica:


stephen-turner:
Seems sound

PR #1607, opened 2017-05-25
Fix the HealthCheck tests

MihaelaStoica:
The problem with these tests was that XenServerHealthCheck/ServerListHelper class is reading the server list form the user config file, and it also saves it there on every update. Although this tests eventually remove the servers added, sometimes (e.g. the test fails midway through) the added server (e.g. "host1") remained in the config file. 
With this fix, we firstly clear the server list.

### DIFF
--- a/XenAdminTests/HealthCheckTests/CredentialTests.cs
+++ b/XenAdminTests/HealthCheckTests/CredentialTests.cs
@@ -59,16 +59,26 @@ namespace XenAdminTests.HealthCheckTests
             string UserName = "User1";
             string Password = "password1";
 
-            //1. Empty list
+            // Empty list
             ServerListHelper.instance.ClearServerList();
             int conSize = 0;
             List<ServerInfo> con = ServerListHelper.instance.GetServerList();
             Assert.IsTrue(con.Count == conSize);
 
-            //2. Send credential and check result
+            //1. Empty credential 
             NamedPipeClientStream pipeClient = new NamedPipeClientStream(".", HealthCheckSettings.HEALTH_CHECK_PIPE, PipeDirection.Out);
             pipeClient.Connect();
-            string credential = EncryptionUtils.ProtectForLocalMachine(String.Join(SEPARATOR.ToString(), new[] { HostName, UserName, Password }));
+            string credential = EncryptionUtils.ProtectForLocalMachine(String.Join(SEPARATOR.ToString(), new[] { HostName, null, null }));
+            pipeClient.Write(Encoding.UTF8.GetBytes(credential), 0, credential.Length);
+            pipeClient.Close();
+            System.Threading.Thread.Sleep(1000);
+            con = ServerListHelper.instance.GetServerList();
+            Assert.IsTrue(con.Count == conSize);
+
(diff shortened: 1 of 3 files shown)


XenAdminTests/HealthCheckTests/CredentialTests.cs
```diff
@@ -58,20 +58,17 @@ public void CredentialOp()
...
             string UserName = "User1";
             string Password = "password1";
-            int conSize = ServerListHelper.instance.GetServerList().Count;
-            //1. Empty credential 
-            NamedPipeClientStream pipeClient = new NamedPipeClientStream(".", HealthCheckSettings.HEALTH_CHECK_PIPE, PipeDirection.Out);
-            pipeClient.Connect();
-            string credential = EncryptionUtils.ProtectForLocalMachine(String.Join(SEPARATOR.ToString(), new[] { HostName, null, null }));
-            pipeClient.Write(Encoding.UTF8.GetBytes(credential), 0, credential.Length);
-            pipeClient.Close();
+
```
kc284: As discussed, I think we may want to keep the original testcase and additionally ensure beforehand that the list is empty. Also, we should add the sleep before the assertion.
Now, we may want to reconsider the way we handle these nulls in the code, because the current behaviour is a bit strange. Perhaps ignore them, if we don't want to explicitly remove them.

MihaelaStoica:
@kc284 I've added some more changes, as we discussed.

XenAdminTests/HealthCheckTests/CredentialTests.cs
```diff
@@ -58,14 +58,21 @@ public void CredentialOp()
...
+            Assert.IsTrue(con.Count == conSize);
+
             //1. Empty credential 
             NamedPipeClientStream pipeClient = new NamedPipeClientStream(".", HealthCheckSettings.HEALTH_CHECK_PIPE, PipeDirection.Out);
             pipeClient.Connect();
             string credential = EncryptionUtils.ProtectForLocalMachine(String.Join(SEPARATOR.ToString(), new[] { HostName, null, null }));
             pipeClient.Write(Encoding.UTF8.GetBytes(credential), 0, credential.Length);
             pipeClient.Close();
-            List<ServerInfo> con = ServerListHelper.instance.GetServerList();
+            System.Threading.Thread.Sleep(1000);
```
Zhengchai: Why sleep hardcoded 1k second?

XenServerHealthCheck/ServerListHelper.cs
```diff
@@ -165,13 +165,28 @@ public bool UpdateServerInfo(ServerInfo server)
...
+        }
+
         public void UpdateServerCredential(string credential)
         {
             log.Info("Receive credential update message");
 
             string decryptCredential = EncryptionUtils.UnprotectForLocalMachine(credential);
             string[] decryptCredentialComps = decryptCredential.Split(SEPARATOR);
 
+            if (decryptCredentialComps.Length != 1 && decryptCredentialComps.Length != 3)
```
Zhengchai: Would this a good practice to hardcode the length inside? I am not sure, just wondering.

PR #1609, opened 2017-05-26
CA-252871 - Updates order is not preserved on Exporting updates to a …

letsboogey:
…file

Applies sort order of DataGridView to the unfiltered list of updates before
exporting all to file, thereby preserving the update order in the file


MihaelaStoica:
Looks good. Merging.

PR #1618, opened 2017-05-31
CA-242134: Remove Host.IsXCP extension

MihaelaStoica:


XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -969,9 +969,6 @@ private void generateLicenseBox()
 
             PDSection s = pdSectionLicense;
 
-            if (host.license_params == null || host.IsXCP)
-                return;
```
stephen-turner: Do we need to retain the host.license_params condition?

stephen-turner:
Thanks, pulling now.

PR #1629, opened 2017-06-06
CA-231518: Manage updates page - the last datagridview column is not …

MihaelaStoica:
…autoresized properly after dismissing all the updates

- Change the AutoSizeMode to ColumnHeader if the grid is empty, to force the column to auto resize

Signed-off-by: Mihaela Stoica <mihaela.stoica@citrix.com>

PR #1628, opened 2017-06-06
CA-231518 - Manage updates page - issues with dismissal

letsboogey:
Signed-off-by: Letsibogo Ramadi <letsibogo.ramadi@citrix.com>

PR #1630, opened 2017-06-07
One more take to CA-231518

kc284:
@letsboogey and @MihaelaStoica, after reviewing PRs #1628  and #1629, I think that we should implement the same method as the other two pages. The proposed solution fixes both the resizing issue and the multiple gridview rebuilding when we dismiss updates. Please have a look and alert me to any caveats that may have escaped me.

### DIFF
--- a/XenAdmin/TabPages/ManageUpdatesPage.cs
+++ b/XenAdmin/TabPages/ManageUpdatesPage.cs
@@ -62,6 +62,7 @@ namespace XenAdmin.TabPages
         private List<string> selectedUpdates = new List<string>();
         private int checksQueue;
         private bool CheckForUpdatesInProgress;
+        private readonly CollectionChangeEventHandler m_updateCollectionChangedWithInvoke;
 
         public ManageUpdatesPage()
         {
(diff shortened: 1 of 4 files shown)


XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -502,8 +516,7 @@ into g
         private void DeleteAllAllertAction_Completed(ActionBase sender)
         {
             Program.Invoke(Program.MainWindow, () =>
-            {                
-                Rebuild();
```
letsboogey: Thank you for the fix @kc284 . I think these changes are good and dismiss updates now works as intended with the columns resizing as necessary. However there is a regression introduced by removing this Rebuild call. It is needed to set the 'No Updates Found' message/warning on the grid after all the alerts are dismissed. Maybe a full rebuild is not necessary, but the bit of code that sets the icon and message are needed, because at the moment it just shows an empty grid with no message.

letsboogey: As a sidenote : An interesting discovery I also made about the above is that when you dismiss all updates from another instance on XenCenter (without the changes), the message is displayed, and the column does not get resized at all

kc284: @letsboogey I've been looking at restoring the message on dismissal, however I'm wondering now whether it's correct to show it: the user has requested dismissal and not a check, so isn't it confusing to show immediately 'No updates found'? Maybe it's ok to have an empty grid and show the message later if the user requests a new check or any other action that requires a refresh?

kc284:
I am aware of this change, the reason I made it is that there was an inconsistency between the state of the 'Check for updates' link and the 'Check for updates' button, which I found strange/confusing. I see your point, in which case I think that if we hide the button we should also hide/disable the link. Any concerns about this?

letsboogey:
Yes I think it would be okay to hide both of them. Don't see any negative effects because they don't work any way when you have dismissed all

letsboogey:
Following the discussion we had in sprint planning, it was determined that the button is necessary so I will now merge this

PR #1636, opened 2017-06-12
Private/falcon lcm/new guest support

minli1:
Backport new linux guest support for falcon LCM.


cheng-z:
Looks good to me. 
Hi @MihaelaStoica , please help to check if the branch I create is correct or not?
Although we don't need to have a hotfix now, it will be save to have the lcm branch with these change.
So I created lcm branch instead of rebrand branch.
Thanks.

cheng-z:
Thanks for change the branch @MihaelaStoica 
Do you think we can merge it now?

PR #1640, opened 2017-06-13
Modified the help script used for downloading dependencies so that the domain…

kc284:
can be passed in as a parameter and authentication is not required. For this
purpose the artifactory download patterns had to be expanded so as to be
understood by wget.

### DIFF
--- a/devtools/download_packages.ps1
+++ b/devtools/download_packages.ps1
@@ -29,41 +29,34 @@
 # SUCH DAMAGE.
 
 # help script to download third party binaries to local dev environment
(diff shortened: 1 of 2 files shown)


kc284:
Apart from the usual reviewers, could @jonludlam also check the modified script suits the dev-vm's needs. Thanks.

letsboogey:
The changes look sound to me, will wait for @jonludlam to comment on suitability to dev-vm's needs before merging.

jonludlam:
Yep, it seems to work very nicely. Thanks!

PR #1641, opened 2017-06-13
CA-220169 - Network heartbeat status initially shows 'Error' after en…

letsboogey:
…abling HA

Network HB initially shown as error because it takes time for the status to be
read from xha by xapi. This adds a delay to allow the status to be updated
before XC displays it.

### DIFF
--- a/XenAdmin/TabPages/HAPage.cs
+++ b/XenAdmin/TabPages/HAPage.cs
@@ -52,11 +52,13 @@ namespace XenAdmin.TabPages
 
         private IXenObject xenObject;
 
-        private bool newPoolHABox;
+        private bool restartHBInitializationTimer;
 
-        private bool HBInitTimeElapsed;
+        private bool initializationDelayElapsed;
 
-        private System.Timers.Timer networkHBInitDelay;
+        private System.Timers.Timer initializationDelayTimer;
+
+        private const int HB_INITIALIZATION_DELAY = 30000;
 
         private readonly CollectionChangeEventHandler Host_CollectionChangedWithInvoke;
         /// <summary>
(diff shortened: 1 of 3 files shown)


XenModel/Messages.resx
```diff
@@ -5760,6 +5760,9 @@ not currently live:
   <data name="HA_HEARTBEAT_SERVERS" xml:space="preserve">
     <value>{0} out of {1} servers</value>
   </data>
+  <data name="HA_HEARTBEAT_SERVERS_INITIALISING" xml:space="preserve">
+    <value>Initialising...</value>
```
MihaelaStoica: This should say "Initializing..." (with a z)

XenAdmin/TabPages/HAPage.cs
```diff
@@ -52,6 +52,12 @@ internal partial class HAPage : BaseTabPage
 
         private IXenObject xenObject;
 
+        private bool newPoolHABox;
```
MihaelaStoica: Could we have a better name for this? 'newPoolHABox' doesn't suggest its usage, which is "if this is true I need to restart the initialising timer"
Also, I suggest the camelCasing for the private fields (HBInitTimeElapsed)

MihaelaStoica: Also, the HBInitTimeElapsed and the timer could be renamed so that they have similat names (e.g. initializationDelayElapsed and initializationDelayTimer)

XenAdmin/TabPages/HAPage.cs
```diff
@@ -444,27 +462,28 @@ private void generatePoolHABox(Pool pool)
                 l.Padding = new Padding(0, 5, 0, 5);
                 l.Font = BaseTabPage.ItemValueFont;
                 l.AutoSize = true;
-                if (members[i].ha_network_peers.Length == members.Count)
+                l.ForeColor = (members[i].ha_network_peers.Length == members.Count && HBInitTimeElapsed) ? Color.Green : l.ForeColor = BaseTabPage.ItemValueForeColor;
```
letsboogey: Oops, yes.. cut & paste error!

letsboogey:
Thanks for the review @MihaelaStoica , I've made the requested changes

PR #1647, opened 2017-06-14
CA-249858: Fix build warning as error

MihaelaStoica:
(The member 'CanExecute()' does not hide an inherited member. The new keyword is not required)


kc284:
Will this compile without warnings? Both classes derive from CrossPoolMigrateCommand which has the same method.

kc284:
Ignore above comment, I've jsut noticed the parent class method has an additional optional parameter. Merging.

PR #1648, opened 2017-06-14
CA-255547: WLB code in XenCenter: Remove all code targetting <= 6.1 XenServers

GaborApatiNagy:
Fixed the Release build errors and removed some more dead code that's just been found


XenAdmin/Controls/Wlb/WlbReportView.cs
```diff
@@ -142,12 +141,6 @@ public IEnumerable<Host> Hosts
             }
         }
 
-        public bool IsCreedenceOrLater
```
MihaelaStoica: I dont't think that IsCredenceOrGreater should be removed, because we are still supporting Clearwater, which is pre-Creedence, so if there is functionality which only applies to Creedence or greater, then we should keep it that way.

PR #1649, opened 2017-06-15
CA-247333 - Remove rdpclient6 code after April 2017 

letsboogey:


XenAdmin/ConsoleView/RdpClient.cs
```diff
@@ -211,17 +179,17 @@ public void UpdateDisplay(int width, int height, Point locationOffset)
 
         private bool Connected
         {
-            get { return rdpControl == null ? false : (rdpClient9 == null ? rdpClient6.Connected == 1 : rdpClient9.Connected == 1); }
+            get { return rdpControl == null ? false : rdpClient9.Connected == 1; }
```
kc284: Style: this could be better written as `rdpControl != null && rdpClient9.Connected == 1`

XenAdmin/ConsoleView/RdpClient.cs
```diff
@@ -78,26 +75,26 @@ internal RdpClient(ContainerControl parent, Size size, EventHandler resizeHandle
...
             {
                 if (parent.Controls.Contains(rdpControl))
                     parent.Controls.Remove(rdpControl);
+                rdpControl.Dispose();
+                rdpControl = null;
                 rdpClient9 = null;
-                rdpControl = rdpClient6 = new MsRdpClient6();
-                RDPConfigure(size);
-                parent.Controls.Add(rdpControl);
+                Log.Error("Adding rdpControl to parent.Controls list caused an exception.", ex);
```
kc284: Can we move the Resize registration inside the try block please ; if something goes wrong and the catch block is hit, rdpControl will be null.

letsboogey: Oops, good catch, I don't know what it was doing out there!

XenAdmin/ConsoleView/RdpClient.cs
```diff
@@ -55,11 +55,8 @@ class RdpClient: IRemoteConsole
         /// </summary>
         private MsRdpClient9 rdpClient9 = null;
 
-        private MsRdpClient6 rdpClient6 = null;
-
         /// <summary>
-        /// This will be equal to rdpClient9, if the DLL that we've got is version 8, otherwise equal to
-        /// rdpClient6.
+        /// This will be equal to rdpClient9
```
MihaelaStoica: I dont't think we need both rdpControl and rdpClient9 members, because now they are always equal.

kc284: Hmm, yes, I was thinking about it, but perhaps we may want to leave it there in case there are new guests in future for which a new "rdpClientx" is required?

kc284:
Looks good to me.

PR #1650, opened 2017-06-15
CA-243675: No Clear error message when proxy server which is configured is not running

GaborApatiNagy:
Mention proxy in the suggestion that is shown when a connection fails


MihaelaStoica:
@GaborApatiNagy These changes look good to me, but I had noticed that Messages.resx became unsorted and therefore new strings are added at the end. Could you please sort it?

PR #1652, opened 2017-06-16
CP-22472: Add new XenCenter metadata file to the server status report 

MihaelaStoica:


XenAdmin/Core/Telemetry.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            var metadata = new XenCenterMetadata
+            {
+                System = new SystemInfo
+                {
+                    Version = Assembly.GetExecutingAssembly().GetName().Version.ToString(),
+                    DotNetVersion = Environment.Version.ToString(4),
+                    Culture = Thread.CurrentThread.CurrentUICulture.EnglishName,
+                    OsVersion = Environment.OSVersion.ToString(),
+                    OsCulture = CultureInfo.InstalledUICulture.EnglishName,
```
stephen-turner: I think we should test this on a foreign OS, to make sure that OsCulture and (XenCenter) Culture are set correctly.

XenAdmin/Core/Telemetry.cs
```diff
@@ -0,0 +1,148 @@
...
+                    IpAddress = GetLocalIPAddress()
```
stephen-turner: This can be null. Are you confident of the behaviour of JavaScriptSerializer.Serialize() in that case?

XenAdmin/Core/HealthCheck.cs
```diff
@@ -125,5 +125,38 @@ private static void actionCompleted(ActionBase sender)
             if (CheckForAnalysisResultsCompleted != null)
                 CheckForAnalysisResultsCompleted(action.Succeeded);
         }
+
+        public static void SendProxySettingsToHealthCheck(bool suppressHistory)
```
stephen-turner: I think we should always suppress history for this

XenModel/Actions/HealthCheck/TransferXenCenterMetadata.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.IO.Pipes;
+using System.ServiceProcess;
+using System.Text;
+using XenAdmin.Core;
+using XenAdmin.Model;
+using XenAPI;
+
+namespace XenAdmin.Actions
+{
+    public class TransferXenCenterMetadata : PureAsyncAction
```
stephen-turner: We now have three different "send stuff to health check" actions. Can we combine them?

XenModel/Actions/HealthCheck/TransferXenCenterMetadata.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+                catch
+                {
+                    System.Threading.Thread.Sleep(1000);
+                    pipeClient = null;
+                    pipeClient = new NamedPipeClientStream(".", HealthCheckSettings.HEALTH_CHECK_PIPE, PipeDirection.Out);
+                }
+            } while (!pipeClient.IsConnected && retryCount-- != 0);
+
+            byte[] metadataBytes = Encoding.UTF8.GetBytes(String.Join(SEPARATOR.ToString(), HealthCheckSettings.XENCENTER_METADATA, metadata));
```
stephen-turner: Is it possible that we have an old health check service installed? What's going to happen if we then send it this new type of data that it doesn't know about?

XenAdmin/Core/HealthCheck.cs
```diff
@@ -125,5 +125,38 @@ private static void actionCompleted(ActionBase sender)
...
+        /// Sends XenCenter metadata to the Health Check service; will not show the action progress.
+        /// </summary>
+        public static void SendMetadataToHealthCheck()
+        {
+            if (Program.Exiting)
+                return;
+            var metadata = XenAdminConfigManager.Provider.GetXenCenterMetadata();
+            if (!string.IsNullOrEmpty(metadata))
+            {
+                new TransferXenCenterMetadata(XenAdminConfigManager.Provider.GetXenCenterMetadata(), true).RunAsync();
```
stephen-turner: This creates the metadata a second time, rather than just using the variable you've already stored it in.

stephen-turner:
Also, there were some new items to add that we discussed

XenServerHealthCheck/ServerListHelper.cs
```diff
@@ -267,5 +267,24 @@ public void UpdateProxy(string proxy)
...
             }
         }
+
+        public void UpdateXenCenterMetadata(string message)
+        {
+            log.Info("Receive XenCenter metadata update message");
+
+            try
+            {
+                string[] metadata = message.Split(SEPARATOR);
```
stephen-turner: Can we prove that none of the metadata can contain this SEPARATOR? If not, we will have to escape it or change it somehow.

XenAdmin/Core/Telemetry.cs
MihaelaStoica: @stephen-turner In the case of null objects, the JavaScriptSerailizer will output `"IpAddress":null`. Is this OK? Otherwise I can change the GetLocalIPAddress() function to return an empty string instead of null.

XenServerHealthCheck/ServerListHelper.cs
MihaelaStoica: Good catch, I can't guarantee that the metadata won't contain the separator, especially as more items could be added in the future. Therefore I changed it to encode the metadata string before sending.

XenModel/Actions/HealthCheck/TransferXenCenterMetadata.cs
MihaelaStoica: An old Health Check service will ignore this new message. It will actually try to parse it as a "server list update" and fail silently, because the service expects a server list update message to be an encrypted string (while the metadata message contains the non-encrypted identifier followed by the encrypted metadata).

MihaelaStoica:
@stephen-turner I had added a new commit with the changes suggeated.

XenAdmin/Core/Telemetry.cs
stephen-turner: Let's have an empty string if possible, that one looks weird.

XenModel/Actions/HealthCheck/TransferXenCenterMetadata.cs
stephen-turner: Could we also change the Health Check service to allow better extensibility in future, because one day there might be a clash?

XenAdmin/Core/Telemetry.cs
MihaelaStoica: Ok, I've changed it in a new commit

XenModel/Actions/HealthCheck/TransferXenCenterMetadataAction.cs
```diff
@@ -0,0 +1,53 @@
...
+        public TransferXenCenterMetadataAction(string metadata, bool suppressHistory)
+            : base(null, Messages.ACTION_TRANSFER_HEALTHCHECK_SETTINGS, Messages.ACTION_TRANSFER_HEALTHCHECK_SETTINGS, suppressHistory)
+        {
+            this.metadata = metadata;
+        }
+
+        protected override string GetMessageToBeSent()
+        {
+            return string.Join(SEPARATOR.ToString(), HealthCheckSettings.XENCENTER_METADATA,
+                EncryptionUtils.ProtectForLocalMachine(metadata));
```
stephen-turner: Could we look at some sample protected data, and try and see whether the SEPARATOR could occur in it, or whether it just uses some restricted character set?

stephen-turner:
On reflection, I think we should include both the XenCenter time and the upload time in the report. For XenCenter uploads, they would be the same. For health check uploads, the XenCenter time would be the time the data was last transferred, but health check would insert the upload time when it uploaded the data.

MihaelaStoica:
@stephen-turner I had added the additional changes in a new commit

XenModel/Actions/HealthCheck/TransferXenCenterMetadataAction.cs
stephen-turner: (Note: the protected data only has "normal" characters in, so this is OK).

XenServerHealthCheck/ServerListHelper.cs
```diff
@@ -280,7 +281,9 @@ public void UpdateXenCenterMetadata(string message)
                 string[] metadata = message.Split(SEPARATOR);
                 if (metadata.Length != 2) 
                     return;
-                Properties.Settings.Default.XenCenterMetadata = EncryptionUtils.UnprotectForLocalMachine(metadata[1]);
+                var metadataString = EncryptionUtils.UnprotectForLocalMachine(metadata[1]);
+                metadataString = metadataString.Replace(HealthCheckSettings.REPORT_TIME_PLACEHOLDER, DateTime.UtcNow.ToString("u"));
```
stephen-turner: This is wrong because the substitution happens at the wrong time. It needs to happen when (and each time) the report is sent to CIS, not when it is received from XenCenter.

MihaelaStoica:
@stephen-turner Please review my last commit

stephen-turner:
Thanks, looks good now, merging.

PR #1653, opened 2017-06-16
CA-177402 - XenCenter doesnt show migrate option on selecting multipl…

letsboogey:
…e Suspended VMs

### DIFF
--- a/XenAdmin/Commands/MoveVMCommand.cs
+++ b/XenAdmin/Commands/MoveVMCommand.cs
@@ -67,7 +67,7 @@ namespace XenAdmin.Commands
         protected override void ExecuteCore(SelectedItemCollection selection)
         {
 
-            if (selection.Count > 1 && new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).CanExecute())
+            if (new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).CanExecute())
             {
                 new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).Execute();
             }
--- a/XenAdmin/Commands/MoveVMCommand.cs
+++ b/XenAdmin/Commands/MoveVMCommand.cs
@@ -67,7 +67,7 @@ namespace XenAdmin.Commands
         protected override void ExecuteCore(SelectedItemCollection selection)
         {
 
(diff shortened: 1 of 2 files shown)


XenAdmin/Commands/MoveVMCommand.cs
```diff
@@ -66,16 +66,22 @@ public MoveVMCommand(IMainWindow mainWindow, VM vm)
 
         protected override void ExecuteCore(SelectedItemCollection selection)
         {
-            VM vm = (VM)selection[0].XenObject;
 
-            if (CrossPoolMoveVMCommand.CanExecute(vm, null))
+            if (selection.AllItemsAre<VM>() && new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).CanExecute())
```
letsboogey: Yep you are right, it's not necessary there. It makes sense to check if selection contains more than one VM to use that execution path

XenAdmin/Commands/MoveVMCommand.cs
```diff
@@ -67,7 +67,7 @@ public MoveVMCommand(IMainWindow mainWindow, VM vm)
         protected override void ExecuteCore(SelectedItemCollection selection)
         {
 
-            if (selection.AllItemsAre<VM>() && new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).CanExecute())
+            if (selection.Count > 1 && new CrossPoolMoveVMCommand(MainWindowCommandInterface, selection).CanExecute())
```
MihaelaStoica: Actually, the count check is not needed, we want the CrossPoolMoveVMCommand even if there is only one VM selected.

letsboogey: If only one VM is selected don't we want to use the MoveVMDialog instead?

kc284: If I'm not mistaken on older servers we launch the MoveVMDialog which moves one VM at a time.

MihaelaStoica: The MoveVMDialog does the intra-pool move of a VM's disks. But if a cross-pool can be done, then we should offer this, even for a single VM.

kc284: Ok, agreed.

letsboogey: Ok okay, will make the change

PR #1661, opened 2017-06-22
CA-224724: VBD edit page: show 'position in use' message on the dropdown items

GaborApatiNagy:
This commit changes the dropdown control to not only show the available
positions, but also mark the ones that are already being used.

### DIFF
--- a/XenAdmin/SettingsPanels/VBDEditPage.cs
+++ b/XenAdmin/SettingsPanels/VBDEditPage.cs
@@ -391,7 +391,7 @@ namespace XenAdmin.SettingsPanels
                     return position;
 
                 if (vdi != null)
-                    return string.Format(Messages.VBD_EDIT_CURRENTLY_IN_USE_BY, position, vdi.ToString());
+                    return string.Format(Messages.VBD_EDIT_CURRENTLY_IN_USE_BY, position, vdi.ToString().Ellipsise(30));
 
                 return string.Format(Messages.VBD_EDIT_CURRENTLY_IN_USE, position);
             }
--- a/XenAdmin/SettingsPanels/VBDEditPage.cs
+++ b/XenAdmin/SettingsPanels/VBDEditPage.cs
@@ -42,6 +42,7 @@ using XenAdmin.Core;
 using XenAdmin.Network;
 using XenAPI;
 using XenAdmin.Dialogs;
+using System.Linq;
 
 
 namespace XenAdmin.SettingsPanels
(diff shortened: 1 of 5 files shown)


XenAdmin/SettingsPanels/VBDEditPage.cs
```diff
@@ -370,5 +363,38 @@ public String SubText
...
+                if (vbd != null && vbd.Connection != null)
+                {
+                    vdi = vbd.Connection.Resolve(vbd.VDI);
+                }
+
+                if (vbd == null)
+                    return position;
+
+                if (vdi != null)
+                    return string.Format(Messages.VBD_EDIT_CURRENTLY_IN_USE_BY, position, vdi.ToString());
```
kc284: We may want to ellipsise this string.

kc284: The check at line 94 is not needed since we have the check at line 97.

XenAdmin/SettingsPanels/VBDEditPage.cs
```diff
@@ -149,16 +150,30 @@ public void UpdateDevicePositions(Session session)
...
 
                     // Make sure the userdevice for the selected VBD is the one selected in the combobox
-                    foreach (String position in devicePositionComboBox.Items)
+                    foreach (DevicePositionComboBoxItem item in devicePositionComboBox.Items)
                     {
-                        if (position != vbd.userdevice)
-                            continue;
+                        if (item.VBD != null && item.VBD.userdevice == vbd.userdevice)
+                        {
+                            devicePositionComboBox.SelectedItem = item;
```
kc284: We may want to select the item in the loop above that populates the combobox to avoid the re-iteration.

kc284:
Also, the drop down is still too long (try it for example with a disk on a VM made off CoreOS which has 255 positions); I think it's not very pretty to drop further than the lower end of the dialog.

XenAdmin/SettingsPanels/VBDEditPage.cs
GaborApatiNagy: The code at line 94-97 was not touched by this PR, I leave code as is, because removing those lines you mentioned would cause a change in functionality, because the "vbd" is a field in declared the class, not just local in the method. (Therefore calling this setter with a non-VBD would null out the previously set one after this change whereas it is not the case now.)

XenAdmin/SettingsPanels/VBDEditPage.cs
GaborApatiNagy: I like it as is, because it may iterate the loop again, but only N times where N is the position of the currently selected item. If we moved this code into the previous block it would make that more complex whereas both are trivial now, so much easier to maintain. Re-iteration of this list is cheap anyway, also the layout is suspended while the execution is here)

GaborApatiNagy:
Thanks for reviewing @kc284, I've made the requested changes except for the two I wouldn't change - I commented on those separately.

kc284:
Looks ok to me now.

PR #1664, opened 2017-06-23
CA-257574: Manage vApps dialog is unusable with a vApp containing a l…

MihaelaStoica:
…arge number of VMs

- In the VMAppliancesDialog, change event handlers to reduce the number of refreshes when something changes: only reload the appliances when the VM_appliance collection change; if a relevant property is changed (in VM_appliance or VM), then update only the affected rows.
- Improve the way we assign VMs to a group (AssignVMsToVMApplianceAction and AssignVMsToPolicyAction): only update the VMs that really changed.
- Also improve the way we update the VM startup options and HA restart priorities on a set of VMs

### DIFF
--- a/XenAdmin/Core/VMGroup.cs
+++ b/XenAdmin/Core/VMGroup.cs
@@ -158,13 +158,7 @@ namespace XenAdmin.Core
             get { return typeof(T) == typeof(VMSS) ? InvisibleMessages.UPSELL_LEARNMOREURL_VMSS : InvisibleMessages.UPSELL_LEARNMOREURL_VM_APPLIANCES; }
         }
 
-        internal static AsyncAction AssignVMsToGroupAction(T group, List<XenRef<VM>> vmRefs, bool suppressHistory)
-        {
-            return typeof(T) == typeof(VMSS) ? (AsyncAction)(new AssignVMsToPolicyAction(group as VMSS, vmRefs, suppressHistory)) :
-                (AsyncAction)(new AssignVMsToVMApplianceAction(group as VM_appliance, vmRefs, suppressHistory));
-        }
-
-        internal static AsyncAction AssignVMsToGroupAction(T group, List<VM> vms, bool suppressHistory)
+        internal static AsyncAction AssignVMsToGroupAction(T group, List<XenRef<VM>> vms, bool suppressHistory)
         {
             return typeof(T) == typeof(VMSS) ? (AsyncAction)(new AssignVMsToPolicyAction(group as VMSS, vms, suppressHistory)) :
                 (AsyncAction)(new AssignVMsToVMApplianceAction(group as VM_appliance, vms, suppressHistory));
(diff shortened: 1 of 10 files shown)


XenAdmin/Core/VMGroup.cs
```diff
@@ -158,7 +158,13 @@ internal static string UpsellLearnMoreUrl
             get { return typeof(T) == typeof(VMSS) ? InvisibleMessages.UPSELL_LEARNMOREURL_VMSS : InvisibleMessages.UPSELL_LEARNMOREURL_VM_APPLIANCES; }
         }
 
-        internal static AsyncAction AssignVMsToGroupAction(T group, List<XenRef<VM>> vms, bool suppressHistory)
+        internal static AsyncAction AssignVMsToGroupAction(T group, List<XenRef<VM>> vmRefs, bool suppressHistory)
```
stephen-turner: You're sometimes passing XenRef<VM>s and sometimes VMs, and we keep parallel functions/constructors several levels deep. It would be worth seeing whether you can resolve the references earlier and always just pass VMs.

XenAdmin/Dialogs/VMAppliances/VMAppliancesDialog.cs
```diff
@@ -323,7 +411,7 @@ private void dataGridView1_SelectionChanged(object sender, EventArgs e)
 
         private void VMAppliancesDialog_FormClosed(object sender, FormClosedEventArgs e)
         {
-            Pool.Connection.Cache.DeregisterBatchCollectionChanged<VM_appliance>(VMApplianceCollectionChanged);
+            Pool.Connection.Cache.DeregisterCollectionChanged<VM_appliance>(VMApplianceCollectionChanged);
             Pool.Connection.Cache.DeregisterCollectionChanged<VM>(VMCollectionChanged);
             foreach (VM vm in Pool.Connection.Cache.VMs)
```
stephen-turner: We need to deregister all the vApp.PropertyChanged's near here too.

XenAdmin/Wizards/NewVMApplianceWizard/NewVMApplianceVMOrderAndDelaysPage.cs
```diff
@@ -210,7 +215,7 @@ private void updateTextBoxes()
         /// Gets the current (uncommitted) VM settings. Must be called on the GUI thread.
         /// </summary>
         /// <returns></returns>
-        public Dictionary<VM, VMStartupOptions> getCurrentSettings()
+        private Dictionary<VM, VMStartupOptions> getCurrentSettings(bool all = true)
```
stephen-turner: Check whether "all" is needed

XenModel/Actions/VMAppliances/AssignVMsToVMApplianceAction.cs
```diff
@@ -57,11 +69,14 @@ protected override void Run()
             Description = Messages.ASSIGNING_VM_APPLIANCE;
             foreach (var xenRef in _vmAppliance.VMs)
             {
-                VM.set_appliance(Session, xenRef, null);
+                var found = _selectedVMs.Any(vm => vm.opaque_ref == xenRef);
```
stephen-turner: This is an O(n^2) operation. I think it must be possible to take the set difference between _selectedVMs and _vmAppliance.VMs more efficiently.

XenAdmin/Core/VMGroup.cs
MihaelaStoica: I will revert these changes, as it will be more convienent to have a list of XenRefs in the action in order to calculate the set difference between two lists more efficiently.

stephen-turner:
Looks good now, thanks.

PR #1665, opened 2017-06-27
CA-254479,CA-223802,XSO-672: distinct the VM IP addresses which are r…

minglumlu:
…etrieved from XAPI.

This change is to make the XenCenter de-duplicate the IPv4 addresses reported from XAPI. 
This is becasue that XAPI has been changed to used new xenstore protocol described in http://xenbits.xen.org/gitweb/?p=xen.git;a=blob;f=docs/misc/xenstore-paths.markdown;hb=HEAD#l491

For the sake of compatibility, XAPI will report one IPv4 address in both formats to XenCenter:
old: 0/ip -> [ip address]
new: 0/ipv4/0 -> [ip address]

This would cause duplicated IP address displayed in XenCenter. This PR aims to check-in the change to fix this issue.

### DIFF
--- a/XenModel/XenAPI-Extensions/VIF.cs
+++ b/XenModel/XenAPI-Extensions/VIF.cs
@@ -63,13 +63,15 @@ namespace XenAPI
                     if (vmGuestMetrics != null)
                     {
                         // PR-1373 - VM_guest_metrics.networks is a dictionary of IP addresses in the format:
-                        // [["0/ip", <IPv4 address>], ["0/ipv6/0", <IPv6 address>], ["0/ipv6/1", <IPv6 address>]]
+                        // [["0/ip", <IPv4 address>], 
+                        //  ["0/ipv4/0", <IPv4 address>], ["0/ipv4/1", <IPv4 address>],
+                        //  ["0/ipv6/0", <IPv6 address>], ["0/ipv6/1", <IPv6 address>]]
 
                         return
                             (from network in vmGuestMetrics.networks
                              where network.Key.StartsWith(string.Format("{0}/ip", this.device))
                              orderby network.Key
-                             select network.Value).ToList();
+                             select network.Value).Distinct().ToList();
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


XenModel/XenAPI-Extensions/VIF.cs
```diff
@@ -63,13 +63,15 @@ public List<string> IPAddresses
                     if (vmGuestMetrics != null)
                     {
                         // PR-1373 - VM_guest_metrics.networks is a dictionary of IP addresses in the format:
-                        // [["0/ip", <IPv4 address>], ["0/ipv6/0", <IPv6 address>], ["0/ipv6/1", <IPv6 address>]]
+                        // [["0/ip", <IPv4 address>], 
+                        //  ["0/ip/0", <IPv4 address>], ["0/ip/1", <IPv4 address>],
```
stephen-turner: I think this should read 0/ipv4/0 and 0/ipv4/1

XenModel/XenSearch/Common.cs
```diff
@@ -1123,7 +1123,9 @@ private static ComparableList<ComparableAddress> IPAddressProperty(IXenObject o)
                 foreach (VIF vif in vifs)
                 {
                     // PR-1373 - VM_guest_metrics.networks is a dictionary of IP addresses in the format:
-                    // [["0/ip", <IPv4 address>], ["0/ipv6/0", <IPv6 address>], ["0/ipv6/1", <IPv6 address>]]
+                    // [["0/ip", <IPv4 address>], 
+                    //  ["0/ip/0", <IPv4 address>], ["0/ip/1", <IPv4 address>],
```
stephen-turner: As above, I think this should read 0/ipv4/0 and 0/ipv4/1

XenModel/XenAPI-Extensions/VIF.cs
minglumlu: Fixed it. Thanks!

XenModel/XenSearch/Common.cs
minglumlu: Fixed it. Thanks!

PR #1667, opened 2017-06-27
CA-249823 - Accept zip files in Updates Wizard

letsboogey:


XenModel/Actions/Updates/UnzipXenServerPatchAction.cs
```diff
@@ -0,0 +1,123 @@
...
+            finally
+            {
+                if (iterator != null)
+                    iterator.Dispose();
+            }
+        }
+
+        void archiveIterator_CurrentFileExtractProgressChanged(object sender, ExtractProgressChangedEventArgs e)
+        {
+            int pc = 95 + (int)(5.0 * e.BytesTransferred / e.TotalBytesToTransfer);
```
MihaelaStoica: This would report the PercentComplete from 95 to 100%. I think we neded something like this:  `pc = (int)(100.0 * e.BytesTransferred / e.TotalBytesToTransfer)`

MihaelaStoica:
As discussed, we should combine the two actions that share the unzipping code.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -64,13 +64,13 @@ public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string ou
             : this(patchName, uri, outputFileName, InvisibleMessages.XEN_UPDATE)
         { }
 
-        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, string updateFileExtension)
+        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, params string[] updateFileExtensions)
             : base(null, string.Format(Messages.DOWNLOAD_AND_EXTRACT_ACTION_TITLE, patchName), string.Empty, false)
```
MihaelaStoica: This action is always used with suppressHistory=false. I think we should expose this parameter and call it with suppressHistory=true from the Select Update page

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -370,9 +373,25 @@ private void BrowseButton_Click(object sender, EventArgs e)
...
                 {
                     if (dlg.ShowDialog(this) == DialogResult.OK && dlg.CheckFileExists)
-                        AddFile(dlg.FileName);
+                    {
+                        if (dlg.FileName.ToLowerInvariant().EndsWith(".zip"))
+                        {
+                            string unzippedFilePath = ExtractPatchAction(dlg.FileName);
+                            if (!string.IsNullOrEmpty(unzippedFilePath))
+                            {
+                                unzippedUpdateFilePath = unzippedFilePath;
```
MihaelaStoica: Is the unziped file going to be removed after installation?

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -167,9 +167,10 @@ private void ExtractFile()
                     {
                         string currentExtension = Path.GetExtension(iterator.CurrentFileName());
 
-                        if (!string.IsNullOrEmpty(updateFileExtension) && currentExtension == "." + updateFileExtension)
+                        if (Array.Exists(updateFileExtensions, item => item == currentExtension))
```
letsboogey: I have removed the "." from currentExtension now

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -414,13 +433,39 @@ public string SelectedNewPatch
...
                 }
-                else return null;
+                else 
+                    return null;
             }
         }
 
+        private string ExtractPatchAction(string zippedUpdatePath)
+        {
+            DownloadAndUnzipXenServerPatchAction unzipAction = new DownloadAndUnzipXenServerPatchAction(Path.GetFileNameWithoutExtension(zippedUpdatePath), null, zippedUpdatePath, UpdateExtension,".iso");
```
letsboogey: Done

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -167,9 +167,10 @@ private void ExtractFile()
                     {
                         string currentExtension = Path.GetExtension(iterator.CurrentFileName());
 
-                        if (!string.IsNullOrEmpty(updateFileExtension) && currentExtension == "." + updateFileExtension)
+                        if (Array.Exists(updateFileExtensions, item => item == currentExtension))
                         {
-                            string path = Path.Combine(Path.GetDirectoryName(outFileName), iterator.CurrentFileName());
+                            string path = address != null ? Path.Combine(Path.GetDirectoryName(zippedFileName), iterator.CurrentFileName())
```
letsboogey: Done.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -380,6 +380,7 @@ private void BrowseButton_Click(object sender, EventArgs e)
                             if (!string.IsNullOrEmpty(unzippedFilePath))
                             {
                                 unzippedUpdateFilePath = unzippedFilePath;
+                                PatchingWizard.PatchesToRemove.Add(unzippedUpdateFilePath);
```
MihaelaStoica: I suggest having the list of unzipped files as a property of this page and then the wizard can simply add this list to its own list of files that will eventually be cleaned up.

In general, in the wizards we do things this way: 
- The page doesn't know anything about the wizard (so it can be added to any wizard); 
- The page publishes some properties, which the wizard can access (because the wizard knows about its pages).

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -60,12 +60,14 @@ public string PatchPath
...
-        { }
+        private bool UnzipOnly
+        {
+            get { return address == null; }
+        }
 
-        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, params string[] updateFileExtensions)
-            : base(null, string.Format(Messages.DOWNLOAD_AND_EXTRACT_ACTION_TITLE, patchName), string.Empty, false)
+        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, bool suppressHist, params string[] updateFileExtensions)
+            : base(null, uri == null ? string.Format(Messages.DOWNLOAD_AND_EXTRACT_ACTION_TITLE, patchName)
```
MihaelaStoica: The action title should be assigned the other way around: if uri is null, then UPDATES_WIZARD_EXTRACT_ACTION_TITLE, else DOWNLOAD_AND_EXTRACT_ACTION_TITLE

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -165,11 +167,11 @@ private void ExtractFile()
 
                     while (iterator.HasNext())
                     {
-                        string currentExtension = Path.GetExtension(iterator.CurrentFileName());
+                        string currentExtension = Path.GetExtension(iterator.CurrentFileName()).Replace(".","");
 
                         if (Array.Exists(updateFileExtensions, item => item == currentExtension))
                         {
-                            string path = address != null ? Path.Combine(Path.GetDirectoryName(zippedFileName), iterator.CurrentFileName())
+                            string path = !UnzipOnly ? Path.Combine(Path.GetDirectoryName(zippedFileName), iterator.CurrentFileName())
```
MihaelaStoica: I know that I suggested the UnzipOnly variable, but seeing now that it is always used as `!UnzipOnly`, maybe having a variable with the opposite meaning would be more efficient? 
(I don't have a strong opinion about it, so if we cannot come up with a better variable name, then it's fine as it is.)

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
letsboogey: Ahh yes, I see the pattern now, thanks for that

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
letsboogey: silly mistake!

XenAdmin/Wizards/PatchingWizard/PatchingWizard.cs
```diff
@@ -339,6 +339,7 @@ private void RemoveDownloadedPatches()
             else
             {
                 listOfDownloadedFiles.AddRange(PatchingWizard_UploadPage.AllDownloadedPatches.Values);
+                listOfDownloadedFiles.AddRange(PatchingWizard_SelectPatchPage.UnzippedUpdateFiles);
```
MihaelaStoica: I think that we need to add the unzip files regardless of the mode the wizard is in when the cleanup is called, because I can, for example, select a zip file and then select the automated mode; if I finish (or cancel) the wizard in the automated mode, the unzipped file won't be cleaned up.
I haven't tried it out, but I believe this is the case.

We should also evaluate if we need to do the same with the downloaded patches (line 341 above).

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -60,17 +60,19 @@ public string PatchPath
...
-            : this(patchName, uri, outputFileName, InvisibleMessages.XEN_UPDATE)
-        { }
+        private bool DownloadedUpdateFromUri
+        {
+            get { return address != null; }
+        }
 
-        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, string updateFileExtension)
-            : base(null, string.Format(Messages.DOWNLOAD_AND_EXTRACT_ACTION_TITLE, patchName), string.Empty, false)
+        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, bool suppressHist, params string[] updateFileExtensions)
```
MihaelaStoica: Just a note on the name of this method: DownloadAndUnzipXenServerPatchAction suggests that it returns an action, when in fact it creates and execute an action. A better name would be DownloadAndUnzipUpdate,

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -60,17 +60,19 @@ public string PatchPath
             get; private set;
         }
 
-        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName)
-            : this(patchName, uri, outputFileName, InvisibleMessages.XEN_UPDATE)
-        { }
+        private bool DownloadedUpdateFromUri
```
MihaelaStoica: I suggest renaming this to downloadUpdate and change it to a private readonly variable which is set once in the constructor (similar to address, zippedFileName, etc). There is no need to have it as a property with a getter if it never changes.

MihaelaStoica:
@letsboogey Have you tested the following scenario? 
I select a valid zip file, via the Browse button. The Next button is enabled. I then change the file name to something invalid by typing in the textbox, Is the filename validated again when I change it or when I press Next?

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
MihaelaStoica: Please ignore the above comment. I was thinking at the ExtractPatchAction function, not this action! Sorry!

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -175,9 +177,19 @@ public override void PageLeave(PageLoadedDirection direction, ref bool cancel)
         {
             if (direction == PageLoadedDirection.Forward)
             {
+                //check if we are installing update user sees in textbox
+                if (Path.GetFileNameWithoutExtension(unzippedUpdateFilePath) != Path.GetFileNameWithoutExtension(fileNameTextBox.Text))
```
MihaelaStoica: We should only go on on this path if the file is a zip.
Also, we should clear unzippedUpdateFilePath if the slected file is not a zip, othewise we need to add the check for zip extension on line 180 below.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -414,13 +442,39 @@ public string SelectedNewPatch
                 else if (selectFromDiskRadioButton.Checked)
                 {
                     return SelectedUpdateType == UpdateType.NewRetail || SelectedUpdateType == UpdateType.ISO
-                              ? fileNameTextBox.Text
-                               : null;
+                        ? isValidFile(unzippedUpdateFilePath) ? unzippedUpdateFilePath : fileNameTextBox.Text : null;
```
MihaelaStoica: We should only return the unzippedUpdateFilePath  if the file extension is zip.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -633,6 +688,7 @@ private void fileNameTextBox_Enter(object sender, EventArgs e)
 
         private void fileNameTextBox_TextChanged(object sender, EventArgs e)
         {
+            zippedUpdateFilePath = fileNameTextBox.Text;
```
MihaelaStoica: I think that zippedUpdateFilePath variable is not needed anymore

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -177,7 +178,22 @@ public override void PageLeave(PageLoadedDirection direction, ref bool cancel)
...
+                            if (unzippedUpdateFilePath == null)
+                                cancel = true;
+
+                            unzippedFiles.Add(unzippedUpdateFilePath);
+                        }
+                    }
+                    else
+                        unzippedUpdateFilePath = null;
+                    
+                    var fileName = isValidFile(unzippedUpdateFilePath) ? unzippedUpdateFilePath : fileNameTextBox.Text.ToLowerInvariant();
```
MihaelaStoica: unzippedUpdateFilePath.ToLowerInvariant()

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -414,13 +440,40 @@ public string SelectedNewPatch
                 else if (selectFromDiskRadioButton.Checked)
                 {
                     return SelectedUpdateType == UpdateType.NewRetail || SelectedUpdateType == UpdateType.ISO
-                              ? fileNameTextBox.Text
-                               : null;
+                        ? isValidFile(unzippedUpdateFilePath) && Path.GetExtension(fileNameTextBox.Text).Equals(".zip")
```
MihaelaStoica: .ToLowerInvariabnt()

PR #1675, opened 2017-06-30
CP-23197: Implement the Update homogeneity Pool Join rule in XenCenter

MihaelaStoica:


XenModel/PoolJoinRules.cs
```diff
@@ -388,6 +397,35 @@ private static bool DifferentServerVersion(Host slave, Host master)
...
+            // Make a list of updates that should be homogeneous
+            var homogeneousUpdates = new List<Pool_update>();
+            foreach (var host in allHosts)
+            {
+                homogeneousUpdates.AddRange(host.AppliedUpdates().Where(update => update.enforce_homogeneity));
+            }
+
+            foreach (var update in homogeneousUpdates)
+            {
+                if (allHosts.Any(host => !host.AppliedUpdates().Exists(u => u.uuid == update.uuid)))
```
kc284: Wouldn't it be more efficient to check `allHosts.Any(host => host.AppliedUpdates().Exists(u => u.uuid != update.uuid)` ?

stephen-turner: I think @kc284's check isn't correct, but there should be a more efficient way to do it. For example, construct a Map<Update, int> showing the number of hosts having each update: for each update, that number should be the same as the number of hosts in the pool.

kc284: Ignore above comment, I wasn't aware of the requirement.

XenModel/PoolJoinRules.cs
```diff
@@ -388,6 +397,35 @@ private static bool DifferentServerVersion(Host slave, Host master)
                 slave.ProductBrand != master.ProductBrand;
         }
 
+        /// <summary>
+        /// Check whether all updates that request homogeneity are in fact homogeneous
+        /// across the proposed pool. This is used in CanJoinPool and prevents the pool from being created
+        /// </summary>
+        private static bool DifferentHomogeneousUpdates(Host slave, Host master)
```
GaborApatiNagy: I think it would be beneficial to check for server versions as all the below doesn't apply to pre-Ely hosts. Doing so would document the code.

MihaelaStoica:
Thanks @stephen-turner and @GaborApatiNagy for your suggestions. Please see the new commit.

GaborApatiNagy:
Looks good to me (at 6950ad2)

GaborApatiNagy:
I've noticed the uuid comparison is case sensitive, but because this code compares only uuids of the same updates (on different hosts), I think it won't make any difference.

stephen-turner:
The bit I commented on previously looks good now. I haven't reviewed the whole PR -- do you want me to?

MihaelaStoica:
@stephen-turner, @GaborApatiNagy I had added a new commit, to check for the platform version instead of product version.

PR #1676, opened 2017-07-03
Backport pool join rules changes to honolulu

MihaelaStoica:


MihaelaStoica:
This is the equivalent on #1675 and commit a46815cc and on honolulu

MihaelaStoica:
I've done some manual tests on pool join: 
1. Servers on version 7.0 and older (pre-Ely): 
- joining diferent versions, e.g. 6.1 and 6.2 is blocked by the API version difference
- joining pre-Ely and Ely version is also blocked due to the API version difference
- joining same version servers with different patching level is still allowed by XenCenter, but if fails in xapi - there is no change here in XenCenter from the current behaviour (XC does not check the equivalence of patches on pre-Ely servers)
2. Servers on Ely (including Honolulu):
- joining Ely servers with different patching level is still allowed by XenCenter if none of the updates changes the API version (XenCenter cannot check the database schema because it is not exposed and the homogeneity flag on updates is not present)
- on Honolulu servers, XenCenter blocks the pool join with different updates based on the the homogeneity flag

MihaelaStoica:
@stephen-turner Please see my previous comment regarding the manual tests I've performed

stephen-turner:
Thanks. I'll pull the changeset now.

PR #1677, opened 2017-07-03
CA-258321: Incorrect CFU user agent

jijiang:
Fixed user agent string for CFU.


stephen-turner:
👍

PR #1679, opened 2017-07-04
CP-23101: Add tests for the metadata transfer functions

MihaelaStoica:

### DIFF
--- a/XenAdminTests/HealthCheckTests/CredentialTests.cs
+++ b/XenAdminTests/HealthCheckTests/CredentialTests.cs
@@ -48,8 +48,6 @@ namespace XenAdminTests.HealthCheckTests
 {
     public class CredentialTests : UnitTester_TestFixture
     {
-        private const char SEPARATOR = '\x202f'; // narrow non-breaking space.
-
         [TestFixtureSetUp]
         public void FixtureSetup()
         {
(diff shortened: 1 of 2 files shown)


XenAdminTests/HealthCheckTests/CredentialTests.cs
```diff
@@ -158,15 +156,91 @@ public void CredentialOp()
...
+            pipeClient.Close();
+            System.Threading.Thread.Sleep(1000);
+            con = ServerListHelper.instance.GetServerList();
+            Assert.IsTrue(con.Count == conSize); 
+            Assert.IsTrue(ServerListHelper.instance.XenCenterMetadata == metadata, "\"Send metadata\" test failed"); 
+
+            // Send metadata containing the separator
+            pipeClient = new NamedPipeClientStream(".", HealthCheckSettings.HEALTH_CHECK_PIPE, PipeDirection.Out);
+            pipeClient.Connect();
+            int separatorPosition = new Random().Next(0, metadata.Length);
```
GaborApatiNagy: I'm a bit worried that using Random without an initial seed makes the test not (or not easily) reproducible. I suppose it is there to bring some randomness into the test -- I was going to suggest pseudo-random number, but that wouldn't really randomize this (because it's being used only once). 

So my suggestion is to keep it as is, but include the random number (position) in the error message.

GaborApatiNagy:
All the other changes look good to me

MihaelaStoica:
@GaborApatiNagy Thanks for reviewing. I had added your suggestion in a new commit

GaborApatiNagy:
Thanks, looks good, merging

PR #1680, opened 2017-07-04
Revert the changes to the XenAPI/Relations

MihaelaStoica:


MihaelaStoica:
This change needs to be reverted because the changes to the PCI class (the virtual_functions property) have not been included in the API bindings.

PR #1682, opened 2017-07-06
CA-259276: Error when trying to install an update via JA XenCenter

MihaelaStoica:


stephen-turner:
👍

PR #1688, opened 2017-07-11
CA-258357: Modified the build script to build the MSI installer

GaborApatiNagy:
Added a download link


MihaelaStoica:
This change also needs to be applied to the  XenCenter.l10n.diff file (because the line changed is included in the diff) and to en-us.wxl.
Also the updated string needs to be translated in zh-cn.wxl and ja-jp.wxl.

MihaelaStoica:
Just noticed that the *.wxl files refer to .Net 4, not 4.6

PR #1689, opened 2017-07-11
CA-258357: Create the output directory before trying to write into it

MihaelaStoica:


mk/package-and-sign.sh
```diff
@@ -52,6 +52,8 @@ mkdir_clean()
   rm -rf $1 && mkdir -p $1
 }
 
+mkdir_clean ${OUTPUT_DIR}
```
kc284: @MihaelaStoica this line should not be necessary because the output folder should have already been created by the Jenkinsfile in the 'Create manifest' stage. Moreover this line clears the created manifest. Did you add it because you noticed a different behaviour?

MihaelaStoica: @kc284 This is in the package-and-sign.sh script which is run to test the unsigned package and it copies the installers in a separate output folder (inside the TestXenAdminUnsigned folder), not the one created by the Jenkinsfile. 
I added it because the build started to fail in this phase ("Testing package-and-sign script") follwing Gabor's changes to switch to the msi.

PR #1691, opened 2017-07-12
CA-258357: Create the output directory before trying to write into it

MihaelaStoica:


MihaelaStoica:
This is the equivalent of  #1689 on master

PR #1692, opened 2017-07-12
CA-259659: Revise the update homogeneity pool join rule

MihaelaStoica:
- only check the updates on the master and the new slave

### DIFF
--- a/XenModel/PoolJoinRules.cs
+++ b/XenModel/PoolJoinRules.cs
@@ -409,22 +409,10 @@ namespace XenAdmin.Core
             if (!Helpers.ElyOrGreater(slave) || !Helpers.ElyOrGreater(master))
                 return false;
 
-            List<Host> hosts = new List<Host> {master, slave};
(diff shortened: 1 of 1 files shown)


stephen-turner:
Now that we are only comparing the lists on two hosts, I suspect that the algorithm is less clear than alternatives. Would some of the suggestions on https://stackoverflow.com/questions/22173762/check-if-two-lists-are-equal work equally well and be more readable?

MihaelaStoica:
@stephen-turner Thanks for the suggestion, Please see the new commit.

stephen-turner:
Yes, thanks, that's much clearer now. The code looks right to me, but what dev testing have you done?

MihaelaStoica:
I couldn't do any testing with homogenoeus updates, because we don't have any yet, but I dev tested the algorithm on a slightly modified code that ignored the `enforce_homogeneity` flag.

stephen-turner:
OK, thanks, pulling the change.

PR #1693, opened 2017-07-13
CA-235459 - G11n:SC: The numeric tag for “Disks” and “Network” is ove…

letsboogey:
…rlap on the search window.

### DIFF
--- a/XenAdmin/Controls/XenSearch/QueryPanel.cs
+++ b/XenAdmin/Controls/XenSearch/QueryPanel.cs
@@ -213,13 +213,15 @@ namespace XenAdmin.Controls.XenSearch
 
             if (s == "disks")
             {
-                int colWidth = Int32.Parse(resources.GetString("DISKS_OVERVIEW_COL_WIDTH"));
+                int colWidth;
+                Int32.TryParse(resources.GetString("DISKS_OVERVIEW_COL_WIDTH"), out colWidth);
                 return NewGridHeaderItem(Messages.OVERVIEW_DISKS, SortOrder.Descending, colWidth, colWidth);
             }
 
             if (s == "network")
             {
-                int colWidth = Int32.Parse(resources.GetString("NETWORK_OVERVIEW_COL_WIDTH"));
+                int colWidth;
+                Int32.TryParse(resources.GetString("NETWORK_OVERVIEW_COL_WIDTH"), out colWidth);
                 return NewGridHeaderItem(Messages.OVERVIEW_NETWORK, SortOrder.Descending, colWidth, colWidth);
             }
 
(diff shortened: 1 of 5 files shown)


XenAdmin/Controls/XenSearch/QueryPanel.cs
```diff
@@ -213,12 +213,12 @@ private static GridItemBase GetDefaultColumn(string s)
 
             if (s == "disks")
             {
-                return NewGridHeaderItem(Messages.OVERVIEW_DISKS, SortOrder.Descending, 120, 120);
+                return NewGridHeaderItem(Messages.OVERVIEW_DISKS, SortOrder.Descending, 138, 138);
```
kc284: This number is hardcoded, which means it's the same for all languages and I'm a bit concerned that it may result to the english counterpart having more white space than needed. Ideally, this number should be somewhere in a localisable resource file (perhaps on the NewGridHeaderItem or parent classes).

Can we check how easy it is to do this - these are completely custom controls and there may be caveats, for example, one thing to watch out for is how that resource manager will be called to avoid memory leaks like we had recently with the images on the command toolstrip items. If it's too big a change, ok, let's leave it.

XenAdmin/Controls/XenSearch/QueryPanel.cs
```diff
@@ -213,12 +213,14 @@ private static GridItemBase GetDefaultColumn(string s)
 
             if (s == "disks")
             {
-                return NewGridHeaderItem(Messages.OVERVIEW_DISKS, SortOrder.Descending, 120, 120);
+                int colWidth = Int32.Parse(resources.GetString("DISKS_OVERVIEW_COL_WIDTH"));
```
kc284: I recommend using
```
int colWidth;
Int32.TryParse(resources.GetString("DISKS_OVERVIEW_COL_WIDTH"), out colWidth);
```
If parsing fails the colWidth wil be zero and the result will be ugly, but at least it doesn't throw exceptions.

XenAdmin/Controls/XenSearch/QueryPanel.cs
```diff
@@ -213,12 +213,14 @@ private static GridItemBase GetDefaultColumn(string s)
...
             {
-                return NewGridHeaderItem(Messages.OVERVIEW_DISKS, SortOrder.Descending, 120, 120);
+                int colWidth = Int32.Parse(resources.GetString("DISKS_OVERVIEW_COL_WIDTH"));
+                return NewGridHeaderItem(Messages.OVERVIEW_DISKS, SortOrder.Descending, colWidth, colWidth);
             }
 
             if (s == "network")
             {
-                return NewGridHeaderItem(Messages.OVERVIEW_NETWORK, SortOrder.Descending, 120, 120);
+                int colWidth = Int32.Parse(resources.GetString("NETWORK_OVERVIEW_COL_WIDTH"));
```
kc284: Same as above.

letsboogey:
Updated @kc284

PR #1696, opened 2017-07-14
Removed duplicate resource.

kc284:
This was introduced by PR #1668 . Fixing it to have correct resources until the next update of the API bindings.

### DIFF
--- a/XenModel/XenAPI/FriendlyErrorNames.resx
+++ b/XenModel/XenAPI/FriendlyErrorNames.resx
@@ -112,10 +112,10 @@
     <value>2.0</value>
   </resheader>
   <resheader name="reader">
-    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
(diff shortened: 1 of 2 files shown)


XenModel/XenAPI/FriendlyErrorNames.resx
```diff
@@ -858,9 +858,6 @@
   <data name="POOL_AUTH_ENABLE_FAILED_DUPLICATE_HOSTNAME" xml:space="preserve">
     <value>Failed to enable external authentication, a duplicate hostname was detected.</value>
   </data>
-  <data name="POOL_AUTH_ENABLE_FAILED_INVALID_ACCOUNT" xml:space="preserve">
```
MihaelaStoica: I think we should remove the instance at line 876, to keep this file sorted.

kc284: We need to keep the instance in line 876 as it was added to fix CA-254480 (https://github.com/xenserver/xenadmin/commit/04cfd02bd0fce3e46f2a99f82bdfd7167ecce663). I'll move it here instead.

PR #1697, opened 2017-07-14
CA-257927: Fixed scalability issue where XenCenter froze momentarily when disconnecting from a host with a big number of alerts

kc284:


XenAdmin/TabPages/AlertSummaryPage.cs
```diff
@@ -585,24 +584,27 @@ private void tsmiDismissSelected_Click(object sender, EventArgs e)
...
-                    RemoveAlertRow(a);
+
+                    var a = e.Element as Alert;
+                    if (a != null)
+                        RemoveAlertRow(a);
+                    else
+                    {
+                        var range = e.Element as List<Alert>;
+                        if (range != null)
+                            Rebuild();
```
MihaelaStoica: I wonder if it's worth checking if there was just one element in the list, and in this case call RemoveAlertRow?

MihaelaStoica:
@kc284 Apart for the question above, these changes look good to me.

XenAdmin/TabPages/AlertSummaryPage.cs
kc284: Can do. Hmm, or alternatively put the check in the Changeable list and fire the event for a single item in line 87?

MihaelaStoica: I like the idea of putting the check in the Changeable list

MihaelaStoica:
Looks good. Merging.

PR #1698, opened 2017-07-17
[updated  - ready for review] L10N: Translation Check Ins CA-259778, CA-238883, CA-259356, CA-259729, CA-259730,CA-236057

vetrivelc:


XenModel/Messages.zh-CN.resx
```diff
@@ -3623,19 +3623,19 @@
...
   <data name="DATEFORMAT_DMY_HMS" xml:space="preserve">
-    <value>yyyy-MM-dd tt h:mm:ss</value>
+    <value>yyyy-MM-dd  HH:mm:ss</value>
   </data>
   <data name="DATEFORMAT_DMY_LONG" xml:space="preserve">
     <value>yyyy年M月d日</value>
   </data>
   <data name="DATEFORMAT_HM" xml:space="preserve">
-    <value>HH:mm</value>
+    <value>tt HH:mm</value>
```
MihaelaStoica: I think this should be "tt h:mm", because if you add the "tt" AM/PM designator, then the hour should be represented in the 12-hour clock format ("h" or "hh"), and not in the 24-hour clock format ("HH").

This also applies to line 3638.

XenModel/Messages.zh-CN.resx
```diff
@@ -3623,19 +3623,19 @@
     <value>yyyy-MM-dd</value>
   </data>
   <data name="DATEFORMAT_DMY_HM" xml:space="preserve">
-    <value>yyyy-MM-dd tt h:mm</value>
+    <value>yyyy-MM-dd HH:mm</value>
```
MihaelaStoica: I don't understand why this has been changed to the 24-hour format, while DATEFORMAT_HM has been changed to 12-hour format on line 3635. 

This change wasn't needed for CA-236057 and will create discrepancies again on the Performance tab and on other windows where both formats are used.

GaborApatiNagy:
Hi @vetrivelc , we've noticed you changed time formats in the Chinese resources from "morning/afternoon+twelve hours" to "24 hours". 
Is there any reason that's been done here?

vetrivelc:
@GaborApatiNagy Yeah, it was done as part of bug fix, but now we have corrected it as suggested by @MihaelaStoica and I committed the fix now f09ae6b607410284c79079646c47a525f29cce35

PR #1701, opened 2017-07-17
Increase XenCenter micro version for UPD-99

MihaelaStoica:


stephen-turner:
Seems sound

PR #1702, opened 2017-07-18
CA-258021 - Change updates retrieval url

letsboogey:


MihaelaStoica:
This change also needs to be applied to branding.sh (https://github.com/xenserver/xenadmin/blob/master/Branding/branding.sh#L100)

letsboogey:
Updated it @MihaelaStoica

PR #1709, opened 2017-07-24
CP-23450: Add "View" btn in updates page toolbar

jijiang:
The View button contains "By Update" and "By Host" options.

### DIFF
--- a/XenAdmin/TabPages/ManageUpdatesPage.Designer.cs
+++ b/XenAdmin/TabPages/ManageUpdatesPage.Designer.cs
@@ -55,6 +55,10 @@
             this.pictureBoxProgress = new System.Windows.Forms.PictureBox();
             this.tableLayouPanel5 = new System.Windows.Forms.TableLayoutPanel();
             this.toolStrip1 = new XenAdmin.Controls.ToolStripEx();
+            this.toolStripSplitButtonView = new System.Windows.Forms.ToolStripSplitButton();
+            this.byUpdateToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.byHostToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.toolStripSeparator2 = new System.Windows.Forms.ToolStripSeparator();
             this.toolStripDropDownButtonServerFilter = new XenAdmin.Controls.FilterLocationToolStripDropDownButton();
             this.toolStripDropDownButtonDateFilter = new XenAdmin.Controls.FilterDatesToolStripDropDownButton();
             this.toolStripSeparator3 = new System.Windows.Forms.ToolStripSeparator();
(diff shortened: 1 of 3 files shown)


XenAdmin/TabPages/ManageUpdatesPage.Designer.cs
```diff
@@ -423,5 +467,10 @@ private void InitializeComponent()
         private System.Windows.Forms.DataGridViewTextBoxColumn ColumnDate;
         private System.Windows.Forms.DataGridViewTextBoxColumn ColumnWebPage;
         private System.Windows.Forms.TableLayoutPanel tableLayouPanel5;
+        private System.Windows.Forms.ToolStripSeparator toolStripSeparator;
+        private System.Windows.Forms.ToolStripSplitButton toolStripSplitButtonView;
```
kc284: The control for this use should be ToolStripDropDownButton, not ToolStripSplitButton. With the ToolStripSplitButton the user selects an item and this item becomes the default, so next time they want to repeat the action they can click it directly without having to open the dropdown again. However, here you only select one of the two items, you don't need to repeat an action, therefore you need the ToolStripDropDownButton.

Also, when you select one of the dropdown items, this item should stay highlighted. Currently, the first item is always highlighted, no matter which one you click.

(We have a similar button on the Snapshots TabPage, you can have a look there to see how it works).

XenAdmin/TabPages/ManageUpdatesPage.Designer.cs
```diff
@@ -203,10 +210,42 @@ private void InitializeComponent()
...
+            this.byUpdateToolStripMenuItem.Checked = true;
+            this.byUpdateToolStripMenuItem.CheckState = System.Windows.Forms.CheckState.Checked;
+            this.byUpdateToolStripMenuItem.Image = global::XenAdmin.Properties.Resources.notif_updates_16;
+            this.byUpdateToolStripMenuItem.Name = "byUpdateToolStripMenuItem";
+            resources.ApplyResources(this.byUpdateToolStripMenuItem, "byUpdateToolStripMenuItem");
+            this.byUpdateToolStripMenuItem.Click += new System.EventHandler(this.byUpdateToolStripMenuItem_Click);
+            // 
+            // byHostToolStripMenuItem
+            // 
+            this.byHostToolStripMenuItem.Image = global::XenAdmin.Properties.Resources._000_TreeConnected_h32bit_16;
```
kc284: The images look a bit stretched and blurry, you probably need to try the other scaling options in their properties.

kc284:
I understand it's a bit difficult to split this big chunk of work into items because they depend on each other, but I think we can put a bit more code in this PR, namely hide/show the relevant buttons in the top toolstrip when the View changes.

kc284:
One more thing: because this is a biggish piece of work and it will be at least 3 tickets/PRs before it is completely functional, I recommend you close this PR and re-submit it to the freshly created branch REQ-534, because the master should in theory be releasable at all times.

jijiang:
As recommended by Tina, this change should be part of REQ-534, and should be pushed to REQ-534 branch instead.

PR #1710, opened 2017-07-24
CA-260364: Pool join in XenCenter: Use homogeneity flag from other_co…

MihaelaStoica:
…nfig for Ely/Honolulu hosts


stephen-turner:
And to me. Merging.

PR #1711, opened 2017-07-24
[Honolulu] CA-260364: Pool join in XenCenter: Use homogeneity flag from other_config for Ely/Honolulu hosts

MihaelaStoica:


stephen-turner:
Agreed. Merging.

PR #1712, opened 2017-07-24
CP-23450: Add "View" btn in Updates page toolbar

jijiang:
1) View button has "By Update" and "By Host" options.
2) "By Host" view hides buttons "Filter by Date", "Dismiss All" and
"Restore Dismissed Updates".

### DIFF
--- a/XenAdmin/Controls/FilterDatesToolStripDropDownButton.cs
+++ b/XenAdmin/Controls/FilterDatesToolStripDropDownButton.cs
@@ -141,5 +141,10 @@ namespace XenAdmin.Controls
             if (FilterChanged != null)
                 FilterChanged();
         }
+
+        public void ResetFilterDates()
+        {
+            DropDownItems[6].PerformClick();
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


kc284:
This looks better now. One more remark: when the application is first launched, one of the views is the default one and is already selected, however neither of the dropdown menu items is selected.

kc284:
This I only considered today that I saw the buttons in action and is more of a design question (which may have to be seen in conjunction with the other tickets in this REQ as part of a different PR): if we are viewing by update we give the option to filter by date, in which case the top right label says "Filters are ON". If we then change to viewing by host, the date filtering button is hidden by design. What does the design prescribe about the filtering in this case? Do we reset it (in which case the label has to change)? And what happens if we go back to the previous view? Is it preserved/re-applied?

jijiang:
Have updated the PR. Now the initial check state is added and date filter is turned OFF by "By Host" view.

kc284:
Now this behaves correctly. One last improvement: the method resetting the filter repeats some of the logic in the event handler OnDropDownItemClicked, so you can simply do the following:
```
public void ResetFilterDates()
{
    DropDownItems[6].PerformClick();
}
```

jijiang:
Thanks for comment. Updated.

PR #1715, opened 2017-07-27
CA-260365: Hide the build number from the General Tab for Ely and new…

MihaelaStoica:
…er hosts (to cover the Honolulu case)

Also restrict the use of build number as an integer to pre-Ely hosts (because a Honolulu host has a non-numeric build number and XenCenter doesn't know the difference between Ely and Honolulu)
And a small update to the pool join rule to match what we have in Honolulu XenCenter (check the database schema regardless of the version)


GaborApatiNagy:
Looks good

PR #1722, opened 2017-08-07
CP-23604: Add IGMP snooping toggle.

huizh:
Add the toggle to enable/disable IGMP snooping.
Display the property if the pool is licensed
and the network backend is OVS.

### DIFF
--- a/XenAdmin/SettingsPanels/NetworkOptionsEditPage.Designer.cs
+++ b/XenAdmin/SettingsPanels/NetworkOptionsEditPage.Designer.cs
@@ -1,0 +1,119 @@
+﻿namespace XenAdmin.SettingsPanels
+{
+    partial class NetworkOptionsEditPage
+    {
+        /// <summary> 
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary> 
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Component Designer generated code
+
+        /// <summary> 
+        /// Required method for Designer support - do not modify 
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(NetworkOptionsEditPage));
+            this.radioButtonDisable = new System.Windows.Forms.RadioButton();
+            this.radioButtonEnable = new System.Windows.Forms.RadioButton();
+            this.labelRubric = new System.Windows.Forms.Label();
+            this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
+            this.iconWarningIGMPSnoopingOption = new System.Windows.Forms.PictureBox();
+            this.labelWarningIGMPSnoopingOption = new System.Windows.Forms.Label();
+            this.warningsTable = new System.Windows.Forms.TableLayoutPanel();
+            ((System.ComponentModel.ISupportInitialize)(this.iconWarningIGMPSnoopingOption)).BeginInit();
+            this.warningsTable.SuspendLayout();
+            this.SuspendLayout();
+            // 
+            // radioButtonDisable
+            // 
+            resources.ApplyResources(this.radioButtonDisable, "radioButtonDisable");
+            this.warningsTable.SetColumnSpan(this.radioButtonDisable, 2);
+            this.radioButtonDisable.Name = "radioButtonDisable";
+            this.radioButtonDisable.TabStop = true;
+            this.radioButtonDisable.UseVisualStyleBackColor = true;
+            // 
+            // radioButtonEnable
+            // 
+            resources.ApplyResources(this.radioButtonEnable, "radioButtonEnable");
+            this.warningsTable.SetColumnSpan(this.radioButtonEnable, 2);
+            this.radioButtonEnable.Name = "radioButtonEnable";
+            this.radioButtonEnable.TabStop = true;
+            this.radioButtonEnable.UseVisualStyleBackColor = true;
+            // 
+            // labelRubric
+            // 
+            resources.ApplyResources(this.labelRubric, "labelRubric");
+            this.labelRubric.BackColor = System.Drawing.Color.Transparent;
+            this.warningsTable.SetColumnSpan(this.labelRubric, 2);
+            this.labelRubric.Name = "labelRubric";
+            // 
+            // tableLayoutPanel1
+            // 
+            resources.ApplyResources(this.tableLayoutPanel1, "tableLayoutPanel1");
+            this.tableLayoutPanel1.Name = "tableLayoutPanel1";
+            // 
+            // iconWarningIGMPSnoopingOption
+            // 
+            resources.ApplyResources(this.iconWarningIGMPSnoopingOption, "iconWarningIGMPSnoopingOption");
+            this.iconWarningIGMPSnoopingOption.Image = global::XenAdmin.Properties.Resources._000_Info3_h32bit_16;
+            this.iconWarningIGMPSnoopingOption.Name = "iconWarningIGMPSnoopingOption";
+            this.iconWarningIGMPSnoopingOption.TabStop = false;
+            // 
+            // labelWarningIGMPSnoopingOption
+            // 
+            resources.ApplyResources(this.labelWarningIGMPSnoopingOption, "labelWarningIGMPSnoopingOption");
+            this.labelWarningIGMPSnoopingOption.Name = "labelWarningIGMPSnoopingOption";
+            // 
+            // warningsTable
+            // 
+            this.warningsTable.AllowDrop = true;
+            resources.ApplyResources(this.warningsTable, "warningsTable");
+            this.warningsTable.Controls.Add(this.labelRubric, 0, 0);
+            this.warningsTable.Controls.Add(this.iconWarningIGMPSnoopingOption, 0, 3);
+            this.warningsTable.Controls.Add(this.labelWarningIGMPSnoopingOption, 1, 3);
+            this.warningsTable.Controls.Add(this.radioButtonEnable, 0, 1);
+            this.warningsTable.Controls.Add(this.radioButtonDisable, 0, 2);
+            this.warningsTable.Name = "warningsTable";
+            // 
+            // NetworkOptionsEditPage
+            // 
+            resources.ApplyResources(this, "$this");
+            this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Dpi;
+            this.Controls.Add(this.warningsTable);
+            this.Controls.Add(this.tableLayoutPanel1);
+            this.Name = "NetworkOptionsEditPage";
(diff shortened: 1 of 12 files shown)


XenAdmin/SettingsPanels/NetworkOptionsEditPage.Designer.cs
```diff
@@ -0,0 +1,88 @@
...
+            this.tableLayoutPanel1.ResumeLayout(false);
+            this.tableLayoutPanel1.PerformLayout();
+            this.ResumeLayout(false);
+            this.PerformLayout();
+
+        }
+
+        #endregion
+
+        private System.Windows.Forms.RadioButton radioButtonDisable;
```
kc284: Radiobuttons are normally used to represent options, while things that have an enabled/disabled state are represented using a checkbox. I would replace the label and the two radiobuttons with a single checkbox reading "Enable IMGP (Internet Group Management Protocol) snooping" (the "in this pool" bit is not needed since the tab only appears on the pool or standalone host properties).

huizh: Hi @kc284, thanks for the comment.

This implementation refers to code of Live Patch (https://github.com/xenserver/xenadmin/blob/a59e4f4ad3c3f590d551ad54848bbfb967119717/XenAdmin/SettingsPanels/LivePatchingEditPage.cs#L54). Does it make sense?

kc284: That implementation was not very correct either. I did not review those commits, so I'm not familiar with their history,  any idea why the original checkbox was changed to radiobuttons?

stephen-turner: I'd like to see the options. Radio buttons can sometimes be useful for binary choices, especially if more explanation is needed.

kc284: I'm not suggesting to change the LivePatchPage now, however on the IMGP page the choice is plainly binary, and two radios saying Enable/Disable are superfluous.

kc284: Forgot to say, if more explanation is needed on what enabling IMGP snooping does, it can be written underneath the checkbox in an indented label.

stephen-turner: Having now seen the design, I agree with Tina that this should be a checkbox not radio buttons. But I also don't want a tab with just one checkbox on it. Is there anywhere else it can fit?

stephen-turner: If there isn't anywhere else, a pair of radio buttons -- but with longer labels -- might be preferable to one checkbox. We've done this on live patching and integrated GPU, for example.

stephen-turner: The Microsoft guidelines at https://msdn.microsoft.com/en-us/library/windows/desktop/dn742436(v=vs.85).aspx say:
"Use radio buttons if you have enough screen space and the options are important enough"

I think we are in this situation: there is loads of space, and there is only one, important option. So I am happy with radio buttons. However, they need more text next to each button to avoid looking stupid.

huizh: Thanks @stephen-turner and @kc284.

I've updated this PR accordingly.
Please could you take a look?

Thanks.

stephen-turner: @huizh I'm a bit confused, I still only see one commit. Where is the new diff?

huizh: Hi @stephen-turner, I forced push the commit (normally for toolstack PR, with update we forced push the commit). Please let me know if it's inappropriate.

Here are the change list:
1). Update radioButton text "Enable" to "Enable IGMP snooping", "Disable" to "Disable IGMP snooping".
2). Update the description to "Enable IGMP (Internet Group Management Protocol) snooping will prevent flooding multicast traffic to all the ports in a broadcast domain. Select whether you wish to enable IGMP snooping."

With the update, the UI is:
image

Does it make sense?
Thanks.

kc284: @huizh force-pushing your commit is fine.
However, I don't see the adjustment making a significant difference to what we had before.

stephen-turner: I understand now. Thank you. But in general, I would prefer a diff on top of the previous change, otherwise the history of the comments is lost.

stephen-turner: From the text on the UI, it sounds as if "Enable" is always the correct option. Why would anyone choose "Disable"?

huizh: This is based on requirement `R4. IGMP Snooping must be disabled by default`. Also to get the benefit, this feature depends on the user to turn on IGMP snooping querier functionality on physical switch (will document and let customer know), so customer should configure their switch then enable it.

stephen-turner: Understood. I'm just wondering how to explain the options.

If there is a configuration needed on the switch, we should mention that on the page too. What will happen if they select "Enable" before configuring their switch. Will it act the same as Disabled, or will network traffic get blocked, or what?

huizh: It will act similar as Disabled and will not block network traffic.

stephen-turner: In that case we should mention the need to configure the switch in the text.

I'm not very happy with the text in general, but I don't know enough about the topic to rewrite it. But it seems to have a lot of jargon in it.

I'm happy with the layout. I think radio buttons are appropriate here.

kc284: I'll concede to using radios if we're writing something to warrant their use (like the example at https://msdn.microsoft.com/en-us/library/windows/desktop/dn742436(v=vs.85).aspx#Labels). Leaving it as it is now, "Enable X"/"Disable X", looks amature.

The argument against the checkbox was that it leaves too much space empty. Since it turns out we need some advisory text with it, why don't we use the checkbox and put the advisory text underneath it with a little info icon in front of it? This will fill the whitespace and it will look exactly like the Host Properties > Multipathing tab, for example:

image

With regards to the exact text I suggest we ask the docs team to lend us a helping hand.

huizh: Hi @stephen-turner and @kc284, thanks for the discussion and suggestions. To proceed with code refine, please could you make a decision that which layout should we take, "raidobutton" or "checkbox"?

Thanks a lot.

stephen-turner: I think we need to decide the exact text and then see which looks better.

huizh: Hi @stephen-turner @kc284, here is the suggested text.
Also please could @robhoes comment for your expertise?

`Enable IGMP (Internet Group Management Protocol) snooping will prevent flooding multicast traffic to all the ports in a broadcast domain. Select whether you wish to enable IGMP snooping.`
`To achieve the benefit, please turn on IGMP snooping querier functionality on the physical switch.`

Thanks.

stephen-turner: I'm not really happy with this text, but I don't know enough to rewrite it accurately.

robhoes: How about something along the lines of:
> IGMP snooping is a feature that makes that use of IPv4 multicast more efficient, by forwarding packets only to the subscribers of multicast streams. It is generally recommended to enable this. Depending on the configuration of your LAN, it may also be necessary to enable an IGMP querier on one of the physical switches. Please consult the documentation of your switch for details.

stephen-turner: That's good, @robhoes, thank you.

Instead of the second sentence, we could add "(Recommended)" to the On radio button as shown at https://msdn.microsoft.com/en-us/library/windows/desktop/dn742436(v=vs.85).aspx#Labels. This would give a reason for using radio buttons instead of a checkbox. (It's a bit weird that the recommended option is not the default one, but we've discussed that).

I think I would add the last two sentences ("Depending on...") underneath the radio buttons with a lightbulb icon.

So the whole thing would look like this:

IGMP snooping is a feature that makes the use of IPv4 multicast more efficient, by forwarding packets only to the subscribers of multicast streams.

( ) Enable IGMP snooping (recommended)
( ) Disable IGMP snooping

💡 Depending on the configuration of your LAN, it may also be necessary to enable an IGMP querier on one of the physical switches. Please consult the documentation of your switch for details.

huizh: Thanks @stephen-turner and @robhoes for the text and layout suggestion.
And thank you @kc284 for help guiding the implementation of the layout.
I've updated accordingly. 
Here is the screenshot:
igmp-snooping-updated
(Since there's no existing lightbubl icon, I use info icon instead.)

Could you review and let me know if there's anymore comment?
Thanks.

stephen-turner: The screenshot looks good to me. (I haven't reviewed the code).

XenAdmin/SettingsPanels/NetworkOptionsEditPage.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            return new DelegatedAsyncAction(pool.Connection, title, description, null,
+               delegate(Session session) { Pool.set_igmp_snooping_enabled(session, pool.opaque_ref, radioButtonEnable.Checked); }, true);
+        }
+
+        public void SetXenObjects(IXenObject orig, IXenObject clone)
+        {
+            pool = Helpers.GetPoolOfOne(clone.Connection);  // clone could be a pool or a host
+
+            if (pool.igmp_snooping_enabled)
```
kc284: I would add a null check on the pool here to be on the safe side.

XenAdmin/SettingsPanels/NetworkOptionsEditPage.cs
```diff
@@ -0,0 +1,126 @@
...
+        public void ShowLocalValidationMessages()
+        { }
+
+        public void Cleanup()
+        { }
+
+        public bool HasChanged
+        {
+            // Server flag is for enabled, so compare to enable radio
+            get { return pool.igmp_snooping_enabled != radioButtonEnable.Checked; }
```
kc284: Better add a null check on the pool here too.

XenAdmin/SettingsPanels/NetworkOptionsEditPage.cs
huizh: Thanks. Updated.

XenAdmin/SettingsPanels/NetworkOptionsEditPage.cs
huizh: Thanks. Updated.

huizh:
As the layout and text have been decided, and the comments have been addressed, I merge the commits to one and force pushed it here.

Please could you merge it there's no other comment?

Thanks.

PR #1727, opened 2017-08-11
Sort pool-host collapsing rows other than name column in different views

jijiang:


kc284:
There is a lot of code repetition here, I think the whole thing can be written in more compact way:

- make one sorter class, and since it's private and lives in this file, call it LocalRowSorter so it's readable
- make its constructor take two parameters, direction and clickedCellIndex, the latter will be stored in a private field which will replace the [lhs_c.VersionCellIndex] and [rhs_c.VersionCellIndex in the PerformSort method (and also check whether this index is within bounds just in case)
- then your private gridview class will be
```
private class UpdatePageByHostDataGridView : CollapsingPoolHostDataGridView {
    protected override void SortAdditionalColumns() {
        SortAndRebuildTree(new LocalRowSorter(direction, columnToBeSortedIndex));
    }
}
```
- remove the extra properties in the UpdatePageDataGridViewRow 

If I haven't missed anything and the whole thing works ok, you can add a second commit with the refactoring for the RollingPoolUpgradeWizard and a third commit to implement it on the UpdateWizard where we have the same bug.

PR #1730, opened 2017-08-15
CP-23455: Add Update btn in Updates page "By Host"

jijiang:

### DIFF
--- a/XenAdmin/TabPages/ManageUpdatesPage.cs
+++ b/XenAdmin/TabPages/ManageUpdatesPage.cs
@@ -82,6 +82,7 @@ namespace XenAdmin.TabPages
             Updates.CheckForUpdatesCompleted += CheckForUpdates_CheckForUpdatesCompleted;
             toolStripSplitButtonDismiss.DefaultItem = dismissAllToolStripMenuItem;
             toolStripSplitButtonDismiss.Text = dismissAllToolStripMenuItem.Text;
+            toolStripButtonUpdate.Visible = false;
         }
 
         public void RefreshUpdateList()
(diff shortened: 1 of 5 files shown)


XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -1307,5 +1312,23 @@ private void byHostToolStripMenuItem_Click(object sender, EventArgs e)
             dataGridViewHosts.Visible = true;
             Rebuild();
         }
+
+        private void toolStripButtonUpdate_Click(object sender, EventArgs e)
+        {
+            Program.Invoke(Program.MainWindow, () =>
```
kc284: Why do we need Invoke here? Is it likely this won't be on the UI thread? (I also see the same in line 919?)

jijiang:
Thanks for comments. Updated the commit. I think it is in the UI thread, so remove invoke function. Also I remove the Program.Invoke() function inside ToolStripMenuItemDownload_Click().

kc284:
These changes are ok. However, I tested the behaviour again with a few more hosts of various versions and now I have mixed feelings about it: If there are no hosts where automated updates apply, the users may find themselves in the 3rd wizard page with all the hosts disabled, which is a bit strange; are they supposed to go back to the 2nd page and try a different mode? Then they should be taken straight to that page instead, or the button should be disabled with a tooltip explaining automatic updates do not apply. It may be a good idea then to have a tooltip even when the button is enabled to explain what it does (if it's supposed to launch the wizard only for automated updates, this is not very clear from the button text).

jijiang:
Thanks for the comment. Now the update button checks the automated updates availability to decide showing tooltip or becoming enabled.

XenModel/Messages.resx
```diff
@@ -7429,6 +7429,9 @@ This will permanently delete and reinitialize all local storage on the servers.
   <data name="MANAGEMENT_NETWORK_WITH_POOL" xml:space="preserve">
     <value>{0} (management) on {1}</value>
   </data>
+  <data name="MANAGE_UPDATES_PAGE_AUTOMATED_UPDATES_NOT_AVAILABLE" xml:space="preserve">
+    <value>Automated updates are not available</value>
```
GaborApatiNagy: Good point @kc284 

Although I would shorten it a bit to:
"Automated updates are not supported for these servers"
what is the plural form of POOL_FORBIDS_AUTOMATED_UPDATES="Automated updates are not supported for this server" that is already being used.

I suggest we should just show that one if there is only one server in the list and also add the plural version that would be shown when there are more than one servers shown.

jijiang:
Updated.

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -669,7 +670,45 @@ private void StoreSelectedUpdates()
...
+
+            // check Automated Updates Availability
+            foreach (IXenConnection connection in connectionList)
+            {
+                var poolOfOne = Helpers.GetPoolOfOne(connection);
+                var pool = Helpers.GetPool(connection);
+
+                if ((poolOfOne != null && poolOfOne.IsAutoUpdateRestartsForbidden) ||
+                    (pool != null && !pool.IsPoolFullyUpgraded) ||
+                    (Updates.GetUpgradeSequence(connection) == null) ||
```
kc284: Here also, the Updates.GetUpgradeSequence(connection) is called three times. It's much more efficient to calculate it once: `var seq =  Updates.GetUpgradeSequence(connection)` and then use this instead. This means you'll probably have to split this big if check to smaller ifs and continues.

jijiang:
The original commit has been modified. Currently update button in "by host" view would lead to the second page of install update wizard. And become disabled and show tooltip if all host(s) are already up to date.

PR #1731, opened 2017-08-16
CP-23967: Fix the same string sorting bug

jijiang:
The bug is that no sorting is performed when the strings are the same.


jijiang:
After discussion with Stephen and Mihaela, the name column should be sorted purely on alphabetic order, not by default sorting. And in cases of same string if sorting "by version" and "by status", it needs further sorting by default, with no direction in reverse case...
I will fix the above comments, and resubmit a new pull request. So close this PR now.

PR #1732, opened 2017-08-16
CA-260366: Adding description when signing

GaborApatiNagy:
Updated dotnet-packages location


MihaelaStoica:
@GaborApatiNagy I think we should update the dotnet-packages location in DOTNET_BUILD_LOCATION_CTXSIGN as well

PR #1733, opened 2017-08-17
CA-248292: Fix btn truncation in VM storage page

jijiang:
The bug is that two button texts are truncated in Japanese version. In
order to keep consistency, increase the width of all buttons to the same size.

### DIFF
--- a/XenAdmin/TabPages/VMStoragePage.ja.resx
+++ b/XenAdmin/TabPages/VMStoragePage.ja.resx
@@ -59,46 +59,46 @@
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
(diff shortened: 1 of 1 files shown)


kc284:
This makes the buttons wider in the english version, but the complaint was about the japanese version, so the modification should be in the *.ja.resx file.

XenAdmin/TabPages/VMStoragePage.resx
```diff
@@ -421,13 +421,13 @@
     <value>1</value>
   </data>
   <data name="DetachButtonContainer.Location" type="System.Drawing.Point, System.Drawing">
-    <value>602, 2</value>
```
stephen-turner: I haven't looked at the control, but this line of the diff looks suspicious. It seems one of the buttons has moved a long distance.

kc284:
It's kind of difficult to review changes in resx files, and in an unknown language on top of it, but I pulled down the code and it looks ok to me. There are several lines that have been removed automatically by the designer, but I think this is ok because it's properties that are the same as in english and the l10n resx file keeps the overriden only. I'm merging this.

PR #1734, opened 2017-08-18
CA-261107: Metadata file osVersion is showing same value for XenCenter on different windows versions

GaborApatiNagy:
Added app manifest to explicitly target XC to all windows versions from Vista to Windows 10 so that the Environment variables will reflect the correct Windows versions. Note that "Manifesting the .exe for Windows 8.1 or Windows 10 will not have any impact when run on previous operating systems." [1]

[1] https://msdn.microsoft.com/en-gb/library/windows/desktop/dn481241.aspx


stephen-turner:
Merging

PR #1735, opened 2017-08-18
CP-23967: Fix and unify sorting in different views

jijiang:
The commit fixes bug and unify sorting rules in ManageUpdatesPage,
PatchingWizard_SelectServers and RollingUpgradeWizardSelectPool data
grid views.

The rules are:
1) Name column sorted purely by name.
2) Other columns sorted by string value, if the same, sorted by default,
which is still in order even in reverse direction.


kc284:
Closing this because the commit is in PR #1739, so it can be reviewed there.

PR #1739, opened 2017-08-22
CP-24032, CP-23967: Collapsing and sorting functionality of Updates Page

jijiang:
CP-23967 fix PR should be submitted before this PR.

### DIFF
--- a/XenAdmin/Controls/DataGridViewEx/CollapsingPoolHostDataGridViewRow.cs
+++ b/XenAdmin/Controls/DataGridViewEx/CollapsingPoolHostDataGridViewRow.cs
@@ -125,12 +125,16 @@ namespace XenAdmin.Controls.DataGridViewEx
         public void SetCollapseIcon()
         {
             _expansionCell.Value = Resources.tree_minus;
+            IsACollapsedRow = false;
         }
 
         public void SetExpandIcon()
         {
             _expansionCell.Value = Resources.tree_plus;
+            IsACollapsedRow = true;
         }
+
+        public bool IsACollapsedRow { get; protected set; }
 
         /// <summary>
         /// Convert row to a standalone host row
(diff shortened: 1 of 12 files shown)


kc284:
I've reviewed  commit 13b9456 (the other one is in PR #1735, so it will be reviewed there and that PRhas to be merged first).
It looks ok to me; an enhancement to store the expanded/collapsed state when the tree is rebuilt would be nice in order to provide a better user experience, because if they have collapsed their hosts and the tree is rebuilt all will be expanded again (as long as this is not too detrimental to performance that is).

XenAdmin/Controls/DataGridViewEx/CollapsingPoolHostDataGridViewRow.cs
```diff
@@ -125,11 +126,18 @@ public bool HasPool
...
+            _isACollapsedRow = false;
         }
 
         public void SetExpandIcon()
         {
             _expansionCell.Value = Resources.tree_plus;
+            _isACollapsedRow = true;
+        }
+
+        public bool IsACollapsedRow
```
kc284: Since there is no logic in the property body, you can make the code more compact by defining it as:
```
 public bool IsACollapsedRow { get; protected set; }
```
and then you don't need the extra field.

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -696,6 +724,17 @@ private void StoreSelectedUpdates()
             }
         }
 
+        private void StoreCollapsedPoolRow()
```
kc284: "StoreCollapsedPoolRows" in plural (they may be many).

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -517,6 +520,31 @@ private void RebuildHostView()
...
+                        if (!row.IsPoolOrStandaloneHost)
+                        {
+                            row.Visible = !row.Visible;
+                            continue;
+                        }
+                        else
+                            checkHostRow = false;
+                    }
+                    Pool pool = row.Tag as Pool;
+                    if (pool != null)
```
kc284: These two ifs can be merged into one:
```
if (pool != null && collapsedPoolRowsList.Contains(pool.uuid))
```

kc284:
Looks good to me now. Merging.

PR #1740, opened 2017-08-22
CP-24087: Parse the error parameters of the UPDATE_PRECHECK_FAILED_UN…

MihaelaStoica:
…KNOWN_ERROR precheck failure

When the update precheck fails with UPDATE_PRECHECK_FAILED_UNKNOWN_ERROR, we try to find the actual error in the failure's parameters (in the form of an xml string containing the error code) and display that error instead of the "unknown error".


stephen-turner:
I'm convinced. Pulling.

PR #1744, opened 2017-08-24
CP-24210: Populate "By Host" view correctly

jijiang:
1) Show required updates in case of versions without minimal patches
2) Hide version text in installed updates when version is Ely or above

### DIFF
--- a/XenAdmin/TabPages/ManageUpdatesPage.cs
+++ b/XenAdmin/TabPages/ManageUpdatesPage.cs
@@ -354,11 +354,11 @@ namespace XenAdmin.TabPages
                     _requiredUpdateCell.Value = "";
                     _installedUpdateCell.Value = "";
 
-                    var poolPatchingStatus = pool.Connection.Cache.Hosts.Any(h => Updates.RecommendedPatchesForHost(h).Count > 0);
-                    _patchingStatusCell.Value = poolPatchingStatus
+                    var outOfDate = pool.Connection.Cache.Hosts.Any(h => RequiredUpdatesForHost(h).Length > 0);
+                    _patchingStatusCell.Value = outOfDate
                         ? Properties.Resources._000_error_h32bit_16
                         : Properties.Resources._000_Tick_h32bit_16;
-                    _statusCell.Value = poolPatchingStatus ? Messages.NOT_UPDATED : Messages.UPDATED;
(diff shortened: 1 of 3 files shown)


XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -400,14 +399,46 @@ private void UpdateDetails()
...
+                    if (!(alert is XenServerPatchAlert))
+                        continue;
+                    if (alert.AppliesTo.Contains(host.Name))
+                        updatesList.Add(alert.Name);
+                }
+
+                updatesList.Sort(StringUtility.NaturalCompare);
+                return string.Join(", ", updatesList.ToArray());
+            }
+
```
GaborApatiNagy: It is good as is, but for better maintainability I'd suggest you put the rest in an else branch, so it will make sure that the return in line 424 can't just be removed and also this would make the code a lot more readable.
Also, the 'result' List is only needed here, let's move it here.

```
if ()
{
unchanged
return unchanged
}
else //new
{
var result....
everything else
return
}
```

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -400,14 +399,50 @@ private void UpdateDetails()
...
+            else
+            {
+                // versions with minimum patches
+                var result = new List<string>();
+                foreach (var patch in requiredUpdates)
+                    result.Add(patch.Name);
+
+                result.Sort(StringUtility.NaturalCompare);
```
MihaelaStoica: Shouldn't we show the minimal patches in the original order, as this is the order that they need to be applied on? ~Also, on the General tab we show them in the original order...

GaborApatiNagy: I see this as just a list of the missing ones. We could use the minimal patch order, but because for older hosts we don't know that, sorting wouldn't be consistent between old and new pools.

If that's what we'd like to have, I'm fine with it, despite I still prefer the alphabetical sorting as I see this rather as a list, not as an ordered install sequence.

stephen-turner: We're already using the minimal list instead of the full list for newer pools, so they're already not consistent. I think we should preserve consistency with the general tab.

PR #1745, opened 2017-08-25
CP-24213: Show CUs as updates

MihaelaStoica:
CUs are available as both updates and upgrades, but they should be presented as updates where possible.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
@@ -400,7 +400,7 @@ namespace XenAdmin.Wizards.PatchingWizard
                         ? Updates.GetUpgradeSequence(pool.Connection, UpdateAlert, ApplyUpdatesToNewVersion && !automatedUpdatesRestricted)
                         : Updates.GetUpgradeSequence(pool.Connection);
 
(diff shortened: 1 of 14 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -400,6 +400,8 @@ private void worker_DoWork(object sender, DoWorkEventArgs e)
                         ? Updates.GetUpgradeSequence(pool.Connection, UpdateAlert, ApplyUpdatesToNewVersion && !automatedUpdatesRestricted)
                         : Updates.GetUpgradeSequence(pool.Connection);
 
+                    log.InfoFormat("Minimal patches for {0}: {1}", pool.Name, string.Join(",", us.MinimalPatches.Select(p => p.Name)));
```
GaborApatiNagy: I think this should definitely be after the null check on 'us' (in next line). Additionally, a null check for the MinimalPatches is also needed here.

GaborApatiNagy: ...or we can just leave it here, but do the null-check inline (to log that case too)

MihaelaStoica: Fixed in the new commit

GaborApatiNagy:
Looks good
Merging

PR #1749, opened 2017-09-05
REQ-420 Preliminary refactoring Pt2

kc284:
Conversion of the remaining extension properties to methods. There are still 2-3 in XenObject left and another 2-3 in the VM class which require much more thought, so for the time being I'll just mark them as non-serializable.


MihaelaStoica:
We discussed the first commit (71301eb) and decided to put the GetAlerts() method call on a background thread as it iterates through a possible large list of messages and is executed every time the policy selection is changed.

The second commit (514c331) looks fine

XenAdmin/Dialogs/ScheduledSnapshots/ScheduledSnapshotsDialog.cs
```diff
@@ -144,90 +185,130 @@ private void RefreshPoolTitle(Pool pool)
...
                                                         ? Messages.VMSS_SCHEDULED_SNAPSHOTS_DEFINED_FOR_POOL
                                                         : Messages.VMSS_SCHEDULED_SNAPSHOTS_DEFINED_FOR_SERVER,
                                                     pool.Name().Ellipsise(45), protectedVMs, realVMs);
         }
 
-        void VMSSCollectionChanged(object sender, EventArgs e)
+        private void VMSSCollectionChanged(object sender, EventArgs e)
         {
-            LoadPolicies();
+            if (!bgWorker.IsBusy)
```
MihaelaStoica: I think we could use a QueuedBackgroundWorker here, so that we don't ignore CollectionChanged events that happen while the bgWorker is busy.

MihaelaStoica:
Looks good now. Merging.

PR #1757, opened 2017-09-12
Merging REQ-420 into master

kc284:
@MihaelaStoica as discussed here is a PR to merge the preliminary changes of REQ-420 into master before more changes come.


MihaelaStoica:
Merging.

PR #1761, opened 2017-09-13
CP-24196: Initial implementation of switching XenCenter to use the JsonRpc backend

kc284:


XenModel/XenAPI/Session.cs
```diff
@@ -315,11 +355,36 @@ private void SetAPIVersion()
                 APIVersion = Helper.GetAPIVersion(host.API_version_major, host.API_version_minor);
                 break;
             }
+
+            //if supported swap endpoints
+            JsonRpcClient = null;
+            bool isELy = (int)APIVersion == (int)API_Version.API_2_6;
```
MihaelaStoica: Shouldn't this be "is Ely or greater"? 
As JsonRpcVersion.v2 is not supported on Falcon, I'm assuming that `gtFalcon `will become `(int)APIVersion > (int)API_Version.API_2_7` when the API version gets bumped for Inverness.
Then line 364 would be  `if (isELyOrGreater)`

kc284: We're not supporting json-rpc at all for falcon, it will be ely or inverness and above (I'll rename gtFalcon to invernessOrGreater once the version is bumped so it's clearer).

MihaelaStoica: OK, it all makes sense now :)

PR #1764, opened 2017-09-14
CA-265684: Fix bug of disabled "Export All" button

jijiang:
The bug is that "Export All" button is disabled when all hosts are up to
date.


MihaelaStoica:
Looks good. Merging.

PR #1780, opened 2017-09-22
CP-24704: Add pool size license check

jijiang:

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -8047,6 +8047,9 @@ You should only proceed if you have verified that these settings are correct.</v
   <data name="NEWPOOL_MASTER_ROLE" xml:space="preserve">
     <value>Your current role on the master is not authorized to add hosts to the master's pool</value>
   </data>
+  <data name="NEWPOOL_MAX_NUMBER_HOST_REACHED" xml:space="preserve">
+    <value>The pool has already reached the maximum number of servers allowed by your license</value>
+  </data>
   <data name="NEWPOOL_POOLINGRESTRICTED" xml:space="preserve">
     <value>Pooling is restricted with this server's license</value>
   </data>
--- a/XenModel/PoolJoinRules.cs
+++ b/XenModel/PoolJoinRules.cs
@@ -59,6 +59,7 @@ namespace XenAdmin.Core
             LicensedHostUnlicensedMaster,
             UnlicensedHostLicensedMaster,
             LicenseMismatch,
+            MasterPoolMaxNumberHostReached,
             DifferentServerVersion,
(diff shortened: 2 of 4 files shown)


XenModel/PoolJoinRules.cs
```diff
@@ -461,6 +467,11 @@ private static bool LicenseMismatch(Host slave, Host master)
             return slaveEdition != Host.Edition.Free && masterEdition != Host.Edition.Free && slaveEdition != masterEdition;
         }
 
+        private static bool UnlicensedMasterMaxNumberHostReached(IXenConnection connection, Host master)
+        {
+            return connection.Cache.HostCount > 2 && master.IsFreeLicense();
```
MihaelaStoica: This check only needs to be performed if the pool has the pool size restriction (the "restrict_pool_size" license key exists and it is true).
Then `master.IsFreeLicense()` is not needed

XenModel/Messages.resx
```diff
@@ -8047,6 +8047,9 @@ You should only proceed if you have verified that these settings are correct.</v
   <data name="NEWPOOL_MASTER_ROLE" xml:space="preserve">
     <value>Your current role on the master is not authorized to add hosts to the master's pool</value>
   </data>
+  <data name="NEWPOOL_MAX_NUMBER_HOST_REACHED" xml:space="preserve">
```
MihaelaStoica: The message should say something on the lines of: the pool already reached the maximum number of servers allowed by your license

XenModel/PoolJoinRules.cs
```diff
@@ -130,6 +131,9 @@ public static Reason CanJoinPool(IXenConnection slaveConnection, IXenConnection
 
             if (LicenseMismatch(slaveHost, masterHost))
                 return Reason.LicenseMismatch;
+
+            if (UnlicensedMasterMaxNumberHostReached(masterConnection, masterHost))
```
MihaelaStoica: If this function take masterConnection as parameter, then masterHost shouldn't be needed

XenModel/PoolJoinRules.cs
```diff
@@ -461,6 +467,11 @@ private static bool LicenseMismatch(Host slave, Host master)
             return slaveEdition != Host.Edition.Free && masterEdition != Host.Edition.Free && slaveEdition != masterEdition;
         }
 
+        private static bool MasterPoolMaxNumberHostReached(Host master)
+        {
+            return Host.RestrictPoolSize(master) && master.Connection.Cache.HostCount > 2;
```
MihaelaStoica: I think we should use FeatureForbidden here.

PR #1782, opened 2017-09-25
CP-24498: RPU wizard to block the selection of unlicensed hosts

MihaelaStoica:
Also includes some improvements to the Select Pools page of the RPU wizard
- do not show the checkbox if the item is disabled
- show the icon for the pool items, not only for hosts (making this consistent with the similar page in the Update Wizard)


GaborApatiNagy:
I think it looks good, I've got just one suggestion for the messages: Could we use 'this server' instead of 'the server'?

PR #1783, opened 2017-09-25
CP-24707: Add license status information to the title bar for pool and hosts

GaborApatiNagy:


XenAdmin/MainWindow.cs
```diff
@@ -2846,18 +2849,60 @@ void SearchPanel_SearchChanged()
...
+                        }
+                    }
+                }
+                else if (xenObject is Host)
+                {
+                    var host = xenObject as Host;
+
+                    if (host.Connection != null && host.Connection.IsConnected && host.Connection.CacheIsPopulated)
+                    {
+                        if (Host.GetEdition(host.edition) == Host.Edition.Free || host.LicenseExpiryUTC() < DateTime.UtcNow - host.Connection.ServerTimeOffset)
```
MihaelaStoica: We should use host.IsFreeLicense()

XenModel/XenAPI-Extensions/Pool.cs
```diff
@@ -95,6 +95,21 @@ public bool IsMasterUpgraded()
...
 
+        public bool IsUnlicensedPool
+        {
+            get
+            {
+
+                bool unlicensed = true;
+                
+                if (Connection != null && Connection.Cache != null && Connection.Cache.Hosts != null
+                    && Connection.Cache.Hosts.All(h => Host.GetEdition(h.edition) != Host.Edition.Free && h.LicenseExpiryUTC() > DateTime.UtcNow - h.Connection.ServerTimeOffset))
```
MihaelaStoica: I don't think that Cache and Cache,Hosts can ever be null, so there is no need for these chacks.

XenAdmin/MainWindow.cs
```diff
@@ -2846,18 +2849,60 @@ void SearchPanel_SearchChanged()
...
                 TitleLabel.Text = HelpersGUI.GetLocalizedSearchName(SearchPage.Search);
+                LicenseStatusTitleLabel.Text = string.Empty;
                 TitleIcon.Image = Images.GetImage16For(SearchPage.Search);
             }
             else if (!SearchMode && SelectionManager.Selection.ContainsOneItemOfType<IXenObject>())
             {
                 IXenObject xenObject = SelectionManager.Selection[0].XenObject;
                 TitleLabel.Text = xenObject.NameWithLocation();
+
+                if (xenObject is Pool)
```
MihaelaStoica: I think we should extract this block into a separate method and maybe try to reduce the code repetition

PR #1789, opened 2017-09-28
CP-24412 & CP-24333: USB device list filter in Attach dialog.

michael2012z:
This PR contains the GUI update for 2 tasks:
CP-24412: Check whether a USB device is valid before adding it to candidate list in Attach dialog.
CP-24333: Apply the constraint of HA.


XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -75,6 +96,31 @@ private void BuildList()
             }
         }
 
+        private void treeUsbList_SelectedIndexChanged(object sender, EventArgs e)
+        {
+            UsbItem item = treeUsbList.SelectedItem as UsbItem;
```
GaborApatiNagy: I think this whole method can be implemented in one line as:

```
buttonAttach.Enabled = treeUsbList.SelectedItem is UsbItem;
```

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -56,16 +56,37 @@ private void BuildList()
...
             {
-                Host[] hosts = _vm.Connection.Cache.Hosts;
-                foreach (Host host in hosts)
+                VM.HA_Restart_Priority SelectedPriority = _vm.HARestartPriority();
+                Pool pool = Helpers.GetPool(_vm.Connection);
+                // Check if HA was enabled on pool and Restart priority was set on VM.
+                bool haEnabled = (pool != null &&
+                    pool.ha_enabled &&
+                    VM.HaPriorityIsRestart(_vm.Connection, SelectedPriority));
+                List<XenRef<Host>> possibleHostRefs = VM.get_possible_hosts(_vm.Connection.Session, _vm.opaque_ref);
```
MihaelaStoica: We shouldn't execute server calls on the UI thread. We can use a DelegateAsyncAction or can create a new action class.

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -56,16 +56,37 @@ private void BuildList()
...
+                List<Host> possibleHosts = new List<Host>();
+                foreach (XenRef<Host> possibleHostRef in possibleHostRefs)
```
MihaelaStoica: `ResolveAll` can be used here:
`List<Host> possibleHosts = _vm.Connection.ResolveAll(possibleHostRef)`

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -75,6 +96,31 @@ private void BuildList()
...
+            }
+            else
+            {
+                buttonAttach.Enabled = false;
+            }
+        }
+
+        private void buttonAttach_Click(object sender, EventArgs e)
+        {
+            this.DialogResult = System.Windows.Forms.DialogResult.OK;
```
MihaelaStoica: The DialogResult can be set directly in the designer and there is no need to call Close() here

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -56,16 +56,37 @@ private void BuildList()
...
                     treeUsbList.AddNode(hostNode);
                     List<PUSB> pusbs = _vm.Connection.ResolveAll(host.PUSBs);
                     foreach (PUSB pusb in pusbs)
                     {
-                        UsbItem usbNode = new UsbItem(pusb);
-                        treeUsbList.AddChildNode(hostNode, usbNode);
+                        // Add a USB in the host to tree list.
+                        // Determin if the USB is valid to attach.
+                        if ((haEnabled == false) &&
+                            (pusb != null) &&
```
MihaelaStoica: pusb cannot be null, so this check is not needed

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -56,16 +56,37 @@ private void BuildList()
             treeUsbList.BeginUpdate();
             try
             {
-                Host[] hosts = _vm.Connection.Cache.Hosts;
-                foreach (Host host in hosts)
+                VM.HA_Restart_Priority SelectedPriority = _vm.HARestartPriority();
+                Pool pool = Helpers.GetPool(_vm.Connection);
+                // Check if HA was enabled on pool and Restart priority was set on VM.
+                bool haEnabled = (pool != null &&
```
MihaelaStoica: The HA check should be done on the form that opens this dialog, where we should disable the USB attach button and display the message stating why cannot attach USB devices.

michael2012z: Good idea. HA-check is on VM, but not on each candidate, should be done on the form. Accepted.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: VM.HaPriorityIsRestart(IXenConnection connection, HA_Restart_Priority haRestartPriority) do not make any communication with server, but compare local fields. Parameter 'connection' is not used in fact. So it's safe to keep this code in UI. We called this method in some other UI pages/forms as well. Do you agree?

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Efficient code! Accepted.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Agree.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Got it. Thanks.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Accepted.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Sorry, I think I misunderstood. You mean get_possible_hosts(), right?

MihaelaStoica: Yes

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -56,16 +56,48 @@ private void BuildList()
...
                     foreach (PUSB pusb in pusbs)
                     {
-                        UsbItem usbNode = new UsbItem(pusb);
-                        treeUsbList.AddChildNode(hostNode, usbNode);
+                        // Add a USB in the host to tree list.
+                        // Determin if the USB is valid to attach.
+                        if (pusb != null)
+                        {
+                            bool attached = false;
+                            USB_group usbGroup = pusb.Connection.Resolve(pusb.USB_group);
```
MihaelaStoica: This can be written as: 
`attached = usbGroup != null && usbGroup.VUSBs != null && usbGroup.VUSBs.Count > 0;`

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -56,16 +56,48 @@ private void BuildList()
...
+                            if ((usbGroup != null) && 
+                                (usbGroup.VUSBs != null) && 
+                                (usbGroup.VUSBs.Count > 0))
+                                attached = true;
+                            else
+                                attached = false;
+
+                            if ((haEnabled == false) &&
+                                (pusb.passthrough_enabled == true) &&
```
MihaelaStoica: This can be written: 
`if (pusb.passthrough_enabled && !attached)` 
 (I ignored haEnabled as we agreed that this dialog won't even be shown is ha is enabled)

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -75,6 +107,30 @@ private void BuildList()
...
+            else
+            {
+                buttonAttach.Enabled = false;
+            }
+        }
+
+        private void buttonAttach_Click(object sender, EventArgs e)
+        {
+            UsbItem item = treeUsbList.SelectedItem as UsbItem;
+            System.Threading.ThreadPool.QueueUserWorkItem((System.Threading.WaitCallback)delegate (object o)
```
MihaelaStoica: Although we disable the button for items that are not UsbItem, we should still add the null check here for `item`

MihaelaStoica: The action will run on a different thread, so there is no need for another thread here.

Same for UsbUsageDialog.cs

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -88,7 +144,7 @@ public HostItem(Host host) :base(false)
 
         private class UsbItem : CustomTreeNode
         {
-            private PUSB _pusb;
+            public PUSB _pusb { get; }
```
MihaelaStoica: Please use PascalCasing for public members (capitalizes the first character, no underscore)

XenModel/Actions/USB/USBPassthrough.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using XenAPI;
+
+namespace XenAdmin.Actions
+{
+    public class USBPassthrough : PureAsyncAction
```
MihaelaStoica: Please rename this action to `SetUsbPassthroughAction` or something similar

XenModel/Actions/USB/CreateVUSBAction.cs
```diff
@@ -0,0 +1,68 @@
...
+        public CreateVUSBAction(PUSB pusb, VM vm) : 
+            base(pusb.Connection, String.Format(Messages.ACTION_VUSB_CREATING,  pusb.Name(), vm.Name()))
+        {
+            _pusb = pusb;
+            _vm = vm;
+        }
+
+        protected override void Run()
+        {
+            try
```
MihaelaStoica: In general there is no need for a try..catch in the action, because the action itself handles exceptions and sets the Description to the exception's message, which is more useful than a generic "action failed" message.

XenModel/Actions/USB/DeleteVUSBAction.cs
```diff
@@ -0,0 +1,71 @@
...
+            base(vusb.Connection, String.Format(Messages.ACTION_VUSB_DELETING, vusb.Name(), vm.Name()))
+        {
+            _vusb = vusb;
+        }
+
+        protected override void Run()
+        {
+            try
+            {
+                VUSB.async_unplug(Session, _vusb.opaque_ref);
```
MihaelaStoica: As `async_unplug` is used, we need to wait for the task to be completed before continuing with the other calls, e.g. `async_destroy`. See how we use `PollToCompletion `in other places, for example in DetachVirtualDiskAction class

XenModel/Messages.resx
```diff
@@ -1209,6 +1227,30 @@
   <data name="ACTION_VM_TEMPLATIZING_TITLE" xml:space="preserve">
     <value>Converting VM '{0}' to template</value>
   </data>
+  <data name="ACTION_VUSB_CREATED" xml:space="preserve">
+    <value>Virtual USB created</value>
```
MihaelaStoica: I don't think we need to expose the term "virtual USB" to the user. From their point of view, they are attaching an USB device to a VM, so this is what the action should say.

XenAdmin/MainWindow.cs
```diff
@@ -1446,10 +1446,9 @@ public void UpdateToolbarsCore()
             ShowTab(TabPageNICs, !multi && !SearchMode && ((isHostSelected && isHostLive)));
             ShowTab(TabPageDockerProcess, !multi && !SearchMode && isDockerContainerSelected && !(SelectionManager.Selection.First as DockerContainer).Parent.IsWindows());
             ShowTab(TabPageDockerDetails, !multi && !SearchMode && isDockerContainerSelected);
-            ShowTab(TabPageUSB, !multi && !searchMode && ((isHostSelected && isHostLive)));
 
             bool isPoolOrLiveStandaloneHost = isPoolSelected || (isHostSelected && isHostLive && selectionPool == null);
-
+            ShowTab(TabPageUSB, !multi && !SearchMode && ((isHostSelected && isHostLive) || isPoolOrLiveStandaloneHost) && Helpers.InvernessOrGreater(selectionConnection) && !Helpers.FeatureForbidden(selectionConnection, Host.RestrictUsbPassthrough));
```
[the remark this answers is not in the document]
michael2012z: OK.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Agree.

michael2012z: Agree.

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: OK.

XenModel/Actions/USB/DeleteVUSBAction.cs
michael2012z: Thanks, I will use PollToCompletion.

XenModel/Messages.resx
michael2012z: Good idea, I'm going to change "virtual USB" into "USB device".

XenModel/Messages.resx
```diff
@@ -8850,6 +8892,9 @@ It is strongly recommended that you Cancel and apply the latest version of the p
   <data name="NOT_AGILE_VM_HAS_VGPU" xml:space="preserve">
     <value>The VM has one or more virtual GPUs. Restart cannot be guaranteed.</value>
   </data>
+  <data name="NOT_AGILE_VM_HAS_VUSB" xml:space="preserve">
+    <value>The VM has one or more virtual USBs. Restart cannot be guaranteed.</value>
```
MihaelaStoica: _The VM has one or more USB devices attached. Restart cannot be guaranteed._

XenModel/Messages.resx
```diff
@@ -1209,6 +1221,24 @@
...
     <value>Converting VM '{0}' to template</value>
   </data>
+  <data name="ACTION_VUSB_CREATED" xml:space="preserve">
+    <value>USB device attached</value>
+  </data>
+  <data name="ACTION_VUSB_CREATING" xml:space="preserve">
+    <value>Attaching USB device '{0}' to VM '{1}'</value>
+  </data>
+  <data name="ACTION_VUSB_DELETED" xml:space="preserve">
+    <value>USB device deleted</value>
```
MihaelaStoica: Although we are internally deleting the VUSB, the message that the user needs to see is _detachig_. 
So:
ACTION_VUSB_DELETED: USB device detached
ACTION_VUSB_DELETING: Detaching USB device '{0}' from VM '{1}'
ACTION_VUSB_DETACH: Detach USB device
ACTION_VUSB_DETACH_CONFIRM: Are you sure you want to detach the selected USB device?

XenModel/Messages.resx
```diff
@@ -972,6 +972,18 @@
   <data name="ACTION_UPLOAD_SERVER_STATUS_REPORT_PROGRESS" xml:space="preserve">
     <value>Uploading report to [Citrix] Insight Services</value>
   </data>
+  <data name="ACTION_USB_PASSTHROUGH_DISABLED" xml:space="preserve">
+    <value>USB device detached.</value>
```
MihaelaStoica: The action's title/description should reflect the UI, so if the button says "enable passthrough", then the action should also say "enabling passthrough".
ACTION_USB_PASSTHROUGH_DISABLED: USB passthrough disabled
ACTION_USB_PASSTHROUGH_DISABLING: Disabling USB passthrough for {0}
ACTION_USB_PASSTHROUGH_ENABLED: USB passthrough enabled
ACTION_USB_PASSTHROUGH_ENABLING: Enabling USB passthrough for {0}

XenAdmin/Dialogs/PropertiesDialog.cs
```diff
@@ -149,7 +149,10 @@ private void Build()
                     ShowTab(VCpuMemoryEditPage = new CPUMemoryEditPage());
                     ShowTab(StartupOptionsEditPage = new BootOptionsEditPage());
                     ShowTab(VMHAEditPage = new VMHAEditPage {VerticalTabs = verticalTabs});
-                    ShowTab(usbEditPage = new USBEditPage());
+                    if (!Helpers.FeatureForbidden(xenObjectCopy, Host.RestrictUsbPassthrough))
```
MihaelaStoica: The USB page should be placed between the GPU and Advanced options pages.
The same for the main window: the USB tab should be placed between GPU and Console

XenModel/Actions/USB/SetUsbPassthroughAction.cs
```diff
@@ -0,0 +1,59 @@
...
+        public SetUsbPassthroughAction (PUSB pusb, bool passthroughEnabled) :
+            base(pusb.Connection, String.Format(passthroughEnabled ? Messages.ACTION_USB_PASSTHROUGH_ENABLING : Messages.ACTION_USB_PASSTHROUGH_DISABLING, pusb.Name()))
+        {
+            _pusb = pusb;
+            _passthroughEnabled = passthroughEnabled;
+        }
+
+        protected override void Run()
+        {
+            PUSB.set_passthrough_enabled(_pusb.Connection.Session, _pusb.opaque_ref, _passthroughEnabled);
```
MihaelaStoica: Please use the action's Session property here, as this could be an elevated session if the logged in user doesn't have sufficient permissions.

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -33,20 +33,35 @@
...
+
+            DelegatedAsyncAction action = new DelegatedAsyncAction(_vm.Connection,
+                string.Format(Messages.FETCH_POSSIBLE_HOSTS, _vm.Name()),
+                string.Format(Messages.FETCHING_POSSIBLE_HOSTS, _vm.Name()),
+                string.Format(Messages.FETCHED_POSSIBLE_HOSTS, _vm.Name()),
+                delegate (Session session)
+                {
+                    List<XenRef<Host>> possibleHostRefs = VM.get_possible_hosts(_vm.Connection.Session, _vm.opaque_ref);
+                    possibleHosts = _vm.Connection.ResolveAll(possibleHostRefs);
+                    Program.Invoke(Program.MainWindow, BuildList);
```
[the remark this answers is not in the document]
michael2012z: Thank you very much for sharing the info.

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -144,12 +125,12 @@ public HostItem(Host host) :base(false)
 
         private class UsbItem : CustomTreeNode
         {
-            public PUSB _pusb { get; }
+            public PUSB Pusb { get; }
```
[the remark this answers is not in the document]
michael2012z: A member with both set and get looks nice. But why is a private set method important? What if the private set is missing?
Thanks.

XenAdmin/SettingsPanels/USBEditPage.cs
```diff
@@ -100,19 +109,28 @@ public AsyncAction SaveSettings()
...
         {
-            Trace.Assert(clone is VM);
+            Trace.Assert(orig is VM);
 
-            _vm = (VM)clone;
+            _vm = (VM)orig;
             if (Connection == null)
                 Connection = _vm.Connection;
 
             if (_vm != null)
```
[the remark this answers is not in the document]
michael2012z: Thanks.

XenAdmin/SettingsPanels/USBEditPage.cs
```diff
@@ -200,7 +223,32 @@ private void buttonAttach_Click(object sender, System.EventArgs e)
 
         private void buttonDetach_Click(object sender, System.EventArgs e)
         {
-
+            if ((selectedRow != null) && (_vm != null))
+            {
+                bool confirmed = false;
+                Program.Invoke(Program.MainWindow, delegate ()
```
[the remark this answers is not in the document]
michael2012z: Thanks.

XenModel/Messages.resx
michael2012z: Thanks.

XenModel/Messages.resx
michael2012z: Accepted.

XenModel/Messages.resx
michael2012z: Thanks.

XenAdmin/Dialogs/PropertiesDialog.cs
michael2012z: Good idea.

XenModel/Actions/USB/SetUsbPassthroughAction.cs
michael2012z: Thanks. It is absolutely an error. :)

XenModel/Actions/USB/DeleteVUSBAction.cs
```diff
@@ -53,15 +53,16 @@ protected override void Run()
             try
             {
                 VUSB.async_unplug(Session, _vusb.opaque_ref);
```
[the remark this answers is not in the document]
michael2012z: Thanks. I checked the implementation of DetachVirtualDiskAction for reference.

XenAdmin/SettingsPanels/USBEditPage.cs
```diff
@@ -100,19 +109,28 @@ public AsyncAction SaveSettings()
...
             if (_vm != null)
             {
                 _vm.PropertyChanged += Vm_PropertyChanged;
-                _vm.Connection.Cache.RegisterBatchCollectionChanged<VUSB>(UsbCollectionChanged);
             }
 
             BuildList();
+
+            // Check if HA was enabled on pool and Restart priority was set on VM.
+            VM.HA_Restart_Priority SelectedPriority = _vm.HARestartPriority();
```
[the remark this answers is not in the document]
michael2012z: Thanks.
On SelectedIndexChanged event of PropertiesDialog, I should update HA selected priority and call ShowHideWarnings() like GpuEditPage.

XenAdmin/SettingsPanels/USBEditPage.cs
```diff
@@ -100,19 +109,28 @@ public AsyncAction SaveSettings()
...
+            Pool pool = Helpers.GetPool(_vm.Connection);
+            bool haEnabled = (pool != null &&
+                pool.ha_enabled &&
+                VM.HaPriorityIsRestart(_vm.Connection, SelectedPriority));
+            labelHAWarning.Text = Messages.ALERT_USB_NOT_ATTACHABLE_HA;
```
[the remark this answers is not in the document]
michael2012z: All handled.

XenAdmin/Dialogs/AttachUsbDialog.cs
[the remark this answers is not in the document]
MihaelaStoica: Public setter is fine as well, but in this particular case we only set the property from inside the class, so private makes more sense.

XenAdmin/MainWindow.cs
```diff
@@ -1448,8 +1448,8 @@ public void UpdateToolbarsCore()
             ShowTab(TabPageDockerDetails, !multi && !SearchMode && isDockerContainerSelected);
 
             bool isPoolOrLiveStandaloneHost = isPoolSelected || (isHostSelected && isHostLive && selectionPool == null);
-            ShowTab(TabPageUSB, !multi && !SearchMode && ((isHostSelected && isHostLive) || isPoolOrLiveStandaloneHost) && !Helpers.FeatureForbidden(selectionConnection, Host.RestrictUsbPassthrough));
             ShowTab(TabPageGPU, !multi && !SearchMode && ((isHostSelected && isHostLive) || isPoolOrLiveStandaloneHost) && Helpers.ClearwaterSp1OrGreater(selectionConnection) && !Helpers.FeatureForbidden(selectionConnection, Host.RestrictGpu));
+            ShowTab(TabPageUSB, !multi && !SearchMode && ((isHostSelected && isHostLive) || isPoolOrLiveStandaloneHost) && !Helpers.FeatureForbidden(selectionConnection, Host.RestrictUsbPassthrough));
```
MihaelaStoica: I know that this code hasn't changed in this PR, but I had just noticed that we add the tab for pools as well, which shouldn't be the case. So we should remove `isPoolOrLiveStandaloneHost` from here

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -75,6 +102,20 @@ private void BuildList()
...
+        {
+            buttonAttach.Enabled = treeUsbList.SelectedItem is UsbItem;
+        }
+
+        private void buttonAttach_Click(object sender, EventArgs e)
+        {
+            UsbItem item = treeUsbList.SelectedItem as UsbItem;
+            if (item != null)
+                new XenAdmin.Actions.CreateVUSBAction(item.Pusb, _vm).RunAsync();
+            this.DialogResult = System.Windows.Forms.DialogResult.OK;
```
MihaelaStoica: Please remove these two lines, as not needed.

XenAdmin/SettingsPanels/USBEditPage.cs
```diff
@@ -200,7 +223,29 @@ private void buttonAttach_Click(object sender, System.EventArgs e)
...
+                {
+                    if (dlg.ShowDialog(Program.MainWindow) == DialogResult.Yes)
+                    {
+                        confirmed = true;
+                    }
+                }
+                if (confirmed)
+                {
+                    // Run this stuff off the event thread, since it involves a server call
+                    System.Threading.ThreadPool.QueueUserWorkItem((System.Threading.WaitCallback)delegate (object o)
```
MihaelaStoica: Teh extra thread is not needed here, as the action runs on a separate thread already.

XenModel/Actions/USB/DeleteVUSBAction.cs
```diff
@@ -52,15 +52,20 @@ protected override void Run()
...
-                VUSB.async_unplug(Session, _vusb.opaque_ref);
-                PollToCompletion(0, 50);
+                if ((_vusb.Connection.Resolve(_vusb.attached) != null) &&
+                    XenAPI.VUSB.get_allowed_operations(Session, _vusb.opaque_ref).Contains(XenAPI.vusb_operations.unplug))
+                {
+                    VUSB.async_unplug(Session, _vusb.opaque_ref);
+                    PollToCompletion(0, 50);
+                }
             }
             catch
```
MihaelaStoica: The `catch` block is redundant here, so it can be removed

XenModel/Actions/USB/DeleteVUSBAction.cs
```diff
@@ -52,15 +52,20 @@ protected override void Run()
         {
             try
             {
-                VUSB.async_unplug(Session, _vusb.opaque_ref);
-                PollToCompletion(0, 50);
+                if ((_vusb.Connection.Resolve(_vusb.attached) != null) &&
+                    XenAPI.VUSB.get_allowed_operations(Session, _vusb.opaque_ref).Contains(XenAPI.vusb_operations.unplug))
+                {
+                    VUSB.async_unplug(Session, _vusb.opaque_ref);
```
MihaelaStoica: PollToCompletion relies on the action's `RelatedTask` property being assigned, so it can poll that task. 
So we need to assign it like `RelatedTask = VUSB.async_unplug(Session, _vusb.opaque_ref);`

Same for the async_destroy call below.

XenAdmin/MainWindow.cs
michael2012z: Indeed, USB is not available for pool. I checked design page.

XenModel/Actions/USB/DeleteVUSBAction.cs
michael2012z: Thank you, I saw DetachVirtualDiskAction has similar usage.

PR #1790, opened 2017-09-28
CA-267553: Fix offline XC showing up to date bug

jijiang:
The bug is that when XC cannot connect to Internet, By Server Update
page shows all the hosts up to date.

### DIFF
--- a/XenAdmin/TabPages/ManageUpdatesPage.cs
+++ b/XenAdmin/TabPages/ManageUpdatesPage.cs
@@ -67,6 +67,7 @@ namespace XenAdmin.TabPages
         private List<string> collapsedPoolRowsList = new List<string>();
         private int checksQueue;
         private bool CheckForUpdatesInProgress;
+        private bool CheckForUpdateSucceeded;
         private readonly CollectionChangeEventHandler m_updateCollectionChangedWithInvoke;
 
         public ManageUpdatesPage()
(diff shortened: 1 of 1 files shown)


GaborApatiNagy:
It looks good to me, but I'm wondering whether we should store the IsFull flags in each row? It is always the same for all rows in the grid.
By the way, isFull is not quite meaningful to me, this name could be misleading in the future.

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -383,16 +407,26 @@ private void UpdateDetails()
...
+                            if (IsFullRow)
+                            {
+                                _patchingStatusCell.Value = outOfDate
+                                    ? Properties.Resources._000_error_h32bit_16
+                                    : Properties.Resources._000_Tick_h32bit_16;
+                                _statusCell.Value = outOfDate ? Messages.NOT_UPDATED : Messages.UPDATED;
+                            }
+                            else
+                            {
+                                _statusCell.Value = "";
```
GaborApatiNagy: It's nicer to have string.Empty instead of ""

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -1361,9 +1401,11 @@ private string GetHostUpdateDetailsCsvQuotes(IXenConnection xenConnection, Host
             Program.Invoke(Program.MainWindow, delegate
             {
                 pool = hasPool ? Helpers.GetPool(xenConnection).Name() : String.Empty;
-                requiredUpdates = RequiredUpdatesForHost(host);
+                requiredUpdates = CheckForUpdateSucceeded ? RequiredUpdatesForHost(host) : String.Empty;
```
MihaelaStoica: I am not sure that the invoking on the UI thread is needed on line 1401

MihaelaStoica:
This looks good now. Merging.

PR #1802, opened 2017-10-05
Update XC XenAPI binding based on XS master branch

jijiang:


jijiang:
We need only FriendlyErrorName.resx file change before XS/CBT branch merging into XS/master, so close the PR, and include the change in the related PR.

PR #1812, opened 2017-10-12
CP-25050: Modify XenCenter to add CPU group config

jishengx:
Just modify the GUI of XenCenter to allow use to configure CPU group. The logics hasn't been added to XenCenter because the XAPI interface hasn't been ready.

### DIFF
--- a/XenAdmin/Commands/ManageCPUGroupCommand.cs
+++ b/XenAdmin/Commands/ManageCPUGroupCommand.cs
@@ -1,0 +1,97 @@
+﻿/* Copyright (c) Citrix Systems, Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer. 
+ * *   Redistributions in binary form must reproduce the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer in the documentation and/or other 
+ *     materials provided with the distribution. 
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+ * CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR 
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, 
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+ * SUCH DAMAGE.
+ */
+
+using System.Collections.Generic;
+using System.Windows.Forms;
+using XenAdmin.Network;
+using XenAPI;
+using XenAdmin.Core;
+using XenAdmin.Dialogs;
+
+namespace XenAdmin.Commands
+{
+    /// <summary>
+    /// Launches the Manage CPU group dialog.
+    /// </summary>
+    internal class ManageCPUGroupCommand : Command
+    {
+        /// <summary>
+        /// Initializes a new instance of this Command. The parameter-less constructor is required if 
+        /// this Command is to be attached to a ToolStrip menu item or button. It should not be used in any other scenario.
+        /// </summary>
+        public ManageCPUGroupCommand()
+        {
+        }
+
+        public ManageCPUGroupCommand(IMainWindow mainWindow, IEnumerable<SelectedItem> selection)
+            : base(mainWindow, selection)
+        {
+        }
+
+        public ManageCPUGroupCommand(IMainWindow mainWindow, IXenConnection connection)
+            : base(mainWindow, Helpers.GetPoolOfOne(connection))
+        {
+        }
+
+        public ManageCPUGroupCommand(IMainWindow mainWindow, Pool pool)
+            : base(mainWindow, pool)
+        {
+        }
+
+        public ManageCPUGroupCommand(IMainWindow mainWindow, Host host)
+            : base(mainWindow, host)
+        {
+        }
+
+        protected override bool CanExecuteCore(SelectedItemCollection selection)
+        {
+            // this check ensures there's no cross-pool
+            return (selection.FirstAsXenObject != null && selection.FirstAsXenObject.Connection != null && selection.FirstAsXenObject.Connection.IsConnected &&
+                        (selection.PoolAncestor != null || selection.HostAncestor != null)) ;
+        }
+
+        protected override void ExecuteCore(SelectedItemCollection selection)
+        {
+            var pool = Helpers.GetPoolOfOne(selection.FirstAsXenObject.Connection);
+            if (pool != null)
+            {
+                this.MainWindowCommandInterface.ShowPerConnectionWizard(pool.Connection, new ManageCPUGroupDialog(pool));
+            }
+        }
+
+        private void ShowUpsellDialog(IWin32Window parent)
+        {
+            using (var dlg = new UpsellDialog(HiddenFeatures.LinkLabelHidden ? Messages.UPSELL_BLURB_HA : Messages.UPSELL_BLURB_HA + Messages.UPSELL_BLURB_HA_MORE,
+                                    InvisibleMessages.UPSELL_LEARNMOREURL_HA))
+                dlg.ShowDialog(Parent);
+        }
+    }
+}
--- a/XenAdmin/Dialogs/ManageCPUGroupDialog.Designer.cs
+++ b/XenAdmin/Dialogs/ManageCPUGroupDialog.Designer.cs
@@ -1,0 +1,299 @@
+﻿using XenAdmin.Commands;
+
+namespace XenAdmin.Dialogs
+{
+    partial class ManageCPUGroupDialog
+    {
+        /// <summary>
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary>
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Windows Form Designer generated code
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle1 = new System.Windows.Forms.DataGridViewCellStyle();
+            this.panel1 = new System.Windows.Forms.Panel();
+            this.toolStrip1 = new System.Windows.Forms.ToolStrip();
+            this.toolStripButtonNew = new System.Windows.Forms.ToolStripButton();
+            this.toolStripButtonDelete = new System.Windows.Forms.ToolStripButton();
+            this.toolStripButtonEdit = new System.Windows.Forms.ToolStripButton();
+            this.tableLayoutPanel1 = new System.Windows.Forms.TableLayoutPanel();
+            this.labelCPUGroupInPool = new System.Windows.Forms.Label();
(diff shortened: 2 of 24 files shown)


XenAdmin/Commands/ManageCPUGroupCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public ManageCPUGroupCommand(IMainWindow mainWindow, Host host)
+            : base(mainWindow, host)
+        {
+        }
+
+        protected override bool CanExecuteCore(SelectedItemCollection selection)
+        {
+            // this check ensures there's no cross-pool and items don't contain VM
+            if (selection.FirstAsXenObject != null && selection.FirstAsXenObject.Connection != null && selection.FirstAsXenObject.Connection.IsConnected &&
```
kc284: When you have a statement of the form
```
if (something)
    return true;
return false;
```
it's simpler to write
```
return something;
```

XenAdmin/Commands/ManageCPUGroupCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+                    return true;
+            return false;
+        }
+
+        protected override void ExecuteCore(SelectedItemCollection selection)
+        {
+            var pool = Helpers.GetPoolOfOne(selection.FirstAsXenObject.Connection);
+            if (pool != null)
+            {
+                ////*************Host.RestrictHA should be replaced with Host.RestrictCPUGroup*************
```
kc284: Please avoid leaving copied-pasted-commented-out code in PRs because it's very easy to forget fixing something afterwards. If you must, it's preferable to use an empty placeholder.

XenAdmin/Commands/ManageCPUGroupCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+            // this check ensures there's no cross-pool and items don't contain VM
+            if (selection.FirstAsXenObject != null && selection.FirstAsXenObject.Connection != null && selection.FirstAsXenObject.Connection.IsConnected &&
+               (selection.PoolAncestor != null || selection.HostAncestor != null) && !itemsContainVM(selection))
+                return true;
+
+            return false;
+        }
+
+
+        private bool itemsContainVM(SelectedItemCollection selection)
```
kc284: No need to implement a new method to check we have a VM. You can use one of the ready predicates in SelectedItemCollection.cs, namely ContainsOneItemOfType<VM>.

XenAdmin/Dialogs/ManageCPUGroupDialog.Designer.cs
```diff
@@ -0,0 +1,300 @@
...
+            // toolStripButtonEddit
+            // 
+            this.toolStripButtonEddit.Font = new System.Drawing.Font("Segoe UI", 9F);
+            this.toolStripButtonEddit.Image = global::XenAdmin.Properties.Resources.edit_16;
+            this.toolStripButtonEddit.ImageTransparentColor = System.Drawing.Color.Magenta;
+            this.toolStripButtonEddit.Name = "toolStripButtonEddit";
+            this.toolStripButtonEddit.Size = new System.Drawing.Size(80, 25);
+            this.toolStripButtonEddit.Text = "&Properties";
+            // 
+            // tableLayoutPanel1
```
kc284: I think tableLayoutPanel1 has 1 row too many, the top row does not seem to be used.
General tip: having extra rows/columns happens very often when they are set to autosize. It may not be a problem in the first instance, but it may make it tricky to add more controls in the panel in the future. One way to avoid this is to set originally all the columns/rows to Percent, add the controls you want, and then set the necessary column/rows to Autosize.

XenAdmin/Dialogs/ManageCPUGroupDialog.Designer.cs
```diff
@@ -0,0 +1,300 @@
...
+            this.ResumeLayout(false);
+
+        }
+
+        #endregion
+        private System.Windows.Forms.Panel panel1;
+        private System.Windows.Forms.ToolStrip toolStrip1;
+        private System.Windows.Forms.ToolStripButton toolStripButtonNew;
+        private System.Windows.Forms.ToolStripButton toolStripButtonDelete;
+        private System.Windows.Forms.ToolStripButton toolStripButtonEddit;
```
kc284: "toolStripButtonEdit"
Typos in variable names don't cause problems to the UI or the compiler, but they may make trouble for a developer searching the solution by text.

XenAdmin/MainWindow.resx
```diff
@@ -2007,6 +2007,15 @@
   <data name="pluginItemsPlaceHolderToolStripMenuItem2.Text" xml:space="preserve">
     <value>PluginItemsPlaceHolder</value>
   </data>
+  <data name="manageCPUGroupToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
+    <value>277, 22</value>
+  </data>
+  <data name="manageCPUGroupToolStripMenuItem.Text" xml:space="preserve">
+    <value>Manage CPU Group...</value>
```
kc284: I believe this needs a hotkey.

XenAdmin/SettingsPanels/CPUMemoryEditPage.resx
```diff
@@ -729,6 +777,159 @@
...
+    <value>3, 6, 3, 6</value>
+  </data>
+  <data name="linkLabelCPUGroup.Size" type="System.Drawing.Size, System.Drawing">
+    <value>139, 13</value>
+  </data>
+  <data name="linkLabelCPUGroup.TabIndex" type="System.Int32, mscorlib">
+    <value>1</value>
+  </data>
+  <data name="linkLabelCPUGroup.Text" xml:space="preserve">
+    <value>Configure CPU group now...</value>
```
kc284: "now" is not that necessary in this context.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.Designer.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        protected System.Windows.Forms.TableLayoutPanel tableLayoutPanel1;
+        private System.Windows.Forms.Label label2;
+        private Controls.SearchTextBox searchTextBox1;
+        private System.Windows.Forms.DataGridView dataGridView1;
+        private System.Windows.Forms.Label label1;
+        private System.Windows.Forms.Label labelCounterVMs;
+        private System.Windows.Forms.Button buttonSelectAll;
+        private System.Windows.Forms.Button buttonClearAll;
+        private System.Windows.Forms.DataGridViewTextBoxColumn CheckBoxColumn;
```
kc284: From what I see on the design, we also need a column for the expand/collapse icon. Now the design is not finalised yet: do we want the headers visible or hidden?

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.resx
```diff
@@ -0,0 +1,456 @@
...
+    <value>1, 6, 1, 6</value>
+  </data>
+  <data name="label1.Size" type="System.Drawing.Size, System.Drawing">
+    <value>478, 25</value>
+  </data>
+  <data name="label1.TabIndex" type="System.Int32, mscorlib">
+    <value>0</value>
+  </data>
+  <data name="label1.Text" xml:space="preserve">
+    <value>Select a checkbox to add a CPU cores to the group; clear its checkbox to remove it from the group.</value>
```
kc284: "a CPU core" in singular.

kc284:
I saw some mixed instances of "CPU group" and "CPU Group". I think the word "group" should be with a lowercase g in all labels, except when it is a dialog title or menu item.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.Designer.cs
```diff
@@ -0,0 +1,175 @@
...
+            this.tableLayoutPanel1.ResumeLayout(false);
+            this.tableLayoutPanel1.PerformLayout();
+            ((System.ComponentModel.ISupportInitialize)(this.dataGridView1)).EndInit();
+            this.ResumeLayout(false);
+
+        }
+
+        #endregion
+
+        protected System.Windows.Forms.TableLayoutPanel tableLayoutPanel1;
```
kc284: The column containing the search box should be autosized (although I admit this is not consistent throughout XenCenter).

XenAdmin/Commands/ManageCPUGroupCommand.cs
jishengx: Thanks, I will simplify it.

XenAdmin/Commands/ManageCPUGroupCommand.cs
jishengx: Ok, I have removed it, thanks.

XenAdmin/Commands/ManageCPUGroupCommand.cs
jishengx: In my opinion, the method named ContainsOneItemOfType predicates only one item of specified type. However, I don't know how the method works when user choose more than one item contain hosts and VMs.

XenAdmin/Dialogs/ManageCPUGroupDialog.Designer.cs
jishengx: Thanks, I have fixed it.

XenAdmin/MainWindow.resx
jishengx: Thanks, I change it into "Manage CPU &groups...". Is it ok?

XenAdmin/SettingsPanels/CPUMemoryEditPage.resx
jishengx: Ok, fixed it.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.resx
jishengx: Sorry for that, fixed it.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.Designer.cs
jishengx: I agree with you, thanks.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.Designer.cs
jishengx: According to other dialogs contain the expand/collapse icon(e.g. Attach Disk Dialog), they don't add  a column specially for the icon. So, I am wondering if it is necessary.

XenAdmin/Dialogs/ManageCPUGroupDialog.Designer.cs
jishengx: Thanks, I am fresh to XenCenter, so there may be some mistakes. Thank you very much.

XenAdmin/Commands/ManageCPUGroupCommand.cs
jishengx: In design of REQ-155,  the GUI won't all user to use the "Manage CPU groups..." button if user select a VM. There is a case that user select more than one item contain VM. In that case, we shouldn't allow user to use the button.

kc284: Ahh indeed, you're right, ContainsOneItemOfType is only for 1 item, my bad! In this case it would be better to maker a new reusable function in that class, AtLeastOneXenObjectIs.
However, Mihaela is right, checking the pool or host ancestor should be enough, and it's not wrong if the item is enabled when a VM in the pool is selected. There are other pool or host items on the main menu that are enabled if a VM with that pool or host ancestor is selected.
One more thing I forgot to mention yesterday is we should probably add the item in the Pool's right click menu. For this you'll need to add the item in the SinglePool class in xenadmin\XenAdmin\Commands\Controls\ContextMenuBuilder.cs

XenAdmin/MainWindow.resx
kc284: Yes, I don't see any other item in this menu using the G, so it's ok. G should be capital though, because in this case it is a menu item.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.Designer.cs
kc284: The tree on the Attach disk dialog is a different control, it is a CustomTreeView <=FlickerFreeListBox <=ListBox. Here you have used a DataGridView, in which case you'll need a separate column. Have a look, for example, on the RollingPoolUpgrade wizard or the ManageUpdates Page to see how these columns work.

XenAdmin/MainWindow.resx
jishengx: Ok, I will change it. Thanks.

XenAdmin/Wizards/NewCPUGroupWizard/NewCPUGroupCPUCoresPage.Designer.cs
jishengx: Ok, I have changed it. I'm sorry to say that I'm both a fresh to XenCenter and C#.

XenAdmin/Commands/ManageCPUGroupCommand.cs
jishengx: Thank you very much. I will remove it.

XenAdmin/MainWindow.Designer.cs
```diff
@@ -166,6 +166,8 @@ private void InitializeComponent()
             this.deleteToolStripMenuItem = new XenAdmin.Commands.CommandToolStripMenuItem();
             this.toolStripSeparator26 = new System.Windows.Forms.ToolStripSeparator();
             this.pluginItemsPlaceHolderToolStripMenuItem2 = new System.Windows.Forms.ToolStripMenuItem();
+            this.manageCPUGroupToolStripMenuItem = new XenAdmin.Commands.CommandToolStripMenuItem();
```
kc284: I'd rather have this under the Manage vApps button, in the same group as the latter.

kc284:
I will close this PR because there is no point in having it open since work on it has stopped and the design has not been finalised yet. Once we have resumed work and we have a clear idea on the design, we can make a new PR.

PR #1815, opened 2017-10-13
CA-268975: Do not allow multiple hosts to join the pool if the new po…

MihaelaStoica:
…ol size will exceed the max size


jijiang:
Good!

PR #1817, opened 2017-10-13
CA-267330: [Update Wizard Improvements] The Update wizard: block the installation of an update on slaves if a reboot or a toolstack restart is pending on master

GaborApatiNagy:


XenAdmin/Diagnostics/Checks/RebootPendingOnMasterCheck.cs
```diff
@@ -0,0 +1,127 @@
...
+                        return new MasterIsPendingRestartHostProblem(this, pool);
+                    }
+                }
+            }
+
+            //check toolstack restart
+            var toolstackRestartRequiredPatches = Host.AppliedPatches().Where(p => p.after_apply_guidance.Contains(after_apply_guidance.restartXAPI) && ((double)Util.ToUnixTime(p.AppliedOn(Host)) > agentStart));
+            foreach (Pool_patch patch in toolstackRestartRequiredPatches)
+            {
+                if (string.IsNullOrEmpty(UpdateUuid)) //automated mode
```
jijiang: Actually, this check will never happen, because it returns already. I don't know if we could put the following code on line 60 to avoid repeated checks:
    if (string.IsNullOrEmpty(UpdateUuid))
        return new MasterIsPendingRestartHostProblem(this, pool);

XenAdmin/Diagnostics/Checks/RebootPendingOnMasterCheck.cs
```diff
@@ -0,0 +1,127 @@
...
+                return new HostNotLiveWarning(this, Host);
+            
+            var elyOrGreater = Helpers.ElyOrGreater(Host);
+
+            double bootTime = Host.BootTime();
+            double agentStart = Host.AgentStartTime();
+
+            if (bootTime == 0.0 || agentStart == 0.0)
+                return null; //fine
+
```
jijiang: I am not sure if we need bootTime and agentStart check for ElyOrGreater host. If not, maybe we could move line 62 to 66 to else block?

XenAdmin/Diagnostics/Checks/RebootPendingOnMasterCheck.cs
GaborApatiNagy: This is there for a reason, we check host-restart requirements first, and only then the toolstack-restart ones. I think we should keep it as is.

XenAdmin/Diagnostics/Checks/RebootPendingOnMasterCheck.cs
GaborApatiNagy: Good finding, I can move it into the pre-Ely blocks so it won't run for Ely+ hosts. Good finding 👍

jijiang:
Looks good! Merge it.

PR #1820, opened 2017-10-17
CP-25081: Set platform parameters.

michael2012z:
Make sure platform dictionary contain pair ["device-model":"qemu-upstream-compat"] before creating VUSB.


XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -72,6 +74,11 @@ private void BuildList()
         {
             Program.AssertOnEventThread();
 
+            labelWarningLine3.Visible = 
+                (platform == null) ||
```
MihaelaStoica: We can put this in a function in the VM class and reuse it in CreateVUSBAction.

XenAdmin/Dialogs/AttachUsbDialog.cs
```diff
@@ -42,11 +42,13 @@ public partial class AttachUsbDialog : XenDialogBase
     {
         private VM _vm;
         private List<Host> possibleHosts;
+        private Dictionary<string, string> platform;
 
         public AttachUsbDialog(VM vm): base(vm.Connection)
         {
             _vm = vm;
             possibleHosts = new List<Host>();
+            platform = _vm.platform;
```
MihaelaStoica: No need to save this, we can use _vm.platform directly where is needed.

XenModel/Actions/USB/CreateVUSBAction.cs
```diff
@@ -41,19 +41,28 @@ namespace XenAdmin.Actions
...
+        private Dictionary<string, string> _platform;
 
-        public CreateVUSBAction(PUSB pusb, VM vm) : 
+        public CreateVUSBAction(PUSB pusb, VM vm, Dictionary<string, string> platform) : 
             base(pusb.Connection, String.Format(Messages.ACTION_VUSB_CREATING,  pusb.Name(), vm.Name()))
         {
             _pusb = pusb;
-            _vm = vm;
+            VM = vm;
+            _platform = platform == null ?
```
MihaelaStoica: We don't need to pass the platform as a parameter to the action, because we have the VM object and can simply check the value of `VM.platform` directly where needed (so we would have `platform` as a local variable in `Run()` and create it in there from `VM.platform`)

XenAdmin/Dialogs/AttachUsbDialog.cs
michael2012z: Good idea!

huizh:
Hi @MihaelaStoica, since we depends on commit https://github.com/xenserver/xenadmin/pull/1820/commits/450dde2c420e392b8d7e499e62e7d99327242e6b to implement XenCenter/iCenter test, for this PR, if there's no big concern, is it possible we merge this PR first (or just that commit if possible)?

If there's remaining minor problems, we could raise another PR to fix.

Does it make sense?

Thanks.

MihaelaStoica:
Looks good now. Merging.

huizh:
Thank you @MihaelaStoica.

PR #1823, opened 2017-10-19
CA-265405: Fix string bug in vmss history dialog

jijiang:

### DIFF
--- a/XenModel/Alerts/PolicyAlert.cs
+++ b/XenModel/Alerts/PolicyAlert.cs
@@ -52,7 +52,7 @@ namespace XenAdmin.Alerts
 
             if(Type == "info")
             {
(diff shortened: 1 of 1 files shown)


XenModel/Alerts/PolicyAlert.cs
```diff
@@ -52,7 +52,7 @@ public PolicyAlert(long priority, string name, DateTime time, string body, strin
 
             if(Type == "info")
             {
-                Text = Message.FriendlyBody(name);
+                Text = string.Format(Message.FriendlyBody(name), policyName);
```
kc284: This one is not really needed because of all the vmss messages we have only one error message is a string format, but it does not cause any error if we leave it.

PR #1816, opened 2017-10-13
Translation Updates

vetrivelc:

PR #1827, opened 2017-10-23
L10N:Translation Updates

vetrivelc:


vetrivelc:
Alternate to https://github.com/xenserver/xenadmin/pull/1816

PR #1843, opened 2017-11-06
[CA-269110] Stop calling assert_can_boot_here after migrate dropdown is closed

mcintyre94:
When we right click migrate a VM we load all the hosts in the pool into a dropdown, and then one-by-one assert whether the selected VM can boot on that host, enabling each option where this is positive. If the dropdown is exited then nothing stops these assert calls from continuing, and looking at the logs we can see that they do.

With a large pool this is problematic because the UI basically locks up until it's completed all these asserts, even if we've already selected a host to migrate to. If we try to migrate another host then we see in the logs that both the old thread and new thread are making `assert_can_boot_here` calls, and the UI for the new thread doesn't function until the old one has finished.

This commit adds a new property `_isDropDownClosed`, which is set by the new `OnDropDownClosed` event handler. In the loop enabling dropdown items, each iteration we check this property - and if it is true then the dropdown is no longer showing so we stop the loop and thus don't call any more `assert_can_boot_here` commands. This stops the UI locking up after the dropdown is closed.

Signed-off-by: Callum McIntyre <callum.mcintyre@citrix.com>

### DIFF
--- a/XenAdmin/Commands/Controls/VMOperationToolStripMenuItem.cs
+++ b/XenAdmin/Commands/Controls/VMOperationToolStripMenuItem.cs
@@ -73,9 +73,18 @@ namespace XenAdmin.Commands
             base.DropDownItems.Add(new ToolStripMenuItem());
         }
 
+        private bool _isDropDownClosed;
+
+        protected override void OnDropDownClosed(EventArgs e)
+        {
+            base.OnDropDownClosed(e);
+            _isDropDownClosed = true;
+        }
+
         protected override void OnDropDownOpening(EventArgs e)
         {
             base.DropDownItems.Clear();
+            _isDropDownClosed = false;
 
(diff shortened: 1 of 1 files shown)


mcintyre94:
I may well have missed something in the WLB case, but as far as I can tell it doesn't do a similar expensive loop - it already has the WLB recommendations and it basically just assigns them to the hosts and sorts them by that. I'm happy to add a similar change if I've missed it, but I can't see where it would be.

kc284:
Ok, I'll merge this PR; (I'll do some testing for the WLB case and raise a different ticket if necessary).

huizh:
Hi @kc284, could you update with the test result of WLB?

Thanks.

mcintyre94:
@huizh Tina raised CA-273151 regarding WLB

PR #1844, opened 2017-11-07
[CA 272150] Cross-pool copy/move/migrate wizard performance issues

mcintyre94:
- Sort the hosts in the target server dropdown using a natural sort
- Use async load for hosts in the target server dropdown, instead of locking the UI waiting for them to calculate
- Fix behaviour when stepping backward to avoid reloading when we don't need to

### DIFF
--- a/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
+++ b/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
@@ -63,7 +63,8 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard
         public override void PageLoaded(PageLoadedDirection direction)
         {
             base.PageLoaded(direction);
-            PopulateComboBox();
+            if (direction.Equals(PageLoadedDirection.Forward) || ComboBoxEmpty())
+                PopulateComboBox();
         }
 
         public override bool EnableNext()
--- a/XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
+++ b/XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
@@ -473,7 +473,7 @@ namespace XenAdmin.Wizards.GenericPages
 	        else
 	        {
 	            /* If we have a value, and it's enabled, then enable next button
-                 * This is hit when we step back (so have a default value), but at inital load all items are disabled. 
+                 * This is hit when we step back (so have a default value), but at initial load all items are disabled. 
                  * So when they're updated and hit this they don't trigger the normal enable because cb.value != null
                  */
 	            var value = (IEnableableComboBoxItem) cb.Value;
--- a/XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
+++ b/XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
@@ -361,6 +361,27 @@ namespace XenAdmin.Wizards.GenericPages
             }
         }
 
+	    private void HostSelectionComboBox_ReasonChanged(DelayLoadingOptionComboBoxItem item, DataGridViewComboBoxCell comboBox)
+	    {
+            int index = comboBox.Items.IndexOf(item);
+            if (index > -1 && index <= comboBox.Items.Count)
+            {
+                Program.Invoke(Program.MainWindow, delegate
+                {
+                    if (updatingDestinationCombobox || updatingHomeServerList)
+                        return;
+
+                    DelayLoadingOptionComboBoxItem tempItem =
+                        comboBox.Items[index] as DelayLoadingOptionComboBoxItem;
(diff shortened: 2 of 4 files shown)


XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
```diff
@@ -63,7 +63,8 @@ public CrossPoolMigrateDestinationPage(Host preSelectedHost, List<VM> selectedVM
         public override void PageLoaded(PageLoadedDirection direction)
         {
             base.PageLoaded(direction);
-            PopulateComboBox();
+            if (direction.Equals(PageLoadedDirection.Forward) || ComboBoxEmpty())
```
kc284: Can we check wether we can do this for ImportSelectHostPage? I noticed it repopulates the combobox too and it may not be necessary now that you have removed the clearing line from the base class's PageLeave method.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -350,7 +354,8 @@ private void RestoreGridHomeServerSelectionFromMapping()
 
                     List<IEnableableXenObjectComboBoxItem> list =
                         cbCell.Items.OfType<IEnableableXenObjectComboBoxItem>().ToList();
-                    IEnableableXenObjectComboBoxItem item =
+
+                    var item =
```
kc284: I think this could be written in one line without the intermediate list as `var item  = cbCell.Items.OfType<IEnableableXenObjectComboBoxItem>().FirstOrDefault(cbi => MatchingWithXenRefObject(cbi, mapping.XenRef));`

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -446,11 +464,24 @@ private void SetComboBoxPreSelection(DataGridViewEnableableComboBoxCell cb)
...
+	                    SetButtonNextEnabled(true);
+	                }
+	            }
 	            else
 	                SetButtonNextEnabled(false); //do not allow to leave the page if a vm has no target
 	        }
+	        else
+	        {
+	            /* If we have a value, and it's enabled, then enable next button
+                 * This is hit when we step back (so have a default value), but at inital load all items are disabled.
```
kc284: initial

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -401,12 +406,25 @@ private void PopulateDataGridView(List<ReasoningFilter> homeserverFilters)
                             }
                         }
 
+                        var items = new List<DelayLoadingOptionComboBoxItem>();
                         foreach (var host in Connection.Cache.Hosts)
                         {
                             var item = new DelayLoadingOptionComboBoxItem(host, homeserverFilters);
-                            item.LoadAndWait();
+                            item.ReasonUpdated +=
```
kc284: 1) Can we rename this event to ReasonChanged, because its handler is DelayLoadedComboBoxItem_ReasonChanged and the firing method is OnReasonChanged and it's both pointless and confusing to have a different name for the event.
2) Where is this event unregistered?

kc284:
I think we may need to change the way the checks on the destination page are performed. When we first populate the destination pool combobox, we do an assert_can_migrate for each VM for each of the pool hosts. Afterwards, when we populate the datagridview underneath, we perform the checks again. Can we avoid this and populate the gridview with the cached results of the first series of checks? Is there a reason why we need to repeat them? This will probably need an amount of rewriting, but I believe it would make a significant difference.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
mcintyre94: 1. Will do
2. I missed this and it'll need some work to pass through the combobox - I'll need to think about this more carefully.

mcintyre94:
I think I've the changes except:
1. Make the hosts load without hovering over them
2. Refactoring to make the `ReasonChanged` event unsubscribable

Keeping fixup commits separate for now, and marked needs updating.

mcintyre94:
I've added a commit to unsubscribe the `ReasonChanged` event. The remaining issues are:

1) The combobox doesn't repaint when open - this seems to be a difference in the combobox controls and I'm worried it may be in the WinForms level. I might have missed something but I'm not sure how to get this working. 
2) These async `Load()` calls don't cancel when the dropdown is closed, and I can't see a way to make that happen. Unlike the right-click host list this is done async and not cancelable. 

I'm tempted to close this for now and let the ticket go on the backlog - this feels like too much work to commit to right now, and I'm not sure there's anything we can take without the above 2 issues being fixed. Thoughts?

mcintyre94:
Closing this for now. I'll update the ticket with a summary of the problems with this.

PR #1851, opened 2017-11-10
CA-271970: Incorrect error when copying a VM on an unlicensed pool

MihaelaStoica:
Instead of showing the cross-pool migration upsell dialog, we now show the wizard, where the user can try to perform an intra-pool copy (the cross-pool option is disabled with an explanatory message displayed)


kc284:
This looks ok to me.
One could add a condition to make the check on the page only when moving forwards, but I'm not fussed about it because it is not an expensive calculation and I don't see any scalability issues, so we can leave it as is.

PR #1855, opened 2017-11-13
CA-273674 Added warning when shutting down pool master

serencorbett1:

### DIFF
--- a/XenAdmin/Commands/ShutDownHostCommand.cs
+++ b/XenAdmin/Commands/ShutDownHostCommand.cs
@@ -38,6 +38,7 @@ using XenAdmin.Core;
 using XenAdmin.Properties;
 using XenAPI;
 using XenAdmin.Dialogs;
+using System.Text;
 
 
 namespace XenAdmin.Commands
@@ -135,6 +136,7 @@ namespace XenAdmin.Commands
                 List<Host> hosts = GetSelection().AsXenObjects<Host>();
                 bool hasRunningVMs = false;
                 var hciHosts = new List<Host>();
+                var poolMasters = new List<Host>();
 
                 foreach (Host h in hosts)
                 {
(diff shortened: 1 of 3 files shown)


XenAdmin/Commands/ShutDownHostCommand.cs
```diff
@@ -143,21 +145,40 @@ protected override string ConfirmationDialogText
...
-
-                if (hasRunningVMs)
-                    return hosts.Count == 1
-                        ? string.Format(Messages.CONFIRM_SHUTDOWN_SERVER, hosts[0].Name())
-                        : Messages.CONFIRM_SHUTDOWN_SERVERS;
-
-                return hosts.Count == 1
-                    ? string.Format(Messages.CONFIRM_SHUTDOWN_SERVER_NO_VMS, hosts[0].Name())
-                    : Messages.CONFIRM_SHUTDOWN_SERVERS_NO_VMS;
+                    sb = hciHosts.Count == 1
```
GaborApatiNagy: The Conditional/Ternary operator returns one of two expressions depending on the condition; in this case, the returned value would always be the same 'sb' object whereas the side-effect would be what we were originally aiming for: to append text to the StringBuilder.

Note that there is no need to return the 'sb', especially to assign it again. Therefore I suggest to have the ternary operator removed from here and have an if-else statement implemented instead.

XenAdmin/Commands/ShutDownHostCommand.cs
```diff
@@ -143,21 +145,40 @@ protected override string ConfirmationDialogText
...
+                        ? sb.AppendFormat(Messages.CONFIRM_SHUTDOWN_HCI_WARN_SERVER, hciHosts[0].Name())
+                        : sb.AppendFormat(Messages.CONFIRM_SHUTDOWN_HCI_WARN_SERVERS, string.Join("\n", hciHosts.Select(h => h.Name())));
+
+                else if (hasRunningVMs)
+                    sb = hosts.Count == 1
```
GaborApatiNagy: see comment on line 156

XenAdmin/Commands/ShutDownHostCommand.cs
```diff
@@ -143,21 +145,40 @@ protected override string ConfirmationDialogText
...
+                        ? sb.AppendFormat(Messages.CONFIRM_SHUTDOWN_SERVER, hosts[0].Name())
+                        : sb.Append(Messages.CONFIRM_SHUTDOWN_SERVERS);
+
+                else sb = hosts.Count == 1
```
GaborApatiNagy: see comment on line 156

XenAdmin/Commands/ShutDownHostCommand.cs
MihaelaStoica: Or we can assign the message to a string variable (and continue to use the ternary operator, avoiding lots of ifs) and append it to the StringBuilder further down, before line 169.

XenAdmin/Commands/ShutDownHostCommand.cs
```diff
@@ -143,21 +145,44 @@ protected override string ConfirmationDialogText
...
-                if (hasRunningVMs)
-                    return hosts.Count == 1
-                        ? string.Format(Messages.CONFIRM_SHUTDOWN_SERVER, hosts[0].Name())
-                        : Messages.CONFIRM_SHUTDOWN_SERVERS;
-
-                return hosts.Count == 1
-                    ? string.Format(Messages.CONFIRM_SHUTDOWN_SERVER_NO_VMS, hosts[0].Name())
-                    : Messages.CONFIRM_SHUTDOWN_SERVERS_NO_VMS;
+                    firstWarning = hciHosts.Count == 1
+                        ? firstWarning = string.Format(Messages.CONFIRM_SHUTDOWN_HCI_WARN_SERVER, hciHosts[0].Name())
```
GaborApatiNagy: "firstWarning = " assignments in line 159 and 160 are not needed the value is assigned in line 158

XenAdmin/Commands/ShutDownHostCommand.cs
```diff
@@ -143,21 +145,44 @@ protected override string ConfirmationDialogText
...
+                        : firstWarning = string.Format(Messages.CONFIRM_SHUTDOWN_HCI_WARN_SERVERS, string.Join("\n", hciHosts.Select(h => h.Name())));
+
+                else if (hasRunningVMs)
+                    firstWarning = hosts.Count == 1
+                        ? firstWarning = string.Format(Messages.CONFIRM_SHUTDOWN_SERVER, hosts[0].Name())
```
GaborApatiNagy: same comment as for line 159 and 160

XenAdmin/Commands/ShutDownHostCommand.cs
```diff
@@ -143,21 +145,44 @@ protected override string ConfirmationDialogText
...
+                        : firstWarning = Messages.CONFIRM_SHUTDOWN_SERVERS;
+
+                else firstWarning = hosts.Count == 1
+                    ? firstWarning = string.Format(Messages.CONFIRM_SHUTDOWN_SERVER_NO_VMS, hosts[0].Name())
```
GaborApatiNagy: same comment as for line 159 and 160

PR #1856, opened 2017-11-13
 CA-271775 and Revert CA-247333

GaborApatiNagy:
* It has been seen that the fallback logic should have been kept: reverted CA-247333.
* Showing info message if this happens: CA-271775.


MihaelaStoica:
Looks good to me. Merging

PR #1870, opened 2017-11-23
Some Enhancements to input stream handling in xva_verify

auron-muc:
Changes:
- Open input file in read mode to allow readonly data
- Allow reading from stdin with filename "-"
- test for compressed data, use decompression if needed.

Needed because we use compress=true in vm-export and have the images on a read-only share.


kc284:
I've tested this and works ok for me. Could you please expand the usage lines at the top of the Main method to describe the new additions. Thanks.

kc284:
This PR has been stale for quite a while now; I will merge this and we can fix the usage lines later.

PR #1871, opened 2017-11-24
CA-266891 and some more corrections.

kc284:
Note that this PR contains a change in DOTNET_BUILD_LOCATION which depends on https://github.com/xenserver/dotnet-packages/pull/64


MihaelaStoica:
Looks good. Merging.

PR #1872, opened 2017-11-27
CA-272606: AD user can not log out when closing XenCenter.

michael2012z:
This PR is created to fix CA-272606.
When user close XenCenter main window, XenCenter go through all existing connections and log out the sessions. As the log-out operation involves server communication, XenCenter start a thread to do that for each session. These threads are assigned "Background" type and low priority. When all the foreground threads exit, all background thread will be killed. So race condition exist, if the foreground threads quit quickly, log-out operations haven't been done.
The solution is to set the logging-out thread as foreground in the case of exiting.


MihaelaStoica:
Looks good. Merging.

PR #1882, opened 2017-12-04
CA-272769: Enable user to cancel URL testing in rolling upgrade wizard.

michael2012z:
Changes brief:
- Once the "Test" button is clicked, the button text is changed to "Stop" instead of being disabled. 
- User can click "Stop" button again to cancel the test.
- The test is also canceled when user type "Previous" and "Cancel" button.
- As the text may change in run, so the text was moved into Messages resource file.
- The "cancel" function also applies for HTTP and FTP.


MihaelaStoica:
Looks good to me. Merging.

PR #1884, opened 2017-12-04
CP-25563

serencorbett1:


XenAdmin/Dialogs/NetworkingPropertiesPage.cs
```diff
@@ -166,15 +167,18 @@ public void RefreshButtons()
...
+                    
                 var clusteringEnabled = network.Connection.Cache.Cluster_hosts.Any(cluster =>
-                    cluster.host.opaque_ref == master.opaque_ref && cluster.enabled);
-                var existingCluster = network.Connection.Cache.Clusters.FirstOrDefault();
-
-                if (clusteringEnabled && existingCluster != null &&
-                    existingCluster.network.opaque_ref == network.opaque_ref)
+                    cluster.host.opaque_ref == host.opaque_ref && cluster.enabled);
+                    
+                if (clusteringEnabled && host.enabled)
```
kc284: It is normally a good idea to check the host for null in case it hasn't resolved.

XenAdmin/Dialogs/NetworkingPropertiesPage.cs
```diff
@@ -166,15 +167,18 @@ public void RefreshButtons()
                 haEnabledRubric.Enabled =
                     true;
             }
-            if (network != null)
+
+            var pif = (PIF) Tag;
```
kc284: Generally it's best to avoid implicit casting and do the following: `var pif = Tag as PIF;` The difference is that this one will not crash if the Tag were not to be a PIF.

XenAdmin/Dialogs/NetworkingPropertiesPage.cs
```diff
@@ -165,6 +167,23 @@ public void RefreshButtons()
...
+            if (pif != null && existingCluster != null && existingCluster.network.opaque_ref == network.opaque_ref)
+            {
+                Host host = network.Connection.Resolve(pif.host);
+                    
+                var clusteringEnabled = network.Connection.Cache.Cluster_hosts.Any(cluster =>
+                    cluster.host.opaque_ref == host.opaque_ref && cluster.enabled);
+                    
+                if (clusteringEnabled && host.enabled)
+                {
+                    DisableControls("DisableForClustering",
```
kc284: Having hard coded strings like this in the code is not a good idea, because if one of these instances changes, things wil break without warning. I would remove all these tags and either use directly the control names, or preferably, disable the parent control on which all of them are residing.

XenAdmin/SettingsPanels/ClusteringEditPage.cs
```diff
@@ -82,7 +82,7 @@ public void SetXenObjects(IXenObject orig, IXenObject clone)
         {
             pool = Helpers.GetPoolOfOne(clone.Connection);
             var master = Helpers.GetMaster(clone.Connection);
-            flowLayoutInfo.Visible = false;
+            tableLayoutInfo.Visible = false;
```
kc284: This panel nees better anchoring, because if the warning text is too long (if not in english, perhaps in l10n) resizing will not happen correctly.
There is also something wrong with the top lable as well (or its parent controls) because resizing the properties dialog causes text truncation.

XenAdmin/Dialogs/NetworkingPropertiesPage.cs
```diff
@@ -178,10 +178,9 @@ public void RefreshButtons()
                 var clusteringEnabled = network.Connection.Cache.Cluster_hosts.Any(cluster =>
                     cluster.host.opaque_ref == host.opaque_ref && cluster.enabled);
```
kc284: This `host.opaqueref` here will also complain if the host has not resolved. I think it would be better to write `pif.host.opaqueref`.

PR #1888, opened 2017-12-06
CA-271863: Can not automatically resolve autostart-enabled VM in RPU.

michael2012z:
In RPU Precheck page, if user type Resolve or Resolve All to disable autostart-enabled VM, the refreshed listed can still contain autostart-enabled problem. Because the server update-and-notification took time, but the page was updated before server ready. 
Now a 1-second waiting is added before updating the Precheck page.
Note, in extreme situation, 1 second may not be enough to collect server response (for bad network, etc.). In that case, the fault can happen and user need to click "Check Again" or resolve again to update.

### DIFF
--- a/XenAdmin/Diagnostics/Problems/VMProblem/AutoStartEnabled.cs
+++ b/XenAdmin/Diagnostics/Problems/VMProblem/AutoStartEnabled.cs
@@ -96,7 +96,7 @@ namespace XenAdmin.Diagnostics.Problems.VMProblem
                            {
                                vm.Locked = false;
                            }
-                           int wait = 1000; // wait up to 1 second for the cache to be updated
+                           int wait = 5000; // wait up to 5 seconds for the cache to be updated
                            while (wait > 0 && vm.GetAutoPowerOn() != autostartValue)
                            {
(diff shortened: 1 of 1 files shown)


XenAdmin/Diagnostics/Problems/VMProblem/AutoStartEnabled.cs
```diff
@@ -95,7 +96,12 @@ private Action<Session> ActionDelegate(bool autostartValue)
                            {
                                vm.Locked = false;
                            }
-
+                           int wait = 1000; // wait up to 1 second for the cache to be updated
```
GaborApatiNagy: I think this looks good, but I think a longer timeout like 5 seconds would be better to be more prepared for slow connections. Note that normally this code would be finished very quickly regardless of the (max)wait value.

michael2012z: Agree. If the connection is really so slow, a waiting up to 5 seconds doesn't make things worse.

GaborApatiNagy:
Thanks, looks good. Merging.

PR #1891, opened 2017-12-07
CP-25564: Changed clusteringEnabled variable

serencorbett1:

### DIFF
--- a/XenAdmin/Wizards/NewSRWizard_Pages/ChooseSrProvisioningPage.cs
+++ b/XenAdmin/Wizards/NewSRWizard_Pages/ChooseSrProvisioningPage.cs
@@ -66,9 +66,7 @@ namespace XenAdmin.Wizards.NewSRWizard_Pages
 
         public override void PopulatePage()
         {
-            var master = Helpers.GetMaster(Connection);
-
-            var gfs2Allowed = !Helpers.FeatureForbidden(Connection, Host.RestrictGfs2) && Connection.Cache.Cluster_hosts.Any(cluster => cluster.host.opaque_ref == master.opaque_ref && cluster.enabled);
+            var gfs2Allowed = !Helpers.FeatureForbidden(Connection, Host.RestrictGfs2) && Connection.Cache.Clusters.Any();
 
             radioButtonGfs2.Enabled = labelGFS2.Enabled = gfs2Allowed;
             tableLayoutInfo.Visible = radioButtonLvm.Checked = !gfs2Allowed;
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/NewSRWizard_Pages/ChooseSrProvisioningPage.resx
```diff
@@ -348,14 +348,44 @@
   <data name="&gt;&gt;pictureBoxInfo.ZOrder" xml:space="preserve">
     <value>1</value>
   </data>
+  <data name="linkLabelPoolProperties.AutoSize" type="System.Boolean, mscorlib">
+    <value>True</value>
+  </data>
+  <data name="linkLabelPoolProperties.Location" type="System.Drawing.Point, System.Drawing">
```
MihaelaStoica: The "Enable clustering" link label needs to be alligned with the warning label.

XenAdmin/Wizards/NewSRWizard_Pages/ChooseSrProvisioningPage.cs
```diff
@@ -66,13 +67,26 @@ public bool IsGfs2
...
+        }
+
+        private void linkLabelPoolProperties_LinkClicked(object sender, System.Windows.Forms.LinkLabelLinkClickedEventArgs e)
+        {
+            var pool = Helpers.GetPoolOfOne(Connection);
+
+            using (PropertiesDialog propertiesDialog = new PropertiesDialog(pool))
+            {
+                propertiesDialog.SelectClusteringEditPage();
+                propertiesDialog.ShowDialog(this);
```
MihaelaStoica: When using this link to enable clustering, after the pool properties is closed and clustering is enabled, we need to refresh this page, so that thib provisioning is allowed and the warning is hidden.

XenAdmin/Wizards/NewSRWizard_Pages/ChooseSrProvisioningPage.cs
```diff
@@ -67,26 +69,58 @@ public bool IsGfs2
 
         public override void PopulatePage()
         {
+            Connection.Cache.RegisterCollectionChanged<Cluster>(Cluster_CollectionChangedWithInvoke);
```
MihaelaStoica: Thinking more about this, we should only register the CollectionChanged event on PageLoaded and deregister it on PageLeave. This way the page won't be refreshed when it's not visible. 
We will also need to reload the page on PageLoaded, not on PagePopulate (which is only called on the fisrt load).
Plus, on each reload we shouldn't change the selection of the radio buttons (currently we select first radio button even on reload).

PR #1895, opened 2017-12-11
CP-26131: IP Address Configuration when clustering is enabled

MihaelaStoica:
When configuring the IP address on the cluster network, we need to temporarily unplug all GFS2 SRs and disable clustering.

This commit also includes a fix to the action progress calculation.


MihaelaStoica:
I am closing this pull request because I need to implement it differently

PR #1897, opened 2017-12-12
CP-26131: IP Address Configuration when clustering is enabled

MihaelaStoica:
When configuring the IP address on the cluster network, we need to temporarily unplug all GFS2 SRs and disable clustering.

This commit also includes a fix to the action progress calculation.

### DIFF
--- a/XenModel/Actions/Network/ChangeNetworkingAction.cs
+++ b/XenModel/Actions/Network/ChangeNetworkingAction.cs
@@ -270,7 +270,7 @@ namespace XenAdmin.Actions
             foreach (var pbd in Connection.ResolveAll(host.PBDs).Where(pbd => pbd.currently_attached))
             {
                 var sr = Connection.Resolve(pbd.SR);
-                if (sr != null && sr.type.ToLowerInvariant() == "gfs2")
+                if (sr != null && sr.GetSRType(true) == SR.SRTypes.gfs2)
                 {
                     gfs2Pbds.Add(pbd);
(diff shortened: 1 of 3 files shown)


XenModel/Actions/Network/ChangeNetworkingAction.cs
```diff
@@ -235,5 +246,75 @@ private string GetIPInRange(string range_start, PIF existing_pif)
...
+
+            var clusterHost = Connection.Cache.Cluster_hosts.FirstOrDefault(c => c.host.opaque_ref == host.opaque_ref);
+            if (clusterHost == null) 
+                return;
+
+            // unplug the GFS2 SRs, saving the list of the PBDs unplugged, to plug back later
+            foreach (var pbd in Connection.ResolveAll(host.PBDs).Where(pbd => pbd.currently_attached))
+            {
+                var sr = Connection.Resolve(pbd.SR);
+                if (sr != null && sr.type.ToLowerInvariant() == "gfs2")
```
GaborApatiNagy: I think "gfs2" should be defined as a constant globally. It's also used in ClusteringEditPage.cs

MihaelaStoica: Good point! Actually, we should use SR.GetSRType(), like `sr.GetSRType(true) == SR.SRTypes.gfs2`

GaborApatiNagy: even better, thanks

MihaelaStoica: @GaborApatiNagy please see the new commit

PR #1903, opened 2017-12-25
CP-26000: Add SR-IOV column in Networking and NIC TabPage

jishengx:
Because the XAPI part hasn't been merged to the feature, this PR doesn't contain the logic. It's just about the GUI part.

### DIFF
--- a/XenAdmin/Controls/NetworkingTab/NetworkList.cs
+++ b/XenAdmin/Controls/NetworkingTab/NetworkList.cs
@@ -1041,7 +1041,7 @@ namespace XenAdmin.Controls.NetworkingTab
                     return;
                 }
             }
-            e.SortResult = StringUtility.NaturalCompare(e.CellValue1.ToString(), e.CellValue2.ToString());
+            e.SortResult = (e.CellValue1 == null || e.CellValue2 == null) ? 0 : (StringUtility.NaturalCompare(e.CellValue1.ToString(), e.CellValue2.ToString()));
             e.Handled = true;
         }
     }
--- a/XenAdmin/Controls/NetworkingTab/NetworkList.cs
+++ b/XenAdmin/Controls/NetworkingTab/NetworkList.cs
@@ -167,7 +167,8 @@ namespace XenAdmin.Controls.NetworkingTab
                         this.AutoColumn,
                         this.LinkStatusColumn,
                         this.NetworkMacColumn,
-                        this.MtuColumn});
+                        this.MtuColumn,
+                        this.NetworkSriovColumn});
 
                 //CA-47050: the Description column should be autosized to Fill, but should not become smaller than a minimum
                 //width, which here is chosen to be the column header width. To find what this width is set temporarily the
--- a/XenAdmin/Controls/NetworkingTab/NetworkList.cs
+++ b/XenAdmin/Controls/NetworkingTab/NetworkList.cs
@@ -1031,17 +1031,25 @@ namespace XenAdmin.Controls.NetworkingTab
         {
             if (XenObject != null && XenObject is VM && e.Column.Index == DeviceColumn.Index)
             {
-                int val1 = 0;
-                int val2 = 0;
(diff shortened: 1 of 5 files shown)


XenAdmin/TabPages/NICPage.Designer.cs
```diff
@@ -56,6 +56,7 @@ private void InitializeComponent()
             this.ColumnDeviceName = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.ColumnBusPath = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.ColumnFCoECapable = new System.Windows.Forms.DataGridViewTextBoxColumn();
+            this.SriovSupportedColumn = new System.Windows.Forms.DataGridViewTextBoxColumn();
```
michael2012z: Other columns were all named like "ColumnXxx", we should align with them. Change the name to "ColumnSriovSupported".

kc284: Also, the header is currently wrapping the text into two lines, while we normally go for single line headers.

XenAdmin/TabPages/NICPage.resx
```diff
@@ -212,14 +212,17 @@
...
     <value>True</value>
   </metadata>
-  <data name="SriovSupportedColumn.HeaderText" xml:space="preserve">
+  <data name="ColumnSriovSupported.HeaderText" xml:space="preserve">
     <value>SR-IOV supported</value>
   </data>
-  <data name="SriovSupportedColumn.Width" type="System.Int32, mscorlib">
-    <value>114</value>
+  <data name="ColumnSriovSupported.MinimumWidth" type="System.Int32, mscorlib">
+    <value>150</value>
```
kc284: The correct way to fix this is by disabling the columns' height resizing mode, which fixes the problem for all columns. Setting a minimum width to this column will not prevent the problem from occurring again if a new column is added to the gridview or even now in a different language. The minimum width does not need to change, as long as it is big enough to prevent the column from disappearing completely and allow the user to easily expand it.

kc284:
Please also fix the method NetworkList.NetworksGridView_SortCompare.

XenAdmin/Controls/NetworkingTab/NetworkList.cs
```diff
@@ -1041,7 +1041,7 @@ private void NetworksGridView_SortCompare(object sender, DataGridViewSortCompare
                     return;
                 }
             }
-            e.SortResult = StringUtility.NaturalCompare(e.CellValue1.ToString(), e.CellValue2.ToString());
+            e.SortResult = (e.CellValue1 == null || e.CellValue2 == null) ? 0 : (StringUtility.NaturalCompare(e.CellValue1.ToString(), e.CellValue2.ToString()));
```
kc284:
```
private void NetworksGridView_SortCompare(object sender, DataGridViewSortCompareEventArgs e)
{
    if (XenObject != null && XenObject is VM && e.Column.Index == DeviceColumn.Index)
    {
        int val1, val2;
        if (e.CellValue1 != null && int.TryParse(e.CellValue1.ToString(), out val1)
            && e.CellValue2 != null && int.TryParse(e.CellValue2.ToString(), out val2))
        {
            e.SortResult = val1.CompareTo(val2);
            e.Handled = true;
            return;
        }
    }

    if (e.CellValue1 == null && e.CellValue2 == null)
        e.SortResult = 0;
    else if (e.CellValue1 == null && e.CellValue2 != null)
        e.SortResult = 1;
    else if (e.CellValue1 != null && e.CellValue2 == null)
        e.SortResult = -1;
    else
        e.SortResult = StringUtility.NaturalCompare(e.CellValue1.ToString(), e.CellValue2.ToString());

    e.Handled = true;
}
```

jishengx: Thanks, it's very useful.

PR #1918, opened 2018-01-18
CA-280287: Fix the null reference exception on the Import wizard

MihaelaStoica:
- Added a check so that we don't attempting to show the license warning if the wizard is open without a preselected host/pool.
- The license warning is now updated whenever the host selection changes; plus, we are not attempting to show it if the wizard is open without a preselected host/pool.

- Also a minor refactoring in the GlobalSelectHost page to fix this issue: when the wizard is open with a pre-selected host, we pass this to the page by assigning SelectedHost property, which in turn sets the m_selectedHost. The getter for SelectedHost  resets m_selectedHost to the selected item in the host list. If the getter is used before the host list selection is updated, then the desired pre-selection is lost, because m_selectedHost is reset to null.

### DIFF
--- a/XenAdmin/Wizards/ImportWizard/GlobalSelectHost.cs
+++ b/XenAdmin/Wizards/ImportWizard/GlobalSelectHost.cs
@@ -43,7 +43,7 @@ namespace XenAdmin.Wizards.ImportWizard
 		private IXenObject m_selectedObject;
         private bool m_buttonNextEnabled;
 
-        public event Action HostSelectionChanged;
+        public event Action<IXenConnection> ConnectionSelectionChanged;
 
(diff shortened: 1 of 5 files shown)


kc284:
This looks ok to me.
Another thing one could do is add the connection as a parameter to the HostSelectionChanged delegate, i.e. `Action<IXenConnection> HostSelectionChanged` (or `ConnectionSelectionChanged`), in which case the check for the chosen item can move into the class firing the event, so the event handler in the wizard can simply call `ShowXenAppXenDesktopWarning(connection)` without having to worry about checks.

MihaelaStoica:
Thanks @kc284 for the suggestion. Please see the new commit.

PR #1919, opened 2018-01-18
CP-26618: Log more at INFO level when downloading an update

MihaelaStoica:

### DIFF
--- a/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
+++ b/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
@@ -194,7 +194,7 @@ namespace XenAdmin.Actions
             }
             catch (Exception e)
             {
-                log.InfoFormat("Exception occurred when extracting downloaded archive: {0}", e.Message);
+                log.ErrorFormat("Exception occurred when extracting downloaded archive: {0}", e.Message);
                 throw new Exception(Messages.DOWNLOAD_AND_EXTRACT_ACTION_EXTRACTING_ERROR);
             }
             finally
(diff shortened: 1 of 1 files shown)


XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -194,7 +194,7 @@ private void ExtractFile()
             }
             catch (Exception e)
             {
-                log.ErrorFormat("Exception occurred when extracting downloaded archive: {0}", e.Message);
+                log.InfoFormat("Exception occurred when extracting downloaded archive: {0}", e.Message);
```
kc284: I'm not sure about this change; maybe we should keep it at error level because, if we do a text search in a big log file for errors, it might be overlooked.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -209,15 +209,16 @@ private void ExtractFile()
             if (string.IsNullOrEmpty(PatchPath) && downloadUpdate)
             {
                 MarkCompleted(new Exception(Messages.DOWNLOAD_AND_EXTRACT_ACTION_FILE_NOT_FOUND));
-                log.DebugFormat("File '{0}.{1}' could not be located in downloaded archive", updateName, updateFileExtensions);
+                log.InfoFormat("No '{0}' file could be located in downloaded archive", string.Concat(updateFileExtensions));
```
kc284: It might be worth keeping the update name as well so it's easier to find it in a big log file with many similar failed actions; for example, could we write something like:
```
log.InfoFormat("The downloaded archive does not contain a file with any of the following names: '{0}'",
    string.Join(", ", extensions.Select(s => string.Format("{0}.{1}", update, s))));
```

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
MihaelaStoica: I'm not sure either, I will revert it

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
MihaelaStoica: We are not really looking for a particular update in the archive, we take the first one we find with a matching extension. That's why I thought that saying ""File '{0}.{1}' could not be located..." was not quite accurate.

kc284: Ah ok, fair enough. But I would still join the strings instead of concatenating them, it makes for a prettier log line.

PR #1924, opened 2018-01-22
CA-275566 plus refactoring.

kc284:
Best review the commits one by one. The changes in `XenAPI\Session.cs` can be best reviewed via
https://github.com/xapi-project/xen-api-sdk/pull/134.
With regards to CA-275566, I preferred the easiest solution, namely use the existing DbProxy implementation and avoid switching to the JsonRpcclient, which would have required implementation of a DbJsonRpcClient (we can still do that in future if we want).


XenModel/XenAPI-Extensions/Session.cs
```diff
@@ -144,9 +144,12 @@ private void LogJsonRequest(string json)
             }
             catch
             {
-                methodName = json;
-                parameters = "";
+                //ignore
             }
+#else
+            methodName = json;
+            parameters = "";
```
MihaelaStoica: As `parameters` is always empty in release mode, it would be nice to reflect this on line 162 below so we don't log "with params:" in release mode

XenModel/Network/SessionFactory.cs
```diff
@@ -39,17 +39,23 @@ public static class SessionFactory
...
         public static Session DuplicateSession(Session session, IXenConnection connection, int timeout)
         {
             if (DbProxy.IsSimulatorUrl(session.Url))
-                return new Session(session, DbProxy.GetProxy(connection, session.Url), connection);
-            else
-                return new Session(session, connection, timeout);
+                return new Session(session, DbProxy.GetProxy(connection, session.Url), connection)
+                {
+                    //do nothing; we don't want to swap backends for simulator Urls
+                    XmlRpcToJsonRpcInvoker = obj => { }
```
MihaelaStoica: Can also do `XmlRpcToJsonRpcInvoker =null` as we always perform a null check before calling the invoker.

PR #1930, opened 2018-02-01
CP-26491: modify PIF.IsPhysical function and Properties dialog according to SR-IOV

jishengx:


XenModel/XenAPI-Extensions/PIF.cs
```diff
@@ -55,6 +55,12 @@ public override string Name()
                 PIF transport_pif = Connection.Resolve(tunnel.transport_PIF);
                 return transport_pif.Name();
             }
+            else if(sriov_logical_PIF_of.Count != 0)
+            {
+                Network_sriov network_s = Connection.Resolve(sriov_logical_PIF_of[0]);
```
kc284: Normally it's a good idea to check for null when resolving a reference (see for example how the `else` blockl is written, even if the `else if` blocks above don't do so).

jishengx: Thanks. I will add the check items. Refer to the XenCenter design of SR-IOV, the MAC address of SR-IOV network won't display(just like the VLAN). This ticket is to figure out the SR-IOV logical pif and some related properties.

PR #1931, opened 2018-02-01
CA-280329: Fix CrossPoolMigrateDestinationPage with Current Server showing

makunterry:
Major modifications:
1. Create new overload of function `CreateTargetServerFilterList` in `SelectMultipleVMDestinationPage.cs` with one more argument, type `List<VM>` ;
2. Rewrite new overload function `CreateTargetServerFilterList` in sub-class `CrossPoolMigrateDestinationPage` ;
3. Call `CreateTargetServerFilterList ` with each `VM` later in function `PopulateDataGridView` ;

### DIFF
--- a/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
+++ b/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
@@ -148,15 +148,15 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard
             return new DelayLoadingOptionComboBoxItem(xenItem, filters);
         }
 
-        protected override List<ReasoningFilter> CreateTargetServerFilterList(IEnableableXenObjectComboBoxItem selectedItem)
+        protected override List<ReasoningFilter> CreateTargetServerFilterList(IEnableableXenObjectComboBoxItem selectedItem, List<VM> vmList)
         {
             var filters = new List<ReasoningFilter>();
 
             if(selectedItem != null)
             {
-                filters.Add(new ResidentHostIsSameAsSelectionFilter(selectedItem.Item, selectedVMs));
-                filters.Add(new CrossPoolMigrateCanMigrateFilter(selectedItem.Item, selectedVMs, wizardMode));
-                filters.Add(new WlbEnabledFilter(selectedItem.Item, selectedVMs));
+                filters.Add(new ResidentHostIsSameAsSelectionFilter(selectedItem.Item, vmList));
+                filters.Add(new CrossPoolMigrateCanMigrateFilter(selectedItem.Item, vmList, wizardMode));
+                filters.Add(new WlbEnabledFilter(selectedItem.Item, vmList));
             } 
 
             return filters;
--- a/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
+++ b/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
@@ -148,12 +148,16 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard
             return new DelayLoadingOptionComboBoxItem(xenItem, filters);
         }
 
-        protected override List<ReasoningFilter> CreateTargetServerFilterList(IEnableableXenObjectComboBoxItem selectedItem, List<VM> vmList)
+        protected override List<ReasoningFilter> CreateTargetServerFilterList(IEnableableXenObjectComboBoxItem selectedItem, List<string> vmOpaqueRefs)
         {
             var filters = new List<ReasoningFilter>();
 
-            if(selectedItem != null)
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -407,8 +410,8 @@ private void PopulateDataGridView(List<ReasoningFilter> homeserverFilters)
                             item.LoadAndWait();
                             cb.Items.Add(item);
 
-                            if ((m_selectedObject != null && m_selectedObject.opaque_ref == host.opaque_ref) ||
-                                (target != null && target.Item.opaque_ref == host.opaque_ref))
+                            if (item.Enabled && ((m_selectedObject != null && m_selectedObject.opaque_ref == host.opaque_ref) ||
```
MihaelaStoica: Why was this change needed?

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -383,6 +383,9 @@ private void PopulateDataGridView(List<ReasoningFilter> homeserverFilters)
                 {
                     var tb = new DataGridViewTextBoxCell {Value = kvp.Value.VmNameLabel, Tag = kvp.Key};
                     var cb = new DataGridViewEnableableComboBoxCell{FlatStyle = FlatStyle.Flat};
+                    List<ReasoningFilter> homeserverFilters = CreateTargetServerFilterList(
```
MihaelaStoica: This is the only place where `CreateTargetServerFilterList`, so using the new overload here breaks the Import wizard case, where we only have an override for the old function (CreateTargetServerFilterList(IEnableableXenObjectComboBoxItem item). See ImportSelectHostPage.cs.
So by using the new overload, we will always have an empty `List<ReasoningFilter>` in the Import Wizard, whne we should have the

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -383,6 +383,9 @@ private void PopulateDataGridView(List<ReasoningFilter> homeserverFilters)
                 {
                     var tb = new DataGridViewTextBoxCell {Value = kvp.Value.VmNameLabel, Tag = kvp.Key};
                     var cb = new DataGridViewEnableableComboBoxCell{FlatStyle = FlatStyle.Flat};
+                    List<ReasoningFilter> homeserverFilters = CreateTargetServerFilterList(
+                                                                selectedItem,
+                                                                new List<VM> { Connection.Resolve(new XenRef<VM>(kvp.Key)) });
```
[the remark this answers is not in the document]
makunterry: Thanks!
Will refine.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
makunterry: Very good point!
I will refine here. Thanks!

makunterry:
@MihaelaStoica  and @michael2012z 
I have updated the usage of `CreateTargetServerFilterList`, please help to take a look.
Thanks!

XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
```diff
@@ -148,12 +148,22 @@ protected override DelayLoadingOptionComboBoxItem CreateDelayLoadingOptionComboB
...
         {
             var filters = new List<ReasoningFilter>();
 
-            if(selectedItem != null)
+            if(selectedItem != null && vmOpaqueRefs != null && selectedVMs != null)
             {
+                List<VM> vmList = new List<VM>();
+                foreach (string _opaqRef in vmOpaqueRefs)
+                {
+                    vmList.AddRange(selectedVMs.FindAll(
```
MihaelaStoica: I don't think we need to call `FindAll` here, because there should be only one VM with a given opaque_ref.
So we can use `Find()` and `vmList.Add()`, e.g. 
```
foreach (string opaqRef in vmOpaqueRefs)
     vmList.Add(selectedVMs.Find(vm => vm.opaque_ref == opaqRef));
```

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
MihaelaStoica: OK, got it.

XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
makunterry: Agree.

makunterry:
@MihaelaStoica 
I have refined code of `vmList.add` part. Can you please help to take a look? 
Thanks!

PR #1933, opened 2018-02-01
CA-281646: Do not download the update again when using an update from disk

MihaelaStoica:
When applying a "new version" update, XenCenter starts an automated updates sequence that by default downloads all the updates. However, when using an update from disk, we shouldn't download it again.

With these changes, when an update from disk is used we save the [update, path] mapping, which we then use in the automated updates sequence, so we don't download that update again.

Also fixed a bug where a zipped update file from disk is not identified as a new version.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
@@ -31,7 +31,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.Eventing.Reader;
 using System.Windows.Forms;
 using XenAdmin.Controls;
 using XenAdmin.Controls.DataGridViewEx;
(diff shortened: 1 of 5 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -179,14 +178,23 @@ public WizardMode WizardMode
...
                 return WizardMode.SingleUpdate;
             } 
         }
 
+        public KeyValuePair<XenServerPatch, string> PatchFromDisk
+        {
+            get {
+                return selectFromDiskRadioButton.Checked && FileFromDiskAlert != null
+                    ? new KeyValuePair<XenServerPatch, string>(FileFromDiskAlert.Patch, SelectedNewPatch)
+                    : new KeyValuePair<XenServerPatch, string>(null, null);
```
kc284: It may be safer to add null checks in the actions for the patchFromDisk field itself; here it is carefully set to a `<null,null>` pair, but this might change in future (actually I don't see why not set it simply to null here too).

MihaelaStoica: KeyValuePair<K,V> can't be null, because it is a struct.

kc284: Oh indeed - my bad, then please ignore this comment.

PR #1934, opened 2018-02-02
CA-280322: Space requirement calculation error in migration wizard.

michael2012z:
Solution brief:
1. Add abstract method IsExtraSpaceNeeded() in SelectVMStorageWithMultipleVirtualDisksPage to calculate whether extra space is needed in the operation. No extra space is required if source and target SR is the same one and the operation is migrating or moving, not required otherwise.
2. In making each combobox item. Legacy behavior was to check if there is enough space on target SR to accommodate source disk(s). Now "is-extra-space-needed" condition is also taken into account.

Note:
1. SelectVMStorageWithMultipleVirtualDisksPage is common used for migrate/copy/move/import wizard.
2. There is interface update.

Signed-off-by: Michael Zhao <michael.zhao@citrix.com>

### DIFF
--- a/XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
+++ b/XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
@@ -54,6 +54,7 @@ namespace XenAdmin.Wizards.GenericPages
 		{
 			public string SysId { get; set; }
 			public ulong RequiredSpace { get; set; }
+			public IStorageResource ResourceData { get; set; }
 		}
 
         private class EnableableSrComboboxItem : IEnableableXenObjectComboBoxItem
@@ -196,7 +197,13 @@ namespace XenAdmin.Wizards.GenericPages
             return true;
         }
 
-        public override void PopulatePage()
+		protected virtual bool IsExtraSpaceNeeded(XenRef<SR> sourceRef, XenRef<SR> targetRef)
+		{
+			return true;
+		}
+
+
+		public override void PopulatePage()
 		{
 			m_totalSpaceRequired = 0;
 			m_dataGridView.Rows.Clear();
@@ -229,7 +236,7 @@ namespace XenAdmin.Wizards.GenericPages
                     var cellVmDisk = new DataGridViewTextBoxCell
                                          {
                                              Tag = new StorageDetail
-                                                     {RequiredSpace = resourceData.RequiredDiskCapacity, SysId = sysId},
+                                                     {RequiredSpace = resourceData.RequiredDiskCapacity, SysId = sysId, ResourceData = resourceData},
                                              Value = FormatDiskValueText(resourceData, disklabel)
 					                 	};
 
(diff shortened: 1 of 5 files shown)


XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
```diff
@@ -409,11 +430,17 @@ private bool CheckStorageRequirements(out string error)
 
 					var storageDetail = (StorageDetail)row.Cells[0].Tag;
 					string uuid = selectedItem.item.uuid;
+					ulong requiredSpace = storageDetail.RequiredSpace;
+
+					if (!IsExtraSpaceNeeded(storageDetail.ResourceData.SR, XenRef<SR>.Create((selectedItem.item.opaque_ref))))
```
MihaelaStoica: This can be written as: 
```
var requiredSpace = IsExtraSpaceNeeded(storageDetail.ResourceData.SR, new XenRef<SR>(toStringWrapper.item))
    ? storageDetail.RequiredSpace : 
    0;
```

XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
```diff
@@ -229,7 +236,7 @@ public override void PopulatePage()
                     var cellVmDisk = new DataGridViewTextBoxCell
                                          {
                                              Tag = new StorageDetail
-                                                     {RequiredSpace = resourceData.RequiredDiskCapacity, SysId = sysId},
+                                                     {RequiredSpace = resourceData.RequiredDiskCapacity, SysId = sysId, ResourceData = resourceData},
```
MihaelaStoica: As the `StorageDetail` now contains the `ResourceData`, we could get rid of the `RequiredSpace` property and use `resourceData.RequiredDiskCapacity` instead

XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
```diff
@@ -284,9 +291,23 @@ public override void PopulatePage()
 
 	    private void PopulateSrComboBox(List<ToStringWrapper<SR>> commonSRs)
 	    {
-	        List<EnableableSrComboboxItem> listToAdd = 
-	            commonSRs.Select(toStringWrapper => new EnableableSrComboboxItem(toStringWrapper, m_totalSpaceRequired)).ToList();
-
+			List<EnableableSrComboboxItem> listToAdd = new List<EnableableSrComboboxItem>();
+			foreach (var toStringWrapper in commonSRs)
```
MihaelaStoica: This calculation obsoletes the calculation for `m_totalSpaceRequired`. We should remove this field as it is assigned but never used. When removing, please note that there is a reference to it in a comment as well.

XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
```diff
@@ -284,9 +291,23 @@ public override void PopulatePage()
...
-
+			List<EnableableSrComboboxItem> listToAdd = new List<EnableableSrComboboxItem>();
+			foreach (var toStringWrapper in commonSRs)
+			{
+				ulong requiredSpace = 0;
+				foreach (var kvp in m_vmMappings)
+				{
+					foreach (IStorageResource resourceData in ResourceData(kvp.Key))
+					{
+						if (IsExtraSpaceNeeded(resourceData.SR, XenRef<SR>.Create((toStringWrapper.item.opaque_ref))))
```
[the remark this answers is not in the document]
michael2012z: It works. But what is different?

XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
michael2012z: Nice!

XenAdmin/Wizards/GenericPages/SelectVMStorageWithMultipleVirtualDisksPage.cs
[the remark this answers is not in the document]
MihaelaStoica: It does exactly the same thing, but it is a shorter syntax, especially that you can use a XenObject directly instead of XenObject.opaque_ref

PR #1937, opened 2018-02-05
CA-274082: Check status of SR uploaded updates during update installa…

jishengx:
…tion

### DIFF
--- a/XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
@@ -40,13 +40,9 @@ namespace XenAdmin.Diagnostics.Checks
 {
     public class PBDsPluggedCheck : Check
     {
-        SR srUploadedUpdates = new SR();
+        SR srUploadedUpdates;
 
-        public PBDsPluggedCheck(Host host):base(host)
-        {
-        }
-
-        public PBDsPluggedCheck(Host host, SR sr) : base(host)
+        public PBDsPluggedCheck(Host host, SR sr = null) : base(host)
         {
             srUploadedUpdates = sr;
         }
@@ -61,11 +57,11 @@ namespace XenAdmin.Diagnostics.Checks
 
             foreach (SR sr in brokenSRs)
             {
-                return new BrokenSR(this, sr);
+                return new BrokenSR(this, sr, Host);
             }
 
-            if (srUploadedUpdates.IsBroken())
-                return new BrokenSR(this, srUploadedUpdates);
+            if (srUploadedUpdates != null && !srUploadedUpdates.CanBeSeenFrom(Host))
+                return new BrokenSR(this, srUploadedUpdates, Host);
 
             return null;
         }
--- a/XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
@@ -40,17 +40,15 @@ namespace XenAdmin.Diagnostics.Checks
 {
     public class PBDsPluggedCheck : Check
     {
-        private Dictionary<Pool_update, Dictionary<Host, SR>> srUploadedUpdates = new Dictionary<Pool_update, Dictionary<Host, SR>>();
-        Pool_update selectedUpdate = null;
+        SR srUploadedUpdates = new SR();
 
         public PBDsPluggedCheck(Host host):base(host)
         {
         }
 
-        public PBDsPluggedCheck(Host host, Pool_update selectedUpdate, Dictionary<Pool_update, Dictionary<Host, SR>> sr) : base(host)
+        public PBDsPluggedCheck(Host host, SR sr) : base(host)
         {
             srUploadedUpdates = sr;
-            this.selectedUpdate = selectedUpdate;
         }
 
         protected override Problem RunCheck()
@@ -66,14 +64,8 @@ namespace XenAdmin.Diagnostics.Checks
                 return new BrokenSR(this, sr);
             }
 
-            if (selectedUpdate != null && srUploadedUpdates != null )
-            {
-                foreach (var dict in srUploadedUpdates)
-                {
-                    if (dict.Key.uuid == selectedUpdate.uuid && dict.Value.ContainsKey(Host) && dict.Value[Host].IsBroken())
-                        return new BrokenSR(this, dict.Value[Host]);
-                }
-            }
+            if (srUploadedUpdates.IsBroken())
+                return new BrokenSR(this, srUploadedUpdates);
 
             return null;
         }
--- a/XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
@@ -40,8 +40,17 @@ namespace XenAdmin.Diagnostics.Checks
 {
     public class PBDsPluggedCheck : Check
     {
+        private Dictionary<Pool_update, Dictionary<Host, SR>> srUploadedUpdates = new Dictionary<Pool_update, Dictionary<Host, SR>>();
+        Pool_update selectedUpdate = null;
+
         public PBDsPluggedCheck(Host host):base(host)
         {
+        }
+
+        public PBDsPluggedCheck(Host host, Pool_update selectedUpdate, Dictionary<Pool_update, Dictionary<Host, SR>> sr) : base(host)
+        {
+            srUploadedUpdates = sr;
+            this.selectedUpdate = selectedUpdate;
         }
 
         protected override Problem RunCheck()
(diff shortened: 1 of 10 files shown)


XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
```diff
@@ -40,10 +40,17 @@ namespace XenAdmin.Diagnostics.Checks
 {
     public class PBDsPluggedCheck : Check
     {
+        public List<SR> srUploadedUpdates = new List<SR>();
```
MihaelaStoica: `srUploadedUpdates  `should be private here, as it is only used in this class and set in the constructor.

XenAdmin/Diagnostics/Checks/PatchPrecheckCheck.cs
```diff
@@ -53,6 +54,7 @@ class PatchPrecheckCheck : Check
         private static Regex LivePatchResponseRegex = new Regex("(<livepatch).+(</livepatch>)");
 
         private readonly Dictionary<string, livepatch_status> livePatchCodesByHost;
+        public List<SR> srUploadedUpdates = new List<SR>();
```
MihaelaStoica: Please use pascal case instead of camel case for public members. See https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions.

XenAdmin/Diagnostics/Checks/PatchPrecheckCheck.cs
```diff
@@ -71,6 +73,14 @@ public PatchPrecheckCheck(Host host, Pool_patch patch, Dictionary<string, livepa
             this.livePatchCodesByHost = livePatchCodesByHost;
         }
 
+        public PatchPrecheckCheck(Host host, Pool_update update, Dictionary<string, livepatch_status> livePatchCodesByHost, List<SR> srUploadedUpdates)
+            : base(host)
```
MihaelaStoica: Here can use existing `PatchPrecheckCheck` constructor to avoid code repetition (i.e. use `this` instead of `base`):
```
public PatchPrecheckCheck(Host host, Pool_update update, Dictionary<string, livepatch_status> livePatchCodesByHost, List<SR> srUploadedUpdates)
      : this(host, update, livePatchCodesByHost)
{
    this.srUploadedUpdates = srUploadedUpdates;
}
```

XenAdmin/Wizards/PatchingWizard/PatchingWizard_UploadPage.cs
```diff
@@ -509,6 +509,7 @@ private void singleAction_Completed(ActionBase sender)
                                 AllIntroducedPoolUpdates.Add(PoolUpdate, SelectedNewPatchPath);
                             }
                         }
+                        srUploadedUpdates = (action as UploadSupplementalPackAction).srUploadedUpdates;
```
MihaelaStoica: By assigning `srUploadedUpdates` like this, it will always contain only the SRs from the last upload action. If I am uploading the update to multiple pools, we will have multiple upload actions here, but with this code only the SRs from one action are saved (from the one that finished last, I assume).
So we will need to add to the list instead of replacing it.

MihaelaStoica: I think we need `srUploadedUpdates` to be a `Dictionary<Host, SR>`  instead of a list containing all the SRs where we uploaded the patch, because the prechecks are run per host and we shouldn't check if all these SRs are broken for each host. We should only check the SRs that are relevant to that host,

XenAdmin/Diagnostics/Checks/PBDsPluggedCheck.cs
```diff
@@ -56,6 +63,13 @@ protected override Problem RunCheck()
             {
                 return new BrokenSR(this, sr);
             }
+
+            foreach (SR sr in srUploadedUpdates)
+            {
+                if(sr.IsDetached())
```
MihaelaStoica: I think we should use `IsBroken()` instead of `IsDetached()`

XenAdmin/Dialogs/RepairSRDialog.cs
```diff
@@ -282,7 +283,8 @@ private void RepairButton_Click(object sender, EventArgs e)
             
             _repairAction.Changed += action_Changed;
             _repairAction.Completed += action_Completed;
-            Grow(_repairAction.RunAsync);
+            if(!isModalMode)
```
MihaelaStoica: The two lines above should also be inside the `if` block because these event handlers will try to update the controls on a dialog that has been closed.

XenModel/Messages.resx
```diff
@@ -11740,6 +11740,9 @@ Verify that the file is a valid {1} export.</value>
   <data name="UPDATES_WIZARD_APPLY_UPDATE" xml:space="preserve">
     <value>&amp;Install update</value>
   </data>
+  <data name="UPDATES_WIZARD_BROKEN_SR_WARNING" xml:space="preserve">
+    <value>{0}: Check skipped because the {1} storage is broken.</value>
```
MihaelaStoica: naybe use "disconnected" instrad of "broken"?

XenAdmin/Wizards/PatchingWizard/PatchingWizard_UploadPage.cs
```diff
@@ -509,6 +509,12 @@ private void singleAction_Completed(ActionBase sender)
                                 AllIntroducedPoolUpdates.Add(PoolUpdate, SelectedNewPatchPath);
                             }
                         }
+
+                        foreach( var dict in (action as UploadSupplementalPackAction).SrUploadedUpdates)
```
MihaelaStoica: As we are casting action to UploadSupplementalPackAction a few times now in this block, we could consider replacing the type check `if (action is UploadSupplementalPackAction)` on line 489 with cast and check for null, like 
```
var uploadAction = action as UploadSupplementalPackAction;
if (uploadAction != null)
```

XenAdmin/Wizards/PatchingWizard/PatchingWizard_UploadPage.cs
```diff
@@ -509,6 +509,12 @@ private void singleAction_Completed(ActionBase sender)
                                 AllIntroducedPoolUpdates.Add(PoolUpdate, SelectedNewPatchPath);
                             }
                         }
+
+                        foreach( var dict in (action as UploadSupplementalPackAction).SrUploadedUpdates)
+                        {
+                            if (!SrUploadedUpdates.ContainsKey(dict.Key))
```
MihaelaStoica: We should replace the values in the dictionary even if the key exists already, because it could mean that we uploaded a different update to a different SR. 
For example: I upload update U1 to SR S1; then I detach the SR; the precheck fails, I go back and select a different update, U2, which will be uploaded to a different SR, say S2 (if S1 is still detached). In this case I want S2 to be saved here, not S1. But then what happens if I go back and select update U1 again?

jishengx: Thanks.I think you're right and replace the values in the dictionary. 
Because I think we can uploaded just one updates for a host once(except the Automated Updates because its logic is different here ). So in the case you mentioned above, if U1 can be uploaded again, then S2 can be updated to S1. If U1 cannot be uploaded again, then S2 shouldn't be updated.

MihaelaStoica: Actually U1 will not be uploaded again, because the wizard knows that it was uploaded already (to S1), and the wizard will allow the user to continue to the precheck step with the SrUploadedUpdates containing S2, which is not correct (as S2 doesn't contain U1). The precheck step will not flag the SR as broken., because it doesn't check S1, it checks S2.

kc284:
If the corrections are completed please change the lable of this PR to `updated` so it can be picked up for review.

MihaelaStoica:
I have reviewed the latest changes, my only remaining comment is on the use case discussed in the comments here: https://github.com/xenserver/xenadmin/pull/1937#discussion-diff-166988141R515

cheng-z:
Hi @MihaelaStoica
Could you please help to have a look again and merge if no other concern?
Thanks.

MihaelaStoica:
@jishengx Could you please rebase your branch on master, as the branch currently has conflicts and cannot be merged

XenAdmin/Wizards/PatchingWizard/PatchingWizard_UploadPage.cs
```diff
@@ -509,6 +509,23 @@ private void singleAction_Completed(ActionBase sender)
                                 AllIntroducedPoolUpdates.Add(PoolUpdate, SelectedNewPatchPath);
                             }
                         }
+
+                        var uploadAction = action as UploadSupplementalPackAction;
+                        if (uploadAction != null)
```
MihaelaStoica: `uploadAction != null` is always true here, because of the check on line 489.
We should move these two lines to replace the check on line 489 and use the new variable, uploadAction , everywhere inside the `if` block.

XenModel/Actions/SupplementalPack/UploadSupplementalPackAction.cs
```diff
@@ -52,6 +52,8 @@ public class UploadSupplementalPackAction : AsyncAction
         private readonly List<Host> servers;
 
         private Pool_update poolUpdate = null;
+
+        public Dictionary<Pool_update, Dictionary<Host, SR>> SrUploadedUpdates = new Dictionary<Pool_update, Dictionary<Host, SR>>();
```
MihaelaStoica: A Dictionary<Host, SR> is sufficient here, as the update is already published as `PoolUpdate`

PR #1941, opened 2018-02-07
CA-272150: Optimize destination page of cross-pool-migrate wizard.

michael2012z:
Solution brief:
1. Add a cache for CrossPoolMigrateCanMigrateFilter to save the result of server call VM.assert_can_migrate().
2. Add cancel method to filter.
3. Don't change the update of server list combo-box from sync to async. No server call would be made in building the server list due to the introduction of cache.
4. Sort server list.

The solution was a little bit different (point 3 mentioned above) from the one described in CA-272150 solution slides on which we have discussed.

### DIFF
--- a/XenAdmin/Wizards/CrossPoolMigrateWizard/Filters/CrossPoolMigrateCanMigrateFilter.cs
+++ b/XenAdmin/Wizards/CrossPoolMigrateWizard/Filters/CrossPoolMigrateCanMigrateFilter.cs
@@ -45,15 +45,27 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard.Filters
         private readonly WizardMode _wizardMode;
         private string disableReason = string.Empty;
         private readonly List<VM> preSelectedVMs;
+        private IDictionary<string, IDictionary<string, string>> cache;
+        private bool canceled = false;
+        private static readonly Object cacheLock = new Object();
 
-        public CrossPoolMigrateCanMigrateFilter(IXenObject itemAddedToComboBox, List<VM> preSelectedVMs, WizardMode wizardMode)
+        public CrossPoolMigrateCanMigrateFilter(IXenObject itemAddedToComboBox, List<VM> preSelectedVMs, WizardMode wizardMode, IDictionary<string, IDictionary<string, string>> cache = null)
             : base(itemAddedToComboBox)
         {
             _wizardMode = wizardMode;
+            if (cache == null)
+                this.cache = new Dictionary<string, IDictionary<string, string>>();
+            else
+                this.cache = cache;
 
             if (preSelectedVMs == null)
                 throw new ArgumentNullException("Pre-selected VMs are null");
             this.preSelectedVMs = preSelectedVMs;
+        }
+
+        public override void Cancel()
+        {
+            canceled = true;
         }
 
         public override bool FailureFound
@@ -73,6 +85,20 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard.Filters
 
                     foreach (VM vm in preSelectedVMs)
                     {
+                        if (canceled)
+                            return excludedHosts.Count == targets.Count;
+
+                        // obtain the cache data for a vm
+                        IDictionary<string, string> vmCache;
+                        lock (cacheLock)
+                        {
+                            if (!cache.ContainsKey(vm.opaque_ref))
+                            {
+                                cache.Add(vm.opaque_ref, new Dictionary<string, string>());
+                            }
+                            vmCache = cache[vm.opaque_ref];
+                        }
+
                         try
                         {
                             //CA-220218: for intra-pool motion of halted VMs we do a move, so no need to assert we can migrate
@@ -91,6 +117,16 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard.Filters
                                 continue;
                             }
 
+                            if (vmCache.ContainsKey(host.opaque_ref))
+                            {
+                                disableReason = vmCache[host.opaque_ref];
+                                if (!disableReason.Equals(string.Empty))
+                                {
+                                    excludedHosts.Add(host.opaque_ref);
+                                }
+                                continue;
+                            }
+
                             PIF managementPif = host.Connection.Cache.PIFs.First(p => p.management);
                             XenAPI.Network network = host.Connection.Cache.Resolve(managementPif.network);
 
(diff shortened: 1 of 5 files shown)


mcintyre94:
Can we move migrate VM wizard option to the top of the right-click menu when we have >16 hosts say? We should seriously consider getting rid of the host list for >16 IMO and just using the wizard, but I think moving the wizard link up is the minimum requirement.

mcintyre94:
Given that we calculate the pools as destinations one-at-a-time and doing this for large pools could be very slow, can we move the pool the host is currently in to the top of the list and compute it first? I'm thinking about opening the wizard on a smaller pool but not being able to select the current pool until after it's computed for a 64-host pool you're connected to but don't care about moving the VMs to.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -404,12 +405,20 @@ private void PopulateDataGridView(IEnableableXenObjectComboBoxItem selectedItem)
...
+                        var items = new List<DelayLoadingOptionComboBoxItem>();
+
                         foreach (var host in Connection.Cache.Hosts)
                         {
                             var item = new DelayLoadingOptionComboBoxItem(host, homeserverFilters);
                             item.LoadAndWait();
-                            cb.Items.Add(item);
+                            items.Add(item);
+                        }
+                        items.Sort(new DelayLoadingOptionComboboxItemCompare());
```
mcintyre94: Can we use a natural compare here please? Currently the order is 1,10,11, etc - we have a natural comparer that handles numbers :) See `StringUtility.NaturalCompare`

MihaelaStoica: Actually, whenever we sort the hosts in XenCenter we should use the default sort for the Host class, which sorts in the following order:
 * Masters first
 * Then connected slaves
 * Then disconnected servers
 Within each group, in NaturalCompare order

kc284:
Please label it as `updated` when corrections have been made, so it can be picked up for review again.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
michael2012z: It worked. And sorting result looks good in this way. 👍

michael2012z:
Hi, @mcintyre94. An easy solution to improve the context menu could be limiting the max length if the menu (like 8 or 16 maybe). I created another CA to track it (CA-284226) to avoid handling 2 different issues in one CA/PR.

XenAdmin/Wizards/CrossPoolMigrateWizard/Filters/CrossPoolMigrateCanMigrateFilter.cs
```diff
@@ -91,6 +117,16 @@ public override bool FailureFound
                                 continue;
                             }
 
+                            if (vmCache.ContainsKey(host.opaque_ref))
+                            {
+                                disableReason = vmCache[host.opaque_ref];
+                                if (!disableReason.Equals(string.Empty))
+                                {
+                                    excludedHosts.Add(host.opaque_ref);
```
MihaelaStoica: We should check if the excludedHosts list does not already contain this host

MihaelaStoica:
Is there anything we can do to improve the Destination combo box loading time when the page is first populated? I believe that the CrossPoolMigrateCanMigrateFilter , when executed for a pool, asserts if all the VMs can be migrated to all the hosts in the pool. And we mark the filter as failed if the excludedHosts contains all the hosts in the pool, meaning that we couldn't find a host for all VMs. We can look into an improvement where we stop the execution early, go example when we found a VM that cannot be migrated, or when we found a host for each VM.

michael2012z:
I figured out an improvement on excludedHosts: By the end of VMs loop, once a host was added into excludedHosts list, we can skip checking remaining VMs on this host and break for next host.
In this way, we will have a little pay-off: Because we haven't complete data in cache, there can be a short freezing in updating server combo-box. 
So, in this way, the benefit we won from caching would be balanced between "destination" list and "server" list.

michael2012z:
I didn't add the modification of speeding up "destination" combo-box I just mentioned. As it didn't improve the overall response time and dialog can freeze when cache is not hit in updating server list. This improvement is not good enough.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -404,16 +405,21 @@ private void PopulateDataGridView(IEnableableXenObjectComboBoxItem selectedItem)
...
+
+                        foreach (var host in sortedHosts)
                         {
                             var item = new DelayLoadingOptionComboBoxItem(host, homeserverFilters);
                             item.LoadAndWait();
                             cb.Items.Add(item);
-
-                            if (item.Enabled && ((m_selectedObject != null && m_selectedObject.opaque_ref == host.opaque_ref) ||
-                                (target != null && target.Item.opaque_ref == host.opaque_ref)))
+                            if (item.Enabled && ((m_selectedObject != null && m_selectedObject.opaque_ref == item.Item.opaque_ref) ||
```
MihaelaStoica: If we revert to using `host.opaque_ref` instead of `item.Item.opaque_ref`, the overall diff will show no changes here.

XenAdmin/Wizards/CrossPoolMigrateWizard/Filters/CrossPoolMigrateCanMigrateFilter.cs
```diff
@@ -91,6 +117,16 @@ public override bool FailureFound
                                 continue;
                             }
 
+                            if (vmCache.ContainsKey(host.opaque_ref))
+                            {
+                                disableReason = vmCache[host.opaque_ref];
+                                if (!disableReason.Equals(string.Empty) && !excludedHosts.Contains(host.opaque_ref))
```
MihaelaStoica: Apologies for not spotting this earlier, but our preferred method for checking for empty strings is using the function `string.IsNullOrEmpty()`, this way covering for null values as well.

michael2012z: Good comment, I forgot about IsNullOrEmpty(). We should use this.

PR #1947, opened 2018-02-16
ArchiveMaintainer performance and workflow refactoring.

kc284:


MihaelaStoica:
These changes look good to me and I've seen the performance improvement on the Performance tab, but when I tried editing an existing graph (or adding a new one), the dialog does not show any datasources in the list.

MihaelaStoica:
The issue on the Edit Graph dialog is caused by the switch to JSON-RPC, so unrelated to the changes in this pull request.

PR #1948, opened 2018-02-19
CA-283697: Fixed regression in OVF import manifested after switching to the JsonRpc SDK.

kc284:
There was a long standing issue with the OVF import whereby it was using the object constructors from hashtables, which were assigning default values which were not the API defaults. When these values were null, the JsonRpc client was not changing them like the XmlRpc proxy, resulting in errors.

This PR replaces where possible the constructors from hashtables (fixing at the same time some bugs due to typos), and also uses new bindings where the constructors do not reset all the object's fields, but only the ones specified in the hashtables, while the rest jeep the values they were initialised with.


XenOvfTransport/Import.cs
```diff
@@ -1226,109 +1226,67 @@ private XenRef<VM> DefineSystem(Session xenSession, VirtualHardwareSection_Type
             }
             #endregion
 
+            var longMemorySize = memorySize > long.MaxValue ? 0 : (long)memorySize;
```
MihaelaStoica: Why is the `longMemorySize `set to 0 if is greater than `long.MaxValue`? Why not `long.MaxValue`?

XenOvfTransport/Import.cs
```diff
@@ -1624,7 +1582,7 @@ private void AddResourceSettingData(Session xenSession, XenRef<VM> vmRef, RASD_T
                                 {
                                     if (rasd.Connection[0].Value.ToLower().Contains("sr="))
                                     {
-                                        string[] vpairs = rasd.Connection[0].Value.Split(new char[] { ',' });
+                                        string[] vpairs = rasd.Connection[0].Value.Split(new char[] {','});
```
MihaelaStoica: This can also be `rasd.Connection[0].Value.Split(',');`

XenOvfTransport/Import.cs
kc284: I was wondering about that too, but I'm not sure what the correct solution is. The max value seems to make more sense, but I left it 0 because this is what the Marshalling.ParseLong function also does.

MihaelaStoica: Ok then.
Speaking of the Marshalling class., there are a lot of Parse functions there that are not used anymore after these changes. Shall we remove them?

kc284: Unfortunately they are public :( I think this was a design bug as I doubt there was the intention to use them externally, but since it happened, we should first deprecate them for a little while.

PR #1949, opened 2018-02-20
CA-283656: Display Multipathing Capabilities of GFS2 SRs

serencorbett1:


XenModel/XenAPI-Extensions/SR.cs
```diff
@@ -829,7 +837,10 @@ public override string ToString()
 
         public bool MultipathCapable()
         {
-            return "true" == Get(sm_config, "multipathable");
+            SM relatedSm = GetSM;
+            bool isAnSmCapability = relatedSm != null && relatedSm.MultipathEnabled;
+            bool isInSmConfig = "true" == Get(sm_config, "multipathable");
```
kc284: You can use `bool isInSmConfig = BoolKey(sm_config, "multipathable")` for this instead.

XenModel/XenAPI-Extensions/SR.cs
```diff
@@ -146,6 +146,14 @@ public SRTypes GetSRType(bool ignoreCase)
             }
         }
 
+        public SM GetSM
```
kc284: Please create a method instead of a property (due to CA-235364).

XenModel/XenAPI-Extensions/SM.cs
```diff
@@ -53,5 +53,14 @@ public override bool IsHidden()
         {
             return BoolKey(other_config, HIDE_FROM_XENCENTER);
         }
+
+        public bool MultipathEnabled
```
kc284: Please create a method instead of a property (due to CA-235364).

PR #1954, opened 2018-02-24
CP-26421: Remove SR-IOV network on host's Networking TabPage

jishengx:

### DIFF
--- a/XenAdmin/Controls/NetworkingTab/NetworkList.cs
+++ b/XenAdmin/Controls/NetworkingTab/NetworkList.cs
@@ -478,7 +478,7 @@ namespace XenAdmin.Controls.NetworkingTab
         {
             foreach (PIF pif in network.Connection.ResolveAll(network.PIFs))
             {
-                if (pif.IsPhysical() && !pif.IsBondNIC() && pif.sriov_logical_PIF_of.Count == 0)
+                if (pif.IsPhysical() && !pif.IsBondNIC())
                     return true;
             }
             return false;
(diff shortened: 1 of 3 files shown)


cheng-z:
Hi @MihaelaStoica 
Could you please help to review this PR please?
Thanks.

XenAdmin/Controls/NetworkingTab/NetworkList.cs
```diff
@@ -478,7 +478,7 @@ private static bool HasPhysicalNonBondNIC(XenAPI.Network network)
         {
             foreach (PIF pif in network.Connection.ResolveAll(network.PIFs))
             {
-                if (pif.IsPhysical() && !pif.IsBondNIC())
+                if (pif.IsPhysical() && !pif.IsBondNIC() && pif.sriov_logical_PIF_of.Count == 0)
```
MihaelaStoica: IsPhysical() already checks `pif.sriov_logical_PIF_of.Count == 0`, so I don't think we need to add it here again.

XenModel/Actions/Network/NetworkAction.cs
```diff
@@ -206,7 +206,15 @@ private void destroyPIFs()
                 {
                     if (!pif.physical)
                     {
-                        VLAN.destroy(Session, pif.VLAN_master_of);
+                        if(pif.VLAN != -1)
+                        {
+                            VLAN.destroy(Session, pif.VLAN_master_of);
+                        }
+                        
+                        if(pif.sriov_logical_PIF_of.Count != 0)
```
MihaelaStoica: In other places where we use `sriov_logical_PIF_of` we also check it is is not null. 
In fact I think it would be nicer if we extracted this into a method and use it everywhere, e.g.
```
public bool IsSrIovLogicalPIF()
{
    return sriov_logical_PIF_of != null && sriov_logical_PIF_of.Count != 0;
}
```

cheng-z:
I think we need to rebase instead of merge?
@jishengx  could you please confirm with @MihaelaStoica  and get this PR merge?

PR #1955, opened 2018-02-24
CA-284226: Server list in context menu "Migrate to Server" builds slowly in 64-host pool

michael2012z:
Omit the server list in "Migrate to Server" menu if the pool size is larger than 16.


MihaelaStoica:
I've marked this PR as blocked until we finalise our discussions on the best solution

michael2012z:
Solution got changed.

PR #1961, opened 2018-02-28
CP-27188: Filter out the SR-IOV network when configuring IP address

jishengx:

As the requirement requested, the Mangement Interface shouldn't use the SR-IOV network because the Dom0 shouldn't use the VF(virtual function).

### DIFF
--- a/XenModel/XenAPI-Extensions/Network.cs
+++ b/XenModel/XenAPI-Extensions/Network.cs
@@ -293,6 +293,11 @@ namespace XenAPI
             return Connection.ResolveAll(PIFs).Find(pif => pif.VLAN >= 0) != null;
         }
 
+        public bool IsSriov()
+        {
+            return Connection.ResolveAll(PIFs).Find(pif => pif.IsSrIovLogicalPIF()) != null;
+        }
+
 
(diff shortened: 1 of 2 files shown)


cheng-z:
Looks good to me. Please @MihaelaStoica  help to have a look again.

XenModel/XenAPI-Extensions/Network.cs
```diff
@@ -293,6 +293,11 @@ public bool IsVLAN()
             return Connection.ResolveAll(PIFs).Find(pif => pif.VLAN >= 0) != null;
         }
 
+        public bool IsSriov()
+        {
+            return Connection.ResolveAll(PIFs).Find(pif => pif.sriov_logical_PIF_of != null && pif.sriov_logical_PIF_of.Count != 0) != null;
```
MihaelaStoica: As I had also said on PR  #1954, please create a PIF.IsSrIovLogicalPIF() function and use it here as well.

cheng-z:
Hi @MihaelaStoica
Could you please help to have a look again and merge if no other concern?
Thanks.

PR #1968, opened 2018-03-06
CA-284125: Automated updates should prioritise a pending host restart to a mandatory xapi restart

kc284:
This PR also contains refactoring.
I think it will be easier to review if each commit is done separately.
Note that the action Titles are currently a bit all over the place, this will be done in a different PR.
One thing requiring attention is the use of ResolvedHost vs CurrentHost vs HostXenRef in the PlanActions.


XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -362,19 +377,19 @@ private void WorkerDoWork(object sender, DoWorkEventArgs doWorkEventArgs)
...
 
                         // any non-restart-alike delayed action needs to be run if:
-                        // - this host is pre-Ely and there isn't any restart plan action among the delayed actions, or
-                        // - this host is Ely or above and bgw.AvoidRestartHosts contains the host's uuid (shows that live patching must have succeeded) or there isn't any restart plan action among the delayed actions
-                        if (!Helpers.ElyOrGreater(h) && !actions.Any(pa => pa.IsRestartRelatedPlanAction())
-                            || Helpers.ElyOrGreater(h) && (bgw.AvoidRestartHosts != null && bgw.AvoidRestartHosts.Contains(h.uuid) || !actions.Any(pa => pa.IsRestartRelatedPlanAction())))
+                        // - this host is pre-Ely and there isn't any delayed restart plan action, or
+                        // - this host is Ely or above and live patching must have succeeded or there isn't any delayed restart plan action
+                        if ((!Helpers.ElyOrGreater(h) && restartActions.Count <= 0)
+                            || (Helpers.ElyOrGreater(h) && (h.updates_requiring_reboot.Count <= 0 || restartActions.Count <= 0)))
```
kc284: I think we may need to resolve the host `h` again here. Overall I was wondering whether the host collection should be keyed on the host's opaque_ref rather than the object itself, because most of the HostPlanActions re-resolve it. In the same spirit, maybe it would be better if the PlanAction.StatusChanged event had the host's ref as a parameter and not the object itself.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -372,19 +377,19 @@ private void WorkerDoWork(object sender, DoWorkEventArgs doWorkEventArgs)
                         }
 
                         // any non-restart-alike delayed action needs to be run if:
-                        // - this host is pre-Ely and there isn't any restart plan action among the delayed actions, or
-                        // - this host is Ely or above and bgw.AvoidRestartHosts contains the host's uuid (shows that live patching must have succeeded) or there isn't any restart plan action among the delayed actions
-                        if (!Helpers.ElyOrGreater(h) && !actions.Any(pa => pa.IsRestartRelatedPlanAction())
-                            || Helpers.ElyOrGreater(h) && (bgw.AvoidRestartHosts != null && bgw.AvoidRestartHosts.Contains(h.uuid) || !actions.Any(pa => pa.IsRestartRelatedPlanAction())))
+                        // - this host is pre-Ely and there isn't any delayed restart plan action, or
+                        // - this host is Ely or above and live patching must have succeeded or there isn't any delayed restart plan action
+                        if ((!Helpers.ElyOrGreater(h) && restartActions.Count <= 0)
```
MihaelaStoica: Can this be simplified to
`restartActions.Count <= 0 || ( Helpers.ElyOrGreater(h) && (h.Connection.TryResolveWithTimeout(new XenRef<Host>(h.opaque_ref)).updates_requiring_reboot.Count <= 0))` ?

MihaelaStoica:
@kc284 Apart from the comment on simplifying the if statement, I am happy with the changes and I am going to merge it now, as PR #1937 needs to be rebased on these changes.

PR #1972, opened 2018-03-09
Fixed regression introduced in the RPU wizard + improvement to prevent similar regressions in future

kc284:


kc284:
@michael2012z and @MihaelaStoica please review.

PR #1980, opened 2018-03-12
CP-26884: add network-sriov info in OVF file when exporting vm

jishengx:

### DIFF
--- a/XenAdmin/Controls/NetworkingTab/NetworkList.cs
+++ b/XenAdmin/Controls/NetworkingTab/NetworkList.cs
@@ -509,7 +509,7 @@ namespace XenAdmin.Controls.NetworkingTab
                     // Cache populating?
                     return;
                 }
-                VIFDialog d = new VIFDialog(vm.Connection, null, VIF.GetDeviceId(vm));
+                VIFDialog d = new VIFDialog(vm.Connection, null, VIF.GetDeviceId(vm), vm);
                 if (d.ShowDialog(this) != DialogResult.OK)
                     return;
 
--- a/XenAdmin/Controls/NetworkingTab/NetworkList.cs
+++ b/XenAdmin/Controls/NetworkingTab/NetworkList.cs
@@ -509,7 +509,8 @@ namespace XenAdmin.Controls.NetworkingTab
                     // Cache populating?
                     return;
                 }
-                VIFDialog d = new VIFDialog(vm.Connection, null, VIF.GetDeviceId(vm), vm);
+
+                VIFDialog d = new VIFDialog(vm.Connection, null, VIF.GetDeviceId(vm), vm.HasSriovRecommendation());
                 if (d.ShowDialog(this) != DialogResult.OK)
                     return;
 
--- a/XenModel/XenAPI-Extensions/VM.cs
+++ b/XenModel/XenAPI-Extensions/VM.cs
@@ -426,6 +426,27 @@ namespace XenAPI
             }
         }
 
+        public bool HasSriovRecommendation()
+        {
+            XmlDocument xd = GetRecommendations();
+
+            if (xd == null)
+                return true;
+
+            try
+            {
+                XmlNode xn = xd.SelectSingleNode(@"restrictions/restriction[@field='allow_network_sriov']");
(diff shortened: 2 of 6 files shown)


cheng-z:
Looks good to me now. @MihaelaStoica  Could you please also help to have a look please

XenOvfTransport/Export.cs
```diff
@@ -329,6 +329,11 @@ private EnvelopeType _export(Session xenSession, string targetPath, string ovfna
                 {
                     OVF.AddOtherSystemSettingData(ovfEnv, vsId, "hardware_platform_version", vm.hardware_platform_version.ToString(), OVF.GetContentMessage("OTHER_SYSTEM_SETTING_DESCRIPTION_1"));
                 }
+
+                if (vm.HasSriovRecommendation())
+                {
+                    OVF.AddOtherSystemSettingData(ovfEnv, vsId, "allow_network_sriov", Messages.TRUE, OVF.GetContentMessage("OTHER_SYSTEM_SETTING_DESCRIPTION_1"));
```
MihaelaStoica: I don't think we should use a resource message here, as these messages are localised. 
We can use bool.TrueString or bool.TrueString.ToLowerInvariant() if we want the lowercase "true"

MihaelaStoica: How is this "allow_network_sriov" value used  in the import?

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -426,6 +426,27 @@ public bool CanHaveGpu()
...
+            if (xd == null)
+                return true;
+
+            try
+            {
+                XmlNode xn = xd.SelectSingleNode(@"restrictions/restriction[@field='allow_network_sriov']");
+                if (xn == null || xn.Attributes == null)
+                    return false;
+
+                return true;
```
MihaelaStoica: Why do we always return true here? Shouldn't we read the value of this field?

XenAdmin/Controls/NetworkingTab/NetworkList.cs
```diff
@@ -509,7 +509,7 @@ private void AddNetworkButton_Click(object sender, EventArgs e)
                     // Cache populating?
                     return;
                 }
-                VIFDialog d = new VIFDialog(vm.Connection, null, VIF.GetDeviceId(vm));
```
MihaelaStoica: Instead of passing the VM object to the dialog, we could just pass a bool (e.g. allowSriov) that is calculated here.

XenOvfTransport/Export.cs
jishengx: They will be used in XAPI part. XAPI will know if the vm allow sriov.

XenModel/XenAPI-Extensions/VM.cs
jishengx: Because if the vm has  "allow_network_sriov", the value will be true. If the vm isn't sriov allowed, it won't have the  "allow_network_sriov" in OVF.

XenOvfTransport/Export.cs
MihaelaStoica: But I don't think that xapi ever reads the ovf file. The OVF import code is all in XenCenter.

XenOvfTransport/Messages.resx
```diff
@@ -112,10 +112,10 @@
     <value>2.0</value>
   </resheader>
   <resheader name="reader">
-    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
MihaelaStoica: The changes to this file are not needed, we should revert them to reduce the diff.

XenOvfTransport/Export.cs
MihaelaStoica: Agreed to do the import part in a separate ticket

PR #1981, opened 2018-03-13
CA-278960: Add ColumnHeaderEx

makunterry:

1. Create ColumnHeaderEx derived from ColumnHeader;
2. Use ColumnHeaderEx in ListViewEx;
3. Set minimal width for each ColumnHeaderEx in PhysicalStoragePage.

### DIFF
--- a/XenAdmin/Controls/ColumnHeaderEx.Designer.cs
+++ b/XenAdmin/Controls/ColumnHeaderEx.Designer.cs
@@ -1,0 +1,36 @@
+﻿namespace XenAdmin.Controls
+{
+    partial class ColumnHeaderEx
+    {
+        /// <summary>
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary> 
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Component Designer generated code
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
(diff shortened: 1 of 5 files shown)


XenAdmin/Controls/ListViewEx.cs
```diff
@@ -43,5 +43,31 @@ public ListViewEx()
             // Magically eliminates flicker when the control is resized
             this.SetStyle(ControlStyles.OptimizedDoubleBuffer, true);
         }
+
+        protected override void OnColumnWidthChanging(ColumnWidthChangingEventArgs e)
+        {
+            base.OnColumnWidthChanging(e);
+
+            ColumnHeaderEx columnex = Columns[e.ColumnIndex] as ColumnHeaderEx;
```
michael2012z: I'm wondering if we can handle ColumnHeaderEx in a general way. 
Now the code is to check if the column is of type ColumnHeaderEx, and then do something if it is. I think it's not very good to handle a concrete type here. 
I haven't figured out a better idea. Any way, this is not a request to modify anything. I just want to discuss if it's possible.

kc284:
Instead of writing extra code for a custom control (which reduces maintainability), why not change it completely and use a datagridview on the physical storage page? We use many more datagridviews throughout XenCenter, and they are much more customizable than listviews - and they have a MinimumWidth property for the columns. If there is no particular reason why we can't use a datagridview here, I would get rid of the ListViewEx completely.

makunterry:
@kc284 
Besides *PhysicalStoragePage*, there are DockerProcessPage, WlbHostExclusionPage, WlbAutomationPage which are still using *ListViewEx*.
And I agree with you that we should use *datagridviewex* instead of *listviewex*, but for this small bug I prefer to make less change as possible to make the bug fixed. Maybe we should do the replacing job in a new CP ticket. What you think?

kc284:
Apologies, perhaps I did not make myself clear: I meant we should get rid of the ListViewEx only on the PhysicalStoragePage for this bug, I did not mean to change the other pages within this ticket.

Yes, making a new ticket for the other pages makes sense, especially if they suffer from the same problem (I haven't checked it yet, but they may do).

makunterry:
@kc284 
Then I am closing this PR and will change the code with DatagridviewEx.

PR #1982, opened 2018-03-13
CP-26363: Display the remaining VF numbers in NIC TabPage

jishengx:

### DIFF
--- a/XenAdmin/TabPages/NICPage.cs
+++ b/XenAdmin/TabPages/NICPage.cs
@@ -205,6 +205,7 @@ namespace XenAdmin.TabPages
             string vendor = Messages.HYPHEN;
             string device = Messages.HYPHEN;
             string busPath = Messages.HYPHEN;
+            string sriovSupported ;
 
             public PIFRow(PIF pif)
             {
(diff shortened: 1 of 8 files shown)


cheng-z:
LGTM. Please @MihaelaStoica  help to have a look.

XenAdmin/Wizards/NewNetworkWizard.cs
```diff
@@ -202,10 +203,27 @@ private void CreateCHIN()
...
+
+            dlg.ShowDialog(this);
+            return dlg.DialogResult;
+        }
+
         private void CreateSRIOV()
         {
             XenAPI.Network network = PopulateNewNetworkObj();
 
+            if (ShowSriovCreationWarning() != DialogResult.OK)
```
MihaelaStoica: This should be the first call in this function, before the line ```XenAPI.Network network = PopulateNewNetworkObj();```. Actually, because there are several `return`s in this function, the network variable should be moved closer to the place where it is used.

XenAdmin/TabPages/NICPage.cs
```diff
@@ -216,13 +217,32 @@ public PIFRow(PIF pif)
...
                 }
-                for (int i = 0; i < 9; i++)
+                for (int i = 0; i < 10; i++)
                 {
                     Cells.Add(new DataGridViewTextBoxCell());
                     updateCell(i);
                 }
 
-            }
+                DelegatedAsyncAction action = new DelegatedAsyncAction(pif.Connection,
```
MihaelaStoica: The action is only needed in the case where we call Network_sriov.get_remaining_capacity. In all the other cases (!pif.SriovCapable(), !pif.IsSriovPhysicalPIF()), the Cell value can be assigned immediatelly.
With this in mind, the function updateCell() seems more suitable for this code.

XenAdmin/TabPages/NICPage.cs
```diff
@@ -216,13 +217,32 @@ public PIFRow(PIF pif)
...
+                                                                        string.Format(Messages.FETCH_POSSIBLE_HOSTS, pif.Name()),
```
MihaelaStoica: As this action is not showed in the Events (suppressHistory is true), we can have empty string for title, startDescription and endDescription.

XenAdmin/TabPages/NICPage.cs
```diff
@@ -205,6 +205,7 @@ public class PIFRow : DataGridViewRow
             string vendor = Messages.HYPHEN;
             string device = Messages.HYPHEN;
             string busPath = Messages.HYPHEN;
+            string sriovSupported ;
```
MihaelaStoica: This can be a local variable in the function where it is used

XenAdmin/TabPages/NICPage.cs
```diff
@@ -254,6 +253,37 @@ private void updateCell(int index)
...
+                            Cells[9].Value = sriovSupported;                        }
+                        else
+                        { 
+                            DelegatedAsyncAction action = new DelegatedAsyncAction(pif.Connection,
+                                string.Format(Messages.FETCH_POSSIBLE_HOSTS, pif.Name()),
+                                string.Format(Messages.FETCHING_POSSIBLE_HOSTS, pif.Name()),
+                                string.Format(Messages.FETCHED_POSSIBLE_HOSTS, pif.Name()),
+                                delegate (Session session)
+                                {
+                                    sriovSupported = string.Format(Messages.REAMININF_VFS, Network_sriov.get_remaining_capacity(pif.Connection.Session, pif.sriov_physical_PIF_of[0].opaque_ref));
```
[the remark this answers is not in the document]
MihaelaStoica: Getting the remaining capacity should be done as a best-effort, so if it fails we should not throw an error. Therefore we should wrap the API call in a try-catch. Maybe if it fails we can just display "Yes"
```
try
{
    var remainingCapacity = Network_sriov.get_remaining_capacity(session, pif.sriov_physical_PIF_of[0].opaque_ref);
    sriovSupported = string.Format(Messages.REAMININF_VFS, remainingCapacity);
}
catch
{
    sriovSupported = Messages.YES;
}
```

PR #1987, opened 2018-03-15
CP-27332: Use latest version of XenAPI to pass build

cheng-z:
Update the API to REQ-46 294


cheng-z:
As it is only change for API, I will merge it directly.

@jishengx and @MihaelaStoica . Sorry for not ask your approve as we are waiting for the build pass.

PR #1989, opened 2018-03-15
CA-271850: Increase the timeout of session for querying MaximumTolera…

michael2012z:
Currently the timeout of session is set to 1 minute for for querying MaximumTolerableFailures.
In case of a large pool (say 64) with a great many VMs (like 4096), the calculation could take more time than that.
So I'm going to enlarge the timeout to a safe enough value. That would be 10 minutes.


mcintyre94:
This looks good to me. We're expecting the performance to be better with the new build anyway so hopefully nobody ever comes close to seeing it take 10 minutes, but it makes more sense to have the overhead than not.

PR #1993, opened 2018-03-19
 CA-286278: VLAN network on top of SRIOV also belongs to SR-IOV 

jishengx:


cheng-z:
Please @MihaelaStoica and @michael2012z help to have a look. :-)

XenModel/XenAPI-Extensions/Network.cs
```diff
@@ -295,6 +295,18 @@ public bool IsVLAN()
 
         public bool IsSriov()
         {
+            if(this.IsVLAN())
+            {
+                foreach (PIF pif in Connection.ResolveAll<PIF>(PIFs))
```
MihaelaStoica: PIFs is resolved in IsVLAN() and then again inside this block. It would be nice to avoid this. 
Also the code to check if the vlan is on sr-iov is repeated in NetworkList.cs, so it is suitable to be extracted into a function. And then this function would be simplified to something like: 
`return Connection.ResolveAll(PIFs).Find(pif => pif.IsSriovLogicalPIF() || pif.IsVlanOnSriov) != null;`

XenAdmin/Wizards/NewNetworkWizard_Pages/NetWDetails.cs
```diff
@@ -172,7 +172,11 @@ private List<int> GetVLANList(PIF nic)
 
             foreach (PIF pif in nic.Connection.Cache.PIFs)
             {
-                if (pif.device == nic.device)
+                XenAPI.Network network = pif.Connection.Resolve(pif.network);
```
MihaelaStoica: This should go inside the `if (pif.device == nic.device)` block, because there is no point in resolving the network for all the PIFs, we onlt need to do it for the one that satisfy this condition.

XenModel/XenAPI-Extensions/Network.cs
jishengx: Thanks. I have changed the logic in Network.cs. But I have no idea to avoid the code repeated in NetworkList.cs because I need to get the logic pif to judge if the sriov network needs host to reboot.

MihaelaStoica: How about this:
- instead of PIF.IsVlanOnSriov() we have a method that gives us the Network_sriov,  e.g.
```
        public XenRef<Network_sriov> NetworkSriov()
        {
            if (IsSriovLogicalPIF())
                return sriov_logical_PIF_of[0];

            if (VLAN < 0)
                return null;
            var vlan = Connection.Resolve(VLAN_master_of);
            if (vlan == null)
                return null;
            var taggedPif = Connection.Resolve(vlan.tagged_PIF);
            return taggedPif != null && taggedPif.IsSriovLogicalPIF() ? taggedPif.sriov_logical_PIF_of[0] : null;
        }
```
- then Netwotk.IsSrIov() becomes:
```
return Connection.ResolveAll(PIFs).Any(pif => pif.NetworkSriov() != null);
```
- and in NetworkList we can have:
```
                var networkSriov = Pif != null ? Pif.NetworkSriov() : null;
                SriovCell.Value = networkSriov == null
                    ? Messages.NO
                    : Pif.Connection.Resolve(networkSriov).requires_reboot
                        ? Messages.HOST_NEEDS_REBOOT_ENABLE_SRIOV 
                        : Messages.YES;
```

MihaelaStoica:
This PR now shows all the initial commits from the REQ-46 branch, probably because that branch has been rebased to master. Also there are conflicts that need to be resolved before merging.

MihaelaStoica:
Closing this, as it has been replaced by #2008  (on master)

PR #1997, opened 2018-03-20
CA-281652: Add VM_REQUIRES_GPU failure reason

jijiang:


XenModel/Messages.resx
```diff
@@ -11806,6 +11806,9 @@ Check your settings and try again.</value>
   <data name="UPDATES_WIZARD_CANNOT_MIGRATE_VM_NO_TOOLS" xml:space="preserve">
     <value>Cannot migrate VM '{0}', because it doesn't have up-to-date drivers installed.</value>
   </data>
+  <data name="UPDATES_WIZARD_CANNOT_MIGRATE_VM_NO_GPU" xml:space="preserve">
+    <value>Cannot migrate VM '{0}', because it requires GPU and no GPU is found in the host.</value>
```
MihaelaStoica: How about this message: 
Cannot migrate VM '{0}' because it has a virtual GPU attached and there is no suitable GPU available on other servers

PR #1998, opened 2018-03-20
CA-286342: SRIOV should be enabled on master when 2 NICs are not the …

jishengx:
…same type

### DIFF
--- a/XenModel/Actions/Network/CreateSriovAction.cs
+++ b/XenModel/Actions/Network/CreateSriovAction.cs
@@ -83,5 +83,10 @@ namespace XenAdmin.Actions
                 Network_sriov.create(Session, thePif.opaque_ref, networkRef);
             }
         }
+
+        protected override void Clean()
+        {
+            Connection.ExpectDisruption = false;
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


XenModel/Actions/Network/CreateSriovAction.cs
```diff
@@ -56,9 +56,31 @@ protected override void Run()
...
+                    continue;
+
+                if (host.IsMaster())
+                {
+                    pifOnMaster = thePif;
+                    break;
+                }
+            }
+
+            Connection.ExpectDisruption = true;
```
MihaelaStoica: We also have to set ExpectDisruption back to false at some point

XenModel/Actions/Network/CreateSriovAction.cs
```diff
@@ -56,9 +56,31 @@ protected override void Run()
...
+
+            // Enable SR-IOV network on Pool Master
+            RelatedTask = Network_sriov.async_create(Session, pifOnMaster.opaque_ref, networkRef);
+            PollToCompletion(0, 100);
+
+            // Enable SR-IOV network on Pool Slaves
+            selectedPifs.Remove(pifOnMaster);
+            foreach (PIF thePif in selectedPifs)
+            {
+                Network_sriov.create(Session, thePif.opaque_ref, networkRef);
```
MihaelaStoica: Why has this been changed to `create from `async_create` ?

jishengx: Because we need to know whether the action is successful. I was wrong before because I thought if we use the async_create we can also know the action status.

PR #2004, opened 2018-03-22
CA-285270: Enable iso file download in install update wizard

jijiang:

### DIFF
--- a/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
+++ b/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
@@ -34,42 +34,54 @@ using System.Net;
 using System.ComponentModel;
 using System.Threading;
 using System.IO;
+using System.Linq;
 using XenCenterLib.Archive;
 
 namespace XenAdmin.Actions
 {
-    internal enum DownloadState { InProgress, Cancelled, Completed, Error };
+    internal enum DownloadState
+    {
+        InProgress,
+        Cancelled,
+        Completed,
+        Error
+    };
 
     public class DownloadAndUnzipXenServerPatchAction : AsyncAction
     {
-        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+        private static readonly log4net.ILog log =
+            log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
         private const int SLEEP_TIME_TO_CHECK_DOWNLOAD_STATUS_MS = 900;
-        private const int MAX_NUMBER_OF_TRIES = 5; //If you consider increasing this for any reason (I think 5 is already more than enough), have a look at nextSleepMs in DownloadFile() as well.
+
+        //If you consider increasing this for any reason (I think 5 is already more than enough), have a look at nextSleepMs in DownloadFile() as well.
+        private const int MAX_NUMBER_OF_TRIES = 5;
+
         private Random random = new Random();
 
         private readonly Uri address;
-        private readonly string zippedFileName;
+        private readonly string outputFileName;
         private readonly string updateName;
-        private readonly string[] updateFileExtensions;
+        private readonly string[] updateFileSuffixes;
         private readonly bool downloadUpdate;
+        private readonly bool doNotExtractUpdate;
         private DownloadState patchDownloadState;
         private Exception patchDownloadError;
 
-        public string PatchPath
-        {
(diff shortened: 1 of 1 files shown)


XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -34,6 +34,7 @@
 using System.ComponentModel;
 using System.Threading;
 using System.IO;
+using System.Web.UI.WebControls;
```
michael2012z: It was not used, should be removed.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -68,7 +70,9 @@ public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string ou
             updateName = patchName;
             address = uri;
             downloadUpdate = address != null;
-            zippedFileName = outputFileName;
+            nonExtractUpdate = downloadUpdate &&
+                               Array.Exists(updateFileExtensions, item => address.ToString().Contains(item));
```
michael2012z: If an URL contains an extension, it must be a file that doesn't need to be extracted. Am I right?
I think we'd better guarantee the URL contains "iso" (for example) as an extension (at the end). If the URL is like "xxxxxisoxxxx", it also pass the check but it was not what we want.
Please correct me if my understanding is incorrect.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -217,9 +221,7 @@ protected override void Run()
         {
             if (downloadUpdate)
             {
-                log.InfoFormat("Downloading update '{0}' (from from '{1}') to '{2}'", updateName, address, zippedFileName);
-
-
+                log.InfoFormat("Downloading update '{0}' (from from '{1}') to '{2}'", updateName, address, outputFileName);
```
kc284: Double `from`.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
kc284: Or simply `updateFileExtensions.Any(item => address.ToString().Contains(item))`

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -232,11 +234,23 @@ protected override void Run()
...
-            log.DebugFormat("Extracting XenServer patch '{0}'", updateName);
-            Description = string.Format(Messages.DOWNLOAD_AND_EXTRACT_ACTION_EXTRACTING_DESC, updateName);
-            ExtractFile();
-            log.DebugFormat("Extracting XenServer patch '{0}' completed", updateName);
-
+            if (nonExtractUpdate)
+            {
+                string newFilePath = Path.Combine(Path.GetDirectoryName(outputFileName), updateName);
+                if (File.Exists(newFilePath))
+                    File.Delete(newFilePath);
```
kc284: Can we include some exception handling for these IO operations (Delete and Move).

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -49,10 +50,11 @@ public class DownloadAndUnzipXenServerPatchAction : AsyncAction
         private Random random = new Random();
 
         private readonly Uri address;
-        private readonly string zippedFileName;
+        private readonly string outputFileName;
         private readonly string updateName;
         private readonly string[] updateFileExtensions;
         private readonly bool downloadUpdate;
+        private readonly bool nonExtractUpdate;
```
kc284: Language improvement: let's rename this to `doNotExtractUpdate`.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -203,23 +220,23 @@ private void ExtractFile()
                     iterator.Dispose();
 
                 if (downloadUpdate)
-                    File.Delete(zippedFileName);
+                    File.Delete(outputFileName);
```
kc284: Apologies, I only saw this now: can we wrap this in a try-catch block too? An empty catch block will suffice:
```
if (downloadUpdate)
{
try { File.Delete(outputFileName); }
catch { }
}
```

jijiang:
Tested. Works fine now.

PR #2010, opened 2018-03-26
CA-286458: Fix latestcr XC requirement when apply CR update

jijiang:
Using minimum_xc_version attribute in xenserver version to
determine which xencenter version is required.

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -503,9 +503,10 @@ namespace XenAdmin.Core
 
 
         /// <summary>
-        /// Returns the latest XenCenter version or null, if the current version is the latest. 
-        /// If a server version is provided, it returns the XenCenter version that is required to work with that server. 
-        /// If no server version is provided it will return the latestCr XenCenter.
+        /// If parameter is null, it returns latestcr XenCenter version if it is greater than current XC version,
+        /// or null, if the current XC version is latestcr.
+        /// If parameter is not null, it returns the minimum XenCenter version if it is greater than the current XC version,
+        /// or null, if the minimum XC version couldn't be found or the current XC version is enough.
         /// </summary>
         /// <param name="serverVersion"></param>
         /// <returns></returns>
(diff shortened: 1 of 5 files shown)


XenAdmin/Core/Updates.cs
```diff
@@ -503,32 +503,25 @@ private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host h
...
+        /// Returns the minimum XenCenter version if it is greater than the current XC version,
+        /// or null, if the minimum XC version couldn't be found or the current XC version is enough
         /// </summary>
         /// <param name="serverVersion"></param>
         /// <returns></returns>
         public static XenCenterVersion GetRequiredXenCenterVersion(XenServerVersion serverVersion)
         {
-            if (XenCenterVersions.Count == 0)
+            var minXcVersion = serverVersion.MinimumXcVersion;
+            if (minXcVersion == null || minXcVersion == string.Empty || XenCenterVersions.Count == 0)
```
MihaelaStoica: Can use `string.IsNullOrEmpty` here

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -245,6 +245,7 @@ private void GetXenServerVersions(XmlDocument xdoc)
                     string buildNumber = "";
                     string patchUuid = "";
                     bool presentAsUpdate = false;
+                    string min_xc_version = "";
```
MihaelaStoica: minXcVersion wouls be a better name here

XenAdmin/Core/Updates.cs
```diff
@@ -503,32 +503,25 @@ private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host h
...
-        /// If no server version is provided it will return the latestCr XenCenter.
+        /// Returns the minimum XenCenter version if it is greater than the current XC version,
+        /// or null, if the minimum XC version couldn't be found or the current XC version is enough
         /// </summary>
         /// <param name="serverVersion"></param>
         /// <returns></returns>
         public static XenCenterVersion GetRequiredXenCenterVersion(XenServerVersion serverVersion)
         {
-            if (XenCenterVersions.Count == 0)
+            var minXcVersion = serverVersion.MinimumXcVersion;
```
MihaelaStoica: These changes have removed the case of null serverVersion ("If no server version is provided it will return the latestCr XenCenter.") and iit will actually throw a null reference exception if serverVersion  is null, which could still be the case when used from XenCenterVersionCheck for the RPU wizard

XenModel/Actions/Updates/XenServerVersion.cs
```diff
@@ -85,6 +86,7 @@ public class XenServerVersion
             BuildNumber = buildNumber;
             PatchUuid = patchUuid;
             PresentAsUpdate = presentAsUpdate;
+            MinimumXcVersion = minXcVersion;
```
MihaelaStoica: I think that MinimumXcVersion  should be of type Version, so that when we compare it with the items in XenCenterVersions we don't do a string comparision (as strings "7.4.0" and "7.4" are different)

PR #2015, opened 2018-03-29
CA-278960: Replace ListViewEx with DataGridViewEx for PhysicalStoragePage

makunterry:
Get rid of ListViewEx for PhysicalStoragePage, and avoid affecting every function for this page.

### DIFF
--- a/XenAdmin/TabPages/PhysicalStoragePage.cs
+++ b/XenAdmin/TabPages/PhysicalStoragePage.cs
@@ -43,6 +43,7 @@ using XenAPI;
 using XenAdmin.Commands;
 using XenAdmin.Dialogs;
 using XenAdmin.Controls.DataGridViewEx;
+using XenCenterLib;
 
 
 namespace XenAdmin.TabPages
@@ -178,22 +179,17 @@ namespace XenAdmin.TabPages
         private void RefreshItem(SR sr=null)
         {
             if (sr == null)
-                foreach (DataGridViewRow row in dataGridViewSr.Rows)
+                return;
+
+            foreach (DataGridViewRow row in dataGridViewSr.Rows)
+            {
+                var srRow = row as SRRow;
+                if (srRow != null && sr.Equals(srRow.SR))
                 {
-                    var srRow = row as SRRow;
-                    if (srRow != null)
-                        srRow.UpdateDetails();
+                    srRow.UpdateDetails();
+                    break;
                 }
-            else
-                foreach (DataGridViewRow row in dataGridViewSr.Rows)
-                {
-                    var srRow = row as SRRow;
-                    if (srRow != null && srRow.SR == sr)
-                    {
-                        srRow.UpdateDetails();
-                        break;
-                    }
-                }
+            }
         }
 
         private void UnregisterHandlers()
@@ -317,6 +313,9 @@ namespace XenAdmin.TabPages
             var sr1 = ((SRRow)dataGridViewSr.Rows[e.RowIndex1]).SR;
             var sr2 = ((SRRow)dataGridViewSr.Rows[e.RowIndex2]).SR;
 
+            string sr1CompStr = null;
+            string sr2CompStr = null;
+
             if (e.Column.Index == columnUsage.Index)
             {
                 int percent1 = sr1.physical_size == 0 ? 0 : (int)(100.0 * sr1.physical_utilisation / (double)sr1.physical_size);
@@ -324,17 +323,45 @@ namespace XenAdmin.TabPages
                 long diff = percent1 - percent2;
                 e.SortResult = diff > 0 ? 1 : diff < 0 ? -1 : 0;
                 e.Handled = true;
+                return;
             }
             else if (e.Column.Index == columnSize.Index)
             {
                 long diff = sr1.physical_size - sr2.physical_size;
                 e.SortResult = diff > 0 ? 1 : diff < 0 ? -1 : 0;
                 e.Handled = true;
+                return;
             }
             else if (e.Column.Index == columnVirtAlloc.Index)
             {
                 long diff = sr1.virtual_allocation - sr2.virtual_allocation;
                 e.SortResult = diff > 0 ? 1 : diff < 0 ? -1 : 0;
+                e.Handled = true;
+                return;
+            }
+            else if (e.Column.Index == columnName.Index)
+            {
+                sr1CompStr = sr1.Name();
+                sr2CompStr = sr2.Name();
+            }
+            else if (e.Column.Index == columnDescription.Index)
+            {
+                sr1CompStr = sr1.Description();
+                sr2CompStr = sr2.Description();
+            }
+
+            if (sr1CompStr != null && sr2CompStr != null)
+            {
+                var descCompare = StringUtility.NaturalCompare(sr1CompStr, sr2CompStr);
+                if (descCompare != 0)
+                {
+                    e.SortResult = descCompare;
+                }
+                else
+                {
+                    var refCompare = string.Compare(sr1.opaque_ref, sr2.opaque_ref, StringComparison.Ordinal);
+                    e.SortResult = refCompare;
+                }
                 e.Handled = true;
             }
         }
--- a/XenAdmin/TabPages/PhysicalStoragePage.cs
+++ b/XenAdmin/TabPages/PhysicalStoragePage.cs
@@ -61,7 +61,6 @@ namespace XenAdmin.TabPages
         {
             InitializeComponent();
 
-            dataGridViewSr.SelectionChanged += new EventHandler(dataGridViewSrs_SelectedIndexChanged);
             base.Text = Messages.STORAGE_TAB_TITLE;
             PBD_CollectionChangedWithInvoke=Program.ProgramInvokeHandler(PBD_CollectionChanged);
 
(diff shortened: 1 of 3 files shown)


makunterry:
@kc284 
Can you please help to take a look? Thanks!

XenAdmin/TabPages/PhysicalStoragePage.resx
```diff
@@ -121,84 +121,102 @@
...
+    <value>60</value>
   </data>
-  <data name="columnHeader3.Text" xml:space="preserve">
+  <metadata name="columnType.UserAddedColumn" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
+    <value>True</value>
+  </metadata>
+  <data name="columnType.HeaderText" xml:space="preserve">
     <value>Type</value>
   </data>
-  <data name="columnHeader3.Width" type="System.Int32, mscorlib">
```
michael2012z: We should keep the Width setting for each column. Otherwise, all columns are set to same width by default (now it is 127 each, which doesn't look good.).

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -169,7 +172,17 @@ private void sr_PropertyChanged(object sender, PropertyChangedEventArgs e)
             }
             else
             {
-                Program.Invoke(this, () => RefreshRowForSr((SR)sender));
+                Program.Invoke(Program.MainWindow, RefreshAllItems);
```
kc284: Why do you need to refresh all items instead of only the relevant row as the original code did?

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -60,9 +61,7 @@ public PhysicalStoragePage()
         {
             InitializeComponent();
 
-            listViewSrs.SmallImageList = Images.ImageList16;
-            listViewSrs.ListViewItemSorter = lvwColumnSorter;
-            listViewSrs.SelectedIndexChanged += new EventHandler(listViewSrs_SelectedIndexChanged);
+            dataGridViewSr.SelectionChanged += new EventHandler(dataGridViewSrs_SelectedIndexChanged);
```
kc284: Event registration can be done without explicitly specifying the handler: `dataGridViewSr.SelectionChanged += dataGridViewSrs_SelectedIndexChanged;`

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -328,74 +282,59 @@ private void newSRButton_Click(object sender, EventArgs e)
             new NewSRCommand(Program.MainWindow, connection).Execute();
         }
 
-        /// <summary>
-        /// Taken from http://support.microsoft.com/kb/319401.
-        /// </summary>
-        /// <param name="sender"></param>
-        /// <param name="e"></param>
-        private void listViewSrs_ColumnClick(object sender, ColumnClickEventArgs e)
+        private void dataGridViewSr_MouseUp(object sender, MouseEventArgs e)
```
kc284: I believe the `newSRButton_Click(object sender, EventArgs e)` method above is not used, hence can be removed.

kc284:
The header row should be one line and not wrap words (disable autosizing for the header height). Also, the width for columns Type, Shared, Usage, Size and Virtual allocation can be set to autosize all cells, because it the contents are relatively short (name and description are ok with Fill as they may contain longer strings).

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -328,74 +282,59 @@ private void newSRButton_Click(object sender, EventArgs e)
...
-        /// <summary>
-        /// Taken from http://support.microsoft.com/kb/319401.
-        /// </summary>
-        /// <param name="sender"></param>
-        /// <param name="e"></param>
-        private void listViewSrs_ColumnClick(object sender, ColumnClickEventArgs e)
+        private void dataGridViewSr_MouseUp(object sender, MouseEventArgs e)
         {
-            // Determine if clicked column is already the column that is being sorted.
-            if (e.Column == lvwColumnSorter.SortColumn)
```
kc284: The logici n this method hasa not been replicated in the datagridview and currently sorting is broken. For example, sorting by ascending Size should give `0B, 1024MB, 423.7GB, 3686.4GB`, but currently `3686.4GB` comes before `423.7GB`.

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: Because other tabpages also refresh like this.

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: Do you mean you prefer to register handler in *Designer*?

kc284: You can register it in the designer - in which case visual studio will add the event handler automatically and you don't need to worry about anything. But my comment was only about style, there isn't anything wrong if you register it as you have, it was only a sugegstion for shorter and more readable code.

XenAdmin/TabPages/PhysicalStoragePage.cs
kc284: Ok, but it does not mean it's the right thing to do; in fact, if it may be a concern for performance if you have many entries in the gridview.

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: This is a very valid point. Thanks for your comment!

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: I will refine this to only refresh particular rows.

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -307,6 +312,33 @@ private void dataGridViewSr_MouseUp(object sender, MouseEventArgs e)
             }
         }
 
+        private void dataGridViewSr_SortCompare(object sender, DataGridViewSortCompareEventArgs e)
+        {
+            var sr1 = ((SRRow)dataGridViewSr.Rows[e.RowIndex1]).SR;
+            var sr2 = ((SRRow)dataGridViewSr.Rows[e.RowIndex2]).SR;
+
+            if (e.Column.Index == columnUsage.Index)
```
kc284: We also need to handle it for the Name and Description column, see for example how it is done on the SrStoragePage.

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -172,18 +171,29 @@ private void sr_PropertyChanged(object sender, PropertyChangedEventArgs e)
             }
             else
             {
-                Program.Invoke(Program.MainWindow, RefreshAllItems);
+                Program.Invoke(Program.MainWindow, () => RefreshItem(sender as SR));
             }
         }
 
-        private void RefreshAllItems()
+        private void RefreshItem(SR sr=null)
```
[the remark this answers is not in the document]
makunterry: Accepted.

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: I tested with current code and it works fine for other columns. I guess other column won't be handled in this function so the *e.Handled* won't be set to *true* so it will be sorted automatically by the string sorting.
So is it necessary for us to handle all columns in this function?

kc284: 1) If you change the storage names or descriptions to include numbers at the beginning, e.g. `1 NFS storage`, `10 NFS storage`, `2 NFS storage`, the sorting should be 1, 2, 10, but now it is 1, 10, 2. 
2) if you change the storage names or descriptions to be exactly the same, e.g. `NFS storage`, `NFS storage`, the sorting should be according to the opaque_ref. Have you tested this?
The SrStoragePage has the necessary logic, all you need is to transfer it here.

kc284:
And a small language correction for the commit messages: it should be "Revised PhysicalStoragePage", not refined :)

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: No I haven't tested this case.
Thanks for your advise, I will "revise" :)

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -337,6 +336,41 @@ private void dataGridViewSr_SortCompare(object sender, DataGridViewSortCompareEv
...
+            {
+                sr1CompStr = sr1.Name();
+                sr2CompStr = sr2.Name();
+            }
+            else if (e.Column.Index == columnDescription.Index)
+            {
+                sr1CompStr = sr1.Description();
+                sr2CompStr = sr2.Description();
+            }
+            else if (e.Column.Index == columnType.Index)
```
kc284: The shared and type columns do not need handling.

XenAdmin/TabPages/PhysicalStoragePage.cs
```diff
@@ -337,6 +336,41 @@ private void dataGridViewSr_SortCompare(object sender, DataGridViewSortCompareEv
...
+                sr1CompStr = sr1.FriendlyTypeName();
+                sr2CompStr = sr2.FriendlyTypeName();
+            }
+            else if (e.Column.Index == columnShared.Index)
+            {
+                sr1CompStr = sr1.shared ? Messages.YES : Messages.NO;
+                sr2CompStr = sr2.shared ? Messages.YES : Messages.NO;
+            }
+
+            if (sr1CompStr != null && sr2CompStr != null)
```
kc284: Instead of relying on the developer setting or not setting this in the previous if blocks, it would be safer if you returned after setting e.Handled in the previous blocks.

XenAdmin/TabPages/PhysicalStoragePage.cs
makunterry: Why ? Because there won't be numbers in the string?

PR #2016, opened 2018-03-29
CP-17099: Do the double probe for iSCSI

MihaelaStoica:
- Do the double probe for iSCSI
- Do the probe_ext for HBA


kc284:
This looks generally ok to me, I have some suggestions for refactoring because the overall logic around the storage wizard is complicated and every little simplification helps, hence I've raised https://github.com/MihaelaStoica/xenadmin/pull/7

PR #2017, opened 2018-03-30
CA-284090: Show a blank page in HTTP(S) plugin tab-page when a new object was selected

michael2012z:
When a user switched from one object to another, Web Browser page could not be updated until the new page was available. The user could be confused, especially when it took a long time before the new page ready. 
The solution of this PR is to show a blank page immediately after a new object is selected. Real page will be requested after the blank page is shown.


XenAdmin/Plugins/Features/TabPageFeature.cs
```diff
@@ -105,6 +105,7 @@ internal class TabPageFeature : Feature
 
         private readonly Dictionary<IXenObject, BrowserState> BrowserStates = new Dictionary<IXenObject, BrowserState>();
         private BrowserState lastBrowserState = null;
+        private bool ResettingPage = false;
```
kc284: Code style: private fields are normally declared with first letter lowercase: `resettingPage` or sometimes underscored `_resettingPage`.

XenAdmin/Plugins/Features/TabPageFeature.cs
```diff
@@ -320,7 +318,8 @@ private void Browser_DocumentCompleted(object sender, WebBrowserDocumentComplete
             if (!XenCenterOnly && lastBrowserState != null)
             {
                 log.DebugFormat("url for '{0}' set to '{1}'", Helpers.GetName(lastBrowserState.Obj), e.Url);
-                lastBrowserState.Urls = new List<Uri> { e.Url };
+                if (ResettingPage == false)
```
kc284: Code style: in C# we normally write: `if (myVariable)` instead of `if (myVariable == true)` and `if (!myVariable)` instead of `if (myVariable ==false)`. (There are another couple of instances of this below).

kc284:
These changes do bring an improvement in the user experience, but I think we need some more work to make it obvious what's happening, because it can still be confusing for the user. Also, after these changes I see null reference exceptions thrown because Browser_DocumentCompleted is hit before Browser_ProgressChanged and MainWindowActionAtNavigateTime is null.

kc284:
Could you also rebase on the latest master. Thanks.

XenAdmin/Plugins/Features/TabPageFeature.cs
michael2012z: Thanks a lot for sharing the tip.

PR #2019, opened 2018-03-30
CA-286293: OVF export missing the recommendations field

jishengx:


XenOvfTransport/Export.cs
```diff
@@ -330,9 +330,9 @@ private EnvelopeType _export(Session xenSession, string targetPath, string ovfna
                     OVF.AddOtherSystemSettingData(ovfEnv, vsId, "hardware_platform_version", vm.hardware_platform_version.ToString(), OVF.GetContentMessage("OTHER_SYSTEM_SETTING_DESCRIPTION_1"));
                 }
 
-                if (vm.HasSriovRecommendation())
+                if (vm.recommendations != null)
```
kc284: Let's check for `string.IsNullOrEmpty(vm.recommendations)`, no need to add it if there is no content.

PR #2021, opened 2018-04-03
CP-17099: Do the double probe for HBA

MihaelaStoica:
- The LVMOHBAWarningDialog class has been changed to display the type of the existing SR and the requested type (that will be used for formatting)


XenAdmin/Wizards/NewSRWizard_Pages/SrWizardHelpers.cs
```diff
@@ -108,18 +108,6 @@ public static String DefaultSRName(String potentialName, IXenConnection connecti
...
-
-            foreach (SR.SRInfo sr in lst)
-            {
-                return sr.UUID;
-            }
-
-            return null;
-        }
-
         public static String GetEntry(Dictionary<String, String> dconf, String Key, String Default)
```
kc284: GetEntry can go as well.

PR #2022, opened 2018-04-03
Add XCP-ng compatiblity

gangsterrapper22:
This adds the XCP-ng compatibility. I don't know if you want to merge this or not.

### DIFF
--- a/XenModel/FriendlyNames.resx
+++ b/XenModel/FriendlyNames.resx
@@ -1841,4 +1841,58 @@
   <data name="Label-host.edition-desktop-cloud" xml:space="preserve">
     <value>XenApp/XenDesktop [Citrix] Cloud</value>
   </data>
+  <data name="Label-host.edition-xcp-ng" xml:space="preserve">
+    <value>XCP-NG</value>
+  </data>
+  <data name="Label-performance.CPU0-avg-freq" xml:space="preserve">
+    <value>Average CPU</value>
+  </data>
+  <data name="Label-performance.CPU1-avg-freq" xml:space="preserve">
+    <value>Average CPU</value>
+  </data>
+  <data name="Label-performance.CPU2-avg-freq" xml:space="preserve">
+    <value>Average CPU</value>
+  </data>
+  <data name="Label-performance.CPU3-avg-freq" xml:space="preserve">
+    <value>Average CPU</value>
+  </data>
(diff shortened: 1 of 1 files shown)


XenModel/FriendlyNames.resx
```diff
@@ -1841,4 +1841,58 @@
   <data name="Label-host.edition-desktop-cloud" xml:space="preserve">
     <value>XenApp/XenDesktop [Citrix] Cloud</value>
   </data>
+  <data name="Label-host.edition-xcp-ng" xml:space="preserve">
+    <value>XCP-NG</value>
+  </data>
+  <data name="Label-performance.CPU0-avg-freq" xml:space="preserve">
```
MihaelaStoica: This can be better achieved with a single resource string and regular expressions. See how we use _Label-performance.cpu_ in https://github.com/xenserver/xenadmin/blob/master/XenModel/Utils/Helpers.cs#L1039.

We would like to take this change, but please update it with the suggestion above.

XenModel/FriendlyNames.resx
```diff
@@ -1841,4 +1841,58 @@
   <data name="Label-host.edition-desktop-cloud" xml:space="preserve">
     <value>XenApp/XenDesktop [Citrix] Cloud</value>
   </data>
+  <data name="Label-host.edition-xcp-ng" xml:space="preserve">
+    <value>XCP-NG</value>
```
MihaelaStoica: Thank you for the PR.

The xenserver/xenadmin repo is currently focused on supporting Citrix XenServer. This means that there is no effort in place for other use-cases, to ensure applicability and avoid regressions going forward. Hence at this point, we don't wish to accept this change.

That said, note that XenCenter is licensed under BSD2 in order to make it possible for people to potentially fork the application for other use-cases, which are not covered here.

XenModel/FriendlyNames.resx
borzel: Solution with RegEx: https://github.com/xcp-ng/xenadmin/commit/abe3964cb6fd4ab546e0a6d35a24e7ed345d3bcb

MihaelaStoica: Thank you @borzel. Would you like to update this pull request with your solution?

MihaelaStoica:
Closing this PR, as it hasn't been updated with the suggested changes.

PR #2032, opened 2018-04-09
CA-287430: Fixed the issue where the "Install Update" page of the pat…

MihaelaStoica:
…ching wizard does not show the information on the host reboot steps.

- use the plan action's ProgressDescription property to display the steps in RestartHostPlanAction (evacuate host, reboot host, bring babies back);
- the plan action should not be visible if the restart is skipped;

- The change to the DownloadAndUnzipXenServerPatchAction is to ensure that the download progress description is updated on the wizard with the correct description (in automated updates mode, via DownloadPatchPlanAction); DownloadPatchPlanAction is using the ActionChanged event to update its description based on the  action's DownloadProgressDescription property, but the ActionChanged event is called when the PercentComplete is updated, so this needs to happen after DownloadProgressDescription has been updated.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PlanActions/RestartHostPlanAction.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PlanActions/RestartHostPlanAction.cs
@@ -29,8 +29,8 @@
  * SUCH DAMAGE.
  */
 
-using System;
 using System.Collections.Generic;
+using System.Text;
 using XenAdmin.Core;
 using XenAPI;
 
@@ -54,8 +54,6 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
 
         protected override void RunWithSession(ref Session session)
         {
-            Visible = true;
-
             var hostObj = GetResolvedHost();
 
             if (Helpers.ElyOrGreater(hostObj))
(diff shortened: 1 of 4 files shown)


XenAdmin/Wizards/PatchingWizard/PlanActions/RestartHostPlanAction.cs
```diff
@@ -71,7 +69,8 @@ protected override void RunWithSession(ref Session session)
                 else if (_restartAgentFallback)
                 {
                     log.Debug("Live patching succeeded. Restarting agent.");
-                    Title = string.Format(Messages.UPDATES_WIZARD_RESTARTING_AGENT, hostObj.Name());
+                    Visible = true;
+                    Title = ProgressDescription = string.Format(Messages.UPDATES_WIZARD_RESTARTING_AGENT, hostObj.Name());
```
kc284: Do we really need to set the ProgressDescription? It seems to me like duplication because within the methods EvacuateHost, RebootHost and BringBabiesback we already set the Title and/or the Status (which is presumably why I had originally thought it would be displayed). Can we use those instead? In fact I think we might be able to lose the PlanAction's ProgressDesctription completely in favour of Status?

MihaelaStoica: Changing the Title does not trigger a change event,  so the change won't be reflected in the UI. We can fix this, but I think that the Title should ideally only be set in the constructor. Also in this case the ProgressDescription is a multi-line text containing a line for each step in this plan action, lines being added as these steps are executed. So I am not sure the Title should be changed to this. 
I haven't considered the Status property because, where it is used, it contains more detailed information than it was needed in this case and also it is only a single line describing the current operations.

MihaelaStoica: However, we have indeed a bit too many related properties in this class and we should look into combining some of them.

kc284: Actually it seems to me the Title may not be needed at all for the plan actions (on the Patching wizard at least), it's probably there only because the plan actions were fashioned after the normal actions - but a plan action only needs to change and notify its current status and it's then up to the caller to accumulate it or do anything else with it. In any case, I agree that any simplification should be left for a different PR due to the size of the changes because they affect 2 wizards.

PR #2036, opened 2018-04-11
CA-287714: Fixed crash in the preformance graphs.

kc284:


kc284:
As an aside, these classes have a problematic design. For example, the DataArchive should not be making assumptions on the threads calling its methods (half of them on the UI thread, half of them off it, concepts used in the UI drawing like the Palette being modified on other threads), it should be up to the callers of a class to invoke if they are on a different thread - but fixing this is a different PR.

PR #2037, opened 2018-04-12
CA-286449: Filter out SR-IOV network when importing OVF

jishengx:


cheng-z:
Hi @MihaelaStoica 
As this is a blocker ticket and priority high, can I ask your help to review this PR and see if any comments? 
Thanks.

XenAdmin/Wizards/GenericPages/SelectMultipleVMNetworkPage.cs
```diff
@@ -255,7 +255,7 @@ protected DataGridViewComboBoxCell FillGridComboBox(object xenRef)
 			return cb;
 		}
 
-        private bool ShowNetwork(Host targetHost, XenAPI.Network network)
+        public virtual bool ShowNetwork(Host targetHost, XenAPI.Network network)
```
kc284: This does not need to be public, it can be protected.

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
```diff
@@ -51,6 +53,32 @@ protected override bool ImplementsIsDirty()
             return true;
         }
 
+        public override bool ShowNetwork(Host targetHost, XenAPI.Network network)
```
kc284: Why are you repeating the base class logic, instead of adding only the Sriov bit and then call return base.ShowNetwork?

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
```diff
@@ -51,6 +53,32 @@ protected override bool ImplementsIsDirty()
...
+            if (targetHost == null && !network.AllHostsCanSeeNetwork())
+                return false;
+
+            return true;
+        }
+
+        private bool Allow_Sriov_Network()
+        {
+            string xml = OVF.ToXml(SelectedOvfEnvelope);
+            return xml.Contains("allow-network-sriov");
```
kc284: If you export a VM into OVF format and open the ovf file in a text editor, you'll see that the `allow-network-sriov` is saved in the envelope with a value; here you're checking only the existence of the string; is the value of no importance?

MihaelaStoica: Also, we can export more than one VM in an OVF file, so in theory we can have some VMs with the allow-network-sriov flag present, others without it. This code will allow the selection of the SR-IOV network if one of the VMs has the flag on.

kc284: To elaborate a bit more on Mihaela's comment, we should not be looking randomly in the xml for this string, but rather look in the specific other configuration data section of the machine. Also, there should be a method in the OVF class to do this - I know ToXml is public, but it should probably not be; we should not be handling the xml in the UI level, we should rather have a method in the OVF class to look for what we need.

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
```diff
@@ -51,6 +54,31 @@ protected override bool ImplementsIsDirty()
             return true;
         }
 
+        protected override bool ShowNetwork(Host targetHost, XenAPI.Network network, string sysId = null)
+        {
+            if (network.IsSriov() && !AllowSriovNetwork(sysId))
+                return false;
+
+            return base.ShowNetwork(targetHost, network);
```
kc284: It's not a very good idea to call the base class method without the optional parameter. If someone changes the default value in the overriding method (not a good idea either, but it might happen), it will cause a bug. In general it might be best to avoid optional parameters in abstract and virtual methods because it's easy for things to go wrong. I would make it a normal parameter.

kc284: One more question that it didn't occur to me earlier: this method has been overriden only for the ImportSelectNetworkPage, however we have one more page, CrossPoolMigrateNetworkingPage that derives from SelectMultipleVMNetworkPage; do we need to filter SR-IOV there?

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
```diff
@@ -51,6 +54,31 @@ protected override bool ImplementsIsDirty()
...
+        private bool AllowSriovNetwork(string sysId)
+        {
+            var vhs = OVF.FindVirtualHardwareSectionByAffinity(SelectedOvfEnvelope, sysId, "xen");
+            var data = vhs.VirtualSystemOtherConfigurationData;
+
+            foreach (var s in data)
+            {
+                if (s.Name == "recommendations")
+                {
+                    if (s.Value.Value.Contains("allow-network-sriov"))
```
kc284: I think we need to read the s.Value.Value as xml and get the actual value of "allow-network-sriov" here, because this will still give the wrong answer if you have `restriction field="allow-network-sriov" value="0"`.
The method VM.HasSriovRecommendation() is not checking for the value either, I think in both cases we should repeat the same logic as in the method VM.CanHaveGpu().

kc284:
Could you also please rebase on master?

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
jishengx: For this problem, we only have the "allow-network-sriov" field for thoes VM which allows sriov. For those VMs which doesn't allow sriov, the "allow-network-sriov" doesn't appear. So we can just check the "allow-network-sriov" value.

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
jishengx: Thanks. For the CrossPoolMigrateNetworkingPage, we don't need to filter SR-IOV because of the requirement.

kc284: I'm afraid this is not clear in the req, indeed @MihaelaStoica posed the question in ticket CA-286129, but I don't see an explicit answer. In any case, and due to the time zone difference, we had a quick chat with @robhoes and it should be filtered out. Which means the check can go in the base class. In the one case, whether sriov is allowed will be taken from the VM, in the other from the OVF. Try seeing if it's better as a design to add this in the VmMapping class or make a method in the SelectMultipleVMNetworkPage that the child classes will override.

jishengx: Thanks. I think I have a wrong memory. I also checked the CA-286129 and found out that there was some problems that time. However, actually, I can't remember what is the problem. I think you're right and we need to filter it out.

kc284:
A correction requested by Mihaela, which I didn't realise yesterday: because this page is used both for live migration and copy, we have to differentiate between the two cases. So when the VM is running, we need to filter SR-IOV out regardless the VM's recommendations, otherwise we need to check the VM recommendations too.
I have some further remarks in making the design safer, to proceed a bit faster I just raised PR https://github.com/jishengx/xenadmin/pull/1 to your fork, you can merge it and then @MihaelaStoica can review the combined commits to ensure we got everything right.

XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateNetworkingPage.cs
```diff
@@ -96,6 +96,18 @@ public override NetworkResourceContainer NetworkData(string sysId)
             return new CrossPoolMigrationNetworkResourceContainer(vifs);
         }
 
+        protected override bool AllowSriovNetwork(XenAPI.Network network, string sysId)
+        {
+            VM vm = Connection.Resolve(new XenRef<VM>(sysId));
+            if (vm == null)
+                return false;
+
+            if (vm.IsRunning())
```
MihaelaStoica: How about a VM that is suspended? Probably needs to be treated the same way are running VMs

XenAdmin/Wizards/GenericPages/SelectMultipleVMNetworkPage.cs
```diff
@@ -255,8 +255,11 @@ protected DataGridViewComboBoxCell FillGridComboBox(object xenRef)
 			return cb;
 		}
 
-        private bool ShowNetwork(Host targetHost, XenAPI.Network network)
+        private bool ShowNetwork(Host targetHost, XenAPI.Network network, string vsId)
         {
+            if (network.IsSriov())
+                return AllowSriovNetwork(network, vsId);
```
MihaelaStoica: Shouldn't this be: 
```if (network.IsSriov() && !AllowSriovNetwork(network, vsId))
    return false;;
```
so that we continue to do the other checks in this function even if the `AllowSriovNetwork` returns true?

kc284: Oops,my bad, will correct it.

XenAdmin/Wizards/ImportWizard/ImportSelectNetworkPage.cs
```diff
@@ -51,6 +56,34 @@ protected override bool ImplementsIsDirty()
...
+
+                        XmlNode xn = xdRecommendations.SelectSingleNode(@"restrictions/restriction[@field='allow-network-sriov']");
+                        if (xn != null && xn.Attributes != null)
+                            return Convert.ToInt32(xn.Attributes["value"].Value) != 0;
+                    }
+                    catch
+                    {
+                        return false;
+                    }                    
+                }
```
MihaelaStoica: We should exit the foreach loop here if the "recommendation" has been found (and we didn't hit any the two `return`s above) , as there shouldn't be another instance of the "recommendation" data

PR #2041, opened 2018-04-17
CA-284234: Updated the dotnet-packages reference.

kc284:


kc284:
With regards to the `a` vs. `an` article in front of the SR abbreviation: I'm still waiting to hear on whether we have any official guidelines, but for the time being I corrected it because in the rest of the code we say "an SR" or "an HA pool", as we read it.

PR #2043, opened 2018-04-18
CA-287856: Add new API calls and update SR repair

serencorbett1:


serencorbett1:
@MihaelaStoica Cluster.pool_resync() should only be called once per pool, hence I have taken it out of hosts for loop.

XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -150,11 +150,19 @@ public static string GetEditionText(Edition edition)
...
+            {
+                return iscsi_iqn;
+            }
             return Get(other_config, "iscsi_iqn") ?? "";
         }
 
         public void SetIscsiIqn(string value)
-        {
+        {   
+            if (Helpers.KolkataOrGreater(this))
```
MihaelaStoica: this should be:
```
if (Helpers.KolkataOrGreater(this))
    iscsi_iqn = value;
else
    other_config = SetDictionaryKey(other_config, "iscsi_iqn", value);
```

PR #2045, opened 2018-04-20
CA-284135: Issues with scheduled uploads in the Health Check service.

michael2012z:
**Root cause:**
The logic here was: when a failed upload happened before a successful upload, scheduled upload would never happen again.
**Solution:**
We only take lastSuccessfulUpload into account when we are considering whether we need a retry after a previous failure:
- If no successful upload exists or the last successful upload is to early, we need to retry.
- If a successful upload existed and it happened after last failure, we don't retry.

Signed-off-by: Michael Zhao <michael.zhao@citrix.com>

### DIFF
--- a/XenServerHealthCheck/RequestUploadTask.cs
+++ b/XenServerHealthCheck/RequestUploadTask.cs
@@ -144,17 +144,15 @@ namespace XenServerHealthCheck
                 {
                     DateTime LastFailedUpload = HealthCheckSettings.StringToDateTime(Get(config, HealthCheckSettings.LAST_FAILED_UPLOAD));
 
-                    if (haveSuccessfulUpload)
-                    {
-                        if (DateTime.Compare(lastSuccessfulUpload, LastFailedUpload) > 0)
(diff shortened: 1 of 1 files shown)


XenServerHealthCheck/RequestUploadTask.cs
```diff
@@ -144,17 +144,15 @@ public static bool Request(IXenConnection connection, Session session)
...
-                        needRetry = true;
+                        if ((!haveSuccessfulUpload) ||
+                            (DateTime.Compare(lastSuccessfulUpload, LastFailedUpload) < 0))
+                        {
+                            log.InfoFormat("Retry since retryInterval{0} - {1} > {2} meeted", LastFailedUpload, DateTime.UtcNow, retryInterval);
+                            needRetry = true;
+                        }
                     }
                     else
                         return false;
```
MihaelaStoica: This `return` also needs to be removed, so that we continue with checking if a scheduled upload is due, even when a retry is not due

PR #2047, opened 2018-04-20
Corrected NetworkWizard warning. I10n of exceptions thrown when objects cannot be resolved.

kc284:


XenModel/Actions/Network/ChangeNetworkingAction.cs
```diff
@@ -235,13 +235,9 @@ private void BringUp(PIF new_pif, PIF existing_pif, int hi)
 
         private string GetIPInRange(string range_start, PIF existing_pif)
         {
-            Host host = Connection.Resolve(existing_pif.host);
-            if (host == null)
-                throw new Failure(Failure.INTERNAL_ERROR, "Host has gone away during IP address calculation!");
-
-            int i = Array.IndexOf(Hosts, host);
+            int i = Array.FindIndex(Hosts, h => h.opaque_ref == existing_pif.host);
```
kc284: I think here we can search by opaque_ref and avoid resolving, thus making things faster.

PR #2050, opened 2018-04-24
CA-287857: judge if the status of pifOnMaster before enable SR-IOV

jishengx:

### DIFF
--- a/XenModel/Actions/Network/CreateSriovAction.cs
+++ b/XenModel/Actions/Network/CreateSriovAction.cs
@@ -73,8 +73,11 @@ namespace XenAdmin.Actions
             Connection.ExpectDisruption = true;
 
             // Enable SR-IOV network on Pool Master
-            RelatedTask = Network_sriov.async_create(Session, pifOnMaster.opaque_ref, networkRef);
-            PollToCompletion(0, 100);
+            if(pifOnMaster != null)
+            {
+                RelatedTask = Network_sriov.async_create(Session, pifOnMaster.opaque_ref, networkRef);
+                PollToCompletion(0, 100);
+            }
 
             // Enable SR-IOV network on Pool Slaves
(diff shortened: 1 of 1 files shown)


XenModel/Actions/Network/CreateSriovAction.cs
```diff
@@ -73,8 +73,11 @@ protected override void Run()
             Connection.ExpectDisruption = true;
 
             // Enable SR-IOV network on Pool Master
-            RelatedTask = Network_sriov.async_create(Session, pifOnMaster.opaque_ref, networkRef);
-            PollToCompletion(0, 100);
+            if(pifOnMaster != null)
+            {
+                RelatedTask = Network_sriov.async_create(Session, pifOnMaster.opaque_ref, networkRef);
+                PollToCompletion(0, 100);
+            }
```
MihaelaStoica: We should move the line `selectedPifs.Remove(pifOnMaster);` inside the `if` block as well, to avoid an unnecessary operation (which is an O(n) operation).

Also (and apologies for not noticing it earlier), in the block above we do an `async_create` for the master while for the slaves we do `create`. We should be consistent. 
The `PollToCompletion(0,100)` call above will set the action's progress to 100%, while that  shouldn't happen, the action should be 100% completed only when all SR-IOV networks on the slaves are created.

XenModel/Messages.resx
```diff
@@ -13758,4 +13758,7 @@ You will need to navigate to the Console on each of the selected VMs to complete
   <data name="YOU_ARE_HERE" xml:space="preserve">
     <value>You are here</value>
   </data>
+  <data name="SRIOV_NETWORK_GONE" xml:space="preserve">
+    <value>Could not find the bond in [XenCenter]'s cache.</value>
```
MihaelaStoica: "SR-IOV network"  not "bond"

PR #2059, opened 2018-04-27
Resort Messages.resx

MihaelaStoica:


kc284:
Could you please re-compile XenCenter so the Designer.cs is also updated.

MihaelaStoica:
The Designer,cs is already correctly sorted.

PR #2062, opened 2018-05-04
CP-26866: Add a specific warning when creating a bond on cluster network

jijiang:

### DIFF
--- a/XenAdmin/Controls/BondDetails.cs
+++ b/XenAdmin/Controls/BondDetails.cs
@@ -236,7 +236,9 @@ namespace XenAdmin.Controls
 
             bool will_disturb_primary = NetworkingHelper.ContainsPrimaryManagement(pifs);
             bool will_disturb_secondary = NetworkingHelper.ContainsSecondaryManagement(pifs);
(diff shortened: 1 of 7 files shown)


XenModel/Network/NetworkingHelper.cs
```diff
@@ -263,6 +263,14 @@ public static bool ContainsSecondaryManagement(List<PIF> PIFs)
                 return p.IsSecondaryManagementInterface(true);
             });
         }
+
+        public static bool ContainsClusteringPif(List<PIF> PIFs)
```
kc284: Simpler:
```
public static bool ContainsClusteringPif(List<PIF> PIFs)
{
    return PIFs.Any(p => p.IsUsedByClustering());
}
```

XenAdmin/Controls/BondDetails.cs
```diff
@@ -279,6 +280,22 @@ internal DialogResult ShowCreationWarning()
                 }
                 return dialogResult;
             }
+
+            if (will_disturb_clustering)
```
kc284: Also, is the order of warnings meaningful (i.e. why has it been inerted between primary and secondary interface check)?

jijiang:
Tested, works fine.

PR #2073, opened 2018-05-14
CA-286291: Missing copyright.

kc284:


kc284:
Merging to unblock the build.

PR #2074, opened 2018-05-14
CA-289150: Use a lock object to ensure that the hosts and connections…

MihaelaStoica:
… lists are safely accessed from separate threads

This change fixes the crash with IndexOutOfRangeException in XenServerUpdateAlert


kc284:
Looks ok to me.
You may also want to fix this usage of DistinctHosts https://github.com/xenserver/xenadmin/blob/171284ff343044b636ea1ee84c5d57115331b736/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs#L233, which can be stored in a local variable to avoid calculcating twice.

PR #2078, opened 2018-05-16
More general hotkey definition

schulz-alexander:
Use "Control+N" instead of "Ctrl+N" to be buildable on non-english operation systems

### DIFF
--- a/XenAdmin/MainWindow.ja.resx
+++ b/XenAdmin/MainWindow.ja.resx
@@ -2304,9 +2304,6 @@
   <data name="HostMenuItem.Text" xml:space="preserve">
     <value>サーバー(&amp;S)</value>
   </data>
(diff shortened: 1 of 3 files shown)


kc284:
This workaround is a bit volatile because if you go afterwards and make any changes to the control's designer, visual studio will automatically reset it back to `Ctrl+N`. Check if you can get it to work by removing the shortcut completely from the resx and assigning it programmatically - and please ensure the solution is tested and working on all the supported language packs (en-us, zh-cn, ja).

borzel:
Will look at that soon...

borzel:
I start now working on that .. was busy last days...

borzel:
> Check if you can get it to work by removing the shortcut completely from the resx and assigning it programmatically

Solution was easy because the suggested solution was already there. Someone just forgot to remove the shortcut definition from the resx-files.

PR #2080, opened 2018-05-17
CA-289618: Clear the created network after enabling SR-IOV failed

jishengx:

### DIFF
--- a/XenModel/Actions/Network/CreateSriovAction.cs
+++ b/XenModel/Actions/Network/CreateSriovAction.cs
@@ -40,6 +40,8 @@ namespace XenAdmin.Actions
 {
     public class CreateSriovAction : PureAsyncAction
     {
+        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+
         XenAPI.Network newNetwork;
         private List<PIF> selectedPifs;
 
@@ -52,11 +54,12 @@ namespace XenAdmin.Actions
         }
 
         protected override void Run()
-        {
-            // Create the new network            
-            XenRef<XenAPI.Network> networkRef = XenAPI.Network.create(Session, newNetwork);
-     
+        {  
             PIF pifOnMaster =null;
+
+            if (selectedPifs.Count == 0)
+                return;
+
             foreach (PIF thePif in selectedPifs)
             {
                 Host host = thePif.Connection.Resolve<XenAPI.Host>(thePif.host);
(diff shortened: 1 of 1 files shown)


XenModel/Actions/Network/CreateSriovAction.cs
```diff
@@ -81,19 +80,41 @@ protected override void Run()
...
+                    lo += inc;
+                }
+            }
+            catch(Exception)
             {
-                RelatedTask = Network_sriov.async_create(Session, thePif.opaque_ref, networkRef);
-                PollToCompletion(lo, lo + inc);
-                lo += inc;
+                DestroyNetwork(networkRef);
+                throw;
```
MihaelaStoica: How about the `Network_sriov` objects that we successfully created in the `foreach (PIF thePif in selectedPifs)` loop above?
Do we need to destroy those as well if we destroy the network?

jishengx: Thank you for reminding. I think we don't need to destroy the created ```Network_sriov``` objects. There should be two situation. The first one is that when some NICs have been enabled SR-IOV and the  ```Network_sriov``` objects have been created. So we also shouldn't destroy the network and the  ```Network_sriov``` objects. The second one is that no NIC has been enabled SR-IOV, so we need to destroy the network and there is no  ```Network_sriov``` object need to be deleted.

kc284: Yes, but what happens if you have say 3 NICs with enabled SR-IOV, you create 2 Network_sriov objects successfully and the third creation fails for some unforseen reason? You destroy the network, but then you have the first 2 Network_sriov objects left without one - should those 2 be deleted too?

jishengx: Thanks. I think the problem you mentioned has been solved in the latest commits. The ```if(lo == inc)``` was used for this. Only when the first time the creation fails, the network will be destroyed.

jishengx: Hi @kc284 and @MihaelaStoica, I have already updated the PR. Please correct me if you have further comments. Thank you very much.

kc284: Ah ok, I hadn't seen the second commit when I wrote the comment.
One more question with regards to the polling, and I apologise if I'm missing something, but why are you using ` int inc = 100 / (selectedPifs.Count + 1); int lo = inc;` instead of ` int inc = 100 / selectedPifs.Count; int lo = 0;` (and checking obviously that Count is not 0)? With the first method the progress reported gives the wrong impression to the user.

kc284: Also, could you please squash the commits. Thanks.

jishengx: Thank you. I will refine it.

XenModel/Actions/Network/CreateSriovAction.cs
```diff
@@ -78,22 +77,45 @@ protected override void Run()
                 selectedPifs.Insert(0, pifOnMaster);
             }
 
-            int inc = 100 / (selectedPifs.Count + 1);
-            int lo = inc;
+            int inc = 100 / selectedPifs.Count;
```
kc284: As I mentioned in my previous comment, we also need to ensure Count is not 0.
Perhaps the best is to check and if 0 return - after all, if we have no pifs, do we need to create the network?

jishengx: I think in NewNetworkWizard.cs we already check the pif is not 0 ```if (sriovSelectedPifs.Count != 0)
                (new CreateSriovAction(xenConnection, network, sriovSelectedPifs)).RunAsync();```
Do we need check here again?

kc284: I recommend we check again; it makes the code much safer if something changes in the way the action is called.

kc284: It's a cheap check anyway without any practical impact on performance.

jishengx: OK, I agree it's safer. Thank you.

PR #2087, opened 2018-06-01
Fixed regression in selection of VM destination; some logic simplification; suggested fix for CA-289721.

kc284:
Please review the commits separately.

### DIFF
--- a/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
+++ b/XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
@@ -164,15 +164,11 @@ namespace XenAdmin.Wizards.CrossPoolMigrateWizard
             if (!cancel)
             {
                 bool targetDisconnected = cancel;
-                Program.Invoke(Program.MainWindow,
-                               delegate
-                                   {
-                                       if (Connection == null || !Connection.IsConnected)
-                                       {
-                                           CrossPoolMigrateWizard.ShowWarningMessageBox(Messages.CPM_WIZARD_ERROR_TARGET_DISCONNECTED);
-                                           targetDisconnected = true;
-                                       }
-                                   });
+                if (Connection == null || !Connection.IsConnected)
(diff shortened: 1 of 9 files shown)


XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
```diff
@@ -64,11 +64,6 @@ public CrossPoolMigrateDestinationPage(Host preSelectedHost, List<VM> selectedVM
             InitializeText();
         }
 
-        public override bool EnableNext()
```
MihaelaStoica: With this change, when moving from this page to previous page and then back again, the Destination selection is lost (this is an existing issue, not caused by these changes), but the Next button remains enabled and I can proceed to the Next page.

XenAdmin/Wizards/GenericPages/SelectMultipleVMDestinationPage.cs
```diff
@@ -171,9 +158,26 @@ protected void ShowWarning(string warningText)
...
+            UnregisterHandlers();
+            CancelFilters();
+            ClearComboBox();
+            ClearDataGridView();
+            ChosenItem = null;
+        }
+
+        protected override void PageLeaveCore(PageLoadedDirection direction, ref bool cancel)
+        {
+            UnregisterHandlers();
```
MihaelaStoica: This code is never reached by the Migrate VM wizard, because the `CrossPoolMigrateDestinationPage `class overrides `PageLeaveCore`, without calling the method in the base class.

XenAdmin/Wizards/CrossPoolMigrateWizard/CrossPoolMigrateDestinationPage.cs
kc284: The latest changes improve slightly the behaviour by restoring selection after moving backwards from and then forwards to the destination page at pool level - not the host mapping though; this requires more changes which I think have started going outside the scope of this PR.

PR #2088, opened 2018-06-05
CA-284914: Prechecks are re-run every time the 'Hide successful prech…

jishengx:
…ecks' is ticked/unticked

Signed-off-by: Jisheng Xing <Jisheng.Xing@citrix.com>

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
@@ -960,21 +960,24 @@ namespace XenAdmin.Wizards.PatchingWizard
 
         private void RefreshDataGridViewList()
         {
-            dataGridView1.Rows.Clear();
-
-            foreach (var row in allRows)
+            lock (_update_grid_lock)
             {
-                if(row is PreCheckHostRow)
+                dataGridView1.Rows.Clear();
+
+                foreach (var row in allRows)
                 {
-                    var r = row as PreCheckHostRow;
-                    if(r.Problem != null || (r.Problem == null && !checkBoxViewPrecheckFailuresOnly.Checked))
+                    var hostRow = row as PreCheckHostRow;
+                    if (hostRow != null)
                     {
-                        AddRowToGridView(r);
+                        if (hostRow.Problem != null || !checkBoxViewPrecheckFailuresOnly.Checked)
+                        {
+                            AddRowToGridView(hostRow);
+                        }
                     }
-                }
-                else
-                {
-                    AddRowToGridView(row);
+                    else
+                    {
+                        AddRowToGridView(row);
+                    }
                 }
             }
         }
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
@@ -58,6 +58,7 @@ namespace XenAdmin.Wizards.PatchingWizard
         public List<Host> SelectedServers = new List<Host>();
         private readonly List<Problem> ProblemsResolvedPreCheck = new List<Problem>();
         private AsyncAction resolvePrechecksAction;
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -954,5 +973,32 @@ private void dataGridView1_CellMouseMove(object sender, DataGridViewCellMouseEve
...
+                    }
+                    else if (row.Key == dataGridView1.Rows.Count)
+                    {
+                        AddRowToGridView(row.Value);
+                    }
+                }
+            }
+            else
+            {
+                foreach (var row in succeessfulCheck.Reverse())
```
jijiang: I quite don't understand .Reverse() here. successfulCheck is a Dictionary, not a List. Is your intention to reverse a list here?

jishengx: This is because the implementation of Dictionary of C#. If successfulCheck has 3 items, and you use `<                foreach (var row in successfulCheck)
                {
                        dataGridView1.Rows.Remove(row.Value);
                }
>` It will delete the item 0 and item2, the item 1 will be left.

MihaelaStoica:
Wouldn't be simpler to store all the rows in a separate list and update the grid from that list when the checkbox is ticked/unticked? For example:
- the list of all rows: `List<DataGridViewRow> allRows`
- when we run the prechecks (so when we create the rows) we add all the rows to `allRows` and only the ones needed (depending on the checkbox) to `dataGridView1.Rows`;
- when the checkbox is ticked/unticked, we clear `dataGridView1.Rows` and add the rows needed (depending on the checkbox) from `allRows`

XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -949,7 +955,28 @@ private void UpdateActionProgress(AsyncAction action)
...
+        private void RefreshDataGridViewList()
+        {
+            dataGridView1.Rows.Clear();
+
+            foreach (var row in allRows)
+            {
+                if(row is PreCheckHostRow)
+                {
+                    var r = row as PreCheckHostRow;
+                    if(r.Problem != null || (r.Problem == null && !checkBoxViewPrecheckFailuresOnly.Checked))
```
MihaelaStoica: `r.Problem == null` in not needed, as it will be always true. So this can be simplified to 
```if (r.Problem != null || !checkBoxViewPrecheckFailuresOnly.Checked)```

File: XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs [hunk 3]
```diff
@@ -949,7 +955,28 @@ private void UpdateActionProgress(AsyncAction action)
...
+            RefreshDataGridViewList();
+        }
+
+        private void RefreshDataGridViewList()
+        {
+            dataGridView1.Rows.Clear();
+
+            foreach (var row in allRows)
+            {
+                if(row is PreCheckHostRow)
```
MihaelaStoica: The preferred way would be:
```
var hostRow = row as PreCheckHostRow;
if (hostRow != null)
....
```

XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -949,7 +955,28 @@ private void UpdateActionProgress(AsyncAction action)
 
         private void checkBox1_CheckedChanged(object sender, EventArgs e)
         {
-            RefreshRechecks();
+            RefreshDataGridViewList();
+        }
+
+        private void RefreshDataGridViewList()
+        {
+            dataGridView1.Rows.Clear();
```
MihaelaStoica: dataGridView1.Rows is used under a lock in other places (see the `AddRowToGridView` function). Is this lock really needed? If so, then we should also put this line inside a lock.

PR #2090, opened 2018-06-06
CA-290316: Handle hotfix homogeneity precheck in mixed pool

jijiang:

### DIFF
--- a/XenAdmin/Diagnostics/Checks/ServerSelectionCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/ServerSelectionCheck.cs
@@ -61,6 +61,9 @@ namespace XenAdmin.Diagnostics.Checks
             if (update == null || !update.EnforceHomogeneity()) 
                 return null;
 
+            if (!pool.IsPoolFullyUpgraded())
+                return new MixedPoolServerSelectionWarning(this, pool);
+            
             if (pool.Connection.Cache.Hosts.Any(h => !update.AppliedOn(h) && !selectedServers.Contains(h)))
                 return new ServerSelectionProblem(this, pool);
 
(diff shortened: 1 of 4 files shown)


XenAdmin/Diagnostics/Checks/ServerSelectionCheck.cs
```diff
@@ -61,7 +61,8 @@ protected override Problem RunCheck()
             if (update == null || !update.EnforceHomogeneity()) 
                 return null;
 
-            if (pool.Connection.Cache.Hosts.Any(h => !update.AppliedOn(h) && !selectedServers.Contains(h)))
+            //The check is hit when there is a patch-non-applied host, whose version is equal to the selected host but not selected
+            if (pool.Connection.Cache.Hosts.Any(h => !update.AppliedOn(h) && h.ProductVersion() == selectedServers[0].ProductVersion() && !selectedServers.Contains(h)))
```
kc284: Shouldn't we be using the method Helpers.productVersionCompare() for the comparison instead of comparing the strings?

XenAdmin/Diagnostics/Checks/ServerSelectionCheck.cs
```diff
@@ -61,7 +61,8 @@ protected override Problem RunCheck()
             if (update == null || !update.EnforceHomogeneity()) 
                 return null;
 
-            if (pool.Connection.Cache.Hosts.Any(h => !update.AppliedOn(h) && !selectedServers.Contains(h)))
+            //The check is hit when there is a patch-non-applied host, whose version is equal to the selected host but not selected
+            if (pool.Connection.Cache.Hosts.Any(h => !update.AppliedOn(h) && Helpers.productVersionCompare(h.ProductVersion(), selectedServers[0].ProductVersion()) == 0 && !selectedServers.Contains(h)))
```
MihaelaStoica: `selectedServers[0]` is not the right server to use here, because `selectedServers` can contain servers from other pools and these pools can be of different version if, for example, the selected update can be applied on both versions.
I think that we need to skip this check if the `pool`has mixed version hosts. so something like
```
var masterVersion = Helpers.HostProductVersion(Host);
if (pool.Connection.Cache.Hosts.Any(h => Helpers.productVersionCompare(Helpers.HostProductVersion(h), masterVersion) != 0))
    return null;
```

PR #2092, opened 2018-06-07
CA-289629: prompt host may need to be rebooted to enable SR-IOV

jishengx:

### DIFF
--- a/XenAdmin/Wizards/NewNetworkWizard.cs
+++ b/XenAdmin/Wizards/NewNetworkWizard.cs
@@ -209,7 +209,7 @@ namespace XenAdmin.Wizards
             var dlg = new ThreeButtonDialog(
                 new ThreeButtonDialog.Details(
                     SystemIcons.Warning, 
-                    Messages.SRIOV_NETWORK_CREATE_WILL_DISTURB_CONNECTION),
+                    Messages.SRIOV_NETWORK_CREATE_WARNING),
                  new ThreeButtonDialog.TBDButton(Messages.SRIOV_NETWORK_CREATE, DialogResult.OK),
                 ThreeButtonDialog.ButtonCancel);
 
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/NewNetworkWizard.cs
```diff
@@ -209,7 +209,7 @@ private DialogResult ShowSriovCreationWarning()
             var dlg = new ThreeButtonDialog(
                 new ThreeButtonDialog.Details(
                     SystemIcons.Warning, 
-                    Messages.SRIOV_NETWORK_CREATE_WILL_DISTURB_CONNECTION),
+                    Messages.SRIOV_NETWORK_CREATE_WARNING),
```
kc284: Creating and launching dialogs should be done within a `using` block so we don't leak resources.

XenModel/Messages.resx
```diff
@@ -11113,17 +11113,16 @@ Do you want to connect to the pool master '{1}'?</value>
...
   </data>
   <data name="SRIOV_NETWORK_CREATE" xml:space="preserve">
     <value>Create &amp;SR-IOV anyway</value>
   </data>
-  <data name="SRIOV_NETWORK_CREATE_WILL_DISTURB_CONNECTION" xml:space="preserve">
-    <value>Creating this SR-IOV network will have effects on network connection status and [XenCenter] connections to the pool will temporarily be disturbed</value>
+  <data name="SRIOV_NETWORK_CREATE_WARNING" xml:space="preserve">
+    <value>Creating this SR-IOV network will have effects on network connection status and [XenCenter] connections to the pool will temporarily be disturbed.
+
+Host(s) may need to be rebooted to enable SR-IOV network.</value>
```
kc284: I think we need to rephrase the message because it is a bit vague.

kc284:
I'm going to merge this now because it has been inactive for too long. I'll fix the dialog launching in a new PR.

PR #2095, opened 2018-06-12
CP-28286: Support resume from a partially upgraded pool via Update wi…

jijiang:
…zard

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -429,8 +429,8 @@ namespace XenAdmin.Core
 
                 foreach (XenServerVersion xenServerVersion in serverVersions)
                 {
-                    XenServerVersion patchApplicableVersion = xenServerVersion;
-                    List<XenServerPatch> patches = xenServerPatches.FindAll(patch => patchApplicableVersion.Patches.Contains(patch));
+                    XenServerVersion version = xenServerVersion;
+                    List<XenServerPatch> patches = xenServerPatches.FindAll(patch => version.Patches.Contains(patch));
 
                     if (patches.Count == 0)
                         continue;
@@ -452,7 +452,7 @@ namespace XenAdmin.Core
 
                         XenServerPatch serverPatch = xenServerPatch;
 
-                        var noPatchHosts = hosts.Where(host => PatchCanBeInstalledOnHost(serverPatch, host, patchApplicableVersion));
+                        var noPatchHosts = hosts.Where(host => PatchCanBeInstalledOnHost(serverPatch, host, version));
         
                         if (noPatchHosts.Count() == hosts.Count)
                             alert.IncludeConnection(xenConnection);
@@ -465,12 +465,12 @@ namespace XenAdmin.Core
             return alerts;
         }
 
-        private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host host, XenServerVersion version)
+        private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host host, XenServerVersion patchApplicableVersion)
         {
             Debug.Assert(serverPatch != null);
             Debug.Assert(host != null);
 
-            if (Helpers.productVersionCompare(version.Version.ToString(), host.ProductVersion()) != 0)
+            if (Helpers.productVersionCompare(patchApplicableVersion.Version.ToString(), host.ProductVersion()) != 0)
                 return false;
 
             // A patch can be installed on a host if:
@@ -617,11 +617,6 @@ namespace XenAdmin.Core
         {
             Debug.Assert(conn != null);
             Debug.Assert(alert != null);
-
-            // the pool has to be homogeneous
-            //var version = GetCommonServerVersionOfHostsInAConnection(conn, XenServerVersions);
-            //if (version == null)
-            //    return null;
 
             var minimalPatches = new List<XenServerPatch> {alert.Patch};
 
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -416,15 +416,21 @@ namespace XenAdmin.Core
                 if (master == null || pool == null)
                     continue;
 
-                var serverVersions = GetServerVersions(master, xenServerVersions);
+                var serverVersions = new List<XenServerVersion>();
+                foreach (Host host in hosts)
+                {
+                    var serverVersion = GetServerVersions(host, xenServerVersions);
+                    serverVersions.AddRange(serverVersion);
+                }
+                serverVersions = serverVersions.Distinct().ToList();
 
                 if (serverVersions.Count == 0)
                     continue;
 
                 foreach (XenServerVersion xenServerVersion in serverVersions)
                 {
-                    XenServerVersion version = xenServerVersion;
-                    List<XenServerPatch> patches = xenServerPatches.FindAll(patch => version.Patches.Contains(patch));
+                    XenServerVersion patchApplicableVersion = xenServerVersion;
+                    List<XenServerPatch> patches = xenServerPatches.FindAll(patch => patchApplicableVersion.Patches.Contains(patch));
 
                     if (patches.Count == 0)
                         continue;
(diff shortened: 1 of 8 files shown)


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -640,7 +658,7 @@ private void dataGridViewHosts_CheckBoxClicked(object sender, EventArgs e)
         {
             foreach (PatchingHostsDataGridViewRow row in dataGridViewHosts.Rows)
             {
-                var host = row.Tag as Host ?? Helpers.GetMaster(row.Tag as Pool);
+                var host = row.Tag as Host;
```
MihaelaStoica: The original code is still needed here, because we also need to call EnableRow for the pool rows. This EnableRow is used to dissallow pre and post Ely servers to be selected at the same time (see commit 33979bb1f2053c9c932ffa52be3c0aa5146dc72a

XenAdmin/Core/Updates.cs
```diff
@@ -459,7 +465,7 @@ private static bool IsSuitableForXenCenterAlert(XenCenterVersion toUse, Version
             return alerts;
         }
 
-        private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host host)
+        private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host host, XenServerVersion version)
```
MihaelaStoica: We should rename the "version" parameter to something like patchApplicableVersion.

XenAdmin/Core/Updates.cs
```diff
@@ -472,6 +478,9 @@ private static bool PatchCanBeInstalledOnHost(XenServerPatch serverPatch, Host h
             var appliedUpdates = host.AppliedUpdates();
             var appliedPatches = host.AppliedPatches();
 
+            if (Helpers.productVersionCompare(version.Version.ToString(), host.ProductVersion()) != 0)
```
MihaelaStoica: This can be moved above the line 477 (so before `bool elyOrGreater = Helpers.ElyOrGreater(host);`)

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -96,7 +96,9 @@ protected override void PageLoadedCore(PageLoadedDirection direction)
         {
             try
             {
-                poolSelectionOnly = WizardMode == WizardMode.AutomatedUpdates || SelectedUpdateAlert != null || FileFromDiskAlert != null;
+                List<IXenConnection> xenConnections = ConnectionsManager.XenConnectionsCopy;
+                bool hasMixedPool = xenConnections.Any(xc => Helpers.GetPool(xc) != null && !Helpers.GetPool(xc).IsPoolFullyUpgraded());
```
MihaelaStoica: Trying to avoid calling `GetPool` twice, this can be written as:
`hasMixedPool = xenConnections.Select(Helpers.GetPool).Any(p => p != null && !p.IsPoolFullyUpgraded());`

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -148,6 +146,26 @@ protected override void PageLoadedCore(PageLoadedDirection direction)
...
                     }
+
+                    if (hasPool)
+                    {
+                        if (poolIndex == -1)
+                            continue;
+                        var poolRow = (PatchingHostsDataGridViewRow) dataGridViewHosts.Rows[poolIndex];
+                        var masterRow = (PatchingHostsDataGridViewRow) dataGridViewHosts.Rows[poolIndex + 1];
+                        poolRow.Enabled = false;
+                        for (int index = poolIndex + 1; index < dataGridViewHosts.RowCount; index ++)
```
MihaelaStoica: We should avoid iterating through the host rows again and instead enable/disable the pool row in the previous loop (when we create the host rows):
- when we add the pool row, disable it, `poolRow.Enabled = false`;
- when we add a host row that is enabled, we should also enable the pool row; 
- after we added all the host rows and the pool row is still disabled, then set its ToolTipText to the master's tooltip text (master row is `dataGridViewHosts.Rows[dataGridViewHosts.Rows.IndexOf(poolRow) + 1]`)

XenModel/Messages.resx
```diff
@@ -11974,6 +11974,9 @@ Check your settings and try again.</value>
   <data name="UPDATES_WIZARD_LOCAL_STORAGE" xml:space="preserve">
     <value>{0}: The VM '{1}' uses local storage and cannot be migrated.</value>
   </data>
+  <data name="UPDATES_WIZARD_MASTER_VERSION_NOT_COMPATIBLE_PROBLEM" xml:space="preserve">
+    <value>{0}: Master version is not compatible for this update. Please include master in the selection and try again.</value>
```
MihaelaStoica: This message is conflicting, it says that the master's version is incompatible with the update but suggests to include the master in the selection.
How about: "The update cannot be applied on slaves without being applied on the master" or "The update needs to also be applied on the master"

XenAdmin/Core/Updates.cs
```diff
@@ -416,15 +416,21 @@ private static bool IsSuitableForXenCenterAlert(XenCenterVersion toUse, Version
...
+                serverVersions = serverVersions.Distinct().ToList();
 
                 if (serverVersions.Count == 0)
                     continue;
 
                 foreach (XenServerVersion xenServerVersion in serverVersions)
                 {
-                    XenServerVersion version = xenServerVersion;
-                    List<XenServerPatch> patches = xenServerPatches.FindAll(patch => version.Patches.Contains(patch));
+                    XenServerVersion patchApplicableVersion = xenServerVersion;
```
MihaelaStoica: When I suggested the rename of the "version" parameter I was referring to the definition of the `PatchCanBeInstalledOnHost` method, not this local variable.

XenAdmin/Diagnostics/Checks/ServerSelectionCheck.cs
```diff
@@ -51,16 +54,47 @@ public ServerSelectionCheck(Pool pool, Pool_update update, List<Host> selectedSe
...
 
         protected override Problem RunCheck()
         {
             if (!Host.IsLive())
                 return new HostNotLiveWarning(this, Host);
 
+            if (isNewVersionUpdate)
+            {
+                //Must select all hosts in a homogeneous pool
+                if (pool.IsPoolFullyUpgraded() && pool.Connection.Cache.Hosts.Any(h => !selectedServers.Contains(h)))
```
MihaelaStoica: Suggestion (to avoid multiple IsPoolFullyUpgraded calls):
```
if (pool.IsPoolFullyUpgraded())
{
     //Must select all hosts in a homogeneous pool
    if (pool.Connection.Cache.Hosts.Any(h => !selectedServers.Contains(h)))
        return new ServerSelectionProblem(this, pool);
}
else
{
    //For mixed pool, precheck blocks update to a new version which is higher than master without selecting it
    //Otherwise, skip the precheck for a mixed pool and issue warning. Because the version update may not be compatible to all hosts.
    if (Helpers.productVersionCompare(alert.NewServerVersion.Version.ToString(), Host.ProductVersion()) > 0 && !selectedServers.Contains(Host))
         return new MasterVersionNotCompatibleProblem(this, pool);
    else
        return new MixedPoolServerSelectionWarning(this, pool);
}
```

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -561,7 +599,7 @@ public bool ApplyUpdatesToNewVersion
             }
         }
 
-        public UpdateType SelectedUpdateType { private get; set; }
+        public static UpdateType SelectedUpdateType { private get; set; }
```
MihaelaStoica: This property doesn't need to be static

PR #2097, opened 2018-06-14
Revert "Disable creation of SR-IOV network if the feature is disabled."

cheng-z:
As SRIOV will release in Lima, remove the experimental flag and roll back the change in XenCenter for the check.
This reverts commit 3fc14c53441f1f4932dad7f00201b2f2f9a6d1b2.


MihaelaStoica:
Don't we still need the experimental feature check in XenCenter for Kolkata servers (where the feature is still experimental)?

cheng-z:
Oh, you are right. So this change needs to be kept.
No change needs to be make in XenCenter.

Thanks @MihaelaStoica

PR #2098, opened 2018-06-14
CP-28426: RPU wizard: add user option for automatically applying updates

MihaelaStoica:
This is only adding the user option in the wizard. The implementation of this feature will be submitted in a future PR.

### DIFF
--- a/XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizard_UpgradeMode.resx
+++ b/XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizard_UpgradeMode.resx
@@ -112,23 +112,23 @@
     <value>2.0</value>
   </resheader>
   <resheader name="reader">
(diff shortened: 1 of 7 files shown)


XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizard_UpgradeMode.resx
```diff
@@ -286,15 +292,114 @@
...
+    <value>True</value>
+  </data>
+  <data name="applyUpdatesCheckBox.ImeMode" type="System.Windows.Forms.ImeMode, System.Windows.Forms">
+    <value>NoControl</value>
+  </data>
+  <data name="applyUpdatesCheckBox.Location" type="System.Drawing.Point, System.Drawing">
+    <value>6, 274</value>
+  </data>
+  <data name="applyUpdatesCheckBox.Margin" type="System.Windows.Forms.Padding, System.Windows.Forms">
+    <value>6, 5, 3, 3</value>
```
kc284: I think I would reset the margins to the default `3,3,3,3` so the checkbox is closer to the label above it, and its left edge is aligned with the left edge of the radio buttons (which do hang slightly outside the text of the labels).

PR #2100, opened 2018-06-19
CP-28279: RPU wizard: change the wizard to apply hotfixes after upgrading

MihaelaStoica:


kc284:
Looks ok to me - might it be worth making the base AutomatedUpdatesBasePage class and the BlurbText abstract?

PR #2101, opened 2018-06-19
The check for hosts that are unlicensed for automated updates should …

kc284:
…be on all pool hosts as there maybe individual restrictions (for example via a feature flag).


jijiang:
looks good.

PR #2102, opened 2018-06-19
CP-28280: Add CFU availability precheck to RPU wizard

jijiang:

### DIFF
--- a/XenAdmin/Diagnostics/Checks/CfuAvailabilityCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/CfuAvailabilityCheck.cs
@@ -29,47 +29,42 @@
  * SUCH DAMAGE.
  */
 
-using System.Collections.Generic;
-using XenAPI;
+using XenAdmin.Core;
 using XenAdmin.Diagnostics.Problems;
+using XenAdmin.Diagnostics.Problems.UtilityProblem;
 
 namespace XenAdmin.Diagnostics.Checks
 {
-    public abstract class Check
+    class CfuAvailabilityCheck : Check
     {
-        protected Check(Host host)
+
+        protected override Problem RunCheck()
         {
-            _host = host;
+            var action = Updates.CreateDownloadUpdatesXmlAction(Updates.CheckForUpdatesUrl);
+
+            try
+            {
+                action.RunExternal(action.Session);
+            }
+            catch
+            {
+                log.WarnFormat("Could not download check for update file.");
+            }
+
+            if (!action.Succeeded)
+                return new CfuNotAvailableProblem(this);
+            else
+                return null;
         }
 
-        protected Check(){ }
-
-        protected abstract Problem RunCheck();
-
-        // By default, most Checks return zero or one Problems: but a
-        // Check can override this to return multiple Problems
-        public virtual List<Problem> RunAllChecks()
+        public override string Description
         {
-            var list = new List<Problem>(1);
-            var problem = RunCheck();
-            if (problem != null)
-                list.Add(problem);
-            return list;
+            get
+            {
+                return Messages.CFU_STATUS_CHECK_DESCRIPTION;
+            }
         }
 
-        public abstract string Description{ get;}
-
-        public virtual string SuccessfulCheckDescription 
-        {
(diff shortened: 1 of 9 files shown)


XenAdmin/Diagnostics/Checks/CfuAvailabilityCheck.cs
```diff
@@ -0,0 +1,77 @@
...
+using XenAdmin.Diagnostics.Problems;
+using XenAdmin.Diagnostics.Problems.HostProblem;
+using XenAdmin.Diagnostics.Problems.UtilityProblem;
+
+namespace XenAdmin.Diagnostics.Checks
+{
+    class CfuAvailabilityCheck : Check
+    {
+
+        public CfuAvailabilityCheck()
```
kc284: Minor comment: this constructor is not necessary since you have added the parameterless constructor in the base class.
The `using XenAdmin.Diagnostics.Problems.HostProblem;` is not necessary either.

XenModel/Messages.resx
```diff
@@ -2527,6 +2527,9 @@ Do you want to assign it to the schedule '{2}' instead?</value>
   <data name="CHECKING_CANEVACUATE_STATUS" xml:space="preserve">
     <value>Checking VM migration status</value>
   </data>
+  <data name="CHECKING_CHECK_FOR_UPDATE_STATUS" xml:space="preserve">
```
kc284: We don't say "check a check", but rather "run a check" or "check something for blah blah", so you would rename this to `CHECKING_UPDATE_FILE_STATUS` and the message to `Checking update file status`. However, I don't think we need to expose this level of detail, but rather say `Searching for available updates`. Similarly CHECK_FOR_UPDATE_AVAILABILITY_CHECK_DESCRIPTION=`Updates availability check`

MihaelaStoica: Please note that we are not exactly checking for "updates availability", because, as this is used in the RPU, we don't know the version the host will be upgraded to, so can't check if any updates are available on that version. This check purely checks if the CFU file can be downloaded, so if there is internet connection and the CFU website is reachable.

XenAdmin/Diagnostics/Checks/CfuAvailabilityCheck.cs
```diff
@@ -0,0 +1,77 @@
...
+            : base()
+        {
+        }
+
+        protected override Problem RunCheck()
+        {
+            var action = Updates.CreateDownloadUpdatesXmlAction(true, true, true, Updates.CheckForUpdatesUrl);
```
kc284: Please call with the first 3 params set to false to avoid the specific version checks (as they are unknown at this stage).

kc284: It might be more elegant if we change these three constructor parameters to optional defaulting to false (reordering needed), in which case here you can call it without them (please also make the checkForUpdatesUrl param required; it's pointless to have it optional since we assert it's not null).

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -142,6 +142,14 @@ protected override List<CheckGroup> GenerateChecks(Pool_patch patch)
         {
             var groups = new List<CheckGroup>();
 
+            //Check for update file check
+            var cfuCheck = new List<Check>();
```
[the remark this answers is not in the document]
jijiang: yes

XenModel/Messages.resx
```diff
@@ -11789,6 +11795,9 @@ Verify that the file is a valid {1} export.</value>
   <data name="UPDATED_PROPERTIES" xml:space="preserve">
     <value>Updating complete.</value>
   </data>
+  <data name="UPDATE_SERVER_NOT_REACHABLE" xml:space="preserve">
+    <value>Update server unreachable or network connectivity error. Please deselect the updates application checkbox on previous page and try again.</value>
```
kc284: `on the previous page`

PR #2103, opened 2018-06-20
CA-289627: SR-IOV on NIC tab is confusing

jishengx:

### DIFF
--- a/XenAdmin/TabPages/NICPage.cs
+++ b/XenAdmin/TabPages/NICPage.cs
@@ -158,13 +158,10 @@ namespace XenAdmin.TabPages
                 {
                     foreach (PIF PIF in host.Connection.ResolveAll(host.PIFs))
                     {
-                        if (!PIF.IsPhysical() && !PIF.IsSriovLogicalPIF())
+                        if (!PIF.IsPhysical())
                             continue;
 
                         RegisterPIFEventHandlers(PIF);
-
-                        if (PIF.IsSriovLogicalPIF())
-                            continue;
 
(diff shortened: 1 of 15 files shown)


XenAdmin/TabPages/NICPage.cs
```diff
@@ -163,6 +163,9 @@ private void updateList()
 
                         RegisterPIFEventHandlers(PIF);
 
+                        if (PIF.IsSriovLogicalPIF())
```
kc284: Shouldn't this move above the event registration? I wouldn't think we need to register events for skipped PIFs.

XenModel/Messages.resx
```diff
@@ -11124,6 +11124,9 @@ Do you want to connect to the pool master '{1}'?</value>
 
 You may need to reboot your server(s) to enable SR-IOV network.</value>
   </data>
+  <data name="SRIOV_NETWORK_SHOULD_BE_CREATED" xml:space="preserve">
+    <value>Yes (SR-IOV network should be created on this NIC to enable SR-IOV)</value>
```
kc284: Let's make this a bit more user oriented: `Yes (To enable SR-IOV create an SR-IOV network on this NIC)`

kc284:
Obviously, showing the new messages inline makes them more visible than putting them on a tooltip, but I'm a bit concerned about the overall appearance (especially in Japanese) because they are intended for an autosizing column and they are a bit long. One suggestion could be changing the sizing mode of that column - any others?

PR #2104, opened 2018-06-20
CP-28279 + CP-28283: Change the RPU wizard to apply hotfixes after upgrading + Retry mechanism

MihaelaStoica:
(The changes to the upgrade page of the RPU wizard to inherit from the common base class also fixes CP-28283)


kc284:
This I missed in the PR that added the page PatchingWizard_AutomatedUpdatesPage and it broke the build: AutoScaleMode should be Dpi (it's correct in the base class).

PR #2105, opened 2018-06-21
CP-28425: Add automated updates license precheck in RPU wizard

jijiang:

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -2020,6 +2020,18 @@ Apply update '{0}' to server '{1}'...
   <data name="AUTHORIZING_USER" xml:space="preserve">
     <value>Authorizing User</value>
   </data>
+  <data name="AUTOMATED_UPDATES_LICENSE_CHECK_DESCRIPTION" xml:space="preserve">
+    <value>License status for automated updates</value>
+  </data>
+  <data name="AUTOMATED_UPDATES_UNLICENSED_WARNING_MORE_INFO" xml:space="preserve">
+    <value>This warning may have various causes. Please consider one of the following solutions and try again:
+
+- License any unlicensed hosts in the pool.
+
+- Ensure automated updates are not restricted on any hosts in the pool.
+
+- Deselect the updates application checkbox on the previous page. [XenCenter] will upgrade the selected servers to the new version without applying any updates.</value>
+  </data>
   <data name="AUTOSTART" xml:space="preserve">
     <value>Auto-start on server boot</value>
   </data>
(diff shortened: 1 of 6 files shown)


XenModel/Messages.resx
```diff
@@ -2020,6 +2020,18 @@ Apply update '{0}' to server '{1}'...
   <data name="AUTHORIZING_USER" xml:space="preserve">
     <value>Authorizing User</value>
   </data>
+  <data name="AUTOMATED_UPDATES_LICENSE_CHECK_DESCRIPTION" xml:space="preserve">
+    <value>License status for automated updates</value>
+  </data>
+  <data name="AUTOMATED_UPDATES_UNLICENSED_PROBLEM_MORE_INFO" xml:space="preserve">
+    <value>This problem may be caused by various reasons. Please consider one of the following solutions and try again:
```
kc284: Some small improvements to portions of the text:
"This problem may have various causes ... "
" ... Ensure automated updates are not restricted on any hosts in the pool ..." (no need to mention feature flags)
"... [XenCenter] will upgrade the selected servers to the new version without applying any updates. " (it's the application applying the updates, not the servers)

XenAdmin/Diagnostics/Problems/PoolProblem/NotLicensedForAutomatedUpdatesProblem.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public override string Description
+        {
+            get { return string.Format(Messages.UPGRADEWIZARD_PROBLEM_HOST_UNLICENSED_FOR_AUTOMATED_UPDATES, Pool); }
+        }
+
+        protected override AsyncAction CreateAction(out bool cancelled)
+        {
+            Program.Invoke(Program.MainWindow, delegate()
```
kc284: Invocation is not needed because this method is called on the UI thread (by SolveImmediately which is called by UI actions; the name is a bit unfortunate because it only creates the action and does not run it). Even if it were needed, design-wise it shouldn't be up to the CreateAction method to invoke, but rather to the method's caller.
There are some more derivatives of the Problem class that do this, those should be corrected too at some point.

XenAdmin/Diagnostics/Problems/PoolProblem/NotLicensedForAutomatedUpdatesProblem.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Drawing;
+using XenAdmin.Actions;
+using XenAdmin.Diagnostics.Checks;
+using XenAdmin.Dialogs;
+using XenAPI;
+
+
+namespace XenAdmin.Diagnostics.Problems.PoolProblem
+{
+    class NotLicensedForAutomatedUpdatesProblem : PoolProblem
```
kc284: As discussed, please derive from Warning rather than Problem.

PR #2114, opened 2018-06-28
declare VMs with SR-IOV VFs as non-agile

jishengx:

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -9079,6 +9079,9 @@ It is strongly recommended that you Cancel and apply the latest version of the p
   <data name="NOT_AGILE_UNKOWN" xml:space="preserve">
     <value>The VM is not agile. Restart cannot be guaranteed.</value>
   </data>
+  <data name="NOT_AGILE_VM_HAS_SRIOV_VIF" xml:space="preserve">
+    <value>The VM has one or more SR-IOV VFs attached. Restart cannot be guaranteed.</value>
+  </data>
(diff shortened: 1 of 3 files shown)


XenModel/Messages.resx
```diff
@@ -9079,6 +9079,9 @@ It is strongly recommended that you Cancel and apply the latest version of the p
   <data name="NOT_AGILE_UNKOWN" xml:space="preserve">
     <value>The VM is not agile. Restart cannot be guaranteed.</value>
   </data>
+  <data name="NOT_AGILE_VM_HAS_SRIOV_VIF" xml:space="preserve">
+    <value>The VM has one or more SR-IOV VF attached. Restart cannot be guaranteed.</value>
```
kc284: Do we use the acronym `VF` elsewhere in XenCenter? I don't recall having seen it much, would it be perhaps clearer for the user if we said the whole thing `SR-IOV virtual functions`?

jishengx: Hi @kc284, we do use VF in SR-IOV Capable column (e.g. Yes(61 VFs remaining) )on NIC Tabpage. Actually, people always combine the SR-IOV with VF concept.

kc284: Ah yes, sorry, I forgot about this! Then please correct only the number: `VFs` instead of `VF` (because we say one or more). Thanks.

jishengx: You're welcome.

PR #2116, opened 2018-06-28
CA-293092: Add HostCheck class to fix check.Host null exception

jijiang:


XenAdmin/Wizards/PatchingWizard/PatchingWizard_AutomatedUpdatesPage.cs
```diff
@@ -130,7 +131,13 @@ protected override void GeneratePlanActions(Pool pool, List<HostPlan> planAction
             }
 
             //add a revert pre-check action for this pool
-            var problemsToRevert = ProblemsResolvedPreCheck.Where(p => hosts.ToList().Select(h => h.uuid).ToList().Contains(p.Check.Host.uuid)).ToList();
+            var problemsToRevert = ProblemsResolvedPreCheck.Where(p =>
+            {
+                var hostCheck = p.Check as HostCheck;
+                if (hostCheck != null)
+                    return hosts.ToList().Select(h => h.uuid).ToList().Contains(hostCheck.Host.uuid);
```
kc284: `ToList()` is not needed here: `hosts.Select(h => h.uuid).Contains(hostCheck.Host.uuid)`

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeUpgradePage.cs
```diff
@@ -103,7 +103,13 @@ protected override void GeneratePlanActions(Pool pool, List<HostPlan> planAction
             }
 
             //add a revert pre-check action for this pool
-            var problemsToRevert = ProblemsResolvedPreCheck.Where(p => hostNeedUpgrade.ToList().Select(h => h.uuid).ToList().Contains(p.Check.Host.uuid)).ToList();
+            var problemsToRevert = ProblemsResolvedPreCheck.Where(p =>
+            {
+                var hostCheck = p.Check as HostCheck;
+                if (hostCheck != null)
+                    return hostNeedUpgrade.ToList().Select(h => h.uuid).ToList().Contains(hostCheck.Host.uuid);
```
kc284: Same here: `return hostNeedUpgrade.Select(h => h.uuid).Contains(hostCheck.Host.uuid);`

XenAdmin/Diagnostics/Checks/HostCheck.cs
```diff
@@ -0,0 +1,50 @@
...
+using XenAPI;
+
+namespace XenAdmin.Diagnostics.Checks
+{
+    public abstract class HostCheck : Check
+    {
+        private readonly Host _host;
+
+        protected HostCheck(Host host)
+        {
```
kc284: I would add a `Debug.Assert(host!=null)` here so we know how we call it.

XenAdmin/Diagnostics/Checks/PoolCheck.cs
```diff
@@ -30,12 +30,11 @@
  */
 
 using XenAdmin.Core;
-using XenAdmin.Diagnostics.Problems;
 using XenAPI;
 
 namespace XenAdmin.Diagnostics.Checks
 {
-    public abstract class PoolCheck : Check
+    public abstract class PoolCheck : HostCheck
```
kc284: PoolCheck should derive from Check, I can't see a need to derive from HostCheck (some of the constructors of its derivatives will need to change; currently they pass in the master, but only because they derive from HostCheck, not because the maste is needed).

PR #2117, opened 2018-06-29
 CA-292908 (fixed progress bar) + CA-293153 (corrected unwind precheck actions) + improvements

kc284:


XenAdmin/Wizards/PatchingWizard/AutomatedUpdatesBasePage.cs
```diff
@@ -201,20 +208,24 @@ private void WorkerProgressChanged(object sender, ProgressChangedEventArgs e)
...
-                UpdateStatusTextBox();
+                UpdateStatus();
             }
         }
 
-        private void UpdateStatusTextBox()
+        private void UpdateStatus()
         {
+            // only increase the progress if the action succeeded
+            var newVal = backgroundWorkers.Sum(b => b.PercentComplete) / backgroundWorkers.Count;
```
MihaelaStoica: When the page is cancelled, this method is called from WorkerCompleted (which is triggered by the cancellation), but `backgroundWorkers.Count` is 0, due to `backgroundWorkers.Clear()` in PageCancelled(), so the progressBar value cannot be calculated in this case.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -642,9 +642,23 @@ public override bool EnableNext()
         public Pool_patch Patch { private get; set; }
         public Pool_update PoolUpdate { private get; set; }
 
+        public List<AsyncAction> GetUnwindChangesActions()
```
MihaelaStoica: Everytime this function is called (when moving away from the prechecks page and also on cancelling the wizard), it creates the unwind actions and the actions that are created on leaving the page are shown in the Events page before they are actually executed. There will also be duplicates when moving back to the prechecks page and then to the next page again (this is possible in the RPU wizard only)

XenAdmin/Wizards/PatchingWizard/PatchingWizard.cs
```diff
@@ -296,8 +284,11 @@ protected override void OnCancel()
         {
             base.OnCancel();
 
-            var subActions = BuildSubActions(PatchingWizard_PrecheckPage.GetUnwindChangesActions,
-                GetRemovePatchActions, GetRemoveVdiActions, GetCleanUpPoolUpdateActions);
+            var subActions = new List<AsyncAction>();
+            subActions.AddRange(GetUnwindChangesActions(PatchingWizard_PrecheckPage.PrecheckProblemsActuallyResolved));
```
MihaelaStoica: AddRange fails with ArgumentNullException if any of these functions (GetUnwindChangesActions, GetRemovePatchActions, etc.) returns null

kc284: oops! 😳

PR #2120, opened 2018-07-02
CA-293099: When installing updates in the automated updates mode (inc…

MihaelaStoica:
…luding new version and RPU), ignore the UPDATE_ALREADY_APPLIED errors and allow the process to continue witth the remaining actions.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
@@ -62,13 +62,25 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
 
             if (mapping != null && (mapping.Pool_patch != null || mapping.Pool_update != null))
             {
-                AddProgressStep(string.Format(Messages.UPDATES_WIZARD_APPLYING_UPDATE, xenServerPatch.Name, host.Name()));
+                try
+                {
+                    AddProgressStep(string.Format(Messages.UPDATES_WIZARD_APPLYING_UPDATE, xenServerPatch.Name,
+                        host.Name()));
 
-                var task = mapping.Pool_patch == null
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
```diff
@@ -62,13 +62,25 @@ protected override void RunWithSession(ref Session session)
...
+                    var task = mapping.Pool_patch == null
+                        ? Pool_update.async_apply(session, mapping.Pool_update.opaque_ref, host.opaque_ref)
+                        : Pool_patch.async_apply(session, mapping.Pool_patch.opaque_ref, host.opaque_ref);
 
-                PollTaskForResultAndDestroy(Connection, ref session, task);
+                    PollTaskForResultAndDestroy(Connection, ref session, task);
+                }
+                catch (Failure f)
+                {
+                    if (f.ErrorDescription.Count > 1 && (f.ErrorDescription[0] == Failure.PATCH_ALREADY_APPLIED || f.ErrorDescription[0] == Failure.UPDATE_ALREADY_APPLIED))
```
kc284: Might it be useful to also AddProgressStep("patch already applied")? Asking because we show the user again the applying update step in line 67.

MihaelaStoica: Shall we do a ReplaceProgressStep? to replace "Installing update" with "already applied"?

kc284: Yes, this sounds much better.
Irrelevant to this PR, but it's just occurred to me that we should probably be doing ReplaceProgress here too: https://github.com/xenserver/xenadmin/blob/REQ-672/XenAdmin/Wizards/PatchingWizard/PlanActions/DownloadPatchPlanAction.cs#L73

PR #2122, opened 2018-07-03
CA-293328: fixed SafeToUpgradeCheck + fixed order of CFU check

kc284:


XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -234,6 +226,13 @@ protected override List<CheckGroup> GenerateChecks(Pool_patch patch)
                     automatedUpdateLicenseChecks));
             }
 
+            //Check for update file availability
+            if (ApplyUpdatesToNewVersion)
```
jijiang: Could we squash this if block with the previous if block (line 219), as they are the same condition? And maybe could first check if c-f-u downloadable and then check license status, based on normal logic order.

XenModel/Messages.resx
```diff
@@ -2557,6 +2557,9 @@ Do you want to assign it to the schedule '{2}' instead?</value>
   <data name="CHECKING_SAFE_TO_UPGRADE" xml:space="preserve">
     <value>Checking partition layout</value>
   </data>
+  <data name="CHECKING_SAFE_TO_UPGRADE_HOST" xml:space="preserve">
+    <value>Host partition layout</value>
```
jijiang: Better to be "Host partition layout check", consistent with other parallel messages.

XenAdmin/Diagnostics/Checks/SafeToUpgradeCheck.cs
```diff
@@ -76,8 +77,8 @@ protected override Problem RunCheck()
         }
 
         public override string Description
-        {
-            get { return string.Empty; }
+        {
+            get { return Messages.CHECKING_SAFE_TO_UPGRADE_HOST; }
```
jijiang: Better to be "SAFE_TO_UPGRADE_HOST_DESCRIPTION", consistent with other parallel macros.

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
kc284: Hmmm, I was wondering about the order, but in the end I thought that the license check may be more important: if, for example, the cfu check fails due to a connectivity error, is there a point in the user taking time to fix it since they will still be unable to install automated updates on their unlicensed pool? @MihaelaStoica could you offer a third opinion on this?

MihaelaStoica: We should do the license check first, as that's what we usually do in other places: if a feature is not available due to the license status, then no other conditions are checked. 

I am thinking that maybe we should combine this checks into one, which will be the "automated updates availability" check, cheking the license +  connectivity. In most cases these checks will succeed, especially the license one (because the RPU is a licensed feature and therefore we won't reach the prechecks page with an unlicensed pool). So having two groups for these checks seems a bit too much.

PR #2124, opened 2018-07-04
CP-28431: Block updates/upgrades from Kolkata if clustering enabled

serencorbett1:


XenAdmin/Diagnostics/Checks/PoolHasKolkataGFS2SR.cs
```diff
@@ -0,0 +1,71 @@
...
+
+using System.Linq;
+using XenAdmin.Core;
+using XenAPI;
+using XenAdmin.Diagnostics.Problems;
+using XenAdmin.Diagnostics.Problems.HostProblem;
+
+namespace XenAdmin.Diagnostics.Checks
+{
+    class PoolHasKolkataGFS2SR : PoolCheck
```
kc284: I'd make the name of the check version agnostic, i.e. PoolHasGFS2SR. Same for the problem further below: PoolHasGfs2Sr or something like this, without v

kc284: Also, in the interest of minimising conflicts when we merge REQ-672 where the checks have been slightly modified, I'd make this check derive from HostCheck (because we check for host liveness before anything else), and make the needed pool a private readonly property.

XenAdmin/Diagnostics/Checks/PoolHasGFS2SR.cs
```diff
@@ -0,0 +1,70 @@
...
+
+
+            return null;
+        }
+
+        public override string Description
+        {
+            get
+            {
+                return string.Format(Messages.GFS2_UPDATE_UPGRADE_CLUSTER_SR_ERROR, Helpers.GetPoolOfOne(Host.Connection));
```
kc284: Aplogies, I didn't notice this the previous time: this messages is not the error, but rather the check description and in most cases it's similar to the group check title; it appears next to the tick when the check is successful and "Hide successful checks" is deselected. We need a new message saying "Clustering status check".

PR #2125, opened 2018-07-06
CA-293158: reworked the logic of selecting servers on the patching wizard

kc284:


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -687,15 +649,44 @@ private void buttonClearAll_Click(object sender, EventArgs e)
 
         private void dataGridViewHosts_CheckBoxClicked(object sender, EventArgs e)
         {
+            OnPageUpdated();
```
MihaelaStoica: The code that was here was meant to prevent the selection on pre and post Ely pools together. This doesn't work anymore.

PR #2126, opened 2018-07-06
REQ-680: Rebranding to Citrix Hypervisor

michael2012z:
Following aspects were covered:
1. Auto-replacement.
2. Connectivity of both XS and CH.
3. Homepage.
4. Installer.


WixInstaller/XenCenter.l10n.diff
```diff
@@ -174,8 +174,8 @@
 @@ -223,7 +247,7 @@
  								  Vital="no"
  								  Name="XenServerHealthCheck"
- 								  DisplayName="[Citrix] [XenServer] Health Check Service"
--								  Description="Enables monitoring of the health of your Citrix XenServer pools by periodically uploading data from the servers to Citrix Insight Services."
+ 								  DisplayName="[Citrix XenServer] Health Check Service"
+-								  Description="Enables monitoring of the health of your Citrix Hypervisor pools by periodically uploading data from the servers to Citrix Insight Services."
```
MihaelaStoica: We should have [Citrix XenServer] here as well (instead of "Citrix Hypervisor")

WixInstaller/XenCenter.wxs
```diff
@@ -223,8 +223,8 @@
 								  Type="ownProcess"
 								  Vital="no"
 								  Name="XenServerHealthCheck"
-								  DisplayName="[Citrix] [XenServer] Health Check Service"
-								  Description="Enables monitoring of the health of your Citrix XenServer pools by periodically uploading data from the servers to Citrix Insight Services."
+								  DisplayName="[Citrix XenServer] Health Check Service"
+								  Description="Enables monitoring of the health of your Citrix Hypervisor pools by periodically uploading data from the servers to Citrix Insight Services."
```
MihaelaStoica: Same as above ([Citrix XenServer])

WixInstaller/XenCenter.wxs
```diff
@@ -293,7 +293,7 @@
             <ComponentRef Id="ProgramFilesShortcut" />
   
            <?if "$(env.Branding)"="XenCenter"?>		
-				<Feature Id="HealthCheckService" Title="[Citrix] [XenServer] Health Check Service" Description="Enables monitoring of the health of your Citrix XenServer pools." Display="expand" Level="1" ConfigurableDirectory="HEALTHCHECKSERVICEDIR" AllowAdvertise="no" InstallDefault="local" Absent="allow" >
+				<Feature Id="HealthCheckService" Title="[Citrix XenServer] Health Check Service" Description="Enables monitoring of the health of your Citrix Hypervisor pools." Display="expand" Level="1" ConfigurableDirectory="HEALTHCHECKSERVICEDIR" AllowAdvertise="no" InstallDefault="local" Absent="allow" >
```
MihaelaStoica: Same as above ([Citrix XenServer])

WixInstaller/XenCenter.l10n.diff
```diff
@@ -201,8 +201,8 @@
              <ComponentRef Id="ProgramFilesShortcut" />
    
             <?if "$(env.Branding)"="XenCenter"?>		
--				<Feature Id="HealthCheckService" Title="[Citrix] [XenServer] Health Check Service" Description="Enables monitoring of the health of your Citrix XenServer pools." Display="expand" Level="1" ConfigurableDirectory="HEALTHCHECKSERVICEDIR" AllowAdvertise="no" InstallDefault="local" Absent="allow" >
-+				<Feature Id="HealthCheckService" Title="[Citrix] [XenServer] Health Check Service" Description="!(loc.HealthCheckDescrShort)" Display="expand" Level="1" ConfigurableDirectory="HEALTHCHECKSERVICEDIR" AllowAdvertise="no" InstallDefault="local" Absent="allow" >
+-				<Feature Id="HealthCheckService" Title="[Citrix XenServer] Health Check Service" Description="Enables monitoring of the health of your Citrix Hypervisor pools." Display="expand" Level="1" ConfigurableDirectory="HEALTHCHECKSERVICEDIR" AllowAdvertise="no" InstallDefault="local" Absent="allow" >
```
MihaelaStoica: Same as above ([Citrix XenServer])

XenAdmin/Wizards/XenWizardBase.resx
```diff
@@ -198,6 +198,84 @@
   <data name="XSHelpButton.Anchor" type="System.Windows.Forms.AnchorStyles, System.Windows.Forms">
     <value>Right</value>
   </data>
+  <data name="XSHelpButton.Image" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
MihaelaStoica: The image shouldn't be embedded in the resx file

XenModel/Messages.resx
```diff
@@ -3055,9 +3055,9 @@ Would you like to carry out the operation at this time?</value>
     <comment>Message shown when user attempts to connect to a second XE Express host from the UI</comment>
   </data>
   <data name="CONNECTION_RESTRICTED_MESSAGE_LONG" xml:space="preserve">
-    <value>Connection to server {0} restricted - a connection to a [Citrix] [XenServer product] Express Edition server ({1}) already exists.
+    <value>Connection to server {0} restricted - a connection to a [Citrix XenServer] Express Edition server ({1}) already exists.
```
MihaelaStoica: We could remove this message (and its references, including the ExpressRestriction and LicenseWarningDialog classes) because it's not applicable anymore.

XenServerHealthCheck/ProjectInstaller.Designer.cs
```diff
@@ -33,9 +33,9 @@ private void InitializeComponent()
             // 
             // XenServerHealthCheckInstaller
             // 
-            this.XenServerHealthCheckInstaller.Description = Branding.COMPANY_NAME_SHORT + " " + Branding.PRODUCT_BRAND + " Health Check";
-            this.XenServerHealthCheckInstaller.DisplayName = Branding.COMPANY_NAME_SHORT + " " + Branding.PRODUCT_BRAND + " Health Check";
-            this.XenServerHealthCheckInstaller.ServiceName = Branding.PRODUCT_BRAND + "HealthCheck";
+            this.XenServerHealthCheckInstaller.Description = "[Citrix XenServer] Health Check";
```
MihaelaStoica: We should use the Branding variables here, because the `designer.cs` files are not getting rebranded, so [Citrix XenServer] will not be replaced when running the branding scripts

devtools/spellcheck/spellcheck.sh
```diff
@@ -39,12 +39,12 @@ src="$( cd -P "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
 
 output=$(/usr/bin/find "$src/XenAdmin" "$src/XenModel" "$src/XenOvfApi" "$src/XenOvfTransport" \( -name \*.cs -o -name \*.resx \) -exec sh "$dir/spellcheck-file.sh" \{\} \;)
 
-output_brand=$(/usr/bin/find "$src/XenAdmin" "$src/XenModel" "$src/XenOvfApi" "$src/XenOvfTransport" \( -name \*.resx \) -exec sh "$dir/spellcheck-brand-file.sh" \{\} \;)
+output_brand=$(/usr/bin/find "$src/XenAdmin" "$src/XenModel" "$src/XenModel/XenAPI" "$src/XenOvfApi" "$src/XenOvfTransport" \( -name \*.resx \) -exec sh "$dir/spellcheck-brand-file.sh" \{\} \;)
```
MihaelaStoica: Why was "$src/XenModel/XenAPI" added here? Is it not covered by "$src/XenModel" (as I assume we run the check in the subfolfers as well)?

WixInstaller/XenCenter.l10n.diff
michael2012z: Good idea.

XenAdmin/Wizards/XenWizardBase.resx
michael2012z: Yes. Interesting, where was it from....

XenServerHealthCheck/ProjectInstaller.Designer.cs
michael2012z: I am going to introduce a "COMPANY_AND_PRODUCT_BRAND" variable in Branding.

For ServiceName, the base code doesn't have any space, but now with "Branding.PRODUCT_BRAND" we will introduce one. Any good idea?

devtools/spellcheck/spellcheck.sh
michael2012z: A mistake :). I thought subfolders were not covered, so I added it to check FriendlyErrorNames.resx. I confirmed, it works on subfolders.

michael2012z:
I realized I made a bad name of the feature branch. It should be feature/REQ-680 instead of REQ-680. 
Will create feature/REQ-680 from REQ-680 after this PR merged.

XenServerHealthCheck/ProjectInstaller.Designer.cs
```diff
@@ -33,9 +33,9 @@ private void InitializeComponent()
             // 
             // XenServerHealthCheckInstaller
             // 
-            this.XenServerHealthCheckInstaller.Description = Branding.COMPANY_NAME_SHORT + " " + Branding.PRODUCT_BRAND + " Health Check";
-            this.XenServerHealthCheckInstaller.DisplayName = Branding.COMPANY_NAME_SHORT + " " + Branding.PRODUCT_BRAND + " Health Check";
-            this.XenServerHealthCheckInstaller.ServiceName = Branding.PRODUCT_BRAND + "HealthCheck";
+            this.XenServerHealthCheckInstaller.Description = Branding.COMPANY_AND_PRODUCT_BRAND + " Health Check";
+            this.XenServerHealthCheckInstaller.DisplayName = Branding.COMPANY_AND_PRODUCT_BRAND + " Health Check";
+            this.XenServerHealthCheckInstaller.ServiceName = string.Join("", Branding.PRODUCT_BRAND.Split(' ')) + "HealthCheck";
```
MihaelaStoica: Please note that the HEALTHCHECKSERVICENAME in https://github.com/xenserver/xenadmin/blob/master/XenModel/Actions/HealthCheck/TransferHealthCheckSettingAction.cs#L47 needs to match this ServiceName, otherwise XenCenter won't find the service to send updates to. 
I don't think we even need to build the ServiceName here, we could just hardcode it in both places to the same value, or we could define a variable in branding and use it in these two places.

XenModel/FriendlyNames.resx
```diff
@@ -1638,10 +1638,10 @@
     <value>Help on Licensing</value>
   </data>
   <data name="Label-host.edition-desktop" xml:space="preserve">
-    <value>XenApp/XenDesktop</value>
+    <value>[Citrix] Virtual Apps/Desktops</value>
```
MihaelaStoica: I think this should be "[Citrix] Virtual Apps and Desktops". Same for the other occurences.

XenServerHealthCheck/ProjectInstaller.Designer.cs
michael2012z: You avoided a big bug here. Thank you. I would hardcode it.
BTW, in WixInstaller/XenCenter.wxs we also specify the service name, what's the relationship between them?

XenModel/Messages.resx
michael2012z: Let's keep it in feature branch. We'd better do the removal on master with a separate commit.

PR #2127, opened 2018-07-06
CA-293311: RPU wizard: In a partially upgraded pool, apply the hotfix…

MihaelaStoica:
…es on the upgraded hosts before upgrading the remaining slaves

### DIFF
--- a/XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
+++ b/XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
@@ -142,22 +142,32 @@ namespace XenAdmin.Wizards.RollingUpgradeWizard
         {
             var groups = new List<CheckGroup>();
 
+            List<Host> hostsToUpgrade = new List<Host>();
+            List<Host> hostsToUpgradeOrUpdate = new List<Host>();
+            foreach (var pool in SelectedPools)
+            {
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -142,22 +142,32 @@ protected override List<CheckGroup> GenerateChecks(Pool_patch patch)
...
+                hostsToUpgradeOrUpdate.Any(host => new Version(Helpers.HostProductVersion(host)) < latestCrVersion.Version))
             {
                 groups.Add(new CheckGroup(Messages.CHECKING_XENCENTER_VERSION, new List<Check> {new XenCenterVersionCheck(null)}));
             }
 
-            //HostMaintenanceModeCheck checks
+            //HostMaintenanceModeCheck checks - for hosts that will be upgraded or updated
             var livenessChecks = new List<Check>();
-            foreach (Host host in SelectedServers)
+            foreach (Host host in hostsToUpgradeOrUpdate)
```
kc284: Can it be a problem that now the check for maintenance mode will be run on the hosts to update too? Previously for those hosts we only checked for liveness.

MihaelaStoica: I will change it so it doesn't check the maintenance mode for th ehosts that only need updated

PR #2130, opened 2018-07-10
CA-292991: Block host joining cluster with cluster network (bonded/unbonded) unless the interface is defined

serencorbett1:

### DIFF
--- a/XenModel/PoolJoinRules.cs
+++ b/XenModel/PoolJoinRules.cs
@@ -34,8 +34,6 @@ using System.Collections.Generic;
 using XenAdmin.Network;
 using XenAPI;
 using System.Linq;
-using System.Runtime.InteropServices.WindowsRuntime;
-using XenAdmin.Actions;
 
 namespace XenAdmin.Core
 {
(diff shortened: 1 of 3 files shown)


XenModel/PoolJoinRules.cs
```diff
@@ -673,5 +687,51 @@ public static bool HasCompatibleManagementInterface(IXenConnection slaveConnecti
...
+            var pifsWithIPAddress = 0;
+
+            List<string> ids = new List<string>();
+
+            if (clusterHostInBond)
+            {
+
+                List<PIF> slaves = new List<PIF>();
+
+                foreach (Bond bond in masterConnection.ResolveAll(clusterHostPif.bond_master_of))
```
kc284: This code re-resolves the bonds for each iteration (you can test this by putting a break point in the ResolveAll method while iterating). Normally the way to do this is to store the resolved bonds in a local variable:
```
var bonds =  masterConnection.ResolveAll(clusterHostPif.bond_master_of);
foreach (var bond in bonds)... etc.
```
Same for line 725.

XenModel/PoolJoinRules.cs
```diff
@@ -34,6 +34,8 @@
 using XenAdmin.Network;
 using XenAPI;
 using System.Linq;
+using System.Runtime.InteropServices.WindowsRuntime;
```
MihaelaStoica: These are not used and can be removed

XenModel/PoolJoinRules.cs
```diff
@@ -673,5 +687,55 @@ public static bool HasCompatibleManagementInterface(IXenConnection slaveConnecti
...
+
+            var clusterHost = masterConnection.Cache.Cluster_hosts.FirstOrDefault();
+
+            if (clusterHost == null)
+            {
+                clusterHostInBond = false;
+                return true;
+            }
+            var clusterHostPif = clusterHost.Connection.Resolve(clusterHost.PIF);
+            clusterHostInBond = clusterHostPif.IsBondNIC();
```
MihaelaStoica: `Resolve` can return null, so we should check clusterHostPif for null before using it.

PR #2132, opened 2018-07-11
CP-28712: Add new message and friendly names

MihaelaStoica:


XenModel/FriendlyNames.resx
```diff
@@ -972,6 +972,12 @@
   <data name="Message.body-alarm" xml:space="preserve">
     <value>Alarm '{0}' was not recognised.</value>
   </data>
+  <data name="Message.body-cluster_host_enable_failed" xml:space="preserve">
+    <value>Server '{0}' failed to join the cluster.</value>
+  </data>
+  <data name="Message.body-cluster_host_fencing" xml:space="preserve">
+    <value> Server '{0}' has fenced.</value>
```
kc284: Fence is intransitive in this context, so `is fenced` or `has been fenced` depending on whether the message appears as a state or as a result of an action; from the conversation on the ticket I get the impression it's the former. Same below: `Server is fenced`.

PR #2133, opened 2018-07-11
CA-292377: Show warning if no cluster network found

jijiang:
1) When create/delete bond, inconsistent cluster network may occur. Show
warning in this case.
2) Refactor UI on cluster network edit page.


MihaelaStoica:
These changes look good, but we have the issue that only pool admins can currently execute `Cluster.get_network`, so until this is resolved we shouldn't merge this PR as it is now.

PR #2134, opened 2018-07-11
CA-293669: In the Updates tab, show the required updates for the slav…

MihaelaStoica:
…es in a partiallly upgraded pool, even when the master's version cannot be found in the c-f-u

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
@@ -279,6 +279,12 @@ namespace XenAdmin.Wizards.PatchingWizard
                         enabled = false;
                         tooltipText = reason;
                     }
+
+                    if (enabled && !Helpers.ElyOrGreater(host) && Helpers.ElyOrGreater(host.Connection)) // host is pre-Ely, but the master is Ely or greater
+                    {
+                        enabled = false;
+                        tooltipText = Messages.PATCHINGWIZARD_SELECTSERVERPAGE_CANNOT_INSTALL_UPDATE_MASTER_POST_7_0;
+                    }
                     return enabled;
 
                 case UpdateType.ISO:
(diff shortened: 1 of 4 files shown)


MihaelaStoica:
This first commit solves problem 1 from CA-293669

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -279,6 +279,12 @@ private bool CanEnableRowNonAutomated(Host host, out string tooltipText)
                         enabled = false;
                         tooltipText = reason;
                     }
+
+                    if (enabled && !Helpers.ElyOrGreater(host) && Helpers.ElyOrGreater(host.Connection)) // host is pre-Ely, but the master is Ely or greater
```
kc284: I think we may be able to move this check further up; do we need to check whether IsHostAmongApplicable if we are to disable it based on version?
Also, now that I'm looking at this code again, I wonder why we're not returning at line 274; it was like this in the original code https://github.com/xenserver/xenadmin/commit/c11b692958c3223c7409b5644557f24d21eefa02#diff-95fe45652d43f1b59b43a73fcab55f0bL306, but now that's untangled something doesn't look right.
Do you think we should be doing the following?
```
case UpdateType.Existing:
    if (Helpers.ElyOrGreater(host))
    {
        tooltipText = Messages.PATCHINGWIZARD_SELECTSERVERPAGE_PATCH_NOT_APPLICABLE;
        return false;
    }
    if (Helpers.ElyOrGreater(host.Connection)) // host is pre-Ely, but the master is Ely or greater
    {
        tooltipText = Messages.PATCHINGWIZARD_SELECTSERVERPAGE_CANNOT_INSTALL_UPDATE_MASTER_POST_7_0;
        return false;
    }
    return IsHostAmongApplicable(host, out tooltipText);
```

MihaelaStoica: I was wondering too why we are not returning in the first `if` block. So the proposed change looks good there.
I added the new check at the end because I thought that if the patch is not applicable, for example because it is already applied, then we want to show that reason. And show the new reason only if it is applicable (i.e it is applicable on the slave, but we cannot actually apply it because the master's version)

PR #2140, opened 2018-07-12
CA-293764 + some modifications

kc284:


XenModel/Messages.resx
```diff
@@ -11823,7 +11817,7 @@ Verify that the file is a valid {1} export.</value>
     <value>Updated performance monitoring configuration</value>
   </data>
   <data name="UPDATED_PROPERTIES" xml:space="preserve">
-    <value>Updating complete.</value>
+    <value>{0}'s properties were saved.</value>
```
MihaelaStoica: Can't we avoid the `'s`? It might look a bit strange if the object's name is multiple words, for example a VM called `Red Hat Enterprise Linux 6.x (latest)`

PR #2143, opened 2018-07-16
CA-293863: Do not update the done and in-progress actions on the worker thread.

kc284:
Also, do not log error every time the output panel is updated.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/AutomatedUpdatesBasePage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/AutomatedUpdatesBasePage.cs
@@ -189,30 +189,35 @@ namespace XenAdmin.Wizards.PatchingWizard
 
         private void WorkerProgressChanged(object sender, ProgressChangedEventArgs e)
         {
-            var actionsWorker = sender as UpdateProgressBackgroundWorker;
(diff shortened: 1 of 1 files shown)


XenAdmin/Wizards/PatchingWizard/AutomatedUpdatesBasePage.cs
```diff
@@ -189,30 +189,33 @@ private void StartNewWorker(string poolName, List<HostPlan> planActions, List<Pl
...
                     else
                     {
-                        if (!actionsWorker.DoneActions.Contains(action))
-                            actionsWorker.DoneActions.Add(action);
-                        actionsWorker.InProgressActions.Remove(action);
+                        if (!bgw.DoneActions.Contains(action))
+                            bgw.DoneActions.Add(action);
+                        bgw.InProgressActions.Remove(action);
+
+                        if (!failedWorkers.Contains(bgw))
```
MihaelaStoica: I think we should only add the worker to the failedWorkers if action.Error is not null (so the worker had failed)

kc284: oops! 😳

PR #2146, opened 2018-07-18
CA-293923: Use a NetworkAvailabilityChanged event to detect changes i…

MihaelaStoica:
…n the network connectivity and mark the download as failed when the network becomes unavailable

### DIFF
--- a/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
+++ b/XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
@@ -35,6 +35,7 @@ using System.ComponentModel;
 using System.Threading;
 using System.IO;
 using System.Linq;
+using System.Net.NetworkInformation;
 using XenCenterLib.Archive;
 
 namespace XenAdmin.Actions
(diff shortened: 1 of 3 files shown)


jijiang:
good. Merge it.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -86,72 +87,81 @@ public class DownloadAndUnzipXenServerPatchAction : AsyncAction, IByteProgressAc
...
-                        client.DownloadProgressChanged -= client_DownloadProgressChanged;
-                        client.DownloadFileCompleted -= client_DownloadFileCompleted;
+                        // logging only, it will retry again.
+                        log.ErrorFormat(
+                            "Error while downloading from '{0}'. Number of errors so far (including this): {1}. Trying maximum {2} times.",
+                            address, errorCount, MAX_NUMBER_OF_TRIES);
+                        log.Error(patchDownloadError ?? new Exception(Messages.ERROR_UNKNOWN));
+
+                        // wait for some randomly increased amount of time after each retry
+                        nextSleepMs += random.Next(5000);
```
kc284: This was not touched by this PR, but it's just occurred me that the sleep time is set a bit weirdly. I suspect the intention might have been to increase it by a random number of seconds, not milliseconds, because nextSleepMs starts at 0 and random.Next will not necessary give values in the thousands, it may as well give 0 or a single digit number, which practically amounts to immediate retry (probably pointless?).
I think it should be `nextSleepMs += random.Next(1,6) * 1000;`, which increases the sleep time by a value in [1,2,3,4,5]sec, or, if we don't want the sleep time to potentially climb up to 25sec by the last retry, `nextSleepMs = random.Next(1,x) * 1000;` where x can be 6 or anything appropriate.

PR #2151, opened 2018-07-23
CA-289960: If a VM does not acknowledge the need to shut down on host reboot...

kc284:
include the VM's name in the message to help the user identify it.
Also, we have more instances of "shut down" than "shutdown".


kc284:
One issue I see with this is that the message is still not very helpful if there are several VMs with the same name; on the other hand we normally show the VMs in the messages by name, not by uuid, hence I went with this solution. Any opinion on this?

MihaelaStoica:
We would have the same problem in various other places if there were several VMs with the same name, but I think that's acceptable. We could add a log entry and include the uuid there, to help the user identify the VM.

PR #2154, opened 2018-07-25
CA-272127 + CA-294365: fixes on the users tab

kc284:


kc284:
Of what the ticket says, the only thing I haven't changed is the group icon, because the existing one looks very cramped when resized. We'll probably need a completely new icon to look right at 16x16.

XenAdmin/TabPages/AdPage.resx
```diff
@@ -145,56 +145,61 @@
...
-    <value>20</value>
-  </data>
   <data name="ColumnExpand.Width" type="System.Int32, mscorlib">
-    <value>20</value>
+    <value>5</value>
   </data>
   <metadata name="ColumnTypeImage.UserAddedColumn" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
     <value>True</value>
   </metadata>
+  <data name="dataGridViewCellStyle3.NullValue" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
MihaelaStoica: Is this image needed and does it need to be embedded in the resx?

kc284: I'm not sure, it was generated by the designer and I only left it because there were more like this in the file. I'll try to remove all of them and see what happens.

PR #2156, opened 2018-07-27
CA-293628 Some RPU prechecks do not mention the host the precheck has…

timliu55:
… run on

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -12061,7 +12061,7 @@ Check your settings and try again.</value>
     <value>No Default SR on Pool '{0}'</value>
   </data>
   <data name="UPDATES_WIZARD_NO_HOSTS" xml:space="preserve">
-    <value>There are no servers available to migrate the VM {0} to.</value>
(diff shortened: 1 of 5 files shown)


kc284:
Please add sign-off to the commit using your work email address.

timliu55:
Done.

XenModel/Messages.resx
```diff
@@ -12061,7 +12061,7 @@ Check your settings and try again.</value>
     <value>No Default SR on Pool '{0}'</value>
   </data>
   <data name="UPDATES_WIZARD_NO_HOSTS" xml:space="preserve">
-    <value>There are no servers available to migrate the VM {0} to.</value>
+    <value>There are no servers available to migrate the VM '{0}' on '{1}' to.</value>
```
MihaelaStoica: I think it would be better to have the host name at the begining, as we have in other problems.
_{1}: There are no servers available to migrate the VM {0} to._

timliu55: Agree.
I've aligned my code with "other problems".

PR #2158, opened 2018-07-30
CA-204242: save the server status report blocks xencenter

liulinC:

The save/upload process/wizard block the usage of xencenter, 
this commit extract the save/upload into an async action, which would run in background.

The async contain two sub-action, namely zip and upload.
The upload action depends on the zip action, so we need to sync the actions.


liulinC:
Test done:

manually verify.

liulinC:
@kc284  your comments make sense and refined accordingly,
could you please help to review it again?

MihaelaStoica:
Apart from my comments above, I was also wondering which actions would be best to show in the Events: the multiple action or the two sub actions? 
For example, if saving the report succeeds but the upload fails, I think it would be better if the user sees this as two separate entries. With the current implementation they will just see that the action "Saving Server Status Report" failed while uploading, without any indication that saving the report actually succeeded.

XenAdminTests/XenAdminTests.csproj
```diff
@@ -485,7 +485,9 @@
       <Name>XenServerHealthCheck</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
```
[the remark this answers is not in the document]
liulinC: I think this one is added by VS automatically, not sure why VS add it, 
will delete it

XenModel/Actions/ZipStatusReportAction.cs
```diff
@@ -62,13 +62,29 @@ public class ZipStatusReportAction : AsyncAction
...
             _inputTempFolder = tempFolder;
             _destFile = destFile;
         }
 
+        /// <summary>
+        /// Construcor, set suppressHistory to ture
+        /// </summary>
+        /// <param name="tempFolder">tempFolder to store the downloaded logs</param>
+        /// <param name="destFile">The target file to store the compress result</param>
+        public ZipStatusReportAction(string tempFolder, string destFile):this(tempFolder,destFile,true)
```
MihaelaStoica: There is no need for a separate constructor, we can set the default vlaue of the suppressHistory parameter to true in the previous constructor: 
```public ZipStatusReportAction(string tempFolder, string destFile, bool suppressHistory = true)```

File: XenModel/Messages.resx [hunk 3]
```diff
@@ -744,6 +744,9 @@
   <data name="ACTION_SAVE_SETTINGS" xml:space="preserve">
     <value>Save settings for '{0}'...</value>
   </data>
+  <data name="ACTION_SAVE_UPLOAD_SERVER_STATUS_REPORT" xml:space="preserve">
```
MihaelaStoica: As we are not showing the multiple action in the Events, the action description won't be visible to the user, so we don't necessarily need new resource strings for it. I think we can reuse Messages.BUGTOOL_SAVING for startDescription and Messages.COMPLETED for endDescription

XenModel/Actions/HealthCheck/UploadServerStatusReportAction.cs
```diff
@@ -77,12 +77,16 @@ protected override void Run()
...
+            if (Cancelling || Cancelled)
+            {
+                throw new CancelledException();
+            }
+              
             else if (string.IsNullOrEmpty(uploadUuid))
             {
                 // Fail to upload the zip to CIS server.
                 log.ErrorFormat("Fail to upload the Server Status Report {0} to CIS server {1}", bundleToUpload, UPLOAD_DOMAIN_NAME);
+                Description = Messages.ACTION_UPLOAD_SERVER_STATUS_REPORT_FAILED;
```
MihaelaStoica: Setting the Description here is not needed, as throwing the exception should be enough.

XenModel/Actions/ZipStatusReportAction.cs
```diff
@@ -197,6 +213,7 @@ protected override void Run()
                 ZipToOutputFile(_inputTempFolder);
                 PercentComplete = 100;
                 log.ErrorFormat("An exception was trapped while creating a server status report: " + exn.Message);
+                Description = Messages.STATUS_REPORT_ZIP_FAILED;
```
MihaelaStoica: Setting the Description here is not needed, as throwing the exception should be enough.

XenModel/Actions/ZipStatusReportAction.cs
```diff
@@ -62,8 +62,14 @@ public class ZipStatusReportAction : AsyncAction
 
         private long bytesToCompress = 1;
 
-        public ZipStatusReportAction(string tempFolder, string destFile)
-            : base(null, Messages.BUGTOOL_SAVING, Messages.BUGTOOL_SAVING, true)
+        /// <summary>
+        ///  Constructor to zip downloaded file action
+        /// </summary>
+        /// <param name="tempFolder">empFolder to store the downloaded logs</param>
```
kc284: "t" missing from tempFolder
"compressed" result
whether "to" suppress history in the "events tab"

XenAdmin/Wizards/BugToolWizardFiles/BugToolWizard.cs
```diff
@@ -120,33 +121,28 @@ protected override void FinishWizard()
...
-            // zip up the report files and save them to the chosen file
-            Actions.ZipStatusReportAction action =
-                new Actions.ZipStatusReportAction(bugToolPageRetrieveData.OutputFolder, bugToolPageDestination1.OutputFile);
-            using (var dialog = new ActionProgressDialog(action, ProgressBarStyle.Blocks))
+            AsyncAction action;
+            if (bugToolPageDestination1.Upload)
             {
-                dialog.ShowCancel = true;
-                dialog.ShowDialog();
+                // The MultipleAction only used as a wrapper, we will suppress its history and expose the sub-actions in the history
```
kc284: The MultipleAction "is" only...

XenModel/Actions/ZipStatusReportAction.cs
liulinC: updated

XenAdmin/Wizards/BugToolWizardFiles/BugToolWizard.cs
liulinC: updated

PR #2160, opened 2018-07-31
CP-27825: Remove restriction on creating >2TB disk

serencorbett1:


XenModel/XenAPI-Extensions/SR.cs
```diff
@@ -492,6 +492,22 @@ public static List<SRInfo> ParseSRList(List<Probe_result> probeExtResult)
...
         }
 
+        public virtual bool VdiCreationCanProceed(long vdiSize)
+        {
+            SM sm = GetSM();
+
+            bool vdiSizeUnlimited = sm != null && Array.IndexOf(sm.capabilities, "LARGE_VDI") != -1;
+            bool isThinlyProvisioned = sm != null && Array.IndexOf(sm.capabilities, "THIN_PROVISIONING") != -1;
+
+            if (vdiSize > 2 * Util.BINARY_TERA && !vdiSizeUnlimited)
```
MihaelaStoica: Let's swap these two conditions, to make clearer that if the vdi size is not unlimited, then we restrict it to 2TB.(and also to optimise the code a bit)

XenModel/XenAPI-Extensions/SR.cs
```diff
@@ -492,6 +492,22 @@ public static List<SRInfo> ParseSRList(List<Probe_result> probeExtResult)
             return results;
         }
 
+        public virtual bool VdiCreationCanProceed(long vdiSize)
+        {
+            SM sm = GetSM();
+
+            bool vdiSizeUnlimited = sm != null && Array.IndexOf(sm.capabilities, "LARGE_VDI") != -1;
+            bool isThinlyProvisioned = sm != null && Array.IndexOf(sm.capabilities, "THIN_PROVISIONING") != -1;
```
MihaelaStoica: this line can be moved a few lines below, just before it is used on line 505

PR #2162, opened 2018-07-31
CP-28926: Block export of VMs with >2TB disks in ova/ovf format

serencorbett1:

### DIFF
--- a/XenAdmin/Wizards/ExportWizard/ExportSelectVMsPage.cs
+++ b/XenAdmin/Wizards/ExportWizard/ExportSelectVMsPage.cs
@@ -120,7 +120,7 @@ namespace XenAdmin.Wizards.ExportWizard
         protected override void PageLeaveCore(PageLoadedDirection direction, ref bool cancel)
 		{
 			if (direction == PageLoadedDirection.Forward && IsDirty)
-					cancel = !PerformCheck(CheckSpaceRequirements);
+					cancel = !PerformCheck(CheckSpaceRequirements, CheckDiskSizeForTransfer);
 		}
 
         public override void PopulatePage()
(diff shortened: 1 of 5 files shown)


XenAdmin/Wizards/ExportWizard/ExportSelectVMsPage.cs
```diff
@@ -120,7 +120,7 @@ protected override void PageLoadedCore(PageLoadedDirection direction)
         protected override void PageLeaveCore(PageLoadedDirection direction, ref bool cancel)
 		{
 			if (direction == PageLoadedDirection.Forward && IsDirty)
-					cancel = !PerformCheck(CheckSpaceRequirements);
+					cancel = !PerformCheck(CheckSpaceRequirements, CheckDiskSizeForTransfer);
```
kc284: I think CheckDiskSizeForTransfer should be the first check to run: if CheckSpaceRequirements fails, there is no point in looking for more space only to find out afterwards that we are still be unable to export.

XenAdmin/Wizards/ExportWizard/ExportSelectVMsPage.cs
```diff
@@ -263,6 +263,22 @@ private bool CheckSpaceRequirements(out string errorMsg)
 			return true;
 		}
 
+	    private bool CheckDiskSizeForTransfer(out string errorMsg)
+	    {
+	        errorMsg = string.Empty;
+
+	        foreach (VM vm in VMsToExport)
+	        {
+	            if (GetTotalVmSize(vm) > 2 * Util.BINARY_TERA && !ExportAsXva)
```
kc284: I would check `!ExportAsXva` first in order to avoid the size calculation if the user has not selected ovf/ova.

XenModel/Messages.resx
```diff
@@ -5266,6 +5266,9 @@ Warning: to prevent data loss you must ensure that the LUN is not in use by any
   <data name="EXPORT_AS_BACKUP" xml:space="preserve">
     <value>Export to File...</value>
   </data>
+  <data name="EXPORT_ERROR_GREATER_THAN_2TB_OVA_OVF" xml:space="preserve">
+    <value>VMs with disks greater than 2TB cannot be exported as an OVF/OVA package.</value>
```
kc284: The message here says 2TB, but the code in the select VMs page is cheking against 2TiB. Can we confirm which of the two it is, TB or TiB, and modify either the message or the code?

Also, let's not hardcode the value in the message, but rather use the placeholder `{0}`; we can use the Utils class methods to get the disk size string from a number . We can even define the constant DISK_MAX_SIZE_FOR_OVF_EXPORT in the ExportSelectVMsPage.cs so that the number is in one place that can be easily updated  if the max size changes in future.

serencorbett1: Just checked and it is 2 TiB that we want.

PR #2163, opened 2018-07-31
L10N Bug Fixes (CA-294292 & CA-294333)

DarrenMatthews776:
CA-294292
CA-294333


kc284:
@DarrenMatthews776 could you please add sign-off notice to both commits using your work email address. Thanks.

PR #2164, opened 2018-08-02
CA-293854: Patching wizard updates list loses selection and resizes c…

timliu55:
…olumns when resorted

Remove the redundant code
Change the AutoSize mode from DisplayedCells to AllCells.


timliu55:
Done.
Great thanks for pointing out the defect imported by my changes.

PR #2165, opened 2018-08-02
CA-292626: Consider the contains-livepatch flag of all the applicable…

MihaelaStoica:
… updates when doing the prechecks in the automated updates mode.

Instead of looking at the minimal patches, we look at all applicable patches for that host (i.e. all the updates on the host's version that are not already installed) and we say that the host won't need a reboot only if all these are marked as contains-livepatch.

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -630,12 +630,10 @@ namespace XenAdmin.Core
         /// <summary>
         /// Gets an upgrade sequence that contains a version upgrade, optionally followed by the minimal patches for the new version
         /// </summary>
-        /// <param name="conn">Connection for the pool</param>
         /// <param name="alert">The alert that refers the version-update</param>
         /// <param name="updateTheNewVersion">Also add the minimum patches for the new version (true) or not (false).</param>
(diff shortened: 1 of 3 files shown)


XenAdmin/Core/Updates.cs
```diff
@@ -649,6 +649,66 @@ public static List<XenServerPatch> GetMinimalPatches(IXenConnection conn, XenSer
...
+        /// </summary>
+        public static List<XenServerPatch> GetAllPatches(IXenConnection conn, XenServerPatchAlert alert, bool updateTheNewVersion)
+        {
+            Debug.Assert(conn != null);
+            Debug.Assert(alert != null);
+
+            var allPatches = new List<XenServerPatch> { alert.Patch };
+
+            // if it's a version updgrade the update sequence will be this patch (the upgrade) and the patches for the new version
+            if (updateTheNewVersion && alert.NewServerVersion != null && alert.NewServerVersion.MinimalPatches != null)
```
kc284: Shouldn't the last bit be `alert.NewServerVersion.Patches != null`? If the MinimalPatches are null, we'll miss the patches from line 691 too. Or perhaps removed that bit completely and only AddRange if GetAllPatches(alert.NewServerVersion) returns non-null?

MihaelaStoica:
Thanks @kc284. Please see the new commit.

PR #2173, opened 2018-08-09
CA-156828: Do not uncompress imported appliance on a modal dialog

kc284:


kc284:
Although the above ticket does not mention it, the appliance download is also done on a modal dialog, which we may want to fix. I'm going to wait for corrections requested on this PR and then raise a new one for the download (becaue it will be similar).

kc284:
I forgot to mention that I did not retain the Retry button that was on the dialog - I think if the user wants to retry they can click Next again (although in order to do so they have to modify the textbox so it becomes enabled, but this is a different improvement for all the checks on this page).

PR #2175, opened 2018-08-10
CA-286882: Shutting down large number of VMs causes XenCenter to hang

MihaelaStoica:
Created a new base class for the Notifications pages (Alerts, Updates, Events). This can be used in the future to host more of the common code between these pages. For now, it implements a common mechanism for showing/hiding the page and ensures that the event handlers are deregistered when the page is hidden (i.e. when the user switches to a different view) and registered again when the page becomes visible

### DIFF
--- a/XenAdmin/MainWindow.cs
+++ b/XenAdmin/MainWindow.cs
@@ -3031,18 +3031,24 @@ namespace XenAdmin
             switch (submodeItem.SubMode)
             {
                 case NotificationsSubMode.Alerts:
+                    if (updatesPage.Visible)
+                        updatesPage.HidePage();
+                    if (eventsPage.Visible)
+                        eventsPage.HidePage();
                     alertPage.ShowPage();
-                    updatesPage.HidePage();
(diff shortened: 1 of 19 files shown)


XenAdmin/MainWindow.cs
```diff
@@ -3028,21 +3028,26 @@ private void navigationPane_TreeViewSelectionChanged()
...
-                updatesPage.RefreshUpdateList();
-
-            if (eventsPage.Visible)
+            switch (submodeItem.SubMode)
             {
-                eventsPage.RefreshDisplayedEvents();
-            }
+                case NotificationsSubMode.Alerts:
+                    alertPage.ShowPage();
+                    updatesPage.HidePage();
```
kc284: Before calling HidePage, do we need to check for page visibilty as we do in navigationPane_NavigationModeChanged? Because only one of the other two pages will be visible at a time; if, for example, the visible page was the updatePage, we can thus avoid the iteration through the actions when hiding the eventsPage which was invisible.

PR #2177, opened 2018-08-13
CA-280340: Currently used memory for a host is slightly different Use…

timliu55:
…d memory

Revert the conservative estimation

### DIFF
--- a/XenModel/XenAPI-Extensions/Host.cs
+++ b/XenModel/XenAPI-Extensions/Host.cs
@@ -1087,33 +1087,12 @@ namespace XenAPI
         }
 
         /// <summary>
-        /// The amount of memory free on the host. For George and earlier hosts, we use to use
-        /// the obvious Host_metrics.memory_free. Since Midnight Ride, however, we use
-        /// the same calculation as xapi, adding the used memory and the virtualisation overheads
-        /// on each of the VMs. This is a more conservative estimate (i.e., it reports less memory
-        /// free), but it's the one we need to make the memory go down to zero when ballooning
-        /// takes place.
+        /// The amount of memory free on the host.
         /// </summary>
         public long memory_free_calc()
         {
             Host_metrics host_metrics = Connection.Resolve(this.metrics);
-            if (host_metrics == null)
-                return 0;
-
-            long used = memory_overhead;
-            foreach (VM vm in Connection.ResolveAll(resident_VMs))
-            {
-                used += vm.memory_overhead;
-                VM_metrics vm_metrics = vm.Connection.Resolve(vm.metrics);
-                if (vm_metrics != null)
-                    used += vm_metrics.memory_actual;
-            }
-
-            // This hack is needed because of bug CA-32509. xapi uses a deliberately generous
-            // estimate of VM.memory_overhead: but the low-level squeezer code doesn't (and can't)
-            // know about the same calculation, and so uses some of this memory_overhead for the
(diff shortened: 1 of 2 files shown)


XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -1087,33 +1087,12 @@ public string GetDatabaseSchema()
...
-                if (vm_metrics != null)
-                    used += vm_metrics.memory_actual;
-            }
-
-            // This hack is needed because of bug CA-32509. xapi uses a deliberately generous
-            // estimate of VM.memory_overhead: but the low-level squeezer code doesn't (and can't)
-            // know about the same calculation, and so uses some of this memory_overhead for the
-            // VM's memory_actual. This causes up to 1MB of double-counting per VM.
-            return ((host_metrics.memory_total > used) ? (host_metrics.memory_total - used) : 0);
+            return host_metrics?.memory_free ?? 0;
```
kc284: I'm afraid we can't use C# 6 syntax because XenAdmin is still using VS 2013 (see CP-27819).

timliu55: Done.

kc284:
Hi @timliu55, apologies for the delay, we needed to do some digging up before resuming the review.
Host_metrics.memory_free cannot be used, it was deprecated in 5.6; unfortunately due to some missing functionality in the SDK generation code, the property has not been marked as such (while the get_memory_free() method has been marked). So, the way to show the same numbers as the Search tab is to use the same mechanism to obtain them; the Search tab gets the numbers from the RRDs using the MetricsUpdater class, which could be re-used here.

timliu55:
It makes sense.
I will modify it later.
Thanks for the suggestion.

timliu55:
Hi @kc284 ,
I made more investigation on this and had some new concerns:
1. get_memory_free() will be deprecated, but memory_free won't.
Because get_memory_free is an encapsulation of an RPC method, if you search '[Deprecated("XenServer  ' in the whole solution, you will find we ever deprecated lots of RPC methods.
However, memory_free is just a public property we created as needed, and the setter will be used by some other methods like the constructor of Host_metrics, UpdateFrom etc. We are not likely to deprecate it currently.
2. RRDs is too expensive to use
When using RRDs, we have to get the entire RRDs data and parse it.
We have several places which use RRDs: Performance Tab, Search Tab, Export Resource data. All of them need the entire RRDs data to build a complex page or export the whole data. But for this ticket's case, we only need one item - the free memory by OnPaint(). It is not worthy to use RRDs.
3. memory_free property has no big difference with RRDs free memory
RRDs gets free memory by RRD daemon on the host, while memory_free property is updated by RPC when OnPaint. In my test case, they are almost the same number.

So I just suggest to simply use the memory_free property, what do you think?

kc284:
Each SDK property corresponds to a xapi field; if the field is deprecated in a version, we should not use the property from tbhis versino onwards, because even if it's present in the API, and even if it's still populated for this version, its value may change wihtout warning (normally to the field's default) and become unreliable in future.

We don't always need to download everything for the RRDs; depending on what we need, we can obtain subsets and not all of the bulk of data using filtering. Now, there is some filtering functionality missing on the server side, and additionally at some point we should switch the RRDs from XM to JSON to reduce the download time, so if there is not too much we can do now about the current ticket, we can block it on the RRD tickets and improve it later.

kc284:
Hi @timliu55 I'm going to close this PR now as we need to work to a direction different from the suggested solution.

PR #2179, opened 2018-08-13
CP-28648 Add the Lima CHMs to the product build

katshann:


kc284:
Here, too, please add sign-off info; thanks.

PR #2180, opened 2018-08-14
CA-293776: The order in which a single update is applied on the slave…

timliu55:
…s is not the same as the order in which they are displayed

Apply the same Hosts sorting rules for all the PatchingWizard Pages .


kc284:
@timliu55 could you please rebase this PR? One of the changes is in a file which has now been removed. Thanks.

timliu55:
Hi Tina,
I just checked, my changes in the removed file have been applied to PatchingWizard_ModePage.GroupServersPerPool().
So the remained 2 changed files are enough now.

PR is updated. Thank you!

XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectServers.cs
```diff
@@ -518,9 +518,9 @@ public List<Host> SelectedServers
 
                     if (WizardMode != WizardMode.SingleUpdate)   
                         //prechecks will fail in automated updates mode if one of the hosts is unreachable
-                        return SelectedPools.SelectMany(p => p.Connection.Cache.Hosts.Where(host => enabledHosts.Contains(host))).ToList();
+                        return SelectedPools.SelectMany(p => p.Connection.Cache.Hosts.OrderBy(host => host).Where(host => enabledHosts.Contains(host))).ToList();
```
kc284: The OrderBy() can be moved after the Where() filter because theoretically we have fewer objects to sort.
Same goes for the line below.

timliu55: Clever!  👍

PR #2183, opened 2018-08-16
CA-293813: Patch pools in parallel when installing a single update/patch or supplemental package.

kc284:


kc284:
Plesae feel free to squash the commits once reviewed.

PR #2184, opened 2018-08-16
CP-28915: Support for a new license type CXS_CCLD_CCS in Havana

ywwseu:
@MihaelaStoica

### DIFF
--- a/XenModel/Utils/Helpers.cs
+++ b/XenModel/Utils/Helpers.cs
@@ -396,6 +396,25 @@ namespace XenAdmin.Core
         }
 
         /// <param name="conn">May be null, in which case true is returned.</param>
+        public static bool HavanaOrGreaterLTSR71(IXenConnection conn)
+        {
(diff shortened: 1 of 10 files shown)


XenModel/Utils/Helpers.cs
```diff
@@ -395,6 +395,25 @@ public static bool ElyOrGreater(Host host)
...
+         /// As Havana platform version is same with Ely and Honolulu, so use product version here
+        /// <param name="host">May be null, in which case true is returned.</param>
+        public static bool HavanaOrGreaterLTSR71(Host host)
+        {
+            if (host == null)
+                return true;
+             string product_version = HostProductVersion(host);
+            return
+                product_version != null && 
+                ((product_version.StartsWith("7.1") && productVersionCompare(product_version, "7.1.2") >= 0) ||
```
MihaelaStoica: Instead of hardcoding the product versions here, we should add them to the branding, similar to what we have for 7.0 for example https://github.com/xenserver/xenadmin/blob/master/Branding/branding.sh#L107

PR #2186, opened 2018-08-20
CP-28676: XC: Allow choosing boot options during new VM wizard

timliu55:

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/NewVMWizard.cs
+++ b/XenAdmin/Wizards/NewVMWizard/NewVMWizard.cs
@@ -167,13 +167,13 @@ namespace XenAdmin.Wizards.NewVMWizard
                                         page_3_InstallationMedia.SelectedPvArgs,
                                         page_3_InstallationMedia.SelectedCD,
                                         page_3_InstallationMedia.SelectedUrl,
+                                        page_3_InstallationMedia.SelectedBootMode,
                                         m_affinity,
                                         page_5_CpuMem.SelectedVcpusMax,
                                         page_5_CpuMem.SelectedVcpusAtStartup,
                                         (long)page_5_CpuMem.SelectedMemoryDynamicMin,
                                         (long)page_5_CpuMem.SelectedMemoryDynamicMax,
                                         (long)page_5_CpuMem.SelectedMemoryStaticMax,
-                                        page_5_CpuMem.SelectedBootMode,
                                         page_6b_LunPerVdi.MapLunsToVdisRequired
                                             ? page_6b_LunPerVdi.MappedDisks
                                             : page_6_Storage.SelectedDisks,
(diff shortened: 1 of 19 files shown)


XenAdmin/Wizards/NewVMWizard/NewVMWizard.cs
```diff
@@ -173,6 +173,7 @@ protected override void FinishWizard()
                                         (long)page_5_CpuMem.SelectedMemoryDynamicMin,
                                         (long)page_5_CpuMem.SelectedMemoryDynamicMax,
                                         (long)page_5_CpuMem.SelectedMemoryStaticMax,
+                                        page_5_CpuMem.SelectedBootMode,
```
kc284: This new functionality has been added to the Cpu and Memory page according to the design, but I have two issues with it:
1) We seem to have space issues on this page. I suppose we could reduce a bit the margins of the existing controls, but if the bottom warning appears, we'll have truncation issues.
2) I don't think the boot mode matches the rest of the content of this page; the Installation media is more relevant.

I'm going to place a request to reconsider this design. Could you please hold on a bit on this comment until I confirm this with all involved stakeholders.

timliu55: Tina, thanks very much for the comments.

I agree that boot mode should not be with CPU & Memory.
And the similar design also happens in CP-28679, so could you please also involve it in your design reconsideration request?

kc284: Hmmm, the ImportWizard is slightly different in that it is more compact than the NewVM wizard and already the VM definition page contains controls from across 3 different pages of the NewVM wizard (Name, CPU-Memory, and Disk). For the time being, and before considering an extra page, let's see first if we can fit the boot options above the .wim option (we may need to remove some of the empty gridlines or other empty space this page has).
Also, one thing you may want to consider is moving the boot options on a custom control which could be used as is on both wizards.

timliu55: It seems that there is not enough space above the .wim option, but let me have a try.
I agree to extract the boot options as a UserControl, and I will do that, thank you.

timliu55:
Ready for review...

XenAdmin/Wizards/NewVMWizard/Page_InstallationMedia.cs
```diff
@@ -64,6 +65,8 @@ public Page_InstallationMedia()
 
         public bool ShowBootParameters { get { return !SelectedTemplate.IsHVM(); } }
```
kc284: On the NewVMWizard we have this line:
`page_3_InstallationMedia.DisableStep = noInstallMedia && !page_3_InstallationMedia.ShowBootParameters;`
Say noInstallMEdia is false; if ShowBootParameters is false, then ShowBootParameters is true, but the user won't see it because the page will be disabled. I think we can remove the line.

I also think it adds complexity if we have two properties that are checking the same thing. I suggest removing ShowBootParameters and using only ShowBootModesControl, and also renaming it to IsSelectedTemplateHVM so it's more generic and not control specific.

timliu55:
Done.
Thank you for the comments.

PR #2187, opened 2018-08-21
CP-28677: XC: Exported OVFs contain UEFI metadata

timliu55:
Extended xenBootOrder to xenBootParams

### DIFF
--- a/XenOvfApi/app.config
+++ b/XenOvfApi/app.config
@@ -67,8 +67,8 @@
    <setting name="xenBootOptions" serializeAs="String">
     <value>BIOS order</value>
    </setting>
-   <setting name="xenBootOrder" serializeAs="String">
-    <value>dc</value>
+   <setting name="xenBootParams" serializeAs="String">
+    <value>order=dc;</value>
    </setting>
    <setting name="xenPVPlatformSetting" serializeAs="String">
     <value>nx=true;acpi=true;apic=true;pae=true;viridian=true;</value>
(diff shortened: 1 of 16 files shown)


XenOvfApi/app.config
```diff
@@ -67,8 +67,8 @@
    <setting name="xenBootOptions" serializeAs="String">
     <value>BIOS order</value>
    </setting>
-   <setting name="xenBootOrder" serializeAs="String">
-    <value>dc</value>
+   <setting name="xenBootParams" serializeAs="String">
+    <value>order=dc;</value>
```
timliu55: Hi  @kc284,
It occurred to me that this change might break the logic if there were some persistent config in the customers' environment.
I just went through the modification history of the file and found no structural changes.
So for my CP ticket, I would remain xenBootOrder as it was, and create a new xenBootMode node to implement my job, but we may suffer from changing it again in future.
Then what should we do if we really want to change the existing config?

kc284:
I have noticed this too, however I had a look in the installer scripts and I don't think these files are actually installed in the user environment. And indeed it is a bit strange that the original developer put these values in the settings, because I don't think the user should be able to configure them, they should probably be resources or constants in the code.

kc284:
The attribute Name of the node wass HVM_boot_params, which hasn't changed. The only problem is that previously only the value was stored, but now it is a collection of keyvalue pairs. So if you want to change it, you could perhaps special-case it for older ovf imports, i.e. check if there is a non-empty string that cannot be parsed as dictionary in which case assign it to "order".

kc284:
It's probably also safe to improve a bit the Info of that node and use one more relevant to the boot params rather than the generic XENSERVER_SPECIFIC_DESCRIPTION.

PR #2188, opened 2018-08-21
[LIMA] L10N: CA-294897 & Translation updates

DarrenMatthews776:


kc284:
@DarrenMatthews776 presumably these changes should also go to master, correct? If so, could you please raise the relevant PR. Thanks.

PR #2189, opened 2018-08-21
CA-289189: Add sole PIF name changing action

jijiang:

### DIFF
--- a/XenAdmin/Dialogs/NetworkingProperties.cs
+++ b/XenAdmin/Dialogs/NetworkingProperties.cs
@@ -438,6 +438,7 @@ namespace XenAdmin.Dialogs
         private void AcceptBtn_Click(object sender, EventArgs e)
         {
             List<PIF> newPIFs = new List<PIF>();
+            List<PIF> newNamePIFs = new List<PIF>();
             List<PIF> downPIFs = new List<PIF>();
 
             foreach (PIF pif in AllPIFs)
@@ -450,7 +451,7 @@ namespace XenAdmin.Dialogs
             {
                 try
                 {
-                    CollateChanges(page, page.Tag as PIF, newPIFs);
+                    CollateChanges(page, page.Tag as PIF, newPIFs, newNamePIFs);
                 }
                 catch (Failure)
                 {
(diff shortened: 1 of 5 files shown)


XenModel/Actions/Network/SetSecondaryManagementPurposeAction.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace XenAdmin.Actions
+{
+    public class SetSecondaryManagementPurposeAction : AsyncAction
+    {
+        private Pool pool;
+        private List<PIF> pifs;
+        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+
+        public SetSecondaryManagementPurposeAction(IXenConnection connection, Pool pool, List<PIF> pifs)
```
kc284: Any RBAC considerations for this action? Renaming does not sound particularly risky as an operation, but perhaps it should be subject to the same restrictions as the other re-configuration operations?

XenAdmin/Dialogs/NetworkingProperties.cs
```diff
@@ -496,6 +497,13 @@ private void AcceptBtn_Click(object sender, EventArgs e)
             // list originally so that they can be used as a filter against downPIFs.
             newPIFs.RemoveAll(delegate(PIF p) { return !p.Changed; });
 
+            newNamePIFs.RemoveAll(delegate(PIF p) { return !p.Changed; });
```
kc284: Let's modernise a bit this code and the 2 lines above it by using lambdas.

XenModel/Messages.resx
```diff
@@ -771,6 +771,9 @@
   <data name="ACTION_SCANNING_SR_FROM" xml:space="preserve">
     <value>Performing SR scan from '{0}'</value>
   </data>
+  <data name="ACTION_SET_SECONDARY_MANAGEMENT_PURPOSE_TITLE" xml:space="preserve">
+    <value>Changing server networking interface name</value>
```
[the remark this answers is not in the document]
jijiang: Is "..." needed at the end? The string will appear on the Events tab -> Message column, not inside the status bar.

XenAdmin/Dialogs/NetworkingProperties.cs
jijiang: will fix it.

XenModel/Actions/Network/SetSecondaryManagementPurposeAction.cs
jijiang: Actually in this action, only PIF.SaveChanges calling to server side is made. Is it enough to add this one to the RBAC precheck?

XenModel/Messages.resx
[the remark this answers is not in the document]
kc284: I can't rememebr what otehr messages on there do, if the others don't have ellipsis, feel free to drop it.

XenModel/Actions/Network/SetSecondaryManagementPurposeAction.cs
kc284: TYes, ok, the actual API call used is pif.set_other_config so let's add this one.

PR #2193, opened 2018-08-23
CP-29230: Change License names (Enterprise to Premium, Free to Express)

michael2012z:
Modify the names in AssignLicenseDialog, FriendlyNames.resx and Messages.resx.

Note, FriendlyNames.Designer.cs and Messages.Designer.cs also include some modification from "[XenServer]" to "server". They were introduced in a preceding commit, but I forgot to commit the *.Designer.cs files.

### DIFF
--- a/XenAdmin/Dialogs/AssignLicenseDialog.cs
+++ b/XenAdmin/Dialogs/AssignLicenseDialog.cs
@@ -55,7 +55,7 @@ namespace XenAdmin.Dialogs
             InitializeComponent();
             licenseServerNameTextBox.TextChanged += licenseServerPortTextBox_TextChanged;
             licenseServerPortTextBox.TextChanged += licenseServerNameTextBox_TextChanged;
-            SetOptionsForClearwaterAndNewer();
+            LoadLicenseOptions();
             UpdateButtonEnablement();
 
             // if all the hosts have the same license server details then populate the textboxes.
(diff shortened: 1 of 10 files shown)


michael2012z:
Hi, @kc284 , in the overall solution of rebranding on XenCenter, we "hard coded" the product name to "Citrix Hypervisor" even if we were operating a server of old version. Does it make sense to let "AssignLicenseDialog" alone be smart to switch between old and new brand? Can we just apply all them to the new names simply?

XenAdmin/Dialogs/AssignLicenseDialog.cs
```diff
@@ -99,25 +99,39 @@ private List<Host> CreateHostsList()
...
-                enterprisePerSocketRadioButton.Text = String.Format(Messages.ENTERPRISE_PERSOCKET_LICENSES_X_REQUIRED,
-                                                          xos.Sum(x => x.Connection.Cache.Hosts.Sum(h => h.CpuSockets())));
-                standardPerSocketRadioButton.Text = String.Format(Messages.STANDARD_PERSOCKET_LICENSES_X_REQUIRED,
-                                                          xos.Sum(x => x.Connection.Cache.Hosts.Sum(h => h.CpuSockets())));
+                enterprisePerSocketRadioButton.Text = String.Format(Messages.LICENSE_EDITION_ENTERPRISE_PERSOCKET, 
+                    xos.Sum(x => x.Connection.Cache.Hosts.Sum(h => h.CpuSockets()))); 
+                enterprisePerUserRadioButton.Text = Messages.LICENSE_EDITION_ENTERPRISE_PERUSER;
+                desktopPlusRadioButton.Text = Messages.LICENSE_EDITION_DESKTOP_PLUS;
+                desktopRadioButton.Text = Messages.LICENSE_EDITION_DESKTOP;
+                desktopCloudRadioButton.Text = Messages.LICENSE_EDITION_DESKTOP_CLOUD;
```
MihaelaStoica: `desktopCloudRadioButton` should be made visible here (as it is not visible by default)

michael2012z: Agree. And I found enterprisePerUserRadioButton had visibility problem too. Going to fix it.

PR #2199, opened 2018-08-24
 CA-294333: NewPolicySnapshotPage corrections + small checkbox bug

kc284:


XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePage.cs
```diff
@@ -43,20 +43,14 @@ namespace XenAdmin.Wizards.NewPolicyWizard
...
-        }
+        private VMSS _policy;
+
+        public List<VM> SelectedVMs { private get; set; }
         
         public NewPolicySnapshotTypePage()
         {
             InitializeComponent();
-            this.labelWarning.Text = string.Format(this.labelWarning.Text, Messages.VMSS_TYPE);
+            labelWarning.Text = string.Format(labelWarning.Text, Messages.VMSS_TYPE);
```
MihaelaStoica: I think that `string.Format` is not needed here anymore, as we always use the same string (Messages.VMSS_TYPE) as the parameter, so we can included in the label's text directly.

kc284: I've rebased the whole PR but only the third commit is new.

PR #2200, opened 2018-08-27
CA-297020: Do not download imported appliance on a modal dialog

kc284:


MihaelaStoica:
These changes look fine to me. Just one comment regarding shwing the error message: If the download fails, the error is displayed on the page, but this is not cleared when I change the URI or when I select a file from disk. (it will only get cleared when a new download begins). We should maybe clear it when the Filename text box is changed?

kc284:
I also moved the URI validation to the error underneath the textbox so we don't need to wait until hitting Next for this (not sure why why it wasn't done so in the first place).

PR #2202, opened 2018-08-29
CP-28678: XC: Import UEFI OVFs correctly

timliu55:

### DIFF
--- a/XenOvfTransport/Import.cs
+++ b/XenOvfTransport/Import.cs
@@ -1279,6 +1279,9 @@ namespace XenOvfTransport
                         case "platform":
                             newVm.platform = MakePlatformHash(xcsd.Value.Value);
                             break;
+	                    case "nvram":
+		                    newVm.NVRAM = SplitStringIntoDictionary(xcsd.Value.Value);
+		                    break;
                         default:
                             hashtable.Add(key, xcsd.Value.Value);
                             break;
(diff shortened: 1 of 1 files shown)


XenOvfTransport/Import.cs
```diff
@@ -1401,7 +1404,7 @@ private void RemoveVDI(Session xenSession, XenRef<VDI> vdi)
 
         private static Dictionary<string, string> SplitStringIntoDictionary(string inputStr)
         {
-	        return inputStr.Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries).Select(o => o.Split('='))
+	        return inputStr.Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries).Select(o => o.Split(new[] { '=' }, 2))
```
kc284: What is this change fixing?

timliu55: This change is not used to fix the existing logic, but is for this new ticket CP-28678.
Because CP-28678 is asking to import base64 string, and base64 string uses '=' as part of its content (padding char uses '=')

For example, the NVRAM may equal
'EFI-variables-backend=xapidb;EFI-variabbles-on-boot=reset;EFI-variables=dGVzdA=='
And in this case, when the Split function works on 'EFI-variables=dGVzdA==', we would like it to stop moving forward after it finds the first '='

kc284: Could you please add this comment in the code (for future reference); thanks.

timliu55: Sure.
Done.

PR #2203, opened 2018-08-29
CP-28679: XC: Allow choosing boot options during disk import

timliu55:

### DIFF
--- a/XenAdmin/Wizards/BootModesControl.Designer.cs
+++ b/XenAdmin/Wizards/BootModesControl.Designer.cs
@@ -64,9 +64,7 @@ namespace XenAdmin.Wizards
             // radioButtonBIOSBoot
             // 
             resources.ApplyResources(this.radioButtonBIOSBoot, "radioButtonBIOSBoot");
-            this.radioButtonBIOSBoot.Checked = true;
             this.radioButtonBIOSBoot.Name = "radioButtonBIOSBoot";
-            this.radioButtonBIOSBoot.TabStop = true;
             this.radioButtonBIOSBoot.UseVisualStyleBackColor = true;
             // 
             // radioButtonUEFIBoot
--- a/XenAdmin/Wizards/BootModesControl.cs
+++ b/XenAdmin/Wizards/BootModesControl.cs
@@ -29,54 +29,61 @@
  * SUCH DAMAGE.
  */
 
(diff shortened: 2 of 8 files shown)


XenAdmin/Wizards/BootModesControl.Designer.cs
```diff
@@ -64,9 +64,7 @@ private void InitializeComponent()
             // radioButtonBIOSBoot
             // 
             resources.ApplyResources(this.radioButtonBIOSBoot, "radioButtonBIOSBoot");
-            this.radioButtonBIOSBoot.Checked = true;
             this.radioButtonBIOSBoot.Name = "radioButtonBIOSBoot";
-            this.radioButtonBIOSBoot.TabStop = true;
```
timliu55: Refine the code
1. Only check the default radio when necessary
2. Remove the unnecessary set on TabStop(radio button will set tabstop to true when checked)

XenAdmin/Wizards/BootModesControl.cs
```diff
@@ -29,54 +29,61 @@
...
+		}
 
-        public VM TemplateVM
-        {
-            get { return _templateVM; }
-            set
-            {
-                if (_templateVM == value)
-                    return;
+		[DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
```
timliu55: When I drag the BootModesControl from Visual Studio Toolbox
The designer will automatically generate a redundant line
>this.bootModesControl1.TemplateVM = null;
After setting this attribute to hidden, the designer won't generate above line

XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
```diff
@@ -81,6 +81,7 @@ public override void PopulatePage()
 			m_upDownMemory.Value = m_upDownMemory.Minimum;
 			m_upDownCpuCount.Value = m_upDownCpuCount.Minimum;
 			m_upDownAddSpace.Value = m_upDownAddSpace.Minimum;
+			bootModesControl1.CheckBIOSBootMode();
```
timliu55: I notice that when I click Previous button and then click Next button, some controls will clear the users' input value.
So I would like the BootModeControl to keep consistent with them.

PR #2209, opened 2018-08-31
Cherry picking from master: CA-297087: Fixed crash occurring when launching the Import Wizard…

kc284:
…withou having preselected a connection on the treeview. Also added a null check to
prevent crash if the connection is suddenly interrupted while moving away
from the select destination host page.


jijiang:
looks good!

PR #2210, opened 2018-08-31
CA-297088: Fixed issue where XenCenter froze if a new server was added…

kc284:
… while we were on the select host page of the Import and Migration Wizards.


jijiang:
I've tested the change. It works well for both import and migrate VM wizards. Before merging it, I would like to ask: why we use asynchronous invoke in CollectionChanged(), while synchronous invokes in xenConnection_CachePopulated() and xenConnection_ConnectionStateChanged()?

MihaelaStoica:
@jijiang In the CollectionChanged event handler we use asynchronous invoke to avoid a deadlock: The XenConnection collection changed event is called under a lock (ConnectionsManager.ConnectionLock , when the new connection is added to ConnectionsManager.XenConnections) and the function invoked in the event handler (PopulateComboBox) also needs to aquire the lock, when trying to access XenConnectionsCopy.

PR #2214, opened 2018-09-11
CA-297216: Remove the upsell message from the memory tab and disable …

MihaelaStoica:
…the DMC settings for VMs in the "Memory Setting" dialog if the server is not licensed.

### DIFF
--- a/XenAdmin/MainWindow.cs
+++ b/XenAdmin/MainWindow.cs
@@ -1423,7 +1423,6 @@ namespace XenAdmin
             // empty (i.e. at startup).
             bool show_home = SelectionManager.Selection.Count == 1 && SelectionManager.Selection[0].Value == null;
             // The upsell pages use the first selected XenObject: but they're only shown if there is only one selected object (see calls to ShowTab() below).
-            bool dmc_upsell = Helpers.FeatureForbidden(SelectionManager.Selection.FirstAsXenObject, Host.RestrictDMC);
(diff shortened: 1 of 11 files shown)


XenAdmin/MainWindow.cs
```diff
@@ -1447,7 +1446,7 @@ private void UpdateToolbarsCore()
             NewTabCount = 0;
             ShowTab(TabPageHome, !SearchMode && show_home);
             ShowTab(TabPageGeneral, !multi && !SearchMode && (isVMSelected || (isHostSelected && (isHostLive || !is_connected)) || isPoolSelected || isSRSelected || isVdiSelected || isDockerContainerSelected));
-            ShowTab(dmc_upsell ? TabPageBallooningUpsell : TabPageBallooning, !multi && !SearchMode && (isVMSelected || (isHostSelected && isHostLive) || isPoolSelected));
+            ShowTab(TabPageBallooning, !multi && !SearchMode && (isVMSelected || (isHostSelected && isHostLive) || isPoolSelected));
```
kc284: Can BallooningUpsellPage class/instance and TabPageBallooningUpsell be removed completely from the MainWindow since they're not used any more?

MihaelaStoica: Yes, I'll do it

MihaelaStoica: @kc284 pleasee see the new commit

PR #2185, opened 2018-08-17
CP-28915: Support for a new license type CXS_CCLD_CCS in Havana

ywwseu:

PR #2216, opened 2018-09-13
CA-297617: Change Citrix Hypervisor Tools to Citrix VM Tools.

michael2012z:
The PR covers:
1. Replaced all "[XenServer product] Tools" with "[Citrix VM Tools]". Modified files were: *.resx files for UI components, FriendlyName files, Messages files. 
2. In Branding/branding.sh, picked the real name of Tools and saved it in variable BRANDING_PV_TOOLS.
3. In mk/re-branding.sh, replaced all occurrences of "[Citrix VM Tools]" with BRANDING_PV_TOOLS value.
4. Ja and zh-CN files were also handled.


XenAdminTests/TabsAndMenus/MainMenuGeorge.cs
```diff
@@ -128,7 +128,7 @@ public void MainMenu_XenCenterNode()
                 new ExpectedTextMenuItem("Ena&ble PVS-Accelerator...", false, false  ),
                 new ExpectedTextMenuItem("Disable P&VS-Accelerator", false, false  ),
 	            new ExpectedSeparator(),
-	            new ExpectedTextMenuItem("Inst&all " + Branding.PRODUCT_BRAND + " Tools...", false, false  ),
+	            new ExpectedTextMenuItem("Inst&all " + "[Citrix VM Tools]" + "...", false, false  ),
```
MihaelaStoica: This file is not getting rebranded during the build (*.cs files shouldn't, generally, be), so the `[Citrix VM Tools]` will never be replaced. That's why we need to use constants declared in branding.cs (which will contain the correctly branded values, as we apply the branding to that file). See also my comment here https://github.com/xenserver/xenadmin/pull/2185#issuecomment-418325028

PR #2212, opened 2018-09-10
CA-210190 further improvements + CA-294286 show memory limitations on NewVM wizard

kc284:

PR #2222, opened 2018-09-17
CP-28916 port fixes to havana

ywwseu:
Hi @MihaelaStoica @michael2012z , could you please help to review this PR, thanks.


kc284:
@ywwseu I recommended on the ticket to leave CA-210190 out of havana; this means we should remove commits `209d82a`, `d004d3f` and `a063dec` from the above PR.
The reason is that, although these commits fixed CA-210190 itself, they caused some layout issues in other controls (there is PR #2212 to fix these still in review at the time of writing).

ywwseu:
Thanks @kc284 , I've removed CA-210190's three commits per your comments, would you please help to review again? And @MihaelaStoica and @michael2012z as well.

PR #2223, opened 2018-09-17
CA-298592; fixed MainWindow designer; refactoring.

kc284:
4 separate commits for easier review; the actual crash is fixed by the last commit, feel free to squash them if you want.


MihaelaStoica:
@kc284, with these changes, XenCenter starts with the Search page selected, instead of the Home page.

kc284:
Since I was in the area I thought I'd simplify a bit the code updating the tab pages because I've always thought it was more complicated than it should be.

PR #2226, opened 2018-09-20
CA-287643: Replaced list view with data grid view to resolve column sizing issue and for consistency with the rest of the product.

AaronRobson:


AaronRobson:
Please feel free to compress these into a single commit.

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -272,7 +206,8 @@ private void LoadTags(List<string> tags, List<string> indeterminateTags)
             {
                 if (!TagPresent(tag))
                 {
-                    TagsListViewItem item = (TagsListViewItem)tagsListView.Items.Add(new TagsListViewItem(tag));
+                    var item = new TagsDataGridViewRow();
```
kc284: Using object initializers is prettier:
```
var item = new TagsDataGridViewRow(){Checked = CheckState.Checked, Text=tag};
 tagsDataGrid.Rows.Add(item);
```
(also elsewhere in the file)

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -302,117 +237,39 @@ private void addButton_Click(object sender, EventArgs e)
...
 
-        private class TagsListViewItem : ListViewItem
+        public class TagsDataGridViewRow : DataGridViewRow
         {
-            public TagsListViewItem(string text)
-                : base(text)
+            public TagsDataGridViewRow()
             {
-                Checked = CheckState.Unchecked;
+                var cellEnabled = new DataGridViewCheckBoxCell { Value = CheckState.Unchecked };
```
kc284: Please also set ThreeState = true on the CheckBoxCell (maybe on the column as well), otherwise the Indeterminate state does not appear correctly. To test this, set, for example, tag Foo on vm1, tag Bar on vm2, then select both VMs and right click Edit tags. Foo and Bar currently appear Checked instead of Indeterminate.

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -82,75 +82,7 @@ private List<string> GetTags(CheckState checkState)
         private void InitDialog()
```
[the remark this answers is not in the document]
AaronRobson: These should all be resolved now.

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -302,117 +237,39 @@ private void addButton_Click(object sender, EventArgs e)
...
-                : base(text)
+            public TagsDataGridViewRow()
             {
-                Checked = CheckState.Unchecked;
+                var cellEnabled = new DataGridViewCheckBoxCell { Value = CheckState.Unchecked };
+                var cellTag = new DataGridViewTextBoxCell();
+                Cells.AddRange(cellEnabled, cellTag);
             }
 
             public void Toggle()
```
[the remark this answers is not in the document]
kc284: This has not been resolved completely.

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -272,7 +206,8 @@ private void LoadTags(List<string> tags, List<string> indeterminateTags)
             {
                 if (!TagPresent(tag))
                 {
-                    TagsListViewItem item = (TagsListViewItem)tagsListView.Items.Add(new TagsListViewItem(tag));
+                    var item = new TagsDataGridViewRow();
+                    tagsDataGrid.Rows.Add(item);
                     item.Checked = CheckState.Checked;
                     item.Text = tag;
                 }
```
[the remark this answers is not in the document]
kc284: The form's accept button has not been set.

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -422,42 +223,79 @@ public void Toggle()
...
+                set { _cellCheckState.Value = value; }
+            }
+
+            public string Text
+            {
+                get { return _cellTag.Value.ToString(); }
+                set { _cellTag.Value = value; }
+            }
+
+            public int CompareTo(TagsDataGridViewRow other)
```
kc284: I've just realised that the gridview's rows are not clickable, which means sorting is not taking place. This seems to have been a bug in the original code too. The way to fix it is to set SortMode to Automatic on each column and handle the datagridvuiew's SortCompare method (see how this is used elsewhere in XenCenter, e.g. the class BugToolPageSelectCapabilities.cs is a good example).

XenAdmin/Dialogs/NewTagDialog.cs
```diff
@@ -162,257 +83,137 @@ private void textBox1_KeyPress(object sender, KeyPressEventArgs e)
...
-        private Comparison<TagsListViewItem> Comparison()
+        private void DisplayList(List<TagsDataGridViewRow> rows)
         {
-            return delegate(TagsListViewItem x, TagsListViewItem y)
-                       {
-                           return x.Text.CompareTo(y.Text);
+            rows.Sort();
 
-                       };
+            tagsDataGrid.Rows.Clear();
```
kc284: Normally when we clear/re-add rows in a datagridview we do it in a try-optinally catch-finally block as follows to prevent display problems:
```
try
{
    tagsDataGrid.SuspendLayout();
    tagsDataGrid.Rows.Clear();

    foreach (var item in rows)
        tagsDataGrid.Rows.Add(item);
}
finally
{
    tagsDataGrid.ResumeLayout();
}
```

PR #2228, opened 2018-09-21
Minor language corrections and removal of duplicate message.

kc284:


michael2012z:
The corrections look good to me.

PR #2232, opened 2018-09-28
CA-298952: Update product web site link.

michael2012z:

### DIFF
--- a/XenModel/InvisibleMessages.Designer.cs
+++ b/XenModel/InvisibleMessages.Designer.cs
@@ -19,7 +19,7 @@ namespace XenAdmin {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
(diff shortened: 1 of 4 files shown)


XenModel/InvisibleMessages.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace XenAdmin {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
```
michael2012z: This change was introduced by VS. Not possible to get rid of it.

kc284: No worries, this always happens when the designer is opened by a version of VS different from the one where the file was created, but it is perfectly ok.
There will be more changes like this once we switch officially to VS2017 (there is a ticket for this).

PR #2233, opened 2018-09-28
CA-297666: Wait for host to become enabled before resuming a VM in the patching …

AaronRobson:
…wizard.


AaronRobson:
I can demonstrate the happy path where the host becomes enabled quickly, however I am not sure how to test the sad path.

PR #2236, opened 2018-09-29
CA-298954: Match server branding and version.

michael2012z:
This PR contains:
1. Remove branding from Build data, Build number and Version in general tab page.
2. Show different branding in Name field of general tab page according to server version. 

### DIFF
--- a/XenModel/XenAPI-Extensions/Host.cs
+++ b/XenModel/XenAPI-Extensions/Host.cs
@@ -178,7 +178,7 @@ namespace XenAPI
         public override string Description()
         {
             if (name_description == "Default install of XenServer" || name_description == "Default install") // i18n: CA-30372, CA-207273
-                return Messages.DEFAULT_INSTALL_OF_XENSERVER;
+                return string.Format(Messages.DEFAULT_INSTALL_OF_XENSERVER, software_version.ContainsKey("product_brand") ? software_version["product_brand"] : Messages.XENSERVER);
             else if (name_description == null)
                 return "";
             else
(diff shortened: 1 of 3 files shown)


XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -178,7 +178,7 @@ public override string ToString()
         public override string Description()
         {
             if (name_description == "Default install of XenServer" || name_description == "Default install") // i18n: CA-30372, CA-207273
-                return Messages.DEFAULT_INSTALL_OF_XENSERVER;
+                return Helpers.NaplesOrGreater(this) ? Messages.DEFAULT_INSTALL_OF_XENSERVER : Messages.DEFAULT_INSTALL_OF_LEGACY_XENSERVER;
```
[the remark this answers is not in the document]
michael2012z: How graceful this way is! 👍 
Thank you.

XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -178,7 +178,7 @@ public override string ToString()
         public override string Description()
         {
             if (name_description == "Default install of XenServer" || name_description == "Default install") // i18n: CA-30372, CA-207273
-                return Messages.DEFAULT_INSTALL_OF_XENSERVER;
+                return string.Format(Messages.DEFAULT_INSTALL_OF_XENSERVER, software_version["product_brand"]);
```
MihaelaStoica: To be on the safe side, we should check the key "product_brand" exists in software_version before calling `software_version["product_brand"]`. If the key doesn't exist, we can default to Branding.PRODUCT_BRAND

michael2012z: Reasonable. But I would prefer Messages.XENSERVER instead of Branding.PRODUCT_BRAND, as Branding is in XenAdmin and Host is in XenModel.

PR #2243, opened 2018-10-04
CP-29276: Add option to specify an update/supp pack in RPU wizard

jijiang:


XenAdmin/Wizards/RollingUpgradeWizard/PlanActions/RpuUploadAndApplySuppPackPlanAction.cs
```diff
@@ -0,0 +1,281 @@
...
+                            problems[0].Description));
+                }
+            }
+            catch (Exception ex)
+            {
+                log.Error(string.Format("Precheck failed on host {0}", host.Name()), ex);
+                throw;
+            }
+
+            if (livePatchStatus[host.uuid] != livepatch_status.ok_livepatch_complete && !hostsThatWillRequireReboot.Contains(host.uuid))
```
MihaelaStoica: livePatchStatus[host.uuid] can throw a KeyNotFoundException, because a host is added to this dictionary in PatchPrecheckCheck only if the check succeeds. If it fails with an error (e.g. patch not applicable ar already applied), then it is not added.

PR #2256, opened 2018-10-09
Disable automatic VDI Integrity check on export and scare the user if turned on

GeraldElder-Vass:
Set verify on export default to false, added a warning when checkbox is checked (true).

Visual Studio seems to have made some minor changes to files automatically, as I'm new I have no idea if this is an issue or not but I've tested the changes work and I can verify that they do.

CP-Ticket: https://issues.citrite.net/browse/CP-29106


kc284:
My apologies, I was not clear above, I meant the ticket number in the commit message:
`CP-29106: Set verify on export default to false, added a warning when checkbox is checked (true).`

GeraldElder-Vass:
Hi @kc284 I've made the requested changes, thanks for the thorough review; this is my first change to XenCenter :)
Let me know if anything else is off or needs changing.

GeraldElder-Vass:
I've made the changes from the patch you've sent me, and I've updated the commit message to include the CP ticket (though JIRA doesn't appear to be picking it up still)

PR #2258, opened 2018-10-10
L10N: Translations updates

DarrenMatthews776:

### DIFF
--- a/XenModel/InvisibleMessages.zh-CN.resx
+++ b/XenModel/InvisibleMessages.zh-CN.resx
@@ -133,7 +133,7 @@
     <value>https://cis.citrix.com/legal/</value>
   </data>
   <data name="HOMEPAGE" xml:space="preserve">
-    <value>https://www.citrix.com/products/citrix-hypervisor</value>
+    <value>https://www.citrix.com.cn/products/citrix-hypervisor/</value>
   </data>
   <data name="HOMEPAGE_FILENAME" xml:space="preserve">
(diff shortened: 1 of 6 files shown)


XenModel/InvisibleMessages.zh-CN.resx
```diff
@@ -133,7 +133,7 @@
     <value>https://cis.citrix.com/legal/</value>
   </data>
   <data name="HOMEPAGE" xml:space="preserve">
-    <value>https://www.citrix.com/products/citrix-hypervisor</value>
+    <value>https://www.citrix.com.cn/products/citrix-hypervisor/</value>
```
kc284: I think the Japanese resx needs also correction to https://www.citrix.co.jp/products/citrix-hypervisor/

DarrenMatthews776: I will look into this. The '.cn' link exists but there is no '.jp' or '.ja' link which exists.

DarrenMatthews776: Found the correct link for japenese. Changes should be coming now.

kc284: The link I pasted above does exist - note it's `.co.jp` not `.com.jp`.

DarrenMatthews776: O yeah, Thank you!

PR #2261, opened 2018-10-11
CA-299065: Duplicate host name in error message in the RPU and Patchi…

AaronRobson:
…ng wizards.


AaronRobson:
Please note that there may be many "Failure" sub-class "Description" property getters which need reviewing as some of them may need to include their respective Host names in it.
I am making this PR in order to discuss them.

MihaelaStoica:
The problems returned by the check in PatchPrecheckCheck are all OK, they all contain the host name in the description.
However, the automated updates mode (class PatchPrecheckOnHostPlanAction) has the same problem, see this [line](https://github.com/xenserver/xenadmin/blob/master/XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrecheckOnHostPlanAction.cs#L86)

PR #2265, opened 2018-10-11
A couple of random corrrections on error handling

kc284:
The one commit is to fix some InvalidOperationExceptions and XenAPI.Failures I kept seeing when debugging and I believe can be avoided. The other is meant to help improve the Actions code (if an action is reporting percentage out of range, it may be because we're calculating it incorrectly).
Feel free to squash the commits.

### DIFF
--- a/XenModel/Network/TaskPoller.cs
+++ b/XenModel/Network/TaskPoller.cs
@@ -98,55 +98,51 @@ namespace XenAdmin.Network
 
         private void poll()
         {
-            try
-            {
-                XenAPI.Task task = GetTask();
-                _action.Tick((int)(task.progress * _scale + _lo),
-                             task.Description() == "" ? _action.Description : task.Description());
-                switch (task.status)
-                {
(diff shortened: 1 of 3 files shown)


XenModel/Network/TaskPoller.cs
```diff
@@ -98,55 +98,51 @@ public void PollToCompletion()
...
-            {
-                XenAPI.Task task = GetTask();
-                _action.Tick((int)(task.progress * _scale + _lo),
-                             task.Description() == "" ? _action.Description : task.Description());
-                switch (task.status)
-                {
-                    case XenAPI.task_status_type.failure:
-                        log.Warn("Action failed due to API failure:\n" + Environment.StackTrace);
-                        throw new XenAPI.Failure(new List<string>(task.error_info));
+            Task task = GetTask();
```
MihaelaStoica: Is it possible that `GetTask()` throws the HANDLE_INVALID failure instead of returning a task with `task_status_type.failure`?

PR #2266, opened 2018-10-12
Random resource correction

kc284:


kc284:
Overall the images are not managed uniformly in XenCenter; the build action for half of them is Content, for the other half is None, while in reality we probably don't need to include them at all in the project because we add them to the Resources - the exception should be images that are used in the installer. But this is a different PR, here I have just made a small correction to conform with current practices.

PR #2269, opened 2018-10-15
CA-297086: Improve the function that uploads a server status report, …

MihaelaStoica:
…to avoit timeout exceptions

The changes included are:
- use a HttpClient object instead of HttpWebRequest
- use the same FileStream for all the chunk uploads
- retry the chunk upload if it fails (max three tries)

### DIFF
--- a/XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
+++ b/XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
@@ -33,7 +33,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net;
-using System.Text;
+using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Web.Script.Serialization;
 using XenAdmin;
 using XenAdmin.Network;
@@ -44,13 +45,17 @@ namespace XenServerHealthCheck
     public class XenServerHealthCheckUpload
     {
         private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
-        public string UPLOAD_URL = "https://rttf.citrix.com/feeds/api/";
-        public const int CHUNK_SIZE = 1 * 1024 * 1024;
-        private JavaScriptSerializer serializer;
-        private int verbosityLevel;
+        private readonly string uploadUrl = "https://rttf.citrix.com/feeds/api/";
 
-        private string uploadToken;
-        private IWebProxy proxy;
+        private const string UPLOAD_CHUNK_STRING = "{0}upload_raw_chunk/?id={1}&offset={2}";
+        private const string INITIATE_UPLOAD_STRING = "{0}bundle/?size={1}&name={2}";
+        private const int CHUNK_SIZE = 1 * 1024 * 1024;
+
+        private JavaScriptSerializer serializer;
+        private HttpClient httpClient;
+
+        private readonly int verbosityLevel;
+        private readonly string uploadToken;
 
         public XenServerHealthCheckUpload(string token, int verbosity, string uploadUrl, IXenConnection connection)
         {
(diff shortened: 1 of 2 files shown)


XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
```diff
@@ -175,62 +159,60 @@ public string UploadZip(string fileName, string caseNumber, System.Threading.Can
...
-                            // This chunk is successfully uploaded
-                            offset += thisChunkSize;
-                            break;
-                        }
-
-                        // Fail to upload the chunk for 3 times so fail the bundle upload.
-                        if (i == 2)
+                    try
+                    {
+                        for (int i = 0; i < 3; i++)
```
kc284: Let's put 3 in a constant and use it further down, so as to have one place to change if we update the number of retries.

XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
```diff
@@ -214,5 +229,12 @@ public string UploadZip(string fileName, string caseNumber, System.Threading.Can
             log.InfoFormat("Succeeded to upload bundle {0}", fileName);
             return uploadUuid;
         }
+
+        #region IDisposable Members
+        public void Dispose()
+        {
+            httpClient.Dispose();
```
kc284: The HttpClientHandler is missing before httpClient (it will have to be promoted from local variable to class field).

XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
```diff
@@ -132,23 +139,31 @@ private bool UploadChunk(string address, FileStream source, long offset, long th
...
+                        if (response.IsSuccessStatusCode)
+                        {
+                            var respString = response.Content.ReadAsStringAsync().Result;
+
+                            // After sending every chunk upload request to server, response will contain a status indicating if it is complete.
+                            var res = (Dictionary<string, object>) serializer.DeserializeObject(respString);
+                            log.InfoFormat("The status of chunk upload: {0}", res.ContainsKey("status") ? res["status"] : "");
+                            return true;
+                        }
+                        log.ErrorFormat("Failed to upload the chunk. The exception was: {0} ({1})", response.StatusCode, response.ReasonPhrase);
```
kc284: Strictly speaking this is not an exception, I'd rephrase the message to `Failed to upload the chunk. POST request returned {0} ({1})`.

XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
```diff
@@ -80,34 +80,41 @@ public XenServerHealthCheckUpload(string token, int verbosity, string uploadUrl,
...
+                        if (response.IsSuccessStatusCode)
+                        {
+                            var respString = response.Content.ReadAsStringAsync().Result;
+
+                            // Get the upload uuid
+                            var res = (Dictionary<string, object>) serializer.DeserializeObject(respString);
+                            if (res.ContainsKey("id"))
+                                return (string) res["id"];
+                        }
+                        log.ErrorFormat("Exception while initiating a new CIS upload. The exception was: {0} ({1})", response.StatusCode, response.ReasonPhrase);
```
kc284: Strictly speaking this is not an exception. I'd rephrase the message to `Failed to initiate a CIS upload. POST request returned {0} ({1)}`.
Also, not sure if it is possible, but if the status code is 200 and the response does not contain the id, the error will still be logged, which is misleading. We can either put the log in an else block, or preferably (to match the upload chunk case) do `return res.ContainsKey("id") ? (string) res["id"] : "";`

XenModel/Actions/HealthCheck/XenServerHealthCheckUpload.cs
MihaelaStoica: The HttpClientHandler is disposed automatically by HttpClient.Dispose(), because the HttpClient constructor used, sets the internal flag `disposeHandler` to true by default. See the constructor implementation in the HttpClient class:
``` 
    /// <summary>Initializes a new instance of the <see cref="T:System.Net.Http.HttpClient" /> class with a specific handler.</summary>
    /// <param name="handler">The HTTP handler stack to use for sending requests. </param>
    [__DynamicallyInvokable]
    public HttpClient(HttpMessageHandler handler)
      : this(handler, true)
    {
    }

    /// <summary>Initializes a new instance of the <see cref="T:System.Net.Http.HttpClient" /> class with a specific handler.</summary>
    /// <param name="handler">The <see cref="T:System.Net.Http.HttpMessageHandler" /> responsible for processing the HTTP response messages.</param>
    /// <param name="disposeHandler">
    /// <see langword="true" /> if the inner handler should be disposed of by Dispose(), <see langword="false" /> if you intend to reuse the inner handler.</param>
    [__DynamicallyInvokable]
    public HttpClient(HttpMessageHandler handler, bool disposeHandler)
      : base(handler, disposeHandler)
    {
    ....
```
However, as this is not clear in the code, I will change it to explicitily use the second constructor, `HttpClient(HttpMessageHandler handler, bool disposeHandler)`.

kc284: Fair enough.

PR #2270, opened 2018-10-15
CP-14807: Delete OEM in ServerVersion

AaronRobson:


AaronRobson:
I agree, I'll let you know when those constants and moved and the classes are cleaned out.

AaronRobson:
I believe the second commit here should cover those unused classes.

PR #2272, opened 2018-10-16
CA-203431: Host name is repeated twice in multipathing alert message in XC

AaronRobson:

### DIFF
--- a/XenAdminTests/UnitTests/AlertTests/MessageAlertTest.cs
+++ b/XenAdminTests/UnitTests/AlertTests/MessageAlertTest.cs
@@ -1,0 +1,97 @@
+﻿/* Copyright (c) Citrix Systems, Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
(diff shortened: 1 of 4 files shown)


AaronRobson:
The first commit is to resolve the initial separator issue and refactor, the second resolve the duplicate issue.
Note: I have an idea about unit testing this functionality which I will discuss in the ticket.
Please feel free to squash into a single commit.

XenAdminTests/UnitTests/AlertTests/MessageAlertTest.cs
```diff
@@ -0,0 +1,97 @@
...
+ 
+using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+using XenAdmin.Alerts;
+
+namespace XenAdminTests.UnitTests.AlertTests
+{
+    [TestFixture, Category(TestCategories.Unit)]
+    public class MessageAlertTest
```
MihaelaStoica: I think this class should be named MultipathAlertTest, because it tests a function specific to the multipath message

PR #2275, opened 2018-10-17
CA-172658: Descriptive error message when failing to suspend a VM fro…

AaronRobson:
…m XenCenter.

PCI passthrough from VGPU is checked first and informing the user that the VM lacks the Suspend feature as a fallback.

### DIFF
--- a/XenAdmin/Commands/SuspendVMCommand.cs
+++ b/XenAdmin/Commands/SuspendVMCommand.cs
@@ -201,9 +201,17 @@ namespace XenAdmin.Commands
             {
                 return Messages.VM_SHUT_DOWN;
             }
-            else if (vm.power_state == vm_power_state.Suspended)
+            if (vm.power_state == vm_power_state.Suspended)
             {
                 return Messages.VM_ALREADY_SUSPENDED;
+            }
+            if (vm.HasGPUPassthrough())
+            {
+                return FriendlyErrorNames.VM_HAS_PCI_ATTACHED;
+            }
+            if (vm.allowed_operations != null && !vm.allowed_operations.Contains(vm_operations.suspend))
+            {
+                return FriendlyErrorNames.VM_LACKS_FEATURE_SUSPEND;
             }
 
             return GetCantExecuteNoToolsOrDriversReasonCore(item) ?? base.GetCantExecuteReasonCore(item);


XenAdmin/Commands/SuspendVMCommand.cs
```diff
@@ -201,10 +201,18 @@ protected override string GetCantExecuteReasonCore(SelectedItem item)
...
                 return Messages.VM_SHUT_DOWN;
             }
-            else if (vm.power_state == vm_power_state.Suspended)
+            if (vm.power_state == vm_power_state.Suspended)
             {
                 return Messages.VM_ALREADY_SUSPENDED;
             }
+            if (vm.HasGPUPassthrough())
+            {
+                return FriendlyErrorNames.VM_HAS_PCI_ATTACHED;
```
MihaelaStoica: FriendlyErrorNames.VM_HAS_PCI_ATTACHED says "This operation could not be performed..." and FriendlyErrorNames.VM_LACKS_FEATURE_SUSPEND says "You attempted an operation...". These messages are not suitable for a **tooltip**, because at that point no operation has been attempted.

XenAdmin/Commands/SuspendVMCommand.cs
```diff
@@ -201,10 +201,18 @@ protected override string GetCantExecuteReasonCore(SelectedItem item)
...
+            }
+            if (vm.allowed_operations != null && !vm.allowed_operations.Contains(vm_operations.suspend))
```
MihaelaStoica: With this change, GetCantExecuteNoToolsOrDriversReasonCore(item) will never be reached, so we will never see the "tools not installed" reason.

XenAdmin/Commands/SuspendVMCommand.cs
AaronRobson: You suggested yesterday using VM.assert_operation_valid, would this be what I should be aiming for here as well?

XenAdmin/Commands/SuspendVMCommand.cs
AaronRobson: I will move the GetCantExecuteNoToolsOrDriversReasonCore(item) call above the newer checks.

XenAdmin/Commands/SuspendVMCommand.cs
AaronRobson: As per our discussion this is resolved for pool-admin and higher user roles using the API call VM.assert_operation_valid without needing to independently figure out what the error could be.

However it will revert back to "Unknown" for roles vm-operator (lowest that is allowed to Suspend) up to but not including pool-admin (assert_operation_valid minimum).

MihaelaStoica: I think that we shouldn't use the assert_operation_valid api call if it does not work for all users (although fixed in master, XenCenter needs a solution for older hosts as well). It's a shame, because it was a good idea. 
So I think we need to go back to the initial idea and capture the two cases (GPU passthrough and vGPU) and present specific "reasons" for those two.

AaronRobson: I have committed for discussion my current best estimate of the changes required. I may need assistance in replicating the issue.

AaronRobson: I have been able to overcome the problems I was having replicating the issue, this would appear to me to be now a workable solution.

PR #2281, opened 2018-10-22
CA-271945: Attempting to resolve the pool's connection and searching …

AaronRobson:
…for it in all open connections for the EnableHAAction in case a host has been destroyed or the master changed.
Preventing an Object (host) has been deleted error when attempting to reenable Maintenance Mode on a Host that has been kicked out of the pool when exiting the RPU Wizard.
Returning null to prevent making orphened tasks that will be stuck at 0% completion if it is impossible to fulfill.

### DIFF
--- a/XenAdmin/Diagnostics/Problems/HostProblem/HostMaintenanceMode.cs
+++ b/XenAdmin/Diagnostics/Problems/HostProblem/HostMaintenanceMode.cs
@@ -63,8 +63,12 @@ namespace XenAdmin.Diagnostics.Problems.HostProblem
 
         public override AsyncAction CreateUnwindChangesAction()
         {
-            Program.MainWindow.CloseActiveWizards(Server.Connection);
-            return new DisableHostAction(Server);
+            var server = Server.Connection.Resolve(new XenRef<Host>(Server.opaque_ref));
+            if (server == null)
+                return null;
+
+            Program.MainWindow.CloseActiveWizards(server.Connection);
+            return new DisableHostAction(server);
         }
 
     }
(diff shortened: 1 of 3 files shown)


AaronRobson:
I've requested review from @kc284 in addition to @MihaelaStoica as I understand you've worked with the Host Maintenance Mode recently and I wanted to double check that this won't break your work.

XenAdmin/Diagnostics/Problems/VMProblem/AutoStartEnabled.cs
```diff
@@ -88,6 +88,9 @@ private Action<Session> ActionDelegate(bool autostartValue)
 
         public override AsyncAction CreateUnwindChangesAction()
         {
+            if (VM.Connection.Resolve(new XenRef<VM>(VM.opaque_ref)) == null) // check if the vm is still in the cache
```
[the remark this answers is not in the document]
AaronRobson: Please see new commit.

XenAdmin/Diagnostics/Problems/HostProblem/HostMaintenanceMode.cs
```diff
@@ -63,8 +63,12 @@ public override string HelpMessage
 
         public override AsyncAction CreateUnwindChangesAction()
         {
-            Program.MainWindow.CloseActiveWizards(Server.Connection);
-            return new DisableHostAction(Server);
+            var server = Server.Connection.Resolve(new XenRef<Host>(Server.opaque_ref));
+            if (server == null)
+                return null;
+
+            Program.MainWindow.CloseActiveWizards(server.Connection);
```
kc284: One more thing that I didn't notice the first time I read the code: I think we want to close the active wizards on all occasions even if we can't resolve the server, in which case this line should move above.

AaronRobson: I understand, I will move it above the Resolve line.

PR #2282, opened 2018-10-23
CA-300368: Sanitize fields values when loading a xapi xml database in…

MihaelaStoica:
… XenCenter

This fixes the bug where importing a xapi database to xencenter fails with "This server version is incompatible with XenCenter" due to the product brand containing a space character.

### DIFF
--- a/XenModel/ServerDBs/StatusReportXmlDocReader.cs
+++ b/XenModel/ServerDBs/StatusReportXmlDocReader.cs
@@ -45,6 +45,15 @@ namespace XenAdmin.ServerDBs
     /// </summary>
     class StatusReportXmlDocReader
     {
+        private static readonly Dictionary<string, string> escapedCharacters = new Dictionary<string, string>
+        {
+            { "%.", " " },
+            { "%t", "\t" },
+            { "%n", "\n" },
+            { "%r", "\r" },
+            { "%%", "%" }
+        };
+
         /// <summary>
         /// Populates the specified Db with the specified XML document.
         /// </summary>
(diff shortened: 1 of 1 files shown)


kc284:
Shouldn't the ticket number be CA-300368?

XenModel/ServerDBs/StatusReportXmlDocReader.cs
```diff
@@ -128,6 +128,24 @@ private string ParsePropertyName(string p, string table)
...
+            var escapedCharacters = new Dictionary<string, string>
+            {
+                { "%.", " " },
+                { "%t", "\t" },
+                { "%n", "\n" },
+                { "%r", "\r" },
+                { "%%", "%" }
+            };
+            string newValue = value;
+            foreach (var scapedCharacter in escapedCharacters)
```
kc284: *e*scapedCharacter

MihaelaStoica:
Yes, sorry, too early in the morning 😴

XenModel/ServerDBs/StatusReportXmlDocReader.cs
```diff
@@ -128,6 +128,24 @@ private string ParsePropertyName(string p, string table)
             }
         }
 
+        private string SanitizePropertyValue(string value)
```
timliu55: The method SanitizePropertyValue() is inside a loop, thus this method has to be static, otherwise every time it is called, a new instance of Dictionary escapedCharacters will be constructed. It is a lot of memory, I think.
B.T.W., would you think it is a better idea to do the replacement before the loop, something like doc.InnerXml = SanitizePropertyValue(doc.InnerXml); I am not quite sure.

MihaelaStoica: Good point about having the escapedCharacters dictionary outside of the function.
I think sanitizing the values is good enough here, as they are the only strings in the XML that would need replacing anyway.

MihaelaStoica:
@kc284 Are you happy with these changes?

PR #2286, opened 2018-10-25
CA-235084, CA-235085, CA-209495: Layout corrections

kc284:


kc284:
@DarrenMatthews776 when this change is merged and you update the resources, please check whether the second column of tableLayoutPabel1 on both controls might need to be bigger in JA than the EN value of 250.

kc284:
@DarrenMatthews776 could you please review the commit for CA-209495? Visual Studio has removed automatically JA/CH resource entries that have the same value as in EN, however I have noticed that on many occasions you leave them in - which practice is better?

DarrenMatthews776:
@kc284 I wasnt aware that Visual Studio does this. We like to have all resource entries that are in EN present in the JA/CH resx files too. Its so that if we do need to change something we can easily alter the correct file. However with our process even if they are automatically removed we will replace them back with 'resyncs' and translations anyways.

DarrenMatthews776:
@kc284 Thank you, I will make sure I check the UI as mentioned above and change the file if needed.

kc284:
Thanks, in this case I'll just leave the commit as is and it can be synced later.

PR #2287, opened 2018-10-25
CA-290940: Versions newer than those we have on record  should be int…

AaronRobson:
…erpreted as being too new for any automatic updates to have been release yet, not that they do not support automated updates.

Returning an empty list instead of null in this case with associated unit tests.


AaronRobson:
I have redone the history to remove two commits related to an older approach (and its subsequent reversion) for clarity of review and merging.

PR #2296, opened 2018-11-02
CA-301193: XenCenter UI looks ugly after installation of latest Win10 CU update

AaronRobson:
Avoid calling ProfessionalColors before we have called Application.EnableVisualStyles as it causes visual styles to be prevented from being enabled and can prevent the program from displayed as expected as the value of Application.InitializeComCtlSupportsVisualStyles is found only one time when first needed and cached. If Application.useVisualStyles is not true as in this case it relies on calling "comctl32.dll" methods to determine if this is the case and after KB4462933 this is no longer the case.

### DIFF
--- a/XenAdmin/Program.cs
+++ b/XenAdmin/Program.cs
@@ -89,10 +89,14 @@ namespace XenAdmin
         public static Font DefaultFontItalic;
         public static Font DefaultFontHeader;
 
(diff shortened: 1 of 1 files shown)


kc284:
Ah wait, one thing I missed: after this change the TitleBarStartColor and TitleBarEndColor will have to be set in the else block of `case 6` to the ProfessionalColours - no gradient will be shown, but the colours need to be the ones of the framework not the default ones we've chosen above.

PR #2298, opened 2018-11-05
CP-27819: Switch to Visual Studio 2017

AaronRobson:
Upgrading solutions and projects including the latest ToolsVersion and tidying up after the upgrades.

### DIFF
--- a/XenAdmin.sln
+++ b/XenAdmin.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.40629.0
+# Visual Studio 15
+VisualStudioVersion = 15.0.28010.2048
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "XenAdmin", "XenAdmin\XenAdmin.csproj", "{70BDA4BC-F062-4302-8ACD-A15D8BF31D65}"
 	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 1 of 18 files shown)


splash/splash.vcxproj
```diff
@@ -117,6 +117,12 @@
   <ItemGroup>
     <ResourceCompile Include="splash.rc" />
   </ItemGroup>
+  <ItemGroup>
```
kc284: Is this needed?

XenAdmin/XenAdmin.csproj
```diff
@@ -14,11 +14,6 @@
     <SignManifests>false</SignManifests>
     <ManifestCertificateThumbprint>4C1DD393E361B8BBCFA1E1D09539C446F88C0A11</ManifestCertificateThumbprint>
```
kc284: Not related to this PR, but I've just noticed the tags ManifestCertificateThumbprint and ManifestTimestampUrl which exist in some of the solution's projects and are ancient; we've been using `mk\sign.bat` for signing for a while now.

splash/splash.vcxproj
AaronRobson: You are right, this does not appear to be required for it to build and run, I'm not sure why the solution/project updater would add it. I will remove it.

XenAdmin/XenAdmin.csproj
AaronRobson: I see, would removing both these nodes from each project they are present in be your recommended course of action?

XenAdmin.sln
```diff
@@ -13,9 +13,6 @@ EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "xva_verify", "xva_verify\xva_verify.csproj", "{2A70D7E7-EAB2-4C36-B3F4-85B79D2384B5}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "splash", "splash\splash.vcxproj", "{AFB19C9D-DD63-478B-A4A3-8452CBD0B9AB}"
-	ProjectSection(ProjectDependencies) = postProject
```
kc284: I've also noticed this bit has been removed.

AaronRobson: What was this bit doing before it was removed? Ensuring that another project identified by this GUID is compiles after the splash project perhaps?

kc284: I did some more googling, it seems this was needed previously to keep track of dependencies and after the upgrade it's moved to the proejcte references - which explains now the change in splash.vcxproj. Ok, my bad, let's leave this here and revert 6af64589103676ed57380eb3b2c2261a375aa779.

PR #2299, opened 2018-11-05
Colour correction; redundant code removal; refactoring

kc284:
WIth the opportunity of reviewing CA-301193 I did some further work:

Removed redundant code from GradientPanel and made it more object oriented;
also moved it out of the containing folder as it was its only resident.
Corrected the TitleBar fore colour as it was hard to read due to small contrast
with the bar's backround colour; normalised the code setting the other colours
so it's more unambiguous.


AaronRobson:
I can confirm that these changes have not caused any obviously noticable visual regression for the changes for CA-301193 on my system. I'd recommend someone else gives it a more thorough consideration.

PR #2306, opened 2018-11-08
Fixed PatchingWizard crash; removed obsolete UpdateType.Existing.

kc284:
Also refactored error handling with reagrs to selecting single patches.


XenAdmin/Wizards/PatchingWizard/PatchingWizard_SelectPatchPage.cs
```diff
@@ -204,96 +203,78 @@ protected override void PageLeaveCore(PageLoadedDirection direction, ref bool ca
...
+                    SelectedNewPatch = FilePath;
 
-                        using (var dlg = new ThreeButtonDialog(
-                            new ThreeButtonDialog.Details(SystemIcons.Warning,
-                                string.Format(Messages.UPDATES_WIZARD_CANNOT_DOWNLOAD_PATCH, SelectedExistingPatch.Connection.Name),
-                                Messages.UPDATES_WIZARD)))
+                    if (!string.IsNullOrEmpty(FilePath) && Path.GetExtension(FilePath).ToLowerInvariant().Equals(".zip"))
+                    {
+                        //check if we are installing update user sees in textbox
+                        if (unzippedUpdateFilePath == null || Path.GetFileNameWithoutExtension(unzippedUpdateFilePath) !=
```
MihaelaStoica: If the unzipped file is not there anymore (for example because I cleaned my Temp folder), the wizard shows the invalid file errror, instead of trying to unzip it again.

MihaelaStoica:
I had noticed another issue, although not caused by these changes: If I select to install an update from disk, but that update is already applied on my host, the wizard still lets me to select the host. It should show the host disabled with the "update already applied" tooltip.

kc284:
I have rebased all on master and the timestamps changed, so to recap: the commit f3d63fa  is unchanged; in 2460fd6 I changed the lines around the comment; the commit 2d2d7a8 is new.

PR #2309, opened 2018-11-13
CA-302791: Rewrote the ChangeFolder dialog to fix usability issues.

kc284:
- Corrected folder selection; reduced clicks required to select folders;
  improved text; added delete button.
- Moved FolderChangeDialogTreeView to the same folder as the other TreeViews;
- Added missing null checks to the MutliSelectTreeView and hid properties
  HScrollPos and VScrollPos from VS's designer (the former was initialised
  to zero every time the designer was updated, causing the treeview images
  to disappear).

### DIFF
--- a/XenAdmin/Commands/NewFolderCommand.cs
+++ b/XenAdmin/Commands/NewFolderCommand.cs
@@ -31,17 +31,12 @@
 
 using System;
 using System.Collections.Generic;
-using System.Text;
 using XenAdmin.Network;
 using XenAdmin.Model;
 using System.Windows.Forms;
-using XenAdmin.Properties;
 using System.Drawing;
 using XenAdmin.Dialogs;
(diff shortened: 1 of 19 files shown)


AaronRobson:
I can confirm that this is a marked improvement in the previous behaviour of this screen.

The only minor issues I have with it are:
1. If deleting the selected folder I would expect it to select the parent if available (preferably expanded if there are other folders available).
2. Have the delete key trigger the folder deletion confirmation.
3. If we could have it remember which folders were expanded when the display is refreshed following a deletion that would help the user understand where they are if there are lots of folders.
4. Clicking the expand icon and then changing my mind resulted in it counting at a double click I presume and Oking the form, I would not have expected this.

Other suggestions:
1. Folder renaming functionality button which can also be triggered with the F2 key.

PR #2314, opened 2018-11-15
CA-301904: Occasional failure to upload status report to CIS

AaronRobson:
The retry mechanism commit is designed to resolve the reported issue.
However I recommend icluding the other commits as one of them cancels the Upload step if the Download/Construction step failed as it would not be able to succeed anyway, resolving a small issue introduced when the code was changed to use MultipleAction in CA-204242 and the other commit is a changing the comment of the MultipleAction to be a summary.


kc284:
Although the changes look reasonable, they don't correct the problem. I'm still seeing the occasional failure, even if I increase the delay to 1sec. Note that the problem occurs with big zips (I can reproduce it with zips of 200 or 300MB, while it never occurs with zips of 50MB).

AaronRobson:
I will attempt to reproduce the issue on my machine with larger zips (200-300MB).

kc284:
I'm closing this PR as the problem seems to require further investigation and potentially a different solution.

PR #2316, opened 2018-11-16
CA-272847: Compilation time is way overestimated (when creating a ser…

AaronRobson:
…ver status report).

As the times are made up of server timeouts they overestimate quite significantly, we are removing the time estimates from the screen.

### DIFF
--- a/XenAdmin/Wizards/BugToolWizardFiles/BugToolPageSelectCapabilities.Designer.cs
+++ b/XenAdmin/Wizards/BugToolWizardFiles/BugToolPageSelectCapabilities.Designer.cs
@@ -43,14 +43,11 @@ namespace XenAdmin.Wizards.BugToolWizardFiles
             this.DescriptionValue = new System.Windows.Forms.Label();
             this.SizeLabel = new System.Windows.Forms.Label();
             this.SizeValue = new System.Windows.Forms.Label();
-            this.TimeLabel = new System.Windows.Forms.Label();
-            this.TimeValue = new System.Windows.Forms.Label();
             this.TotalSizeLabel = new System.Windows.Forms.Label();
             this.TotalSizeValue = new System.Windows.Forms.Label();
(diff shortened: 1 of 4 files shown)


XenAdmin/Wizards/BugToolWizardFiles/BugToolPageSelectCapabilities.Designer.cs
```diff
@@ -43,14 +43,11 @@ private void InitializeComponent()
...
             this.SizeLabel = new System.Windows.Forms.Label();
             this.SizeValue = new System.Windows.Forms.Label();
-            this.TimeLabel = new System.Windows.Forms.Label();
-            this.TimeValue = new System.Windows.Forms.Label();
             this.TotalSizeLabel = new System.Windows.Forms.Label();
             this.TotalSizeValue = new System.Windows.Forms.Label();
-            this.TotalTimeLabel = new System.Windows.Forms.Label();
-            this.TotalTimeValue = new System.Windows.Forms.Label();
             this.tableLayoutPanel3 = new System.Windows.Forms.TableLayoutPanel();
             this.label1 = new System.Windows.Forms.Label();
```
kc284: We need to also update the text of label1 because it still says you can see the estimated time.

kc284:
The following are not caused by this PR but we may want to fix them:
- the tab indices of the controls  are not in the right order.
- I think the total size label looks a bit untidy; its bottom should be aligned with the bottom of the select/clear buttons; we can move it because there is plent of empty space above it.

The above are related to the containers of the page being a bit odd. There is a split container for which I don't see any reason: it is not visible to the users, they can only find out if they move the mouse accidentally above it, and if they move it far to the left the link label is wrapped and looks ugly. TableLayoutPanel3 should probably be a 3x2 panel with label1 and linklabel1 occupying the top and bottom rows respectively and TableLayoutPanels 3 and 4 occupying each of the cells of the middle row.

PR #2318, opened 2018-11-16
CA-300384: XC unable to create (GFS2) VDI of virtual-size larger than…

AaronRobson:
… available storage when creating new VM.

Calling SR.VdiCreationCanProceed on the total space required of each SR as it takes into account that Thinly Provisioned SRs may overprovision in the same way that the NewDiskDialog does by effectively ignoring the free space limit.
IsSufficientFreeSpaceAvailableOnSrForVdi was not taking into account Thin Provisioning when checking if the virtual disk would fit on the SR and as a consequence would attempt to find another SR to place it on silently; either choosing "Local Storage" or "<no suitable storage>" depending on the size requested relative to the SR sizes.

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
+++ b/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
@@ -261,7 +261,7 @@ namespace XenAdmin.Wizards.NewVMWizard
                 if (item.Disk.SR.opaque_ref != sr.opaque_ref)
                     continue;
 
-                if (sr.FreeSpace() < totalDiskInitialAllocation[sr.opaque_ref])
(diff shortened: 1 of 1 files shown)


kc284:
By inline I mean a warning like the error on the NewDiskDialog: a togglable panel on the control as opposed to a pop-up. By non-blocking I mean that displaying the warning should not be accompanied by disabling the next button (which is the error case).

MihaelaStoica:
A warning is a good idea, and I think we already have this capability on this page, see `DiskOverCommit.Warning`

kc284:
Ah true, in which case the above change needs some modification to show the right overcommit state.

AaronRobson:
Agreed, I will work on hooking into the warning mechanism of the grid to inform the user that over-provisioning is occurring, but not block them from doing so if they choose to do so.

PR #2322, opened 2018-11-19
CA-280340: Currently used memory for a host is slightly different fro…

timliu55:
…m Used memory

Use RRds to calculate used memory
Passingly fix the wrong layout

### DIFF
--- a/XenAdmin/Controls/Ballooning/HostMemoryControls.cs
+++ b/XenAdmin/Controls/Ballooning/HostMemoryControls.cs
@@ -110,7 +110,6 @@ namespace XenAdmin.Controls.Ballooning
         public void UpdateMemoryByMetricUpdater()
         {
             if (MetricUpdater == null) return;
-            MetricUpdater.Prod();
             var free = MetricUpdater.GetValue(host, "memory_free_kib");
             var total = MetricUpdater.GetValue(host, "memory_total_kib");
             var used = (total - free) * Util.BINARY_KILO;
--- a/XenAdmin/Controls/Ballooning/HostMemoryControls.cs
+++ b/XenAdmin/Controls/Ballooning/HostMemoryControls.cs
@@ -98,11 +98,14 @@ namespace XenAdmin.Controls.Ballooning
 
             // Set the text values
             valueTotal.Text = Util.MemorySizeStringSuitableUnits(total, true);
-            UpdateMemoryByMetricUpdater();
             valueAvail.Text = Util.MemorySizeStringSuitableUnits(avail, true);
             valueTotDynMax.Text = Util.MemorySizeStringSuitableUnits(tot_dyn_max, true);
             labelOvercommit.Text = string.Format(Messages.OVERCOMMIT, overcommit);
             valueControlDomain.Text = Util.MemorySizeStringSuitableUnits(dom0, true);
+            if (isMetricUpdating)
+                valueUsed.Text = valueTotDynMax.Text;
+            else
+                UpdateMemoryByMetricUpdater();
         }
 
         public MetricUpdater MetricUpdater { get; set; }
@@ -122,10 +125,33 @@ namespace XenAdmin.Controls.Ballooning
                 this.Refresh();
         }
 
+        static bool isMetricUpdating;
+        Timer metricUpdaterTimer;
         void host_PropertyChanged(object sender, PropertyChangedEventArgs e)
         {
             if (e.PropertyName == "memory_overhead")
                 this.Refresh();
+            else if (e.PropertyName == "resident_VMs" && MetricUpdater != null)
+            {
+                isMetricUpdating = true;
+                this.Refresh();
+                if (metricUpdaterTimer == null)
+                {
+                    metricUpdaterTimer = new Timer();
+                    metricUpdaterTimer.Interval = 15000;
+                    metricUpdaterTimer.Tick += MetricUpdaterTimer_Tick;
+                }
+                metricUpdaterTimer.Enabled = true;
+            }
+        }
+
+        private void MetricUpdaterTimer_Tick(object sender, EventArgs e)
+        {
+            if (MetricUpdater == null) return;
+            MetricUpdater.Prod();
+            isMetricUpdating = false;
+            this.Refresh();
+            metricUpdaterTimer.Enabled = false;
         }
 
         void vm_metrics_PropertyChanged(object sender, PropertyChangedEventArgs e)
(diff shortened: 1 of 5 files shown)


timliu55:
If you test the RRDs update by "start/shutdown VM", you will see a long time delay.
How about calculating used memory by "total memory minus available memory"?

XenAdmin/Controls/Ballooning/HostMemoryControls.cs
```diff
@@ -104,13 +98,25 @@ protected override void OnPaint(PaintEventArgs e)
...
             labelOvercommit.Text = string.Format(Messages.OVERCOMMIT, overcommit);
             valueControlDomain.Text = Util.MemorySizeStringSuitableUnits(dom0, true);
         }
 
+        public MetricUpdater MetricUpdater { get; set; }
+
+        public void UpdateMemoryByMetricUpdater()
+        {
+            if (MetricUpdater == null) return;
+            MetricUpdater.Prod();
```
MihaelaStoica: Calling `Prod` here causes another immediate update from the MetricUpdater, resulting in this page being continuously refreshed. This will happen on every paint event and on every other PropertyChanged event that we handle on this page. Also, because the `UpdateMemoryByMetricUpdater` is called from the MetricsUpdated event handler itself, it will again cause an immediate update of the metrics.

XenAdmin/Controls/Ballooning/HostMemoryRow.cs
```diff
@@ -64,5 +65,16 @@ internal void UnregisterHandlers()
             memoryRowLabel.UnsubscribeEvents();
             hostMemoryControls.UnregisterHandlers();
         }
+
+        public MetricUpdater MetricUpdater
```
MihaelaStoica: I think we should have the MetricUpdater as a read-only property in this class, add a metricUpdater parameter to the constructor, where we assign it directly to hostMemoryControls.MetricUpdater

MihaelaStoica:
1. I am not sure that we can avoid the delay if we are using the RRDs. I have tried calling `MetricUpdater.Prod();` before the `Refresh()` calls we do on host and VM PropertyChanged event. But these refreshes still didn't get the final value for the used memory, only an intermediate update of that value, so in the end the used memory was refreshed with the correct value only after 30 seconds.

2. We cannot use available memory because it it not the same as free memory. See this comment comment in code:
```
/// <summary>
/// The amount of available memory on the host. This is not the same as the amount of free memory, because
/// it includes the memory that could be freed by reducing balloonable VMs to their dynamic_minimum memory.
/// </summary>
public long memory_available_calc()
```

XenAdmin/Controls/Ballooning/HostMemoryControls.cs
```diff
@@ -117,10 +125,33 @@ void vm_PropertyChanged(object sender, PropertyChangedEventArgs e)
...
+                    metricUpdaterTimer.Tick += MetricUpdaterTimer_Tick;
+                }
+                metricUpdaterTimer.Enabled = true;
+            }
+        }
+
+        private void MetricUpdaterTimer_Tick(object sender, EventArgs e)
+        {
+            if (MetricUpdater == null) return;
+            MetricUpdater.Prod();
```
MihaelaStoica: It would probably be safer to disable the timer before we do anything else in this method (i.e. move `metricUpdaterTimer.Enabled = false;` here)

XenAdmin/Controls/Ballooning/HostMemoryRow.cs
```diff
@@ -47,6 +48,13 @@ public HostMemoryRow(Host host)
             this.host = host;
         }
 
+        public HostMemoryRow(Host host, MetricUpdater metricUpdater)
```
MihaelaStoica: The `HostMemoryRow(Host host)` constructor in not used anymore, so it can be removed (as it has been replaced by this one)

XenAdmin/Controls/Ballooning/HostMemoryControls.cs
```diff
@@ -117,10 +125,33 @@ void vm_PropertyChanged(object sender, PropertyChangedEventArgs e)
...
+        static bool isMetricUpdating;
+        Timer metricUpdaterTimer;
         void host_PropertyChanged(object sender, PropertyChangedEventArgs e)
         {
             if (e.PropertyName == "memory_overhead")
                 this.Refresh();
+            else if (e.PropertyName == "resident_VMs" && MetricUpdater != null)
+            {
+                isMetricUpdating = true;
+                this.Refresh();
```
MihaelaStoica: `this` qualifier is redundant

XenAdmin/Controls/Ballooning/HostMemoryControls.cs
```diff
@@ -117,10 +125,33 @@ void vm_PropertyChanged(object sender, PropertyChangedEventArgs e)
                 this.Refresh();
         }
 
+        static bool isMetricUpdating;
```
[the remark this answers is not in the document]
timliu55: When we power off a VM, the memory tab page will be rebuilt for several times, and the HostMemoryControls will be re-constructed again and again, which means, the instance of HostMemoryControls which creates a Timer will be lost when the Timer's tick event triggers. And at that moment you will always see a new instance of HostMemoryControls with a false isMetricUpdating (if it is not static)

timliu55:
Mihaela, I am sorry for my rough code, and next time I will refactor my code as much as possible.

XenAdmin/Controls/Ballooning/HostMemoryControls.cs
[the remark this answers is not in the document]
MihaelaStoica: Oh, I didn't realise that the controls are re-constructed. 
The problem with the static field is that, if I have multiple hosts in a pool, on the pool's memory tab (where we show all the hosts) we show the calculated used memory for all hosts, even though just one of the hosts is changing when I start/stop a VM.

kc284:
@timliu55 could you please separate the designer changes from this PR and submit them as a new PR that will be merged immediately. The rest of the changes will be cherry-picked later after the controls are restructured. Thanks.

timliu55:
Got it. 
A new PR is created as
https://github.com/xenserver/xenadmin/pull/2342

kc284:
Thanks. PR#2342 is merged now. I have exported this PR to a patch and attached it to the ticket so we can close this PR until the control restructuring is done.

PR #2328, opened 2018-11-23
CA-301907: Snapshot schedules should be showing local times.

AaronRobson:


AaronRobson:
I do have a concern that although I have dev tested changing my time zone I'm not sure how to go about changing the server's time zone.

XenModel/Messages.resx
```diff
@@ -10960,6 +10960,10 @@ Click Keep Waiting if the upgrade is still running or Cancel if an error occurre
   <data name="SERVER_TIME" xml:space="preserve">
     <value>The current time on the pool master is: {0}</value>
   </data>
+  <data name="SERVER_TIME_LOCAL_TIME" xml:space="preserve">
+    <value>The current time on the pool master is: {0}
```
kc284: When we say "the current time on the master is", we mean the time on the server in the server's time zone, not the client time corresponding to it, so this message is slightly confusing.
Beyond this, and as mentioned on the ticket, the intention is to make the behaviour a bit more user-centric, i.e. I don't find it very useful to know the current server time (which is not current either, because it doesn't tick, it's rather the time when the dialog was first launched). I think it's probably more useful to know what the time on the server will be when the next schedule snapshot runs (which means we need some sort of info message/mini-panel underneath to be updated on selection).
I think the same should happen on the NewPolicySnapshotFrequencyPage (used on the NewPolicyWizard and the policy Properties dialog), where, instead of the current server time, we should be showing another message with the local time the schedule will be first run and the server time this corresponds too.
This ticket may then need some design+discussion first.

kc284:
Closing this PR because it is being reworked and will also need rebasing. Please raise a new PR with squashed changes (totally or grouped in fewer commits) once done. Thanks.

PR #2330, opened 2018-11-23
XOP-953: Cannot change dom0 memory via XenCenter (XC 7.5 RTM on XS 7.…

AaronRobson:
…5 RTM).

Allowing the user to proceed (not set the value and min/max to 0.0GB) even if the memory static min and max is in the wrong order.
Adding further detail to the error message in case an out of bounds value is sent to the API which could potentially happen due to the range widening that occurs if the min/max would invalidate the existing value.

### DIFF
--- a/XenModel/Actions/Host/ChangeControlDomainMemoryAction.cs
+++ b/XenModel/Actions/Host/ChangeControlDomainMemoryAction.cs
@@ -57,7 +57,23 @@ namespace XenAdmin.Actions
         {
             VM vm = Host.ControlDomainZero();
 
-            XenAPI.VM.set_memory(Session, vm.opaque_ref, memory);
+            try
+            {
+                XenAPI.VM.set_memory(Session, vm.opaque_ref, memory);
+            }
+            catch (Exception e)
+            {
+                var f = e as Failure;
+                if (f != null && f.ErrorDescription[0] == Failure.MEMORY_CONSTRAINT_VIOLATION
+                    && memory < vm.memory_static_min)
+                {
+                    throw new Failure(string.Format(Messages.ACTION_CHANGE_CONTROL_DOMAIN_MEMORY_VALUE_TOO_LOW,
+                        Util.MemorySizeStringSuitableUnits(memory, true),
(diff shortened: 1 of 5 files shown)


XenModel/Actions/Host/ChangeControlDomainMemoryAction.cs
```diff
@@ -57,7 +57,23 @@ protected override void Run()
         {
             VM vm = Host.ControlDomainZero();
 
-            XenAPI.VM.set_memory(Session, vm.opaque_ref, memory);
+            try
+            {
+                XenAPI.VM.set_memory(Session, vm.opaque_ref, memory);
+            }
+            catch (Exception e)
```
kc284: `catch (Failure f)`  so we don't need the cast and null check

XenModel/Messages.resx
```diff
@@ -129,6 +129,9 @@
   <data name="ACTION_CHANGE_CONTROL_DOMAIN_MEMORY" xml:space="preserve">
     <value>Changing control domain memory settings on '{0}'</value>
   </data>
+  <data name="ACTION_CHANGE_CONTROL_DOMAIN_MEMORY_VALUE_TOO_LOW" xml:space="preserve">
+    <value>The memory {0} has to be at least the value of the Static minimum ({1})</value>
```
kc284: Static -> static; I don't think the other messages use it capitalised.

XenModel/Messages.resx
```diff
@@ -129,6 +129,9 @@
   <data name="ACTION_CHANGE_CONTROL_DOMAIN_MEMORY" xml:space="preserve">
     <value>Changing control domain memory settings on '{0}'</value>
   </data>
+  <data name="ACTION_CHANGE_CONTROL_DOMAIN_MEMORY_VALUE_TOO_LOW" xml:space="preserve">
+    <value>The memory {0} has to be at least the value of the static minimum ({1})</value>
```
kc284: One more comment which did not occur to me earlier: `The memory 1GB has to be at least the value of the static minimum (2GB)` sounds a bit strange. How about  `The memory specified (1GB) was too low. It has to be at least the value of the static minimum (2GB)`

PR #2331, opened 2018-11-26
CA-303517: In XenCenter, if a VM has more than 1 IP address, it cannot be found using one of its IPs in the search box

AaronRobson:


XenModel/XenAPI-Extensions/Address.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace XenAPI
+{
+    public abstract class Address
```
kc284: Why is this class abstract? Did you rather mean static?

AaronRobson: You are correct, I meant abstract. I will change it.

XenModel/XenAPI-Extensions/Address.cs
```diff
@@ -0,0 +1,26 @@
...
+            // PR-1373 - VM_guest_metrics.networks is a dictionary of IP addresses in the format:
+            // [["0/ip", <IPv4 address>], 
+            //  ["0/ipv4/0", <IPv4 address>], ["0/ipv4/1", <IPv4 address>],
+            //  ["0/ipv6/0", <IPv6 address>], ["0/ipv6/1", <IPv6 address>]]
+
+            return
+                (from network in networks
+                    where network.Key.StartsWith(string.Format("{0}/ip", device))
+                    orderby network.Key
+                    select network.Value.Split(new[] {"\n", "%n"}, StringSplitOptions.None)).SelectMany(x => x).Distinct().ToList();
```
kc284: I think it's safer to use StringSplitOptions.RemoveEmptyEntries; let's also add a relevant test.

XenModel/XenAPI-Extensions/Address.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace XenAPI
+{
+    public static class Address
```
kc284: On second thought, do we need a dedicated class for this? It would make sense if we could gather here other methods dealing with IPs, but if it's only one method, we could perhaps add it to XenModel\Utils\Helper.cs (which is normally where we look for utility methods)?

PR #2332, opened 2018-11-28
CP-29856: Support minimum number of vCPUs in guest templates

MihaelaStoica:
- Show a warning if the user wants to set a smaller number of vCPUs than the minimum specified in the template

### DIFF
--- a/XenAdmin/SettingsPanels/CPUMemoryEditPage.cs
+++ b/XenAdmin/SettingsPanels/CPUMemoryEditPage.cs
@@ -56,6 +56,7 @@ namespace XenAdmin.SettingsPanels
         private decimal _OrigVCPUWeight;
         private decimal _CurrentVCPUWeight;
         private bool isVcpuHotplugSupported;
+        private int minVCPUs;
 
         // Please note that the comboBoxVCPUs control can represent two different VM properties, depending whether the VM supports vCPU hotplug or not: 
         // If vCPU hotplug is supported, comboBoxVCPUs represents the maximum number of vCPUs (VCPUs_max). And the initial number of vCPUs is represented in comboBoxInitialVCPUs (which is only visible in this case) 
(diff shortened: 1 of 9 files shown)


XenAdmin/SettingsPanels/CPUMemoryEditPage.cs
```diff
@@ -460,28 +462,50 @@ private void ShowMemError(bool showAlways, bool testValue)
...
+        private void ValidateVCPUSettings()
         {
             if (vm == null || !comboBoxVCPUs.Enabled)
                 return;
             Host selectedAffinity = vm.Home();
             if (selectedAffinity == null && vm.Connection.Cache.Hosts.Length == 1)
                 selectedAffinity = vm.Connection.Cache.Hosts[0];
 
-            if (showAlways || (testValue && (selectedAffinity != null && selectedAffinity.host_CPUs.Count < (long)comboBoxVCPUs.SelectedItem)))
+            if (selectedAffinity != null && selectedAffinity.host_CPUs.Count < (long)comboBoxVCPUs.SelectedItem)
```
kc284: Judging by the null check further down, it is presumed likely that the SelectedItem might be null, but here we cast it directly to long, shall we add a check?

XenAdmin/SettingsPanels/CPUMemoryEditPage.cs
```diff
@@ -460,28 +462,50 @@ private void ShowMemError(bool showAlways, bool testValue)
...
+            {
+                VCPUWarningLabel.Visible = true;
+                VCPUWarningLabel.Text = Messages.VM_CPUMEMPAGE_VCPU_WARNING;
+            }
+            else if (comboBoxVCPUs.SelectedItem != null && SelectedVcpusMax < minVCPUs)
             {
                 VCPUWarningLabel.Visible = true;
+                VCPUWarningLabel.Text = string.Format(Messages.VM_CPUMEMPAGE_VCPU_MIN_WARNING, minVCPUs);
```
kc284: Just to be fussy, let's set the text first and then make it visible (lest eagle-eyed users catch a text flickering 😄  )
Same below.

MihaelaStoica:
I don't think we should add  minimum recommended vcpus on the General tab, because it is just a recommendation, not a restriction and most of the VMs will probably have 1 there anyway. Plus, we don't currently show the _maximmum recommended vcpus_

PR #2338, opened 2018-12-04
CA-296490: Upload single update to multiple pools in parallel.

kc284:
The changes make use of the existing functionality of the AutomatedUpdatesBasePage.
(I haven't implemented the Skip button because as a workaround the user can go back and deselect problematic pools).


XenAdmin/Wizards/PatchingWizard/AutomatedUpdatesBasePage.cs
```diff
@@ -572,7 +573,7 @@ private void buttonSkip_Click(object sender, EventArgs e)
                 if (!uploadedPatches.Contains(patch))
                 {
                     planActionsPerHost.Add(new DownloadPatchPlanAction(host.Connection, patch, AllDownloadedPatches, patchFromDisk));
-                    planActionsPerHost.Add(new UploadPatchToMasterPlanAction(host.Connection, patch, patchMappings, AllDownloadedPatches, patchFromDisk));
+                    planActionsPerHost.Add(new UploadPatchToMasterPlanAction(this, host.Connection, patch, patchMappings, AllDownloadedPatches, patchFromDisk));
```
MihaelaStoica: The `UploadPatchToMasterPlanAction` action should be called with `skipDiskSpaceCheck = true` here, because in the automated update mode we do the disk space check on the Prechecks page.

XenAdmin/Wizards/PatchingWizard/PlanActions/UploadPatchToMasterPlanAction.cs
```diff
@@ -31,99 +31,234 @@
...
-            var poolUpdates = new List<Pool_update>(session.Connection.Cache.Pool_updates);
+            this.invokingControl = invokingControl;
+            this.updateFilePath = updateFilePath;
+            this.mappings = mappings;
+            this.skipDiskSpaceCheck = skipDiskSpaceCheck;
+            this.selectedServers = selectedServers;
+        }
 
+        protected override void RunWithSession(ref Session session)
+        {
```
MihaelaStoica: This function is now quite long and it is hard hard to follow the code with all the nested `if`s. I wonder if we can extract some of the code in functions, to aid readability. For example, the first section, where we determine if there is an existing mapping, can be put in a separate function.

PR #2339, opened 2018-12-04
 CA-287648: Do not show the user the licence status before its value has been retrieved.

kc284:
Plus correction of the general section expanded state and one or two inefficiencies.


kc284:
The suggested solution for CA-287648 is the easiest fix. It does cause a minor flickering when the actual status is updated, but I think it is preferable showing a waiting message than showing an empty string which might raise questions if the licence takes a while to be updated.
The ticket desctription suggested caching for the status, but I didn't think it's a good idea; I can investigate, though, if people think differently.

AaronRobson:
I agree r.e. avoiding caching of the license information. This extra state can become out of date and it isn't obvious when it should be updated.

PR #2344, opened 2018-12-11
 CA-304787 plus some UX modifications on the Updates page.

kc284:
This PR includes the fix to CA-304787, some minor layout corrections, and some suggested changes that I think make for a better and clearer user experience (happy to discuss).


MihaelaStoica:
👍 for the UX enhancement

PR #2347, opened 2018-12-13
CA-305383: Correct Helper.AreEqual2 which considered identical (but n…

AaronRobson:
…ot reference equal) generic lists or dictionaries as being different.

PropertiesDialog was saving unchanged Snapshot Schedule information back to the server unnecessarily. Noticed while working on CA-301907.


AaronRobson:
Closing as I ought to change the API generator repository to generate it correctly and then copy the changes here as a new PR.

PR #2349, opened 2018-12-18
CP-30251: Check version and feature flag to determine UI state

MihaelaStoica:


kc284:
This looks ok to me. I have a couple of minor discussion points (which however are not blocking this PR):
- if both UEFI and secure UEFI are disabled, we're left with a slightly odd looking groupbox containing the single, checked BIOS radio, which is the default option anyway, so I'm wondering whether we might want to hide the groupbox completely in this case.
- the warnings shown when the UEFI modes are experimental are very similar in terms of wording (indeed when only one of them is experimental, the difference is one word), which means the user should read them carefully to get right what is experimental each time; I feel it would be clearer, even at the cost of being repetitive, if the warning were closer to the radio it refers to, for example, directly underneath it or a shorter message on the right of the radio.

PR #2352, opened 2018-12-21
CA-305383: Do not save unchanged Snapshot Schedule information.

AaronRobson:
With the new API bindings reflecting this change to the Helper.AreEqual2 function: https://github.com/xapi-project/xen-api-sdk/pull/156
The PR includes the new Bindings and a second commit which unit tests the change.

### DIFF
--- a/XenAdminTests/XenModelTests/AreEqual2Tests.cs
+++ b/XenAdminTests/XenModelTests/AreEqual2Tests.cs
@@ -21,6 +21,18 @@ namespace XenAdminTests.XenModelTests
         }
 
         [Test]
+        public void A_null_list_is_considered_equal_to_an_empty_one()
+        {
+            Assert.IsTrue(Helper.AreEqual2(null, new List<string>()));
+        }
+
+        [Test]
+        public void An_empty_list_is_considered_equal_to_a_null_one()
+        {
+            Assert.IsTrue(Helper.AreEqual2(new List<string>(), null));
+        }
+
+        [Test]
         public void A_list_is_considered_equal_to_itself()
         {
             var given = new List<string> {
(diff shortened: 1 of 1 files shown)


XenAdminTests/XenModelTests/AreEqual2Tests.cs
```diff
@@ -0,0 +1,108 @@
...
+namespace XenAdminTests.XenModelTests
+{
+    [TestFixture, Category(TestCategories.Unit)]
+    public class AreEqual2Tests
+    {
+        [Test]
+        public void A_null_is_considered_equal_to_itself()
+        {
+            List<string> given = null;
+            Assert.That(Helper.AreEqual2(given, given), Is.True);
```
kc284: Simplification suggestion: you can use `Assert.IsTrue`, `Assert.IsFalse`.

XenAdminTests/XenModelTests/AreEqual2Tests.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        [Test]
+        public void Separate_nulls_are_considered_equal_to_each_other()
+        {
+            Assert.That(Helper.AreEqual2((List<string>)null, (List<string>)null), Is.True);
```
kc284: Let's add one more assertion for null Dictionaries since they are a class used heavily in the bindings.
Also, an important test case missing is comparison of null List with empty List and null Dictionary with empty Dictionary (which is what differentiates the method AreEqual2 from AreEqual).

AaronRobson: Agreed, its better to be thorough.

PR #2362, opened 2019-01-08
REQ-638: Support checkpoints for Nvidia VGPU enabled VMs

minli1:
CP-30405: 
- Refine conditions when creating VM scheduled snapshots.
- Change the order of checks for checkpoint.

Signed-off-by: Min Li <min.li1@citrix.com>

### DIFF
--- a/XenAdmin/Dialogs/VmSnapshotDialog.cs
+++ b/XenAdmin/Dialogs/VmSnapshotDialog.cs
@@ -178,8 +178,6 @@ namespace XenAdmin.Dialogs
                 tt = Messages.FIELD_DISABLED;
             else if (_VM.power_state != vm_power_state.Running)
                 tt = Messages.INFO_DISKMEMORY_MODE_POWER_STATE.Replace("\\n", "\n");
-            else if (_VM.HasVGPUs())
-                tt = Messages.INFO_DISKMEMORY_MODE_GPU.Replace("\\n", "\n");
             else if (!_VM.GetVirtualisationStatus().HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
                 tt = (_VM.HasNewVirtualisationStates() ? Messages.INFO_DISKMEMORY_MODE_NO_IO_DRIVERS : Messages.INFO_DISKMEMORY_MODE_NO_TOOLS).Replace("\\n", "\n");
             else
--- a/XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePage.resx
+++ b/XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePage.resx
@@ -227,7 +227,7 @@
 
 1. The VM is running or suspended at the time the snapshot is scheduled.
 2. The VM has [Citrix VM Tools] installed.
-3. The VM does NOT have dedicated GPU assigned to it.</value>
+3. If the VM has a dedicated GPU assigned to it, this GPU supports disk and memory snapshots.</value>
   </data>
   <data name="&gt;&gt;labelWarning.Name" xml:space="preserve">
     <value>labelWarning</value>
--- a/XenAdmin/Commands/SuspendVMCommand.cs
+++ b/XenAdmin/Commands/SuspendVMCommand.cs
@@ -217,11 +217,6 @@ namespace XenAdmin.Commands
                 return Messages.VM_HAS_GPU_PASSTHROUGH;
             }
 
-            if (vm.HasVGPUs())
-            {
-                return Messages.VM_HAS_VGPUS;
-            }
-
             return base.GetCantExecuteReasonCore(item);
         }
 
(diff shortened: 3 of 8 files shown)


minli1:
Hi, @MihaelaStoica 
Please help to review the PR, thanks!
Since it's minor changes, so I create the PR from a private branch.

XenAdmin/Dialogs/VmSnapshotDialog.cs
```diff
@@ -178,10 +178,10 @@ private void CheckpointInfoPictureBox_Click(object sender, EventArgs e)
                 tt = Messages.FIELD_DISABLED;
             else if (_VM.power_state != vm_power_state.Running)
                 tt = Messages.INFO_DISKMEMORY_MODE_POWER_STATE.Replace("\\n", "\n");
-            else if (_VM.HasVGPUs())
-                tt = Messages.INFO_DISKMEMORY_MODE_GPU.Replace("\\n", "\n");
             else if (!_VM.GetVirtualisationStatus().HasFlag(VM.VirtualisationStatus.IO_DRIVERS_INSTALLED))
                 tt = (_VM.HasNewVirtualisationStates() ? Messages.INFO_DISKMEMORY_MODE_NO_IO_DRIVERS : Messages.INFO_DISKMEMORY_MODE_NO_TOOLS).Replace("\\n", "\n");
+            else if (_VM.HasVGPUs())
```
MihaelaStoica: I don't know what other conditions might restrict the checkpoint operations, but the fact that we have the INFO_DISKMEMORY_MODE_MISC message suggests that it might be other reasons why the checkpoints are restricted. 
I suggest to either get rid of the INFO_DISKMEMORY_MODE_GPU message completely (so the message will be INFO_DISKMEMORY_MODE_MISC  even if the vGPU is the actual reason), or enhance the GPU check to also include a version check and the vGPU type (i.e. show the  INFO_DISKMEMORY_MODE_GPU if _VM.HasVGPUs and the server version is pre-Naples or the gpu type is not nvidia)

XenAdmin/Wizards/NewPolicyWizard/NewPolicySnapshotTypePage.resx
```diff
@@ -227,7 +227,7 @@
 
 1. The VM is running or suspended at the time the snapshot is scheduled.
 2. The VM has [Citrix VM Tools] installed.
-3. The VM does NOT have dedicated GPU assigned to it.</value>
+3. The VM does NOT have dedicated Intel/AMD GPU assigned to it(Nvidia GPU is supported).</value>
```
MihaelaStoica: This is only true for Naples or greater hosts, so a version check will be needed here. 
Also I suggest that we don't specify the vendor names here, maybe we can change the message to something neutral, like: The VM does NOT have dedicated GPU assigned that does not support disk and memory snapshots.

kc284: General point: there should be a space between a word and the following parenthesis, i.e. instead of `to it(` it should be `to it (`.
Alternative message: `3. If the VM has a dedicated GPU assigned to it, this GPU supports disk and memory snapshots.`

minli1: Thank you all for the comprehensive comments.

I think the neutral way is good, like @kc284 wrote: `3. If the VM has a dedicated GPU assigned to it, this GPU supports disk and memory snapshots.`  This is nice.

XenAdmin/Dialogs/VmSnapshotDialog.cs
minli1: Thanks.
Now I added new function to check the VGPU type is `Nvidia` or not and check the server version.
Please help to check again. @MihaelaStoica 
Also, after vgpu_type check and version check, I think the check order should still be checking the I/O drivers firstly rather than check VGPU.

MihaelaStoica: I don't think it is correct that the HasVGPUs() function returns false if the host is NaplesOrGreater and vGPU type is Nvidia ; it doesn't return what the name suggests.

Anyway, I think that the check on the server side (for allowing suspend and checkpoint operations) is a bit more complex, as only *some* Nvidia GPUs support suspending and checkponting a VM (by providing compatibility metadata).

Given this complexity, I don't think it's worth having version and vGPU type checks here, and we can simply display the generic message (INFO_DISKMEMORY_MODE_MISC). Please note that we have done the same for vGPU migration (see CA-270505).
And I believe we also have to remove the message for VM suspend (https://github.com/xenserver/xenadmin/blob/master/XenAdmin/Commands/SuspendVMCommand.cs#L222), for the same reason - some vGPU types allow VM suspend.

minli1: Understood. 
I've removed the specific error message to generic message for VM checkpoint and suspend with two separate commits. 
Please take a look, thank you! :) @MihaelaStoica

PR #2366, opened 2019-01-13
Modifications to fix failing tests NewVMWizardTtest and RollingPoolUpgradeTest.

kc284:
Note that the WebBrowser2Tests was not working either, but the problem had been swallowed and only became obvious when the MWWaitFor logic was corrected.


XenAdminTests/MiscTests/WebBrowser2Tests.cs
```diff
@@ -157,10 +141,20 @@ public void TestMultipleUrisWithNoneValid()
...
+                new Uri("http://127.0.0.1/blah"),
+                new Uri("http://127.0.0.1/blahblah")
+            };
+
             foreach (var uri in uris)
             {
-                var curUri = uri;
+                curUri = uri;
+
+                navigated = navigated = navError = false;
```
MihaelaStoica: was this supposed to include `navigating` (instead of one of the `navidated`) ?

kc284: Oops, yes 😳

PR #2369, opened 2019-01-16
L10N : Translation and resx resync

ghost:


kc284:
I think VS is adding automatically the entries `mimetype="application/x-microsoft.net.object.bytearray.base64" ` - are they needed?

DarrenMatthews776:
Those entries are the same as what is in the english resx file. I left them in because I didn't want to cause any unwanted side effects that was different from the English version. If they are auto generated and not needed then they wont be needed.

MatthewsD776:
Are the <metadata> tags not needed too? Or are just the mimetype bytearray data tags the only ones not needed?

DarrenMatthews776:
Are the metadata tags not needed too? Or just remove the mimetype data tags?

PR #2371, opened 2019-01-23
CP-30335 - Part 2: Add a new precheck + Part 3: Enhance the error message displayed in case of migration failure during the RPU

MihaelaStoica:
(also fixing the tests which got broken in the previous commits)


XenAdmin/Diagnostics/Problems/WarningWithMoreInfo.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Drawing;
```
kc284: Copyright missing.

XenAdmin/Diagnostics/Checks/HostMemoryPostUpgradeCheck.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System.Collections.Generic;
```
kc284: Copyright missing.

XenAdmin/Diagnostics/Checks/HostMemoryPostUpgradeCheck.cs
```diff
@@ -0,0 +1,102 @@
...
+                if (Helpers.NaplesOrGreater(upgradePlatformVersion))
+                {
+                    // we know that they are upgrading to Naples or greater, so add specific warning
+                    return new HostMemoryPostUpgradeWarning(this, Host, 0, upgradeProductVersion);
+                }
+            }
+            return null;
+        }
+
+        public bool TryGetDom0MemoryPostUpgrade(out long dom0MemoryAfterUpgrade)
```
kc284: These two try methods can be private; the `if-else block` at lines 69 and 90 can be compacted to `return`.

XenAdmin/Diagnostics/Checks/HostMemoryPostUpgradeCheck.cs
```diff
@@ -0,0 +1,102 @@
...
+                var result = Host.call_plugin(Host.Connection.Session, Host.opaque_ref, "prepare_host_upgrade.py", "getVersion", installMethodConfig);
+                var serializer = new JavaScriptSerializer();
+                var res = (Dictionary<string, object>)serializer.DeserializeObject(result);
+                platformVersion = res.ContainsKey("platform-version") ? (string)res["platform-version"] : null;
+                productVersion = res.ContainsKey("product-version") ? (string)res["product-version"] : null;
+                if (platformVersion == null && productVersion == null)
+                    return false;
+                return true;
+            }
+            catch (Failure failure)
```
kc284: Is it likely to have an exception thrown at deserialization, in which case we should be catching a general exception as well after the Failure.

XenAdmin/Diagnostics/Checks/HostMemoryPostUpgradeCheck.cs
```diff
@@ -0,0 +1,102 @@
...
+                // we don't know the Dom0 memory after the upgrade, so add generic warning if they are upgrading from pre-Naples 
+                if (Helpers.NaplesOrGreater(Host))
+                    return null;
+
+                string upgradePlatformVersion = null;
+                string upgradeProductVersion = null;
+                if (installMethodConfig != null)
+                    TryGetUpgradeVersion(out upgradePlatformVersion, out upgradeProductVersion);
+                
+                if (string.IsNullOrEmpty(upgradePlatformVersion))
```
kc284: Wouldn't it be more generic and safer if we did
```
if (Helpers.NaplesOrGreater(upgradePlatformVersion))
{
    // we know that they are upgrading to Naples or greater, so add specific warning
    return new HostMemoryPostUpgradeWarning(this, Host, 0, upgradeProductVersion);
}
else
{
    // we don't know the upgrade version, so add generic warning (this is the case of the manual upgrade or when the rpu plugin doesn't have the function)
        return new HostMemoryPostUpgradeWarning(this, Host);
}
```
unless it's completely unlikely to have garbage returned in the json or we don't want to show any warning in this case.

XenAdmin/Diagnostics/Problems/HostProblem/HostMemoryPostUpgradeWarning.cs
```diff
@@ -0,0 +1,47 @@
+﻿using XenAdmin.Core;
```
kc284: Copyright missing.

XenAdmin/Diagnostics/Checks/HostMemoryPostUpgradeCheck.cs
```diff
@@ -0,0 +1,102 @@
...
+                // we know the Dom0 memory after the upgrade, check if it will greater then the current value
+                if (dom0MemoryPostUpgrade > currentDom0Memory)
+                {
+                    // add warning that the the Dom0 memory will be changed after the upgrade 
+                    return new HostMemoryPostUpgradeWarning(this, Host, dom0MemoryPostUpgrade);
+                }
+            }
+            else
+            {
+                // we don't know the Dom0 memory after the upgrade, so add generic warning if they are upgrading from pre-Naples
```
kc284: This comment belongs in reality after line 39.

XenAdmin/Diagnostics/Checks/HostMemoryPostUpgradeCheck.cs
MihaelaStoica: This was intended to also cover the case where we know the upgrade platform version and it is pre-Naples, in which case we shouldn't display a warning. But actually, we shouldn't get here in that case, due to lines 38-39

PR #2377, opened 2019-01-29
CA-308212: Turned the prepare_host_upgrade test to a precheck

kc284:


XenAdmin/Diagnostics/Problems/HostProblem/HostPrepareToUpgradeProblem.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly string _shortMessage;
+        private readonly string _longMessage;
+
+        public HostPrepareToUpgradeProblem(Check check, Host host, string friendlyErrorKey = null)
+            : base(check, host)
+        {
+            InitialiseMessagesFromErrorKey(friendlyErrorKey, out _shortMessage, out _longMessage);
+        }
+
+        public override bool IsFixable => false;
```
kc284: One thing I'd like to draw the attention to here is IsFixable being set to false: this results to the Resolve All button being disabled. Strictly speaking this is not wrong - we cannot resolve _all_ since not all are fixable, some have to be manually fixed. It poses the question though whether we might want to change Resolve All to mean resolve all fixable.

PR #2384, opened 2019-02-11
CP-30602: Add boot mode info to Naples XC

jijiang:

### DIFF
--- a/XenModel/Actions/VM/CreateVMAction.cs
+++ b/XenModel/Actions/VM/CreateVMAction.cs
@@ -56,13 +56,13 @@ namespace XenAdmin.Actions.VMActions
             switch (x)
             {
                 case BootMode.BIOS_BOOT:
-                    return "BIOS Boot";
+                    return Messages.BIOS_BOOT;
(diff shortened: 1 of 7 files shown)


XenModel/Actions/VM/CreateVMAction.cs
```diff
@@ -56,13 +56,13 @@ public static string StringOf(this BootMode x)
...
+                    return Messages.BIOS_BOOT;
                 case BootMode.UEFI_BOOT:
-                    return "UEFI Boot";
+                    return Messages.UEFI_BOOT;
                 case BootMode.UEFI_SECURE_BOOT:
-                    return "UEFI Secure Boot";
+                    return Messages.UEFI_SECURE_BOOT;
                 default:
-                    return "Not Available";
+                    return Messages.GPU_UNAVAILABLE;
```
MihaelaStoica: Resuing existing messages is good, but the name of this message suggests that it is GPU-specific and we may change it in the future without realising that it's used for Boot mode as well. So we could either rename it to something generic like UNAVAILABLE, or add a new string for the boot mode.

PR #2392, opened 2019-02-22
CA-310980: Show successful completed status on the patching page when no action is to be taken.

kc284:


XenAdmin/Wizards/PatchingWizard/AutomatedUpdatesBasePage.cs
```diff
@@ -434,7 +439,7 @@ private void WorkerCompleted(object sender, RunWorkerCompletedEventArgs e)
         {
             if (e.Cancelled)
             {
-                Status = Status.Completed;
+                Status = Status.Cancelled;
```
kc284: This change is not related to the current bug, however I think it was a mistake that we were setting it to Completed.

PR #2393, opened 2019-02-22
CP-30597: Remove RPU005. Use a repackaged RPU004 instead.

kc284:


kc284:
Merging by myself to avoid fruther delays.

PR #2396, opened 2019-02-28
CA-309758: Optimisations to the way we assert if Vms can be migrated …

MihaelaStoica:
…to a destination pool in the Cross Pool Migrate wizard

- When asserting if a VM can be migrated to a pool we don't have to check all the hosts in the pool at this point, we can enable the pool when we find the first host where migration is possible; we will check the rest of the hosts only if that pool is selected.
- If the destination pool is older than the source, then we don't need to do any server calls because we know that migration to an older host is not allowed.


kc284:
It looks ok to me. One more improvement we can make (even if it only affects a smaller number of users) is to add the CrossPoolMigrateCanMigrateFilter last in the list (after the WlbEnabledFilter which only accesses the cache) so that if WLB is enabled no checks will run unnecessarily.

PR #2398, opened 2019-03-01
Refactored the VIFDialog to fix various problems:

kc284:
- Launch the dialog in using blocks.
- Event GotFocus is not meant to be handled directly in the code.
- Populate the dialog in the OnLoad override instead of the constructor as this causes problems at design time.
- Popping up an error message for duplicate MAC on combobox item selection or textbox text change is not optimal UX. Perform this check once when closing the form. Additionally the parent form of the pop up should be the dialog, not the main window.
- The validation messages are not obvious to the user because they are hidden in the tooltip of a disabled button. Show them inline instead.
- There is no need to create commands for VIF creation/change; use the relevant actions instead.


XenAdmin/Dialogs/VIFDialog.cs
```diff
@@ -126,80 +96,94 @@ private void LoadDetails()
                     promptTextBoxQoS.Text = ExistingVif.LimitString();
                     checkboxQoS.Checked = ExistingVif.qos_algorithm_type == VIF.RATE_LIMIT_QOS_VALUE;
                 }
-                flowLayoutPanelQoS.Enabled = checkboxQoS.Enabled = true;
 
-                panelLicenseRestriction.Visible = false;
+                tableLayoutPanel3.Enabled = checkboxQoS.Enabled = true;
+                labelInfo.Text = Messages.VIF_LICENSE_RESTRICTION;
```
kc284: Unless I'm missing something, is this message ever visible?

MihaelaStoica: This was related to a license restriction that has been removed long ago (8ea7e73b55d3e75bd1c75cadfb6def6ddcd9add0), so we can drop it.

XenAdmin/Controls/NetworkingTab/NetworkList.cs
```diff
@@ -732,31 +758,40 @@ private void launchVmNetworkSettingsDialog()
...
         }
 
+        private void updateVIFAction_Completed(ActionBase sender)
+        {
+            sender.Completed -= updateVIFAction_Completed;
+            var action = (AsyncAction)sender;
+
+            Program.Invoke(Program.MainWindow, () =>
+            {
+                if (action.Result == false.ToString())
```
MihaelaStoica: I think we should show the hotplug error after the page is updated (as per old behaviour)

XenAdmin/Controls/NetworkingTab/NetworkList.cs
```diff
@@ -732,31 +758,40 @@ private void launchVmNetworkSettingsDialog()
...
+            {
+                if (d.ShowDialog() != DialogResult.OK)
+                    return;
+
+                proxyVIF = d.GetNewSettings();
+            }
 
-            var proxyVIF = d.GetNewSettings();
-            UpdateVIFCommand command = new UpdateVIFCommand(Program.MainWindow, vm, vif, proxyVIF);
+            var action = new UpdateVIFAction(vm, vif, proxyVIF);
```
MihaelaStoica: When a VIF is updated, the Events view shows 3 actions: updating, deleting and creating, because the update action runs the other two. We could hide the deletion and creation from the history. (this is existing behaviour, but we can improve it, since we look at this area)

kc284:
Updated; commits can be squashed.

PR #2400, opened 2019-03-06
CA-309844: Fix meddling actions

MihaelaStoica:
- Refactored MeddlingAction and Task classes: moved the logic for building a MeddlingAction outside the Task class and switched to using the vm_operations enum to identify which tasks are suitable for MeddlingAction


kc284:
Looks ok to me; the imports/exports are not shown, but does not seem to be a regression.

Issue #2403, opened 2019-03-08
View Consoles when XenCenter is connected by SSH tunnel


ander089:
When I'm at work and I want to connect to my XenServer I'm using an SSH tunnel.
With XenCenter I can do everything instead of view console.
I find the the way changing the Console.location with the console.Connection.HostnameWithPort.
I will do a branch with this.

Another way can be implement Socks Proxy into XenCenter, but is more

ander089:
I'm not able to push anything... How can I show you what I've done ?

kc284:
Direct push to upstream repos is disabled for obvious reasons. Please fork the repo, push the changes into your fork, and then create a pull request to this one. Thanks.

kc284:
There is a server side toolstack issue whereby the consoles are not refreshed when the IP changes, so the issue should be fixed there instead of providing a client side workaround. Currently the team is working on this. I'm closing this issue and PR #2404, will also add a reference to the fix once it is submitted.

kc284:
Fix in https://github.com/xapi-project/xen-api/pull/4176.

PR #2404, opened 2019-03-08
change console.location for enable remote consoles

ander089:
Issue #2403

### DIFF
--- a/XenAdmin/ConsoleView/XSVNCScreen.cs
+++ b/XenAdmin/ConsoleView/XSVNCScreen.cs
@@ -1085,6 +1085,8 @@ namespace XenAdmin.ConsoleView
                 throw new Failure(Failure.INTERNAL_ERROR, Messages.HOST_GONE);
             }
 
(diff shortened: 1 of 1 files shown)


XenAdmin/ConsoleView/XSVNCScreen.cs
```diff
@@ -1085,6 +1085,8 @@ private void ConnectHostedConsole(VNCGraphicsClient v, Console console)
                 throw new Failure(Failure.INTERNAL_ERROR, Messages.HOST_GONE);
             }
 
+            console.location = System.Text.RegularExpressions.Regex.Replace(console.location, @"((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.)){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?){1}(:([0-6][0-5][0-5][0-3][0-5]|[0-5]\d{1,4}|\d{1,4}))?", console.Connection.HostnameWithPort);
```
AaronRobson, line 1088: @ander089 after studying your regex I have found:

1. That some valid port numbers are excluded such as `1.1.1.1:65509` where the last digit will not be included in the match and therefore after replacement will include it twice.
2. I'm not sure if this is supported, but the regex won't handle IPv6 addresses.
3. I am struggling to think of a scenario where replacing the IP:port with Hostname:port would help make the connection more reliable to set up.
4. I can think of scenarios where the IP:port would be more reliable; differing DNS setups available to the local client computer and the server for example.

kc284:
Closing as this is being fixed on the server side.

PR #2405, opened 2019-03-11
CA-309152: Remote logging is offered in addition to local and not instead of it

kc284:
The radiobuttons local-remote gave the impression the two types of logging
were mutually exclusive. Use a checkbox instead to make it obvious that
the logs are directed to a remote server in addition to being stored locally.

### DIFF
--- a/XenAdmin/SettingsPanels/LogDestinationEditPage.cs
+++ b/XenAdmin/SettingsPanels/LogDestinationEditPage.cs
@@ -75,8 +75,8 @@ namespace XenAdmin.SettingsPanels
             get
             {
                 return checkBoxRemote.Checked
-                    ? string.Format(Messages.LOCAL_AND_REMOTE, RemoteServer)
-                    : Messages.LOCAL;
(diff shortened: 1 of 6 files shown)


XenAdmin/SettingsPanels/LogDestinationEditPage.cs
```diff
@@ -30,168 +30,140 @@
...
+        public Image Image => Properties.Resources.log_destination_16;
+
+        public string SubText
         {
-            TheHost = clone as Host;
-            Repopulate();
+            get
+            {
+                return checkBoxRemote.Checked
+                    ? string.Format(Messages.LOCAL_AND_REMOTE, RemoteServer)
```
MihaelaStoica: We might want to show the same text on the general tab (Log destination entry there shows "Local" or the remote server

kc284: Done; commits can be squashed when merging.

PR #2411, opened 2019-03-18
CP-29310 & CP-30765: added support for the new xxHash-based VM imports and exports

RoryBG:
Xapi is moving from SHA1 to xxHash for VM imports and exports. These changes allow XenCenter to verify this new format whilst also maintaining backwards compatability. The xxHash code comes from [here ](https://github.com/differentrain/YYProject.XXHash) and is under an MIT license. 

I have manually tested these changes by running all combinations of imports and exports across new and old versions of Xapi.


kc284:
I believe we'll need to add xxHash to the legal notices https://github.com/xenserver/xenadmin/blob/master/XenAdmin/Dialogs/LegalNoticesDialog.resx

PR #2415, opened 2019-03-26
CA-311238 refix: when creating a session from another Session object,…

kc284:
… copy over its properties.

Also, removed unused private static constructor.


XenModel/XenAPI-Extensions/Session.cs
```diff
@@ -78,13 +79,17 @@ public Session(Session session, Proxy proxy, IXenConnection connection)
         /// </summary>
         public Session(Session session, IXenConnection connection, int timeout)
         {
-            if (session.JsonRpcClient != null)
+            APIVersion = session.APIVersion;
```
MihaelaStoica: I think this constructor can call the existing Session(Session session, int timeout). This way we would avoid duplicated code.

PR #2420, opened 2019-04-02
CP-31040: Revert RPU003 to version 6.0 (master)

MihaelaStoica:


MihaelaStoica:
Cancel this for now, as we are trying a different approach.

PR #2425, opened 2019-04-24
 CP-31063: Add info regarding multiple vGPU support. 

MihaelaStoica:
Also, update the text referring to vGPUs and hide the PCI slot number column on older hosts

### DIFF
--- a/XenAdmin/SettingsPanels/GpuEditPage.resx
+++ b/XenAdmin/SettingsPanels/GpuEditPage.resx
@@ -139,7 +139,7 @@
   </data>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
   <data name="addButton.Location" type="System.Drawing.Point, System.Drawing">
-    <value>607, 41</value>
+    <value>778, 41</value>
   </data>
   <data name="addButton.Size" type="System.Drawing.Size, System.Drawing">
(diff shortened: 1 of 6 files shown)


XenAdmin/SettingsPanels/GpuEditPage.resx
```diff
@@ -259,7 +256,19 @@
     <value>50</value>
   </data>
   <data name="VideoRAMColumn.Width" type="System.Int32, mscorlib">
-    <value>72</value>
+    <value>90</value>
+  </data>
+  <metadata name="MultiColumn.UserAddedColumn" type="System.Boolean, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089">
+    <value>True</value>
+  </metadata>
+  <data name="MultiColumn.HeaderText" xml:space="preserve">
```
michael2012z: I think adding this column doesn't provide additional information to a user. After some vGPUs have been added, all values in this column will show "Yes". If it's "No", no more items can be added. You have no chance to see mixing values in this column.
You may concern how do we tell the user whether a vGPU type supports multiple or not. The plan is to add text in entries of the combo-box of the "adding" dialog. But it hasn't been done yet.

MihaelaStoica:
> 
> 
> Another comment to the table:
> "Virtual GPUs per GPU" column is wide, but the content is number (very short), should we make the name to "vGPUs per GPU" or something even shorter?

We don't use the term vGPU in the UI

PR #2427, opened 2019-04-25
CA-310177: Check if the selected iso file contains a valid ISO file s…

MihaelaStoica:
…ystem.

Also removed unnecessary call to OnPageUpdated.


MihaelaStoica:
Good idea, I will do that

PR #2440, opened 2019-05-21
Show bios strings as a new panel in the hosts general tab

borzel:
It would be great if we could see the BIOS strings in xenadmin, to know which mainboard/bios version is used.

grafik

Reference: https://github.com/xcp-ng/xenadmin/issues/147


kc284:
Hi, thanks for raising this PR, however we will not merge it at this time, because by design we don't expose the host BIOS strings in XenCenter and there have been no requests from Citrix users to do otherwise.

However, I appreciate this may be useful to your users, hence I have a couple of suggestions which you may want to implement in your fork:
- Move the new panel further down the page, after the License/Version/Updates panels, so it's closer to more relevant host details.
- All the other panels use user friendly names in the first column instead of the raw strings. To match this, and to avoid having to come up with names for all strings or even account for unknown strings, you may want to implement it in the same way as the VM BIOS strings, i.e. add only one entry in the BIOS panel with a title such as "BIOS strings" in the left column and the list of key-value pairs in the right.

PR #2441, opened 2019-05-22
CA-316236 Rename the export option from Snapshot page

MihaelaStoica:


kc284:
I'm thinking that "E&xport Snapshot as a Template..." is very similar to the other item "&Save as a Template...".
The latter means save it on the server as opposed to the local machine, which is not very obvious before actually clicking it, so how about renaming that too to "New Template from &Snapshot..."? This will match the item "&New VM from Snapshot...", hence it will be unambiguous where each object is created.

PR #2443, opened 2019-05-23
CA-318198 + CA-251400: Consolidation of build artifact folders 

kc284:


kc284:
I have not updated the dotnet-packages build number, as we'll need to take account of CP-31132, which is not necessary yet.

PR #2446, opened 2019-05-31
CA-320182: Error in adding vGPU in New VM wizard.

michael2012z:

### DIFF
--- a/XenAdmin/Dialogs/AddVGPUDialog.cs
+++ b/XenAdmin/Dialogs/AddVGPUDialog.cs
@@ -91,7 +91,7 @@ namespace XenAdmin.Dialogs
                         comboBoxTypes.Items.Add(new GpuTuple(gpu_group, vgpuType, disabledTypes.ToArray())); // GPU_type item
 
                     if (commonTypes.Count == 1 && !disabledTypes.Contains(commonTypes[0]))
-                        comboBoxTypes.SelectedItem = comboBoxTypes.Items[1]; // Items[0] is group
+                        comboBoxTypes.SelectedItem = comboBoxTypes.Items[0];
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


XenModel/XenAPI/FriendlyErrorNames.resx
```diff
@@ -1959,7 +1959,7 @@ Authorized Roles: {1}</value>
     <value>The VGPU is not compatible with any PGPU in the destination.</value>
   </data>
   <data name="VGPU_TYPE_NOT_COMPATIBLE" xml:space="preserve">
-    <value>You try to create a VGPU that is not compatible with existing types on the VM.</value>
+    <value>You tried to create a VGPU that is not compatible with existing types on the VM.</value>
```
MihaelaStoica: We should avoid using "VGPU" in the user-facing messages, so here we should say "virtual GPU" .
My suggestion for this message was (note "virtual GPU" and "the existing types"):
"You tried to create a virtual GPU that is not compatible with the existing types on the VM."

XenAdmin/Dialogs/AddVGPUDialog.cs
```diff
@@ -91,7 +91,7 @@ private void PopulateComboBox()
                         comboBoxTypes.Items.Add(new GpuTuple(gpu_group, vgpuType, disabledTypes.ToArray())); // GPU_type item
 
                     if (commonTypes.Count == 1 && !disabledTypes.Contains(commonTypes[0]))
-                        comboBoxTypes.SelectedItem = comboBoxTypes.Items[1]; // Items[0] is group
+                        comboBoxTypes.SelectedItem = comboBoxTypes.Items[0];
```
MihaelaStoica: So what gets selected here if Items[0] is a group?

michael2012z: If host driver was not installed, Items[0] is a device (no group).
If host driver was installed, Items[0] is the group, but the combobox handles this, focus slip to the next enabled item.

XenModel/XenAPI/FriendlyErrorNames.resx
michael2012z: I've submit a PR in XAPI to update the text.

michael2012z:
@MihaelaStoica Any more comments for this PR? We can update Xapi error text with another PR when it's ready in XAPI build.

PR #2463, opened 2019-07-04
CA-322735: Keep XenCenter showing the real device number used in the VM

liulinC:

This commit change the valid `device` id range for a  VM from `[11,31]` to `[0,20]` for the backward compatibility

### DIFF
--- a/XenAdmin/SettingsPanels/GpuEditPage.cs
+++ b/XenAdmin/SettingsPanels/GpuEditPage.cs
@@ -292,7 +292,10 @@ namespace XenAdmin.SettingsPanels
                         return;
 
                     VGPU_type type = tuple.VgpuTypes[0];
+                    // Set vGPU device to null. vGPU creation will check whether a valid device is assigned
+                    // to the vGPU, if not then 0 is used to declare the first availabe slot in VM
                     var vGpu = new VGPU();
+                    vGpu.device = null;
                     vGpu.GPU_group = new XenRef<GPU_group>(tuple.GpuGroup.opaque_ref);
                     vGpu.type = new XenRef<VGPU_type>(type.opaque_ref);
                     vGpu.Connection = vm.Connection;
@@ -325,6 +328,10 @@ namespace XenAdmin.SettingsPanels
         private readonly DataGridViewTextBoxCell maxResolutionColumn = new DataGridViewTextBoxCell();
         private readonly DataGridViewTextBoxCell maxDisplaysColumn = new DataGridViewTextBoxCell();
         private readonly DataGridViewTextBoxCell videoRamColumn = new DataGridViewTextBoxCell();
+        // Xapi reserves device numbers [0,20] for vGPU for backwards compatibility.
+        // The guest PCI bus slots lie within [11,31], hence XenCenter needs to
+        // shift the device number by 11 in order to show the actual PCI slot number used in the VM
+        private const int VGPU_PCI_SLOT_NUMBER_SHIFT = 11;
 
         public VGPU VGpu { get; }
 
@@ -338,8 +345,8 @@ namespace XenAdmin.SettingsPanels
 
         private void SetCells()
         {
-            if (int.TryParse(VGpu.device, out var device) && device != 0)
-                deviceColumn.Value = device;
+            if (int.TryParse(VGpu.device, out var device))
+                deviceColumn.Value = Helpers.PlymouthOrGreater(VGpu.Connection) ? device + VGPU_PCI_SLOT_NUMBER_SHIFT : device;
             else
                 deviceColumn.Value = Messages.HYPHEN;
 
--- a/XenModel/Actions/VM/GpuAssignAction.cs
+++ b/XenModel/Actions/VM/GpuAssignAction.cs
@@ -34,7 +34,6 @@ using System.Collections.Generic;
 using System.Text;
 
 using XenAdmin.Core;
-
 using XenAPI;
 
 namespace XenAdmin.Actions
@@ -64,7 +63,7 @@ namespace XenAdmin.Actions
 
             // New added vGPUs haven't opaque_ref
             foreach (var vGpu in vGpus.FindAll(x => x.opaque_ref == null))
-                AddGpu(vm.Connection.Resolve(vGpu.GPU_group), vm.Connection.Resolve(vGpu.type), vGpu.device);
+                AddGpu(vm.Connection.Resolve(vGpu.GPU_group), vm.Connection.Resolve(vGpu.type), vGpu.device ?? "0");
         }
 
         private void AddGpu(GPU_group gpuGroup, VGPU_type vGpuType, string device = "0")
--- a/XenModel/XenAPI/VGPU.cs
+++ b/XenModel/XenAPI/VGPU.cs
@@ -656,7 +656,7 @@ namespace XenAPI
         private XenRef<GPU_group> _GPU_group = new XenRef<GPU_group>(Helper.NullOpaqueRef);
 
         /// <summary>
-        /// Guest PCI slot (a value of 0 means auto-assign to first empty slot, valid slot is in range of [11,31] for multi-VGPU purpose)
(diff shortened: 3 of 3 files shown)


XenModel/XenAPI/VGPU.cs
```diff
@@ -50,6 +50,13 @@ public partial class VGPU : XenObject<VGPU>
         public VGPU()
         {
         }
+        /// <summary>
```
kc284: All files in `XenModel/XenAPI` come from the SDK and are autogenerated. No edits should be done in XenCenter because they will be overwritten next time we update the bindings.

XenAdmin/SettingsPanels/GpuEditPage.cs
```diff
@@ -292,7 +292,9 @@ private void addButton_Click(object sender, EventArgs e)
                         return;
 
                     VGPU_type type = tuple.VgpuTypes[0];
-                    var vGpu = new VGPU();
+                    // Set vGPU device to VGPU_UNASSIGNED_DEVICE, does not assign any devcie ID to it.
+                    // When create vGPU, 0 is used to declare the first availabe slot in VM
+                    var vGpu = new VGPU (GpuAssignAction.VGPU_UNASSIGNED_DEVICE);
```
kc284: Since the constructors of the API classes are autogenerated and a new one cannot be added manually, this should be instead
```
var vGpu = new VGPU();
vGpu.device = GpuAssignAction.VGPU_UNASSIGNED_DEVICE;
```

XenModel/XenAPI/VGPU.cs
MihaelaStoica: We don't really need this constructor. We can assign the device after the object is created, as we do with other properties.

XenAdmin/SettingsPanels/GpuEditPage.cs
MihaelaStoica: When we create a new vGPU object, the device field is an empty string, I believe. Can we not simple use the empty string as an indication that there is no device ID assigned yet? I don't think we need an extra string for this.

kc284: The device field is indeed initialised with `0`. If we want it empty string we should set it explicitly afterwards.

kc284: Or it may be even simpler to set it to `null` so that afterwards the check in GpuAssignAction can be `vGpu.device ?? "0"`.

XenAdmin/SettingsPanels/GpuEditPage.cs
```diff
@@ -325,6 +327,9 @@ public class VGpuDetailRow : DataGridViewExRow
         private readonly DataGridViewTextBoxCell maxResolutionColumn = new DataGridViewTextBoxCell();
         private readonly DataGridViewTextBoxCell maxDisplaysColumn = new DataGridViewTextBoxCell();
         private readonly DataGridViewTextBoxCell videoRamColumn = new DataGridViewTextBoxCell();
+        // Xapi reserve slot number [0,20] for vGPU for backward compatibility, and pass [11,31] to VM.
```
kc284: Please use the following comment for clarity:
```
Xapi reserves device numbers [0,20] for vGPU for backwards compatibility. The guest PCI bus slots lie within [11,31],
hence XenCenter needs to shift the device number by 11 in order to show the actual PCI slot number used in the VM
```

XenAdmin/SettingsPanels/GpuEditPage.cs
```diff
@@ -338,10 +343,10 @@ public VGpuDetailRow(VGPU vGpu)
 
         private void SetCells()
         {
-            if (int.TryParse(VGpu.device, out var device) && device != 0)
-                deviceColumn.Value = device;
+            if (int.TryParse(VGpu.device, out var device))
+                deviceColumn.Value = device + VGPU_PCI_SLOT_NUMBER_SHIFT;
```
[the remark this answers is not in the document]
kc284: For older versions I don't think it matters because the code populating the GridView further up hides the column, but if we want to be consistent we may want to write:
```
if (int.TryParse(VGpu.device, out var device) &&
    (Helpers.PlymouthOrGreater(VGpu.Connection) || device != 0))
    deviceColumn.Value = device + VGPU_PCI_SLOT_NUMBER_SHIFT;
else
    deviceColumn.Value = Messages.HYPHEN;
```

XenAdmin/SettingsPanels/GpuEditPage.cs
liulinC: ```The device field is indeed initialised with 0. If we want it empty string we should set it explicitly afterwards```
No, it is initialised  with `VGPU_UNASSIGNED_DEVICE` now, which is showed as `-`. but it may be a good idea to assign it to `null` instead.

kc284: No. When a class is initialised, its fields are initialised before the constructor is called. If the constructor sets a value on a field, this value overwrites the value the field has been initialised with. The device field is initialised as `private string _device = "0";` and the hyphen you assign to it overwrites the original value. But as discussed above, we don't need to add a new constructor, but rather set the property after we call `var vGpu = new VGPU();`.

XenAdmin/SettingsPanels/GpuEditPage.cs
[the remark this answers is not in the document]
liulinC: But for new versions, in case the device nubmer used [0,1,2] will be shown as [-,12,13], just to confirm, does that make sense?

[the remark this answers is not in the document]
liulinC: but the `PlymouthOrGreater` judge is a good point, perhaps we can write someting like
```if (int.TryParse(VGpu.device, out var device) )
    deviceColumn.Value = Helpers.PlymouthOrGreater(VGpu.Connection)? device + VGPU_PCI_SLOT_NUMBER_SHIFT: device;
else
    deviceColumn.Value = Messages.HYPHEN;
```

XenModel/Actions/VM/GpuAssignAction.cs
```diff
@@ -64,7 +63,7 @@ protected override void Run()
 
             // New added vGPUs haven't opaque_ref
             foreach (var vGpu in vGpus.FindAll(x => x.opaque_ref == null))
-                AddGpu(vm.Connection.Resolve(vGpu.GPU_group), vm.Connection.Resolve(vGpu.type), vGpu.device);
+                AddGpu(vm.Connection.Resolve(vGpu.GPU_group), vm.Connection.Resolve(vGpu.type), vGpu.device != null? vGpu.device:"0");
```
kc284: You can use `vGpu.device ?? "0"` , which is the same as `vGpu.device != null ? vGpu.device : "0"`, but much more compact.

XenAdmin/SettingsPanels/GpuEditPage.cs
```diff
@@ -338,8 +345,8 @@ public VGpuDetailRow(VGPU vGpu)
 
         private void SetCells()
         {
-            if (int.TryParse(VGpu.device, out var device) && device != 0)
-                deviceColumn.Value = device;
+            if (int.TryParse(VGpu.device, out var device))
+                deviceColumn.Value = Helpers.PlymouthOrGreater(VGpu.Connection) ? device + VGPU_PCI_SLOT_NUMBER_SHIFT : device;
```
kc284: What happens if the host is earlier than plymouth and the device is 0? In that case 0 means unassigned, shouldn't we use the hyphen?

liulinC: In that case, `0` actually means `11`, which is hard coded.
But per my test, when XC connect to a older XS with only 1 vgpu, the device id is never shown at all.

liulinC:
@kc284 @MihaelaStoica May I know is there any more concern about this PR? 
If no, can I ask to merge this PR?

PR #2465, opened 2019-07-09
 CA-323248: Allow VDI move to a local SR if the latter can be seen from the home hosts of the VMs the VDI is attached to.

kc284:

### DIFF
--- a/XenAdmin/Controls/SrPickerItem.cs
+++ b/XenAdmin/Controls/SrPickerItem.cs
@@ -76,10 +76,7 @@ namespace XenAdmin.Controls
             }
         }
 
-        protected override bool UnsupportedSR
-        {
-            get { return false; }
-        }
+        protected override bool UnsupportedSR => false;
     }
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


XenAdmin/Controls/SrPickerItem.cs
```diff
@@ -91,35 +88,19 @@ public SrPickerMigrateItem(SR sr, Host aff, long diskSize, VDI[] vdis) : base(sr
...
-        }
-
         /// <summary>
-        /// If the VM has a home host, we can move VDIs to a local SR only
-        /// if the latter belongs to the VM's home host
+        /// We can move VDIs to a local SR only if the VDI is attached to VMs
+        /// that have a home server that can see the SR
         /// </summary>
-        private bool HomeHostCanSeeTargetSr(VDI vdi)
+        private static bool HomeHostCanSeeTargetSr(VDI vdi, SR targetSr)
```
MihaelaStoica: Why was this method made static? Currently it is only used inside this class and with TheSR passed to the targetSr parameter.

kc284: Long ago that one used to be instance method `SrIsLocalToTheHostOnWhichTheVmIsRunning(SR sr, VDI vdi)` and I made it instance method `HomeHostCanSeeTargetSr(VDI vdi)` for this very reason. But now I've changed my mind and I think the name is more descriptive of what it does if it is `HomeHostCanSeeTargetSr(VDI vdi, SR targetSr)`. It doesn't need to be static, I only made it static to avoid people using TheSR in it since we pass targetSr.

MihaelaStoica: OK

borzel:
works fine :-) thanks!

PR #2466, opened 2019-07-11
CA-320203: Fixed the update wizard failure when installing two cumula…

MihaelaStoica:
…tive updates

The issue was that when installing 2 new versions in one go (e.g. two cumulative updates), the installation succeeds on master but it cannot continue on slaves, because it cannot find the first CU update anymore.
The fix is to refresh the update record before applying the update, with the record found in the cache. Also, if there is no update in the cache, try to re-introduce the update.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
@@ -76,12 +76,20 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
                 AddProgressStep(string.Format(Messages.UPDATES_WIZARD_APPLYING_UPDATE, xenServerPatch.Name,
                     host.Name()));
 
+                PatchPrecheckOnHostPlanAction.RefreshUpdate(host, mapping, session);
+
                 XenRef<Task> task = null;
 
                 if (mapping is PoolPatchMapping patchMapping)
+                {
+                    log.InfoFormat("Allpying patch on `{0}`. Update = `{1}` (uuid = `{2}`; opaque_ref = `{3}`)", host.Name(), patchMapping.Pool_patch.Name(), patchMapping.Pool_patch.uuid, patchMapping.Pool_patch.opaque_ref);
                     task = Pool_patch.async_apply(session, patchMapping.Pool_patch.opaque_ref, host.opaque_ref);
+                }
                 else if (mapping is PoolUpdateMapping updateMapping)
+                {
+                    log.InfoFormat("Applying update on `{0}`. Update = `{1}` (uuid = `{2}`; opaque_ref = `{3}`)", host.Name(), updateMapping.Pool_update.Name(), updateMapping.Pool_update.uuid, updateMapping.Pool_update.opaque_ref);
                     task = Pool_update.async_apply(session, updateMapping.Pool_update.opaque_ref, host.opaque_ref);
+                }
 
                 PollTaskForResultAndDestroy(Connection, ref session, task);
             }
--- a/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
@@ -82,12 +82,12 @@ namespace XenAdmin.Wizards.PatchingWizard.PlanActions
 
                 if (mapping is PoolPatchMapping patchMapping)
                 {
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
```diff
@@ -76,12 +76,20 @@ protected override void RunWithSession(ref Session session)
                 AddProgressStep(string.Format(Messages.UPDATES_WIZARD_APPLYING_UPDATE, xenServerPatch.Name,
                     host.Name()));
 
+                PatchPrecheckOnHostPlanAction.RefreshUpdate(host, mapping, session);
+
                 XenRef<Task> task = null;
 
                 if (mapping is PoolPatchMapping patchMapping)
+                {
+                    log.InfoFormat("Allpying patch on `{0}`. Update = `{1}` (uuid = `{2}`; opaque_ref = `{3}`)", host.Name(), patchMapping.Pool_patch.Name(), patchMapping.Pool_patch.uuid, patchMapping.Pool_patch.opaque_ref);
```
kc284: `Applying` `Patch = '{1}'`
Why are you using grave accents instead of single quots?

XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrecheckOnHostPlanAction.cs
```diff
@@ -120,5 +125,29 @@ protected override void RunWithSession(ref Session session)
...
+                    host.Name(), poolUpdateMapping.Pool_update.Name(), poolUpdateMapping.Pool_update.uuid,
+                    poolUpdateMapping.Pool_update.opaque_ref);
+                try
+                {
+                    var newUpdateRef = Pool_update.introduce(session, poolUpdateMapping.Pool_update.vdi.opaque_ref);
+                    session.Connection.WaitForCache(newUpdateRef);
+                }
+                catch (Exception e)
+                {
+                    log.Error("Failed to introduce update", e);
```
kc284: I'd log this as "re-introduce" for clarity.
Shouldn't we rethrow the exception to prevent it from continuing unnecessarily since the precheck will eventually fail?

XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrecheckOnHostPlanAction.cs
```diff
@@ -82,15 +83,19 @@ protected override void RunWithSession(ref Session session)
             {
                 AddProgressStep(string.Format(Messages.UPDATES_WIZARD_RUNNING_PRECHECK, xenServerPatch.Name, host.Name()));
 
+                RefreshUpdate(host, mapping, session);
+
                 List<Problem> problems = null;
 
                 if (mapping is PoolPatchMapping patchMapping)
                 {
+                    log.InfoFormat("Running patch precheck on `{0}`. Update = `{1}` (uuid = `{2}`; opaque_ref = `{3}`)", host.Name(), patchMapping.Pool_patch.Name(), patchMapping.Pool_patch.uuid, patchMapping.Pool_patch.opaque_ref);
```
kc284: `Patch = '{1}'`

XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
MihaelaStoica: No particular reason, probably copy-pasted from somewhere else

XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrecheckOnHostPlanAction.cs
MihaelaStoica: Yes, that's a good idea, and we can show a meaningfull message.

XenAdmin/Wizards/PatchingWizard/PlanActions/ApplyXenServerPatchPlanAction.cs
```diff
@@ -76,12 +76,20 @@ protected override void RunWithSession(ref Session session)
                 AddProgressStep(string.Format(Messages.UPDATES_WIZARD_APPLYING_UPDATE, xenServerPatch.Name,
                     host.Name()));
 
+                PatchPrecheckOnHostPlanAction.RefreshUpdate(host, mapping, session);
+
                 XenRef<Task> task = null;
 
                 if (mapping is PoolPatchMapping patchMapping)
+                {
+                    log.InfoFormat("Allpying patch on '{0}'. Patch{1}' (uuid = '{2}'; opaque_ref = '{3}')", host.Name(), patchMapping.Pool_patch.Name(), patchMapping.Pool_patch.uuid, patchMapping.Pool_patch.opaque_ref);
```
kc284: "Allpying" -> "Applying"

MihaelaStoica: Thanks @kc284 , I've updated the last commit

PR #2481, opened 2019-08-07
CP-30958: Initial implementation of the Conversion Manager in XenCenter.

kc284:


kc284:
I think the error display may still be a bit confusing depending on the timing of the errors' arrival. I wonder whether we should consider turning the conversion client requests (minus the heartbeat) to delegated async actions so they are logged on the events tab.

XenAdmin/Dialogs/ConversionDialog.resx
```diff
@@ -0,0 +1,852 @@
...
+    <value>934, 461</value>
+  </data>
+  <data name="$this.Font" type="System.Drawing.Font, System.Drawing">
+    <value>Segoe UI, 9pt</value>
+  </data>
+  <data name="$this.MinimumSize" type="System.Drawing.Size, System.Drawing">
+    <value>850, 500</value>
+  </data>
+  <data name="$this.Text" xml:space="preserve">
+    <value>Manage Conversions</value>
```
MihaelaStoica: It would be useful to show the pool name in the dialog title, as we can open a dialog for each connected pool

PR #2482, opened 2019-08-08
CA-293375: Fixed inconsistencies in the items related to removing servers from a pool

kc284:


XenModel/Actions/DelegatedAsyncAction.cs
```diff
@@ -39,31 +39,18 @@
...
+    /// Anything more complicated should be turned into a proper action of its own.
+    /// </summary>
     public class DelegatedAsyncAction : AsyncAction
     {
         private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
         private readonly Action<Session> invoker;
         private readonly string endDescription;
 
-        public DelegatedAsyncAction(IXenConnection connection, string title, string startDescription, string endDescription,
```
kc284: The compiler understands these constructors, however the latest Resharper issues warnings, so I removed it as a code style improvement.

XenAdmin/Commands/Controls/PoolRemoveServerToolStripMenuItem.cs
```diff
@@ -0,0 +1,106 @@
+﻿/* Copyright (c) Citrix Systems, Inc.
```
kc284: We can discuss if we want to take this change, but generally I think this is what the user would expect to see on the pool menu. I think the pevious item was confusing because if the user selected a server, I doubt they would go to the pool menu to remove it.

MihaelaStoica: I like these changes. Another inconsistency is that the context menu for the host doesn't show the "Remove from Pool" item for master, but the main menu shows it and allows it.

PR #2487, opened 2019-08-12
Basic travis set-up to run dev checks on PRs.

kc284:
Moved roaming check to NUnit tests and removed obsolete check.


.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: c
```
kc284: Unfortunately we can't build and run tests on travis because at the moment they only support Mono and .NET Core, hence the basic setup for c to simply run the dev checks.

PR #2496, opened 2019-08-28
Corrections and enhancements.

kc284:
Fixed version compatibility check. Added context menu to the conversion list.


kc284:
Please squash everything when done.

PR #2497, opened 2019-08-29
CP-32070: Hide the quiesce checkbox based on the version of the server

MihaelaStoica:


XenAdmin/Core/VMGroup.cs
```diff
@@ -194,9 +194,9 @@ internal static Predicate<Host> FeatureRestricted
             get { return typeof(T) == typeof(VMSS) ? ((Predicate<Host>)Host.RestrictVMSnapshotSchedule) : (Predicate<Host>)Host.RestrictVMAppliances; }
         }
 
-        internal static bool isQuescingSupported
+        internal static bool IsQuescingSupported(IXenConnection connection)
```
kc284: Typo (not caused by this PR) "Quiescing"

kc284:
As discussed, I think when we launch the NewPolicySnapshotTypePage as a Properties dialog page and there is an existing quiesced snapshot we should show a message to warn the user that they have to disable it because it won't be working any more. Otherwise the user may be confused because when the NewPolicySnapshotTypePage is launched as a wizard page the checkbox does not appear at all.

PR #2499, opened 2019-08-30
CA-322529: Do not add nulls to the sub-actions list for a multiple ac…

MihaelaStoica:
…tion

- This fixes the issue where we create a multiple action to delete the update installation files, but the sub-actions list contains null values because the connection to the host has been lost


XenAdmin/Wizards/PatchingWizard/PatchingWizard.cs
```diff
@@ -281,6 +281,7 @@ private void CleanUploadedPatches(bool forceCleanSelectedPatch = false)
                 }
             }
 
+            list.RemoveAll(a => a == null);
```
kc284: Instead of adding nulls and removing them afterwards, wouldn't it be better to check first and only add the action to the list if it is not null?

MihaelaStoica: Thanks @kc284, I've replace the commit.

PR #2501, opened 2019-08-31
Obsolete code and spelling corrections.

kc284:


kc284:
Some plug-in spelling corrections will be also coming via the bindings when https://github.com/xapi-project/xen-api/pull/3945 is merged.

PR #2511, opened 2019-09-06
CA-272770: Check whether the selected ISO SR supports VDI creation before copying the Fixup ISO on it.

kc284:
Also, minor layout correction, and usability enhancement: do not disable
the controls under the radio button when the latter is deselected.

### DIFF
--- a/XenAdmin/Wizards/ImportWizard/ImportOptionsPage.cs
+++ b/XenAdmin/Wizards/ImportWizard/ImportOptionsPage.cs
@@ -117,10 +117,13 @@ namespace XenAdmin.Wizards.ImportWizard
         /// Performs certain checks on the pages's input data and shows/hides an error accordingly
         /// </summary>
         /// <param name="checks">The checks to perform</param>
-        private void PerformCheck(params CheckDelegate[] checks)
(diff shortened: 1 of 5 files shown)


kc284:
Note that this change cannot account for cases where there is no write permission on the SR but the SM manager allows vdi_create (which is the case of the referenced ticket, but this is SM's issue).

XenAdmin/Wizards/ImportWizard/ImportOptionsPage.cs
```diff
@@ -374,24 +396,13 @@ private void m_comboBoxISOLibraries_SelectedIndexChanged(object sender, EventArg
...
             if (m_radioButtonDontRunOSFixups.Checked)
-            {
-                m_labelLocationFixupISO.Enabled = false;
-                m_comboBoxISOLibraries.Enabled = false;
-                m_pictureBoxInfo.Visible = false;
-                m_labelFixupISOInfo.Visible = false;
-                m_buttonNextEnabled = true;
-                OnPageUpdated();
-            }
+                ValidateSelection();
```
MihaelaStoica: With this change, the Next button doesn't get enabled if DontRunOSFixup is checked

PR #2525, opened 2019-10-11
CP-32033 Add PV guests precheck in Xencenter

xueqingz:
Signed-off-by: Xueqing Zhang <Xueqing.Zhang@citrix.com>
Add the PV guests precheck in both 'Install update' and 'RPU'. This precheck will give a warning when host have pv guess. 
The warning link is not ready yet , a likely link given for example.
The PV guests want to skip this check can add one key in vm.other_config named "pvcheckpass".

### DIFF
--- a/XenAdmin/Diagnostics/Checks/PVGuestsCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/PVGuestsCheck.cs
@@ -1,0 +1,86 @@
+﻿/* Copyright (c) Citrix Systems, Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer. 
+ * *   Redistributions in binary form must reproduce the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer in the documentation and/or other 
+ *     materials provided with the distribution. 
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+ * CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR 
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, 
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+ * SUCH DAMAGE.
+ */
+
+using System;
+using System.Linq;
+using XenAdmin.Core;
+using XenAPI;
+using XenAdmin.Diagnostics.Problems;
+using XenAdmin.Diagnostics.Problems.PoolProblem;
+using System.Collections.Generic;
+using System.Web.Script.Serialization;
+
+namespace XenAdmin.Diagnostics.Checks
+{
+    class PVGuestsCheck : HostPostLivenessCheck
+    {
+        private readonly Pool _pool;
+        private readonly bool _upgrade;
+        private readonly Dictionary<string, string> _installMethodConfig;
+        private readonly bool _manualUpgrade;
+        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+
+        public PVGuestsCheck(Pool pool, bool upgrade, Dictionary<string, string> installMethodConfig = null, bool manualUpgrade = false)
+            : base(Helpers.GetMaster(pool.Connection))
+        {
+            _pool = pool;
+            _upgrade = upgrade;
+            _installMethodConfig = installMethodConfig;
+            _manualUpgrade = manualUpgrade;
+        }
+
+        protected override Problem RunHostCheck()
+        {
+            string upgradePlatformVersion;
+            if (!_pool.Connection.Cache.VMs.Any(vm => vm.IsPvVm()))
+                return null;
+            if (!_upgrade || _manualUpgrade)
+                return new PoolHasPVGuestWarningUrl(this, _pool, _upgrade);
+            try
+            {
+                var result = Host.call_plugin(Host.Connection.Session, Host.opaque_ref, "prepare_host_upgrade.py", "getVersion", _installMethodConfig);
+                var serializer = new JavaScriptSerializer();
+                var version = (Dictionary<string, object>)serializer.DeserializeObject(result);
+                upgradePlatformVersion = version.ContainsKey("platform-version") ? (string)version["platform-version"] : null;
+            }
+            catch (Exception exception)
+            {
+                log.WarnFormat("Plugin call prepare_host_upgrade.getVersion on {0} failed with {1}", Host.Name(), exception.Message);
+                return new PoolHasPVGuestWarningUrl(this, _pool, _upgrade);
+            }
+            if (Helpers.QuebecOrGreater(upgradePlatformVersion))
+                return new PoolHasPVGuestWarningUrl(this, _pool, _upgrade);
+            return null;
+        }
+
+        public override string Description => Messages.PV_GUESTS_CHECK_DESCRIPTION;
+    }
+}
--- a/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
@@ -489,6 +489,19 @@ namespace XenAdmin.Wizards.PatchingWizard
                 
             }
 
+            //PVGuestsCheck checks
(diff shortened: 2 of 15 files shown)


XenAdmin/Diagnostics/Problems/HostProblem/HostHasPVGuest.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        public HostHasPVGuestWarning(Check check)
+            : base(check)
+        {
+        }
+        public override string Title => Description;
+        public override string Description
+        {
+            get { return string.Format(Messages.HOST_HAS_PV_GUEST_WARNING); }
+        }
```
kc284: We're trying to use newer C# syntac where possible, so please replace
```
public override string Description
{
    get { return string.Format(Messages.HOST_HAS_PV_GUEST_WARNING); }
}
```
with
```
public override string Description => string.Format(Messages.HOST_HAS_PV_GUEST_WARNING); 
```
Same goes for other properties elsewhere in this PR.

XenModel/Messages.resx
```diff
@@ -6380,6 +6383,9 @@ Click Configure HA to alter the settings displayed below.</value>
   <data name="HOST_GONE" xml:space="preserve">
     <value>Could not find host in [XenCenter]'s cache.</value>
   </data>
+  <data name="HOST_HAS_PV_GUEST_WARNING" xml:space="preserve">
+    <value>If you are upgrading to [XenServer] 8.1.0 or higher, some PV OS templates will be removed.</value>
```
kc284: Versions should not be hardcoded. Please use the variable BRANDING_XC_PRODUCT_8_1_VERSION (defined in `xenadmin.git/Branding/branding.sh`.

XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -866,6 +866,15 @@ public List<XenRef<VM>> GetRunningPvVMs()
             return vms.ToList();
         }
 
+        public List<VM> GetPvVMs()
```
kc284: I think it will make for a neater design if we add a method in `XenAPI-Extensions/VM.cs`, e.g.
```
public bool IsPvVm()
{
    return is_a_real_vm() && !IsHVM() && !other_config.ContainsKey("pvcheckpass");
}
```
and call that one from where it's needed.

The PVGuestsCheck  has been added as a host check, but the current method is not host specific since it's looking for VMs at the connection level - or do you rather mean VMs with the specific host as Home()? Also, do we care about real VMs only or custom templates and snapshots too? I think we need to clarify the specification/design a bit.

XenAdmin/Diagnostics/Checks/PVGuestsCheck.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    class PVGuestsCheck : HostCheck
+    {
+        public PVGuestsCheck(Host host)
+            : base(host)
+        {
+        }
+        protected override Problem RunCheck()
+        {
+            //return new HostHasPVGuestWarning(this);
```
kc284: Please remove commented out code.

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -165,6 +165,21 @@ protected override List<CheckGroup> GenerateChecks(Pool_patch patch)
                 groups.Add(new CheckGroup(Messages.CHECKING_XENCENTER_VERSION, new List<Check> {new XenCenterVersionCheck(null)}));
             }
 
+            //Checking PV guests - for hosts that have any PV guests and warn the user before the update/upgrade.
```
kc284: This check should move after the liveness checks; I think a better place would be towards the end after the post-upgrade memory check.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -470,6 +470,22 @@ protected virtual List<CheckGroup> GenerateCommonChecks(List<Host> applicableSer
...
+                    pvChecks.Add(new PVGuestsCheck(host)); 
+                    break;
+                }   
+            }
+             if (pvChecks.Count > 0)
+             {
+                 groups.Add(new CheckGroup(Messages.CHECKING_PV_GUESTS, pvChecks));
+             }
+            
+
```
kc284: This check should only run if we're updating to a new version. Currently it's run for any hotfix or suppack the user applies.

XenModel/Messages.resx
kc284: We need 2 different messages, one for updating and one for upgrading. Also, the message should not describe the code change, but rather the consequences for the user. For example, something like:
`Your PV guests may not function properly after the upgrade/update because support for PV guests is dropped as of [XenServer] [BRANDING_XC_PRODUCT_8_1_VERSION].`

XenModel/Messages.resx
```diff
@@ -10364,6 +10370,9 @@ Press OK to continue the wizard and return to the server and follow the instruct
   <data name="PROXY_SERVERS_NOT_SUPPORTED" xml:space="preserve">
     <value>Proxy servers are not supported.</value>
   </data>
+  <data name="PV_GUESTS_CHECK_DESCRIPTION" xml:space="preserve">
+    <value>PV guest check.</value>
```
kc284: More descriptive for the user: "Guest compatibility check"

XenModel/Messages.resx
```diff
@@ -2607,6 +2607,9 @@ Do you want to assign it to the schedule '{2}' instead?</value>
   <data name="CHECKING_PREPARE_TO_UPGRADE_DESCRIPTION" xml:space="preserve">
     <value>Host upgrade readiness check</value>
   </data>
+  <data name="CHECKING_PV_GUESTS" xml:space="preserve">
+    <value>Checking PV guests on hosts</value>
```
kc284: More descriptive for the user "Guest compatibility check".

XenModel/Messages.resx
kc284: Ah, I forgot to mention that the version the user is upgrading to can be found using the prepare_host_upgrade.py plugin. Please see the logic in HostMemoryPostUpgradeCheck.cs for details.

XenAdmin/Diagnostics/Problems/HostProblem/HostHasPVGuest.cs
```diff
@@ -0,0 +1,82 @@
...
+        public override string HelpMessage
+        {
+           get { return LinkText; }
+        }
+        public override string LinkText
+        {
+            get { return Messages.PATCHING_WIZARD_WEBPAGE_CELL; }
```
kc284: A more appopriate message is "Learn more" or "More info".

XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
kc284: Also, the content of the check (and the same in the RPU precheck page) can be written more efficiently. First of all, we need to decide whether it's a pool or host check. If we assume that it is a pool check (because this is what the logic in GetPvVMs() says) you can write it as:
```
var pvChecks = new List<Check>();
foreach (Host host in SelectedMasters.Where(h => !Helpers.QuebecOrGreater(h.Connection)))
{
    if (host.Connection.Cache.VMs.Any(vm => vm.IsPvVm()))
        pvChecks.Add(new PVGuestsCheck(host));
}

if (pvChecks.Count > 0)
    groups.Add(new CheckGroup(Messages.CHECKING_PV_GUESTS, pvChecks));
```

XenAdmin/XenAdmin.csproj
```diff
@@ -85,6 +85,9 @@
     <Reference Include="System.Drawing" />
     <Reference Include="System.Management" />
     <Reference Include="System.Security" />
+    <Reference Include="System.ValueTuple, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
```
MihaelaStoica: Why was this reference added to the project?

XenModel/Messages.resx
```diff
@@ -6380,6 +6383,14 @@ Click Configure HA to alter the settings displayed below.</value>
   <data name="HOST_GONE" xml:space="preserve">
     <value>Could not find host in [XenCenter]'s cache.</value>
   </data>
+  <data name="HOST_HAS_PV_GUEST_UPDATE_WARNING" xml:space="preserve">
+    <value>{0}: Your PV guests may not function properly after the update.
```
MihaelaStoica: We need a white space character after the full stop. (in both messages)

XenModel/Messages.resx
```diff
@@ -10364,6 +10378,9 @@ Press OK to continue the wizard and return to the server and follow the instruct
   <data name="PROXY_SERVERS_NOT_SUPPORTED" xml:space="preserve">
     <value>Proxy servers are not supported.</value>
   </data>
+  <data name="PV_GUESTS_CHECK_DESCRIPTION" xml:space="preserve">
+    <value>Guest compatibility check.</value>
```
MihaelaStoica: Please remove the full stop character.

XenModel/Messages.resx
MihaelaStoica: This still needs to be done

XenModel/Messages.resx
```diff
@@ -6380,6 +6383,12 @@ Click Configure HA to alter the settings displayed below.</value>
   <data name="HOST_GONE" xml:space="preserve">
     <value>Could not find host in [XenCenter]'s cache.</value>
   </data>
+  <data name="HOST_HAS_PV_GUEST_UPDATE_WARNING" xml:space="preserve">
+    <value>{0}: Support for paravirtualized (PV) guests is dropped as of [XenServer] [BRANDING_XC_PRODUCT_8_1_VERSION]. </value>
```
MihaelaStoica: I've only noticed this now, but here you should use BRANDING_VERSION_8_1

XenModel/InvisibleMessages.zh-CN.resx
```diff
@@ -168,6 +168,9 @@
   <data name="PRIVACY" xml:space="preserve">
     <value>http://citrix.com/English/aboutCitrix/legal/privacyStatement.asp?ntref=hp_nav_US</value>
   </data>
+  <data name="PV_GUESTS_CHECK_URL" xml:space="preserve">
+    <value>https://docs.citrix.com/en-us/citrix-hypervisor/system-requirements/guest-os-support.html</value>
```
MihaelaStoica: This should be "zh-cn" (instead of en-us)

XenModel/InvisibleMessages.ja.resx
```diff
@@ -168,6 +168,9 @@
   <data name="PRIVACY" xml:space="preserve">
     <value>http://citrix.com/English/aboutCitrix/legal/privacyStatement.asp?ntref=hp_nav_US</value>
   </data>
+  <data name="PV_GUESTS_CHECK_URL" xml:space="preserve">
+    <value>https://docs.citrix.com/en-us/citrix-hypervisor/system-requirements/guest-os-support.html</value>
```
MihaelaStoica: This should be "ja-jp" (instead of en-us)

PR #2530, opened 2019-10-21
CA-328267; application exit bug; other code improvements

kc284:

### DIFF
--- a/XenAdmin/Network/XenConnectionUI.cs
+++ b/XenAdmin/Network/XenConnectionUI.cs
@@ -62,8 +62,7 @@ namespace XenAdmin.Network
             if (interactive)
             {
                 // CA-214953 - Focus on this connection's dialog, if one exists, otherwise create one
-                ConnectingToServerDialog dlg;
-                if (connectionDialogs.TryGetValue(connection, out dlg))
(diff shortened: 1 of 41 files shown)


XenAdmin/Network/XenConnectionUI.cs
```diff
@@ -79,61 +79,42 @@ public static void BeginConnect(IXenConnection connection, bool interactive, For
...
+                using (var dialog = new AddServerDialog(connection, true))
+                    result = dialog.ShowDialog(Program.MainWindow) == DialogResult.OK;
+            });
             return result;
         }
 
         private static void RegisterEventHandlers(IXenConnection connection)
         {
             UnregisterEventHandlers(connection);
-            connection.ConnectionResult += Connection_ConnectionResult;
```
MihaelaStoica: With this change the connection result is not handled when the connection is attempted in the non-interactive mode, for example from ConnectAllHostsCommand

kc284: To recap: the 3rd commit was reworked. A couple of additional commits were pushed to fix a typo and simplify event signatures.

PR #2533, opened 2019-10-24
CA-187943: Show a warning when a new VM's disks cannot be accommodated on the template or default SR due to lack of space.

kc284:
This one does not need to go into master immediately. Also, I've made 4 commits for easier review, it may be better to squash them later.

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
+++ b/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
@@ -184,7 +184,7 @@ namespace XenAdmin.Wizards.NewVMWizard
                 return;
 
             AddButton.Enabled = DisksRadioButton.Checked && DisksGridView.Rows.Count < Template.MaxVBDsAllowed() - 1;
-            PropertiesButton.Enabled = DisksRadioButton.Checked && DisksGridView.SelectedRows.Count > 0;
+            EditButton.Enabled = DisksRadioButton.Checked && DisksGridView.SelectedRows.Count > 0;
             DeleteButton.Enabled = DisksRadioButton.Checked && DisksGridView.SelectedRows.Count > 0 && ((DiskGridRowItem)DisksGridView.SelectedRows[0]).CanDelete;
             DisksGridView.Enabled = DisksRadioButton.Checked;
             DisklessVMRadioButton.Enabled = Template.IsHVM() && InstallMethodIsNetwork;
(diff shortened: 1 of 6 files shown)


XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
```diff
@@ -167,88 +170,79 @@ private void LoadDisks()
...
                 return defaultSR;
+            }
 
             foreach (SR sr in connection.Cache.SRs)
             {
                 if (sr.CanCreateVmOn() &&
                     sr.CanBeSeenFrom(affinity) && sr.VdiCreationCanProceed(diskSize))
+                {
+                    if (suggestedSR != null && defaultSR != null)
+                        tooltip = Messages.NEWVMWIZARD_STORAGEPAGE_NOSPACE;
```
MihaelaStoica: When the suggestedSR is the default SR, the more suitable message would be NEWVMWIZARD_STORAGEPAGE_SUGGESTED_NOSPACE.

XenModel/Messages.resx
```diff
@@ -9226,9 +9226,18 @@ Review these settings, then click Previous if you need to change anything. Other
...
     <value>Created by template provisioner</value>
   </data>
+  <data name="NEWVMWIZARD_STORAGEPAGE_DEFAULT_NOSPACE" xml:space="preserve">
+    <value>The default SR '{0}' does not have enough free space for the new VM's disks. [XenCenter] has selected a different SR for you.</value>
+  </data>
   <data name="NEWVMWIZARD_STORAGEPAGE_NAME" xml:space="preserve">
     <value>Storage</value>
   </data>
+  <data name="NEWVMWIZARD_STORAGEPAGE_NOSPACE" xml:space="preserve">
+    <value>Neither the default SR nor the suggested by the template SR have enough free space for the new VM's disks. [XenCenter] has selected a different SR for you.</value>
```
MihaelaStoica: I think that "the SR suggested by the template" sounds better here

XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
kc284: I realised there is one more problem in the logic, namely when the VM has no home server.

kc284:
I have rebased the whole thing on master, please review only the last commit with the corrections.

PR #2537, opened 2019-11-04
CA-329435: Multiple vGPU settings are not preserved when importing OV…

liulinC:
Add support to export/import a VM with multiple vGPU in ovf/ova format

### DIFF
--- a/XenOvfApi/OVF.cs
+++ b/XenOvfApi/OVF.cs
@@ -1047,9 +1047,9 @@ namespace XenOvf
             log.DebugFormat("OVF.AddOperatingSystemSection completed {0}", vsId);
         }
 
-		public static string AddOtherSystemSettingData(EnvelopeType ovfObj, string vsId, string name, string value, string description)
+		public static string AddOtherSystemSettingData(EnvelopeType ovfObj, string vsId, string name, string value, string description, bool permitMultiple=false)
         {
-            return AddOtherSystemSettingData(ovfObj, vsId, Properties.Settings.Default.Language, name, value, description);
+            return AddOtherSystemSettingData(ovfObj, vsId, Properties.Settings.Default.Language, name, value, description, permitMultiple);
         }
         /// <summary>
         /// Add XEN Specific configuration Items.
@@ -1059,7 +1059,8 @@ namespace XenOvf
         /// <param name="name">Name of Parameter: is:  HVM-boot-policy (case sensitive)</param>
         /// <param name="value">value for the parameter</param>
         /// <param name="description">Description of parameter</param>
-		public static string AddOtherSystemSettingData(EnvelopeType ovfObj, string vsId, string lang, string name, string value, string description)
+        /// <param name="permitMultiple">Whether the setting data permit multiple values, default false</param>
+        public static string AddOtherSystemSettingData(EnvelopeType ovfObj, string vsId, string lang, string name, string value, string description, bool permitMultiple=false)
         {
             VirtualHardwareSection_Type[] vhsArray = FindVirtualHardwareSection(ovfObj, vsId);
             VirtualHardwareSection_Type vhs = null;
@@ -1087,8 +1088,9 @@ namespace XenOvf
             {
                 foreach (Xen_ConfigurationSettingData_Type xencsd in vhs.VirtualSystemOtherConfigurationData)
                 {
-                    // if we already have the item skip it here, new replaces old.
-                    if (xencsd.Name.ToLower() != name.ToLower())
+                    // If permitMultiple, the old items are kept, and the new item will be added later,
+                    // otherwise, the old item is skipped here and will be overridden by the new one
+                    if (permitMultiple || xencsd.Name.ToLower() != name.ToLower())
                     {
                         xencfg.Add(xencsd);
                     }
--- a/XenOvfTransport/Import.cs
+++ b/XenOvfTransport/Import.cs
@@ -334,18 +334,28 @@ namespace XenOvfTransport
 
                 #region Set vgpu
                 
-                GPU_group gpuGroup;              
-                VGPU_type vgpuType;
-                FindGpuGroupAndVgpuType(xenSession, vhs, out gpuGroup, out vgpuType);
-
-                if (gpuGroup != null)
+                List<Tuple<GPU_group, VGPU_type> >vgpus = FindGpuGroupAndVgpuType(xenSession, vhs);
+                foreach (var item in vgpus)
                 {
-                    var other_config = new Dictionary<string, string>();
+                    GPU_group gpuGroup = item.Item1;
+                    VGPU_type vgpuType = item.Item2;
+                    if (gpuGroup != null)
+                    {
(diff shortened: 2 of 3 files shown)


XenOvfTransport/Import.cs
```diff
@@ -1293,6 +1295,12 @@ private XenRef<VM> DefineSystem(Session xenSession, VirtualHardwareSection_Type
 	                    case "nvram":
 		                    newVm.NVRAM = SplitStringIntoDictionary(xcsd.Value.Value);
 		                    break;
+                        case "vgpu":
```
MihaelaStoica: Adding the vgpus to the hashtable is not needed here, because they are not going to be imported from the hashtable directly (we process them explicitly, in `#region Set vgpu` above)

liulinC: yes,  As the `UpdateFrom` match the key `VGPUs`, not `vgpu`,
I add vgpus to hashtable just to keep the original behavior (add vgpus to the hashtable),
But it make sense to skip vGPUs here. I will update it.

XenOvfTransport/Import.cs
```diff
@@ -1317,49 +1325,55 @@ private XenRef<VM> DefineSystem(Session xenSession, VirtualHardwareSection_Type
...
             var data = system.VirtualSystemOtherConfigurationData;
             if (data == null)
-                return;
+                return vgpus;
 
-            var datum = data.FirstOrDefault(s => s.Name == "vgpu");
+            var datum = data.Where(s => s.Name == "vgpu");
             if (datum == null)
-                return;
+                return vgpus;
```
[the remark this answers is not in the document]
liulinC: Just to keep the original code behavior to avoid potential risk, 
but your comments make sense, and will update

XenOvfTransport/Import.cs
```diff
@@ -1317,49 +1325,55 @@ private XenRef<VM> DefineSystem(Session xenSession, VirtualHardwareSection_Type
...
 
-            var vgpuTypes = VGPU_type.get_all_records(xenSession);
-            var vgpuKey = vgpuTypes.FirstOrDefault(v =>
-                v.Value.vendor_name == vendorName && v.Value.model_name == modelName);
+                if (vgpuKey.Equals(default(KeyValuePair<XenRef<VGPU_type>, VGPU_type>)))
+                    continue;
 
-            if (vgpuKey.Equals(default(KeyValuePair<XenRef<VGPU_type>, VGPU_type>)))
-                return;
+                vgpuType = vgpuKey.Value;
```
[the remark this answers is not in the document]
liulinC: I would like to keep this, as it provide a default value `null` and there is a chance that  this function may return a pair of `gpuGroup: valid value, vgpuType:null`

XenOvfTransport/Import.cs
```diff
@@ -1317,49 +1323,52 @@ private XenRef<VM> DefineSystem(Session xenSession, VirtualHardwareSection_Type
...
 
-            if (vgpuKey.Equals(default(KeyValuePair<XenRef<VGPU_type>, VGPU_type>)))
-                return;
+                var vgpuTypes = VGPU_type.get_all_records(xenSession);
+                var vgpuKey = vgpuTypes.FirstOrDefault(v =>
+                    v.Value.vendor_name == vendorName && v.Value.model_name == modelName);
 
-            vgpuType = vgpuKey.Value;
-            vgpuType.opaque_ref = vgpuKey.Key.opaque_ref;
+                if (!vgpuKey.Equals(default(KeyValuePair<XenRef<VGPU_type>, VGPU_type>)))
```
kc284: Can vgpuKey be null? This needs a null check.

XenOvfTransport/Import.cs
```diff
@@ -1293,6 +1295,10 @@ private XenRef<VM> DefineSystem(Session xenSession, VirtualHardwareSection_Type
 	                    case "nvram":
 		                    newVm.NVRAM = SplitStringIntoDictionary(xcsd.Value.Value);
 		                    break;
+                        case "vgpu":
+                            // Skip vGPUs here and not put vGPUs into the hashtable,
```
kc284: "and do not put"

XenOvfTransport/Import.cs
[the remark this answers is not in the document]
kc284: The original code uses `FirstOrDefault` which can return null, hence the check is needed. `Where` does not return null, so by keeping it you don't gain anything. What is needed though, is the null check on datum _before_ calling `Where`. If for some reason `VirtualSystemOtherConfigurationData` is not set, `Where` will crash.

XenOvfApi/OVF.cs
```diff
@@ -1087,8 +1088,9 @@ public static string AddOtherSystemSettingData(EnvelopeType ovfObj, string vsId,
             {
                 foreach (Xen_ConfigurationSettingData_Type xencsd in vhs.VirtualSystemOtherConfigurationData)
                 {
-                    // if we already have the item skip it here, new replaces old.
-                    if (xencsd.Name.ToLower() != name.ToLower())
+                    // If permitMultiple, the old items are kept, and the new item will be added later,
+                    // otherwise, the old item is skipped here and will be override by the new one
```
kc284: "overridden"

XenOvfTransport/Import.cs
kc284: Ignore, I read it more carefully: it's a struct, it can't be null.

XenOvfTransport/Import.cs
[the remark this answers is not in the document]
liulinC: ah, I deleted the null check by mistake,  Will update it.

XenOvfTransport/Import.cs
```diff
@@ -334,18 +334,20 @@ public void Process(Session xenSession, EnvelopeType ovfObj, string pathToOvf, s
...
+                    VGPU_type vgpuType = item.Item2;
+                    if (gpuGroup != null)
+                    {
+                        var other_config = new Dictionary<string, string>();
 
-                    if (Helpers.FeatureForbidden(xenSession, Host.RestrictVgpu))
-                        VGPU.create(xenSession, vmRef.opaque_ref, gpuGroup.opaque_ref, "0", other_config);
-                    else if (vgpuType != null)
-                        VGPU.create(xenSession, vmRef.opaque_ref, gpuGroup.opaque_ref, "0", other_config, vgpuType.opaque_ref);
+                        if (Helpers.FeatureForbidden(xenSession, Host.RestrictVgpu))
```
MihaelaStoica: If the vGPU feature is restricted, then we create passthrough vGPUs, but we shouldn't create more than one, as the passthrough type doesn't support multiple instances

liulinC: Yes, `passthrough` does not support  multiple at this moment,  the first call will success and the others would just fail. 

It make sense to limit only one vGPU now, and we may need to release this limitation if we can support multiple passthrough. 

I will add a log message stating the vGPU is forbidden and then break the loop tomorrow (I cannot connect to my workstation as the VPN is broken)

PR #2539, opened 2019-11-07
CA-329399: Improve the error message displayed if the update wizard i…

MihaelaStoica:
…f a host cannot be evacuated

If a host cannot be evacuated because other hosts in the pool are maintenance mode and pending a reboot following an update installation, then include instructions to reboot those hosts and then press Retry.
Also updated the text for the Manual mode, removing the reference to installing other updates and doing a reboot manually at the end, because that won't be always possible, due to the fact that now we put the hosts in maintenance mode before installing the update.

### DIFF
--- a/XenAdmin/Wizards/PatchingWizard/PlanActions/HostPlanAction.cs
+++ b/XenAdmin/Wizards/PatchingWizard/PlanActions/HostPlanAction.cs
@@ -31,6 +31,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using XenAPI;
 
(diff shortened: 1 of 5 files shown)


XenAdmin/Wizards/PatchingWizard/PlanActions/HostPlanAction.cs
```diff
@@ -90,6 +91,13 @@ protected void EvacuateHost(ref Session session)
                         log.WarnFormat("Host {0} cannot be avacuated: {1}", hostObj.Name(), f.Message);
                         throw new Exception(string.Format(Messages.PLAN_ACTION_FAILURE_NOT_ENOUGH_MEMORY, hostObj.Name()), f);
                     }
+
+                    if (f.ErrorDescription.Count > 0 && f.ErrorDescription[0] == Failure.NO_HOSTS_AVAILABLE)
+                    {
+                        log.WarnFormat("Host {0} cannot be avacuated: {1}", hostObj.Name(), f.Message);
```
ramakrishp: typo ? avacuated ->  `evacuated` ?

MihaelaStoica: Thanks Rama, I'll fix it

XenAdmin/Wizards/PatchingWizard/PlanActions/HostPlanAction.cs
```diff
@@ -90,6 +91,13 @@ protected void EvacuateHost(ref Session session)
                         log.WarnFormat("Host {0} cannot be avacuated: {1}", hostObj.Name(), f.Message);
                         throw new Exception(string.Format(Messages.PLAN_ACTION_FAILURE_NOT_ENOUGH_MEMORY, hostObj.Name()), f);
                     }
+
+                    if (f.ErrorDescription.Count > 0 && f.ErrorDescription[0] == Failure.NO_HOSTS_AVAILABLE)
+                    {
+                        log.WarnFormat("Host {0} cannot be avacuated: {1}", hostObj.Name(), f.Message);
+                        if (hostObj.Connection.Cache.Hosts.Any(h=>h.updates_requiring_reboot.Count > 0 && h.enabled == false))
```
kc284: Shorter: `!h.enabled`

PR #2553, opened 2019-11-25
CA-330890: update storage select wizard in XCM

xueqingz:
Signed-off-by: Xueqing Zhang <xueqing.zhang@citrix.com>
Add two kinds of warning in SR select wizard in Conversion Manager. One for the limit in Xapi that 'MAX_VHD_SIZE = 2040 * 1024 * 1024 * 1024' in sm/drivers/vhdutil.py , another for supporting to select a gfs2 SR smaller than request.

### DIFF
--- a/XenAdmin/Wizards/ConversionWizard/SrSelectionPage.cs
+++ b/XenAdmin/Wizards/ConversionWizard/SrSelectionPage.cs
@@ -45,7 +45,6 @@ namespace XenAdmin.Wizards.ConversionWizard
     {
         private bool _buttonNextEnabled;
         private long _requiredDiskSize;
-
         public SrSelectionPage()
         {
             InitializeComponent();
@@ -139,14 +138,20 @@ namespace XenAdmin.Wizards.ConversionWizard
 
         private void UpdateButtons()
         {
-            _buttonNextEnabled = comboBoxSr.SelectedItem is SrWrapper wrapper &&
-                                 wrapper.AvailableSpace >= _requiredDiskSize;
+            _buttonNextEnabled = comboBoxSr.SelectedItem is SrWrapper wrapper && wrapper.CanUse;
             OnPageUpdated();
         }
 
         private void ShowError(string msg)
         {
             pictureBoxError.Image = Images.StaticImages._000_error_h32bit_16;
+            labelError.Text = msg;
+            tableLayoutPanelError.Visible = true;
+        }
+
+        private void ShowWarning(string msg)
+        {
+            pictureBoxError.Image = Images.StaticImages._000_Alert2_h32bit_16;
             labelError.Text = msg;
             tableLayoutPanelError.Visible = true;
         }
@@ -229,13 +234,30 @@ namespace XenAdmin.Wizards.ConversionWizard
             if (comboBoxSr.SelectedItem is SrWrapper wrapper)
             {
                 SelectedSR = wrapper.item;
-
-                if (wrapper.AvailableSpace >= _requiredDiskSize)
-                    tableLayoutPanelError.Visible = false;
-                else
+                SM sm = SelectedSR.GetSM();
+                bool vdiSizeUnlimited = sm != null && Array.IndexOf(sm.capabilities, "LARGE_VDI") != -1;
+                bool isThinlyProvisioned = sm != null && Array.IndexOf(sm.capabilities, "THIN_PROVISIONING") != -1;
+                if (_requiredDiskSize > SR.DISK_MAX_SIZE && !vdiSizeUnlimited)
+                {
+                    ShowError(string.Format(Messages.SR_DISKSIZE_EXCEEDS_DISK_MAX_SIZE, Util.DiskSizeString(SR.DISK_MAX_SIZE, 0)));
+                    wrapper.CanUse = false;
+                }
+                else if (_requiredDiskSize > wrapper.AvailableSpace && !isThinlyProvisioned)
+                {
                     ShowError(Messages.CONVERSION_STORAGE_PAGE_SR_TOO_SMALL);
+                    wrapper.CanUse = false;
+                }
+                else if (_requiredDiskSize > wrapper.AvailableSpace)
+                {
+                    ShowWarning(Messages.CONVERSION_STORAGE_PAGE_SR_OVERCOMMIT);
+                    wrapper.CanUse = true;
+                }
+                else
+                {
+                    tableLayoutPanelError.Visible = false;
+                    wrapper.CanUse = true;
+                }
             }
-
             UpdatePieChart();
             UpdateButtons();
         }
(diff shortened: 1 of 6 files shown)


XenModel/Messages.resx
```diff
@@ -3422,6 +3422,12 @@ This action cannot be undone. Are you sure you want to continue?</value>
   <data name="CONVERSION_STORAGE_PAGE_REQUIRED_SPACE" xml:space="preserve">
     <value>Required space {0}</value>
   </data>
+  <data name="CONVERSION_STORAGE_PAGE_SR_OVERCOMMIT" xml:space="preserve">
+    <value>The SR is overcommitted, overcommitment can lead to filling up the virtual disk and blocking the virtual machine</value>
```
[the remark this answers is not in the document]
xueqingz: Ok, changed.

XenAdmin/Wizards/ConversionWizard/SrSelectionPage.cs
```diff
@@ -229,11 +237,27 @@ private void comboBoxSr_SelectedIndexChanged(object sender, EventArgs e)
             if (comboBoxSr.SelectedItem is SrWrapper wrapper)
             {
                 SelectedSR = wrapper.item;
-
                 if (wrapper.AvailableSpace >= _requiredDiskSize)
-                    tableLayoutPanelError.Visible = false;
+                    if (_requiredDiskSize > _limitsize && ((IList)_srtypes).Contains(SelectedSR.GetSRType(false).ToString()))
```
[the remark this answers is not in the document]
xueqingz: Little change of this logic, there are four kind of results here.
`
if (available >=  required)
     if(vdi size over limit){disk size exceed error}
     else {right}
else if (no vdi size limit)(overcommitted warning)
else {sr too small error}
`

XenAdmin/Wizards/ConversionWizard/SrSelectionPage.cs
```diff
@@ -229,13 +234,22 @@ private void comboBoxSr_SelectedIndexChanged(object sender, EventArgs e)
...
+                    else
+                    {
+                        tableLayoutPanelError.Visible = false;
+                        wrapper.CanUse = true;
+                    }
+                else if (Array.IndexOf(SelectedSR.GetSM().capabilities, "LARGE_VDI") != -1)
+                    {
+                        ShowWarning(Messages.CONVERSION_STORAGE_PAGE_SR_OVERCOMMIT);
+                        wrapper.CanUse = true;
+                    }
```
kc284: In the else-if you probably mean to check for `THIN_PROVISIONING` rather than `LARGE_VDI`,right?

I find this logic slightly more complex, plus that SelectedSR.VdiCreationCanProceed will run the thin provisioning check unnecessarily since it's nested under the `wrapper.AvailableSpace >= _requiredDiskSize` case.

One more thing that I'm thinking now is that, using VdiCreationCanProceed as is, is probably incorrect because it runs the thin provisioning check on the SR's free space, not the available space as calculated by the conversion manager. How about this one:

```
SM sm = SelectedSR.GetSM();
bool vdiSizeUnlimited = sm != null && Array.IndexOf(sm.capabilities, "LARGE_VDI") != -1;
bool isThinlyProvisioned = sm != null && Array.IndexOf(sm.capabilities, "THIN_PROVISIONING") != -1;

if (_requiredDiskSize > SR.DISK_MAX_SIZE && !vdiSizeUnlimited)
{
    ShowError(string.Format(Messages.SR_DISKSIZE_EXCEEDS_DISK_MAX_SIZE, Util.DiskSizeString(SR.DISK_MAX_SIZE, 0)));
    wrapper.CanUse = false;
}
else if (_requiredDiskSize > wrapper.AvailableSpace && !isThinlyProvisioned)
{
    ShowError(Messages.CONVERSION_STORAGE_PAGE_SR_TOO_SMALL);
    wrapper.CanUse = false;
}
else if (_requiredDiskSize > wrapper.AvailableSpace)
{
    ShowWarning(Messages.CONVERSION_STORAGE_PAGE_SR_OVERCOMMIT);
    wrapper.CanUse = true;
}
else
{
    tableLayoutPanelError.Visible = false;
    wrapper.CanUse = true;
}
```

xueqingz: Thanks Konstantina, this logic is better clear. And I mean for  'overcommit' situation, we need to check both 'vdiSizeUnlimited' and 'isThinlyProvisioned'.

kc284: Why do we need to check for vdiSizeUnlimited? I'm not aware of this requirement and indeed the corresponding code is SR.VdiCreationCanProceed does not related the two.

xueqingz: This is because in xcm when convert vm >2TB, we can only support gfs2 storage(vdiSizeUnlimited and isThinlyProvisioned) now. Besides the function SR.VdiCreationCanProceed can't be used to decide whether a SR can be overcommitted. For example, SR with availableSpace=100G, requiredDiskSize=200G:
For SR like 'smb' ,'ext' and 'nfs', vdiSizeUnlimited= false , isThinlyProvisioned= true and  SR.VdiCreationCanProceed return true, but it'll fail in vdi create process in a convert job actually. So I need to both check for vdiSizeUnlimited and isThinlyProvisioned in overcommitted situation.

kc284: In the example you have mentioned with availableSpace=100G, requiredDiskSize=200G, vdiSizeUnlimited= false , isThinlyProvisioned= true, you end up in the third block of the logic I have written above, which is the warning for overcommit. LARGE_VDI is not relevant, it only says whether the SR supports disk size >2TB.

xueqingz: Year, what I want to say is, the overcommit situation is just a warning, the user can continue to submit this job and the job will success. But if we in the example above, I'm sure the job will fail. So I think it should go to Error.CONVERSION_STORAGE_PAGE_SR_TOO_SMALL and I change this logic to now. 
               if (_requiredDiskSize > SR.DISK_MAX_SIZE && !vdiSizeUnlimited)
                {
                    ShowError(string.Format(Messages.SR_DISKSIZE_EXCEEDS_DISK_MAX_SIZE, Util.DiskSizeString(SR.DISK_MAX_SIZE, 0)));
                }
               else if (_requiredDiskSize > wrapper.AvailableSpace && isThinlyProvisioned && vdiSizeUnlimited)
                {
                    ShowWarning(Messages.CONVERSION_STORAGE_PAGE_SR_OVERCOMMIT);
                    wrapper.CanUse = true;
                }
                else if (_requiredDiskSize > wrapper.AvailableSpace)
                {
                    ShowError(Messages.CONVERSION_STORAGE_PAGE_SR_TOO_SMALL);
                }
                else
                {
                    tableLayoutPanelError.Visible = false;
                    wrapper.CanUse = true;
                }

kc284: Have you actually tested that the job will fail? And why is XCM different from the new disk creation?

xueqingz: Oh, I find I make a mistake here. I always thought about we should allow to convert a VM more than 2TB, so I add the vdiSizeUnlimited check in overcommit. You are right, in normal case, just need to check isThinlyProvisioned. I add all the test results in tickets.

PR #2555, opened 2019-11-28
CA-331420: Do not refresh the licence status on the general tab if the selected object is not a pool/host.

kc284:
- Removed AsyncServerTime class as it adds complexity and the risk of  getting an outdated time.
- Simplified event signatures and compacted nested method calls.
- Fixed typo and removed unused code.


kc284:
Rebased this to resolve conflicts after merging PR#2530.

PR #2562, opened 2019-12-23
CA-333236:  Calculate local time for HotfixEligibilityAlert properties on the UI thread.

kc284:
The Title and Description of HotfixEligibilityAlert  involve calculation of local
time which should be done on the UI thread, otherwise the time is not localised
correctly, and additionally the application crashes when dismissing or sorting
these alerts.
Also, stop spamming the logs with each alert removal; code efficiency correction.


jijiang:
Looks good to me.

PR #2586, opened 2020-01-13
Corrected node label.

kc284:


kc284:
Merging this by myself as it's straightforward.

PR #2598, opened 2020-01-26
CA-271455, CA-328267 regressions

kc284:
Also, remove obsolete lines that delete touch files (they are not created any more).


kc284:
With regards to the change for CA-328267, I think that code was there to fix something else and when I made changes to fix CA-328267 it became obsolete and additionally resulted to regressed behaviour. I think it should be removed.

PR #2604, opened 2020-02-13
CA-329461: VM is visible when in conversion

xueqingz:
This change to make the VM operation invisible while the conversion is in progress. 
Signed-off-by: Xueqing Zhang <xueqing.zhang@citrix.com>


kc284:
In the end we won't take in this change. The user generally knows that hidden objects are not available for normal interaction, but disabling actions completely is more drastic and may not be desired (even if we know that the actions might fail). As long as the VPX hides the VMs under conversion, the behaviour is consistent with existing practice in XenCenter (see transferVM or VMs in the process of being imported from an OVF/OVA package).

PR #2611, opened 2020-02-17
CP-32947 (vSwitch controller changes) + CA-192746 (MTU NumericUpDown validation)

kc284:


XenModel/Messages.resx
```diff
@@ -14377,6 +14395,9 @@ Are you sure you want to enable automated power management for this Host?</value
   <data name="XENSERVER_8_1" xml:space="preserve">
     <value>[XenServer] {0}</value>
   </data>
+  <data name="XENSERVER_8_2" xml:space="preserve">
```
MihaelaStoica: Can we use a generic message here, e.g. XENSERVER_VERSION = [XenServer] {0},
to replace XENSERVER_8_1 and XENSERVER_8_2?

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -199,6 +199,11 @@ protected override List<CheckGroup> GenerateChecks(Pool_patch patch)
                 groups.Add(new CheckGroup(Messages.CHECKING_SAFE_TO_UPGRADE, safeToUpgradeChecks));
             }
 
+            //protocol check - for each pool
+            var vswitchChecks = (from Host server in SelectedMasters
```
MihaelaStoica: The check is not needed if the current version is Stockholm or greater (as it will always return "OK") so we shouldn't add it here.

XenAdmin/Wizards/PatchingWizard/PatchingWizard_PrecheckPage.cs
```diff
@@ -501,6 +501,15 @@ protected virtual List<CheckGroup> GenerateCommonChecks(List<Host> applicableSer
                 }
                 if (pvChecks.Count > 0)
                     groups.Add(new CheckGroup(Messages.CHECKING_PV_GUESTS, pvChecks));
+
+                //protocol check - for each pool
+                var vSwitchChecks = (from Pool pool in SelectedPools.Where(p => Helpers.NaplesOrGreater(p.Connection))
```
MihaelaStoica: The check is not needed if the current version is Stockholm or greater (as it will always return "OK") so we can exclude Stockholm and greater hosts here.

XenModel/Messages.resx
kc284: Actually, since the product name-version is the same in non-English locales, I was thinking of removing them completely from the Messages.resx and moving them as XENSERVER_8_1 ="Citrix Hypervisor 8.1", XENSERVER_8_2="Citrix Hypervisor 8.2" to Branding.resx. They would replace the PRODUCT_VERSION_xxx constants because I don't think we use the numbers alone anywhere and we can avoid nested string.Format() calls.

kc284: In any case, I'll leave this change for a different PR on master (as the last installment of CP-31105).

XenAdmin/Diagnostics/Checks/VSwitchControllerCheck.cs
```diff
@@ -0,0 +1,91 @@
...
+            string upgradePlatformVersion = null;
+
+            if (_installMethodConfig != null)
+                Host.TryGetUpgradeVersion(Host, _installMethodConfig, out upgradePlatformVersion, out _);
+                
+            // we know they are upgrading to Stockholm or greater, so block them
+            if (Helpers.StockholmOrGreater(upgradePlatformVersion))
+                return new VSwitchControllerProblem(this, _pool);
+
+            // we don't know the upgrade version, so add generic warning
```
MihaelaStoica: If we know the platform version (upgradePlatformVersion is not null) and it is not Stockholm or greater, then we shouldn't add the warning (for example upgrading to 8.1)

XenModel/Messages.resx
```diff
@@ -10556,6 +10571,15 @@ Please reconnect the host and try again</value>
   <data name="PROBLEM_HOSTPROBLEM_TITLE" xml:space="preserve">
     <value>Server '{0}'</value>
   </data>
+  <data name="PROBLEM_VSWITCH_CONTROLLER_DESCRIPTION" xml:space="preserve">
+    <value>{0}: Support for the vSwitch Controller is dropped as of {1}</value>
+  </data>
+  <data name="PROBLEM_VSWITCH_CONTROLLER_HELP" xml:space="preserve">
```
kc284: This one I'll probably have to remove, it may not be possible to deconfigure it via XenCenter.

XenModel/Messages.resx
```diff
@@ -10479,7 +10494,7 @@ Please reconnect the host and try again</value>
     <value>This pool has no shared storage</value>
   </data>
   <data name="POOL_HAS_PV_GUEST_WARNING" xml:space="preserve">
-    <value>{0}: Support for paravirtualized (PV) guests is dropped as of [XenServer] {1}. Click "Learn more" to see the list of supported guest operating systems. </value>
+    <value>{0}: Support for paravirtualized (PV) guests has been removed in [XenServer] {1}. Click Learn more to see the list of supported guest operating systems. </value>
```
MihaelaStoica: While testing this I've noticed XenCenter doesn't show this warning because of a format exception introduced in 
https://github.com/xenserver/xenadmin/commit/f7391e991b7f53c67084c48cc2d18d2117a4a4e3#diff-759b56e08e904e5eed5cac3cfc9b5fdbL53

kc284: See https://github.com/xenserver/xenadmin/pull/2615

PR #2615, opened 2020-02-24
 CP-32844: Drop legacy security protocols

kc284:


XenAdmin/Diagnostics/Checks/PoolLegacySslCheck.cs
```diff
@@ -0,0 +1,91 @@
...
+            string upgradePlatformVersion = null;
+
+            if (_installMethodConfig != null)
+                Host.TryGetUpgradeVersion(Host, _installMethodConfig, out upgradePlatformVersion, out _);
+                
+            // we know they are upgrading to Stockholm or greater, so block them
+            if (Helpers.StockholmOrGreater(upgradePlatformVersion))
+                return new LegacySslProblem(this, _pool);
+
+            // we don't know the upgrade version, so add generic warning
```
MihaelaStoica: If we know the platform version (upgradePlatformVersion is not null) and is not Stockholm or greater, then we shouldn't add the warning (for example upgrading to 8.1)

PR #2620, opened 2020-03-05
Script path correction.

kc284:


kc284:
Merging by myself to fix the build where the scripts failed silently to run.

PR #2627, opened 2020-03-24
Refactored three button dialog

kc284:
... because the constructors were more complicated than they should.
If reviewed separately, the commits are pretty quick to get through. But they should be squashed before merging.


kc284:
Rebased on master after the merge of feature/REQ-814.

kc284:
Rebased on master again after the merge of feature/REQ-811.

PR #2631, opened 2020-03-26
CA-337287: Skip vSwitch controller and PV guest checks with a warning if the RPU hotfix is missing

kc284:


XenAdmin/Diagnostics/Checks/PVGuestsCheck.cs
```diff
@@ -45,39 +46,41 @@ class PVGuestsCheck : HostPostLivenessCheck
...
-                var version = (Dictionary<string, object>)serializer.DeserializeObject(result);
-                upgradePlatformVersion = version.ContainsKey("platform-version") ? (string)version["platform-version"] : null;
-            }
-            catch (Exception exception)
-            {
-                log.Warn($"Plugin call prepare_host_upgrade.getVersion on {Host.Name()} threw an exception.", exception);
+
+            var hotfix = HotfixFactory.Hotfix(Host);
+            if (hotfix != null && hotfix.ShouldBeAppliedTo(Host))
+                return new HostDoesNotHaveHotfixWarning(this, Host);
```
MihaelaStoica: This warning is now displayed for th eManual upgrade mode, but in this mode we do not require the RPU hotfixes to be installed prior to the upgrade.

(Same for the vSwitch controller check)

PR #2646, opened 2020-04-10
CP-32687, CP-32688, CP-32689 Tools disaggregation

kc284:


XenAdmin/Commands/InstallToolsCommand.cs
```diff
@@ -298,54 +297,32 @@ public bool ConfirmAndExecute()
...
+            //whether RBAC allows connection to the VM's console
+            return vm.Connection.Session != null &&
+                   Role.CanPerform(new RbacMethodList("http/connect_console"), vm.Connection, out _, false);
         }
 
         protected override bool CanExecuteCore(SelectedItemCollection selection)
         {
-            return selection.AllItemsAre<VM>() && selection.AtLeastOneXenObjectCan<VM>(CanExecute);
+            return selection.Count > 0 &&
+                   selection.All(v => v.XenObject is VM vm &&
```
kc284, line 321: I changed the "at least one" to "all" because for cross pool selection of pre- and post- stockholm pools the experience will be confusing for the user.

PR #2651, opened 2020-04-20
CA-338184: Check for null before accessing host properties.

kc284:


XenAdmin/Commands/RebootVMCommand.cs
```diff
@@ -77,107 +76,44 @@ protected override bool CanExecute(VM vm)
...
-
-        private static bool EnabledTargetExists(Host host, IXenConnection connection)
-        {
-            if (host != null)
-                return host.enabled;
 
-            foreach (Host h in connection.Cache.Hosts)
-            {
-                if (h.enabled)
+                var hosts = selection[0].Connection.Cache.Hosts;
```
MihaelaStoica, line 83: Shall we also check that `selection[0].Connection` is not null?

PR #2664, opened 2020-04-29
Merge feature/REQ-716 into master

kc284:


XenAdmin/Commands/InstallToolsCommand.cs
```diff
@@ -84,38 +84,12 @@ public InstallToolsCommand(IMainWindow mainWindow, VM vm)
...
-            }
-
-            return null;
-        }
-
         /// <summary>
         /// Attempts to install tools on the vm
         /// </summary>
         /// <param name="vm"></param>
         /// <returns>null if user cancels or an AsyncAction. This is either the InstallPVToolsAction or the CreateCdDriveAction if the VM needed a DVD drive.</returns>
```
AaronRobson, line 91: There is no longer a return value here, the comment is out of date.

XenAdmin/Commands/InstallToolsCommand.cs
```diff
@@ -132,38 +106,41 @@ private AsyncAction SingleVMExecute(VM vm)
...
                 }
-            }
-            return null;
         }
 
         /// <summary>
         /// Attempts to install tools on several VMs
         /// </summary>
         /// <param name="vms"></param>
         /// <returns>Whether the action was launched (i.e., the user didn't Cancel)</returns>
```
AaronRobson, line 142: This return value comment is also out of date.

PR #2665, opened 2020-05-03
CA-333443 + some layout tweaks

kc284:


XenModel/Actions/Pool/AddHostToPoolAction.cs
```diff
@@ -91,24 +92,31 @@ protected override void Run()
                 bool fixedCpus = FixCpus(Pool, _hostsToCpuMask, AcceptNTolChanges);
                 if (fixedCpus)
                     Session = NewSession();  // We've rebooted the server, so we need to grab the new session
-                RelatedTask = XenAPI.Pool.async_join(Session, Pool.Connection.Hostname, Pool.Connection.Username, Pool.Connection.Password);
+
+                var master = Pool.Connection.TryResolveWithTimeout(Pool.master);
+                var address = master != null ? master.address : Pool.Connection.Hostname;
```
AaronRobson, line 97: The following would I think be equivalent and arguably clearer:
`var address = master?.address ?? Pool.Connection.Hostname;`

XenModel/Actions/Pool/AddHostToPoolAction.cs
```diff
@@ -91,24 +92,31 @@ protected override void Run()
...
                 {
-                    Session[] sessions = new Session[] { Session, Pool.Connection.Session };
-                    // Special parse to cope with multiple connections.
-                    Failure.ParseRBACFailure(f, sessions);
-                    // Will not get RBAC parsed again after the throw as we have altered the error description in ParseRBACFailure
-                    throw f;
+                    var sessions = new[] {Session, Pool.Connection.Session};
+                    var authRoles = Role.ValidRoleList(f.ErrorDescription[1], Session.Connection);
+
+                    var output = string.Join(", ", sessions.Select(s => string.Format(Messages.ROLE_ON_CONNECTION,
```
AaronRobson, line 111: +1 for `string.Join` over the old way of including the separator and removing the final one later.

XenModel/Actions/Pool/AddHostToPoolAction.cs
kc284: The two expressions are not exactly equivalent, namely `var address = master != null ? master.address : Pool.Connection.Hostname` will pass null into Pool.async_join if address is null, while `var address = master?.address ?? Pool.Connection.Hostname` will pass in the connection hostname. Not sure which behaviour is the best - the former will cause Pool.async_join to fail, but maybe the failure is desirable because the address shoud not be null. What do you think?

PR #2666, opened 2020-05-03
CA-314005: Proceed to setting the memory if the issued shutdown has been interrupted by a different shutdown task

kc284:


XenModel/Actions/VM/ChangeMemorySettingsAction.cs
```diff
@@ -73,21 +74,9 @@ public class ChangeMemorySettingsAction : AsyncAction
...
+        private bool staticChanged => static_min != VM.memory_static_min || static_max != VM.memory_static_max;
 
-        private bool needReboot
-        {
-            get
-            {
-                return (staticChanged && VM.power_state != vm_power_state.Halted);
-            }
-        }
+        private bool needReboot => staticChanged && VM.power_state != vm_power_state.Halted;
```
AaronRobson, line 79: +1 for using the nice short form for these properties while in the neighbourhood.

XenModel/Actions/VM/ChangeMemorySettingsAction.cs
```diff
@@ -103,7 +92,22 @@ protected override void Run()
...
+                {
+                    action.RunExternal(Session);
+                }
+                catch
+                {
+                    if (VM.power_state == vm_power_state.Halted || VM.current_operations.Any(op =>
+                        op.Value == vm_operations.clean_shutdown || op.Value == vm_operations.hard_shutdown ||
+                        op.Value == vm_operations.shutdown))
+                    {
+                        //ignore if it got already powered off or there are other shutting down tasks in progress
```
AaronRobson, line 106: If there are other shutting down tasks in progress but that haven't happened yet, will the code below (setting the memory) cope with this, does it need to wait until the shutdown has occurred (perhaps with rechecking, in case the other task is cancelled or fails)?

kc284: It won't cope, it will reboot the VM with the same memory settings as before. I have to say I didn't manage to reproduce this scenario, but I can't exclude it might happen. Ok, Iet me revisit this.

XenModel/Network/XenConnection.cs
```diff
@@ -642,7 +642,7 @@ public void WaitFor(Func<bool> predicate, Func<bool> cancelling)
             }
             try
             {
-                while (true)
+                for (int i = 0; i < 120; i++)
```
kc284, line 645: I changed this one to behave like the WaitForCache calls and time out after 1min instead of relying on the user to cancel it. The only other places where WaitFor is used is for renaming tags and setting folders, which normally finish quite fast - according to my testing at least - so the change should be ok.

PR #2668, opened 2020-05-06
CA-339237 (XSI-687) + a couple of random corrections

kc284:


XenModel/Utils/Helpers.cs
```diff
@@ -1050,7 +1050,7 @@ private static string GetFriendlyDataSourceName_(string name, IXenObject iXenObj
                 VBD vbd = FindVBD(iXenObject, m.Groups[1].Value);
                 return vbd == null
                            ? null
-                           : FormatFriendly(string.Format("Label-performance.vbd_{0}{1}", m.Groups[3].Value, m.Groups[4].Value), vbd.userdevice);
+                           : FormatFriendly(string.Format("Label-performance.vbd{0}{1}", m.Groups[3].Value, m.Groups[5].Value), vbd.userdevice);
```
AaronRobson, line 1053: Though I am convinced by the `read`/`write` and `latency` sections being optional and therefore their underscores being optional.
But it looks to me that the `vbd` needs to be followed by an underscore as it did before.

The regexes and the helper code here aren't clearly followed. If it would be at all practical, I think that this area could do with an example or two of the expected behaviour as unit tests.

kc284: Ok. This is going to make the PR a bit longer than I had planned though, so I think what I'll do is remove this commit and put it with the unit tests in a different PR.

kc284: You can approve this one if the other commits are ok.

PR #2669, opened 2020-05-07
CA-339271: Improved action description.

kc284:
Also, added cancelling ability to MoveVM action; made some simplifications.

### DIFF
--- a/XenModel/Actions/VM/VMMoveAction.cs
+++ b/XenModel/Actions/VM/VMMoveAction.cs
@@ -40,20 +40,21 @@ namespace XenAdmin.Actions.VMActions
 {
     public class VMMoveAction : AsyncAction
     {
-        public Dictionary<string, SR> StorageMapping;
+        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+
+        private Dictionary<string, SR> _storageMapping;
 
         public VMMoveAction(VM vm, Dictionary<string, SR> storageMapping, Host host)
-            : base(vm.Connection, Messages.ACTION_VM_MOVING)
+            : base(vm.Connection, string.Format(Messages.ACTION_VM_MOVING, vm.Name()))
         {
-            this.Description = Messages.ACTION_PREPARING;
             this.VM = vm;
             this.Host = host;
             this.Pool = Helpers.GetPool(vm.Connection);
             if (vm.is_a_template)
                 this.Template = vm;
 
-            StorageMapping = storageMapping;
-            SR = StorageMapping.Values.FirstOrDefault();
+            _storageMapping = storageMapping;
+            SR = _storageMapping.Values.FirstOrDefault();
 
             PopulateApiMethodsToRoleCheck();
         }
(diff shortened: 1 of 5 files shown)


XenModel/Actions/VM/VMMoveAction.cs
```diff
@@ -95,68 +96,77 @@ private void PopulateApiMethodsToRoleCheck()
...
-            {
-                this.Description = string.Format(Messages.MOVE_CANCELLED, VM.Name());
-                throw;
-            }
-            this.Description = Messages.MOVED;
+
+            if (SR != null)
+                VM.set_suspend_SR(Session, VM.opaque_ref, SR.opaque_ref);
+
+            if (exceptions.Count > 0)
```
AaronRobson, line 166: Purely a style thing but `if (exceptions.Any())` I think looks a bit clearer.

XenModel/Actions/VM/VMMoveAction.cs
```diff
@@ -95,68 +96,77 @@ private void PopulateApiMethodsToRoleCheck()
 
         protected override void Run()
```
AaronRobson, line 97: I think I understand the intention around having the messages refer to the specific VM that is moving.
I'm a bit in the dark around the other changes, which appear to be:
1. Allowing errors in destroying VDIs to not stop the processing of others (and be reported later).
2. The removal of Cancel action code, is this handled in generic action handling code somewhere else these day?

I'll leave this one for the time being and look at other PRs, I probably just need to come back later when I've got my head back into it.

XenModel/Actions/VM/VMMoveAction.cs
kc284: Well, it's not just a style thing. If the sizes are not too huge there is no practical difference, but theoretically speaking, `Count > 0` performs better than `Any()` because `exceptions` is a `List<T>`. If it were an `IEnumerable<T>`, `Any()` would be a better choice over `Count() > 0`. Of course, I'm not going to pretend that the usage of the one vs the other has been taken carefully into account throughout XenCenter; we should probably perform an audit at some point because linq is great for conciseness, but has caveats...

XenModel/Actions/VM/VMMoveAction.cs
kc284: 1. The other changes were aimed at making the code a bit more efficient (for example, do not calculate or resolve anything before/unless it is needed) and simpler to understand (I'm trying to get rid of the base method BestEffort() and its passing around refs and delegates, because it is unnecessary complexity in the code; there are still a few instances of it that can be removed in future PRs). Otherwise, the changed code does the same as before: once the VM is copied, the VDIs at the old location should be destroyed, but if one fails, there is no need to stop and prevent the rest in the list from being destroyed, hence the late failure.
2.  In this particular case, catching and rethrowing the CancelException serves for showing a specialised "move cancelled" message. Generally, catching and re-throwing exceptions has some performance impact, so I thought it's not worth it since the base class does handle CancelException, albeit with a slightly more generic "cancelled" message. There are many other actions, particularly the ones more recently written, that prefer the latter.

PR #2670, opened 2020-05-07
CA-339312: Explain why a host cannot be removed from the pool.

kc284:


XenAdmin/Commands/RemoveHostFromPoolCommand.cs
```diff
@@ -121,38 +120,40 @@ protected override void ExecuteCore(SelectedItemCollection selection)
...
+        {
+            if (item is Host host)
+            {
+                if (host.IsMaster())
+                    return Messages.MESSAGEBOX_POOL_MASTER_REMOVE;
+
+                if (!host.IsLive())
+                    return Messages.HOST_UNREACHABLE;
+
+                if (host.resident_VMs != null && host.resident_VMs.Count > 1)
```
AaronRobson, line 133: Could consider `host.resident_VMs.Any()` instead of `host.resident_VMs.Count > 1`.

kc284: `Any()` means `>0`, but it has to be `>1` because the first resident VM is dom0 which need to be excluded from the check.

AaronRobson: Good point, I missed that.

PR #2672, opened 2020-05-09
CA-339331: Skip uses deferred execution and needs enumeration.

kc284:


kc284:
@AaronRobson I forgot to mention on this PR that there is the SDK counterpart at https://github.com/xapi-project/xen-api-sdk/pull/173. If you could please have a look at that one too; thanks in advance.

PR #2673, opened 2020-05-09
CA-338830 + CA-339306

kc284:


XenModel/PoolJoinRules.cs
```diff
@@ -71,6 +71,7 @@ public enum Reason
             NonCompatibleManagementInterface,
             WrongRoleOnMaster,
             WrongRoleOnSlave,
+            HasClusteringEnabled,
```
AaronRobson, line 74: I'm assuming that the order of the elements of the enum is of no programatic consequence i.e. there is no storing of its index or anything like this.

kc284: It is of logical consequence, i.e. the order is used by the IComparable implementation in ConnectionWrapperWithMoreStuff to decide how the servers will be sorted on the New Pool dialog, and servers with stronger reason to be disabled are listed first (after the allowed ones). One has to ensure the checks in CanJoinPool() are performed in the same order as in the enum.

PR #2680, opened 2020-05-21
CA-339384: Take account of thin provisioning when placing a disk on a target SR.

kc284:
Also, set the required disk size to zero when the disk spinner value is invalid.

### DIFF
--- a/XenAdmin/Controls/SrPickerItem.cs
+++ b/XenAdmin/Controls/SrPickerItem.cs
@@ -39,8 +39,8 @@ namespace XenAdmin.Controls
 {
     public class SrPickerLunPerVDIItem : SrPickerVmItem
     {
-        public SrPickerLunPerVDIItem(SR sr, Host aff, long diskSize, VDI[] vdis)
-            : base(sr, aff, diskSize, vdis)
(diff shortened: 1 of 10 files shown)


XenAdmin/Controls/SrPickerItem.cs
```diff
@@ -231,12 +231,13 @@ public abstract class SrPickerItem : CustomTreeNode, IComparable<SrPickerItem>
...
         protected readonly VDI[] existingVDIs;
 
-        protected SrPickerItem(SR sr, Host aff, long diskSize, VDI[] vdis)
+        protected SrPickerItem(SR sr, Host aff, VDI[] vdis)
         {
             existingVDIs = vdis ?? new VDI[0];
             TheSR = sr;
             Affinity = aff;
-            DiskSize = diskSize;
+            DiskSize = existingVDIs.Sum(vdi =>
```
kc284, line 239: One thing I'm wondering about is whether this calculation should move closer to where SR.VdiCreationCanProceed is called.

AaronRobson (approved):

MihaelaStoica (approved):

PR #2681, opened 2020-05-22
CP-32921: Use the full path for the MSBuild executable. Temporary new location for the third party libraries.

kc284:


MihaelaStoica:
👍

PR #2682, opened 2020-05-26
Test corrections. Fixes build failure.

kc284:
- Exclude VNC key map from l10n tests.
- Added parameterless constructors to the archive classes for use by the tests.
- Scalability tests were removed.

### DIFF
--- a/XenCenterLib/Archive/ZipArchiveWriter.cs
+++ b/XenCenterLib/Archive/ZipArchiveWriter.cs
@@ -38,17 +38,17 @@ namespace XenCenterLib.Archive
 {
     class DotNetZipZipWriter : ArchiveWriter
     {
-        private ZipOutputStream zip = null;
+        private ZipOutputStream zip;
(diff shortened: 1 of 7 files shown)


XenCenterLib/Archive/ZipArchiveWriter.cs
```diff
@@ -38,17 +38,17 @@ namespace XenCenterLib.Archive
...
     {
-        private ZipOutputStream zip = null;
+        private ZipOutputStream zip;
         private bool disposed;
 
-        public DotNetZipZipWriter(Stream outputStream) : this()
+        public DotNetZipZipWriter(Stream outputStream)
         {
             zip = new ZipOutputStream( outputStream ) {EnableZip64 = Zip64Option.AsNecessary};
+            disposed = false;
```
kc284, line 47: This is not really needed, but I left it there for symmetry with what the constructors in the other archive classes do.

PR #2686, opened 2020-06-02
CA-337270: Perform authentication on a background thread.

kc284:
- Typo.
- Added controls to display the error in case authentication fails.
- Use the same control (LinkLabel) to show the rubric either when new authentication is needed or an existing one can be used.
- Moved privacy policy statement next to the checkbox because it looked crammed.

Commits best reviewed separately. It's also recommended to check out the code and quickly run it due to the inherent difficulty of diffing designer changes.


kc284:
@AaronRobson Ithis will need review again as new changes cancel earlier approvals.

PR #2688, opened 2020-06-04
CP-33153: Added friendly title and description for GFS2_CAPACITY alert. Corrected scan action description.

kc284:
The handling of this alert is a bit unorthodox because it is not listed among the known Message types since it is coming from a xapi plugin and the API is not obliged to know about it.

### DIFF
--- a/XenAdmin/Alerts/Types/MessageAlert.cs
+++ b/XenAdmin/Alerts/Types/MessageAlert.cs
@@ -114,7 +114,7 @@ namespace XenAdmin.Alerts
                         }
                         break;
 
-                    // applies to is hosts, vms and pools where only the name is required
+                    // applies to hosts, vms and pools where only the name is required
                     case Message.MessageType.HA_HEARTBEAT_APPROACHING_TIMEOUT:
                     case Message.MessageType.HA_HOST_FAILED:
                     case Message.MessageType.HA_HOST_WAS_FENCED:
@@ -240,6 +240,11 @@ namespace XenAdmin.Alerts
                         VMSS vmss = Helpers.XenObjectFromMessage(Message) as VMSS;
                         var policyAlertVMSS = new PolicyAlert(Message.priority, Message.name, Message.timestamp, Message.body, (vmss == null) ? "" : vmss.Name());
                         return policyAlertVMSS.Text;
+
+                    case Message.MessageType.unknown when Message.name == "GFS2_CAPACITY":
+                        if (XenObject != null)
+                            return string.Format(Message.FriendlyBody(Message.name), XenObject.Name());
+                        break;
                 }
 
                 return Message.body;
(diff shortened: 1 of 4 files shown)


XenAdmin/Alerts/Types/MessageAlert.cs
```diff
@@ -240,6 +240,11 @@ public override string Description
                         VMSS vmss = Helpers.XenObjectFromMessage(Message) as VMSS;
                         var policyAlertVMSS = new PolicyAlert(Message.priority, Message.name, Message.timestamp, Message.body, (vmss == null) ? "" : vmss.Name());
                         return policyAlertVMSS.Text;
+
+                    case Message.MessageType.unknown when Message.name == "GFS2_CAPACITY":
```
AaronRobson, line 244: I wasn't aware of the `when` keyword till now. Learn something new every day (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when).

XenAdmin/Alerts/Types/MessageAlert.cs
```diff
@@ -240,6 +240,11 @@ public override string Description
                         VMSS vmss = Helpers.XenObjectFromMessage(Message) as VMSS;
                         var policyAlertVMSS = new PolicyAlert(Message.priority, Message.name, Message.timestamp, Message.body, (vmss == null) ? "" : vmss.Name());
                         return policyAlertVMSS.Text;
+
+                    case Message.MessageType.unknown when Message.name == "GFS2_CAPACITY":
+                        if (XenObject != null)
+                            return string.Format(Message.FriendlyBody(Message.name), XenObject.Name());
```
AaronRobson, line 246: Are there any reasonable situations where we could get this error but not a set `XenObject` for which we would want to raise the error anyway but with `Unknown` as a stand-in for the `XenObject`?

XenModel/FriendlyNames.resx
```diff
@@ -1001,6 +1001,9 @@
 
 {1}</value>
   </data>
+  <data name="Message.body-gfs2_capacity" xml:space="preserve">
+    <value>Space utilization on SR '{0}' has exceeded 80%.</value>
+  </data>
```
AaronRobson, line 1006: For my better understanding what is the distinction between the `body` and `name` messages?

kc284: On the server side name is the key of the alert and body to message that appears to CLI users, e.g.:
```
uuid ( RO)         : xxxxx
         name ( RO): GFS2_CAPACITY
     priority ( RO): 1
        class ( RO): SR
     obj-uuid ( RO): xxxxx
    timestamp ( RO): 20200605T10:30:24Z
         body ( RO): SR utilisation above 80.0% for SR xxxx
```
In XenCenter we change this a bit and by name we refer to the short title of the alert when it first appears on the Alerts tab, while by body to the longer descritpion that appears once you expand the alert.

XenAdmin/Alerts/Types/MessageAlert.cs
kc284: Throughout XenCenter objects can happen to be null, often when we concincide with a cache update, but in this case we just avoid doing things with them rather than saying something occurred on an unknown object.

PR #2690, opened 2020-06-08
CA-187276: re-enable "Open SSH Console" button after VM restart

AaronRobson:

### DIFF
--- a/XenAdmin/ConsoleView/VNCTabView.cs
+++ b/XenAdmin/ConsoleView/VNCTabView.cs
@@ -529,7 +529,7 @@ namespace XenAdmin.ConsoleView
 
                 UpdateOpenSSHConsoleButtonState(); //guest_metrics change when there is an IP address change on a VIF
             }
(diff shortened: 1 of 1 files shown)


kc284:
This did not work for me when I tested it. I looked a bit further into it and I think I have fixed it with this change (which also contains a calculation efficiency improvement): https://github.com/kc284/xenadmin/commit/9fc3ea41c3013ea069314340f22ff8b935f8e5c9

I added a Trace() in Server_PropertyChanged() and my testing showed that domid was not the last property changed when rebooting. Allowed_operations changed last, which did call UpdateOpenSSHConsoleButtonState(), however the button was still disabled. Drilling further down I found out that no IP could be obtained for the SSH console because the VM's guest_metrics had not been fully populated yet and had 0 networks. Hooking up the "networks" property to update the button does re-enable it.

AaronRobson:
Thank you @kc284 for looking into it. I will confirm with latest master (including your commit).

PR #2696, opened 2020-06-12
Bring branch dotnet48 up to date with master

kc284:


XenAdmin/Commands/InstallCertificateCommand.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            MainWindowCommandInterface.ShowForm(typeof(InstallCertificateDialog), new object[] {host});
+        }
+
+        protected override bool CanExecuteCore(SelectedItemCollection selection)
+        {
+            if (!selection.AllItemsAre<Host>() || selection.Count > 1)
+                return false;
+
+            var host = selection.AsXenObjects<Host>().First();
```
AaronRobson, line 85: Shouldn't this be `FirstOrDefault` to ensure that it doesn't crash with an empty list. Or is the empty list case already prevented elsewhere?

XenAdmin/Commands/InstallToolsCommand.cs
```diff
@@ -132,38 +106,41 @@ private AsyncAction SingleVMExecute(VM vm)
...
                 }
-            }
-            return null;
         }
 
         /// <summary>
         /// Attempts to install tools on several VMs
         /// </summary>
         /// <param name="vms"></param>
         /// <returns>Whether the action was launched (i.e., the user didn't Cancel)</returns>
```
AaronRobson, line 142: This method now no longer returns anything (`void`), please update the annotation to reflect that.

XenAdmin/Commands/RemoveHostFromPoolCommand.cs
```diff
@@ -121,38 +120,40 @@ protected override void ExecuteCore(SelectedItemCollection selection)
...
+        }
+
         public static bool CanExecute(Host host)
         {
             if (host != null && host.Connection != null)
             {
                 Pool pool = Helpers.GetPool(host.Connection);
-                return pool != null && host.opaque_ref != pool.master && host.resident_VMs != null && host.resident_VMs.Count < 2 && host.IsLive();
+                return pool != null && host.opaque_ref != pool.master && host.IsLive() &&
+                       host.resident_VMs != null && host.resident_VMs.Count <= 1;
```
AaronRobson, line 146:
```c#
host.resident_VMs?.Count <= 1
```
Could be a clearer way of implementing:
```c#
host.resident_VMs != null && host.resident_VMs.Count <= 1
```

XenAdmin/Commands/RemoveHostFromPoolCommand.cs
```diff
@@ -121,38 +120,40 @@ protected override void ExecuteCore(SelectedItemCollection selection)
...
-                    if (commonPool != null && !pool.Equals(commonPool))
-                    {
-                        return false;
-                    }
-                    commonPool = pool;
-                }
-                return true;
-            }
-            return false;
+            // all selected items must be hosts and in the same pool
```
AaronRobson, line 153: I presume `CanExecute` itself now takes care of the "in the same pool" requirement. Its a little difficult to follow.

XenAdmin/Commands/InstallCertificateCommand.cs
kc284: Theoretically, yes. Practically, no; the logic is indeed kind of difficult to follow, what's happening is that ExecuteCore is not called for the commands unless CanExecuteCore is successful and the latter checks for empty collection.

XenAdmin/Commands/RemoveHostFromPoolCommand.cs
kc284: The same pool requirement is the `selection.Select(s => s.Connection).Count() == 1` bit in the CanExecuteCore. If this fails either as an empty or as e multi-pool selection, it will not proceed to evaluate the second bit which calls CanExecute on each host.

XenAdmin/Commands/InstallCertificateCommand.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        public InstallCertificateCommand(IMainWindow mainWindow, IEnumerable<SelectedItem> selection)
+            : base(mainWindow, selection)
+        {
+        }
+
+        protected override void ExecuteCore(SelectedItemCollection selection)
+        {
+            var host = selection.AsXenObjects<Host>().First();
```
AaronRobson, line 60: If there are no suitable elements found this could crash out.
Assuming this can happen in real world usage, I'd recommend this pattern:
```c#
var host = selection.AsXenObjects<Host>().FirstOrDefault();
if (host == null)
    return;
```

AaronRobson: I see, @kc284 this case would be prevented by `CanExecuteCore`.

XenAdmin/Commands/InstallToolsCommand.cs
```diff
@@ -84,38 +84,12 @@ public InstallToolsCommand(IMainWindow mainWindow, VM vm)
...
-            }
-
-            return null;
-        }
-
         /// <summary>
         /// Attempts to install tools on the vm
         /// </summary>
         /// <param name="vm"></param>
         /// <returns>null if user cancels or an AsyncAction. This is either the InstallPVToolsAction or the CreateCdDriveAction if the VM needed a DVD drive.</returns>
```
AaronRobson, line 91: This no longer returns anything.

XenAdmin/Dialogs/CopyVMDialog.cs
```diff
@@ -119,68 +130,21 @@ private void MoveButton_Click(object sender, EventArgs e)
         {
             if (!tableLayoutPanelSrPicker.Enabled || CloneRadioButton.Checked)
             {
-                new VMCloneAction(TheVM,NameTextBox.Text, DescriptionTextBox.Text).RunAsync();
+                new VMCloneAction(_vm,NameTextBox.Text, DescriptionTextBox.Text).RunAsync();
```
AaronRobson, line 133: Nit: There should be a space after the comma.

PR #2704, opened 2020-06-25
CA-334863: decimal network and disk access rate unit calculations

AaronRobson:
Though this was shown in units like kBps and MBps (decimal units) it was calculated in
binary terms i.e. 1MBps was treated as 1024kBps instead of the expected 1000kBps.


kc284:
I think the ticket number should be CP-34000 rather than CA-334863, but I'm not insisting.

kc284:
Could you also change the values of `VAL_{TER|GIG|MEG|}RATE` to be `TB/s` etc. instead of `TBps` etc. as per CP-34000. The latter ticket contains more work, but we can make a small step to unit normalisation since we are in the area. Otherwise it looks ok to me.

AaronRobson:
Please see: https://github.com/xenserver/xenadmin/pull/2706

PR #2706, opened 2020-06-29
CP-34000: decimal network and disk access rate unit calculations

AaronRobson:
Though this was shown in units like kB/s and MB/s (decimal units) it was calculated in
binary terms i.e. 1MB/s was treated as 1024kB/s instead of the expected 1000kB/s.

Updated translations for the units to be of the form 'MB/s' instead of 'MBps'.


XenModel/Messages.ja.resx
```diff
@@ -13161,7 +13161,7 @@ SR-IOV ネットワークを有効にするには、サーバーの再起動が
     <value>kW</value>
   </data>
   <data name="VAL_KILRATE" xml:space="preserve">
-    <value>Kbps</value>
+    <value>KB/s</value>
```
kc284, line 13164: Lowercase 'k'.

XenModel/Messages.zh-CN.resx
```diff
@@ -13158,7 +13158,7 @@ VM 克隆使用文件管理器的快照和克隆功能来实现高性能，并
     <value>kW</value>
   </data>
   <data name="VAL_KILRATE" xml:space="preserve">
-    <value>KBps</value>
+    <value>KB/s</value>
```
kc284, line 13161: Lowercase 'k'.

XenModel/Messages.ja.resx
AaronRobson: Good spot, corrected.

XenModel/Messages.zh-CN.resx
AaronRobson: Also corrected.

PR #2729, opened 2020-09-08
CP-34821: OVF/OVA/disk image import/export using the API's http actions

kc284:
Important: please disable whitespace diff when reviewing this PR.
It's probably easier to review the commits separately.


XenModel/Actions/OvfActions/Import.cs
```diff
@@ -121,158 +70,96 @@ public static void Process(IXenConnection connection, EnvelopeType ovfObj, strin
 
         	XenRef<VM_appliance> applRef = null;
 			if (applianceName != null)
```
AaronRobson, line 72: Nit: please replace the tab indentation here with spaces.

kc284: I'll include it in the next PR.

MihaelaStoica (changes requested):

XenModel/Actions/VM/ImportVmAction.cs
```diff
@@ -422,6 +316,9 @@ public void EndWizard(bool start, List<VIF> vifs)
 			m_startAutomatically = start;
 			m_VIFs = vifs;
 
+            if (m_startAutomatically)
+                ApiMethodsToRoleCheck.Add("vm.start");
```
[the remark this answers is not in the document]
kc284: You're right. I also think it would be better if we launched the VMStartAction here instead of calling directly VM.start. I have done this for CA-64500, so if that's ok, I'll leave this fix out of this PR and include it to the upcoming one.

XenOvfTransport/Export.cs
```diff
@@ -48,112 +48,58 @@ public class Export
...
-                    }
-                    else
-                    {
-                        vmtype = string.Format(typeformat, "xen", "3.0", "unknown");
-                    }
-                }
-                OVF.AddVirtualSystemSettingData(ovfEnv, vsId, vhsId, vm.name_label, OVF.GetContentMessage("VSSD_CAPTION"), vm.name_description, Guid.NewGuid().ToString(), vmtype);
+                #region ADD VirtualSystemType identification
+
+                var pv = vm.IsHVM() ? "hvm" : "xen";
```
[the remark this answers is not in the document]
kc284: This is an interesting one. I think the old logic for HVM was wrong. First of all, for `string.IsNullOrEmpty(vm.domarch)` it gave `hvm-3.0-unknown` and for `!string.IsNullOrEmpty(vm.domarch)` it gave `{vm.domarch}-3.0-unknown`, which looks a bit weird, especially when compared with the PV case where we have `xen-3.0-unknown` and `xen-3.0-{vm.domarch}` respectively. Then, I'm not sure why we need to check additionally that `HVM_boot_policy` is `BIOS order` (these settings are not configurable) in order to decide whether it's HVM; it made snse to me to use the same check as elsewhere in XenCenter.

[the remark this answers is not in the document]
MihaelaStoica: I think your change is correct.
According to https://github.com/xenserver/xenadmin/blob/a2596ec201c6760f703833fdff1f1ea8a55548f3/XenOvfApi/app.config#L230 `{vm.domarch}-3.0-unknown` is not a valid VirtualSystemType, so a file containing this would produce a warning on validation in https://github.com/xenserver/xenadmin/blob/bfc8141391bb892608d6a6d16150ad1401e68c2e/XenOvfApi/Validation.cs#L138. Have we seen such warning?

MihaelaStoica (approved):

XenOvfTransport/Export.cs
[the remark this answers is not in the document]
kc284: No, because none of xenserver's templates or VMs seem to have a value for this field (with the exception of dom0, which is not exportable). Not sure if this is a toolstack bug or not. (By the way, the knownVirtualSystemTypes needs updating).

XenModel/Actions/OvfActions/ImportImageAction.cs
```diff
@@ -43,33 +43,26 @@
...
-		protected override void Run()
-		{
-		    base.Run();
+            Title = string.Format(Messages.IMPORT_DISK_IMAGE, ovfEnv.Name, Helpers.GetName(connection));
+        }
 
-			Debug.Assert(m_vmMappings.Count == 1, "There is one VM mapping");
+        protected override void RunCore()
+        {
+            Debug.Assert(m_vmMappings.Count == 1, "There is one VM mapping");
```
AaronRobson, line 65: Nit: The assertion failure message ought to read `"There must be only one VM mapping"`.
This was the case before the changes here however.

XenAdminTests/StringExtensionsTests/TruncateStringTest.cs
```diff
@@ -61,4 +64,50 @@ public string Utf16SurrogatePairsTest(int size)
...
+        [Test]
+        public void TestSplitToDictionary()
+        {
+            string input = "EFI-variables-backend=xapidb;EFI-variables-on-boot=reset;EFI-variables=dGVzdA==";
+            
+            var output = new Dictionary<string, string>
+            {
+                {"EFI-variables-backend", "xapidb"},
+                {"EFI-variables-on-boot", "reset"},
+                {"EFI-variables", "dGVzdA=="}
```
AaronRobson, line 107: 👍 for automatically confirming that `SplitToDictionary` handles values with `=` in correctly.

AaronRobson (approved):

XenModel/Actions/OvfActions/Export.cs
```diff
@@ -0,0 +1,344 @@
...
+                            },
+                            () => Cancelling, XenAdminConfigManager.Provider.GetProxyTimeout(true),
+                            Connection.Hostname, XenAdminConfigManager.Provider.GetProxyFromSettings(Connection),
+                            diskPath, taskRef, Connection.Session.opaque_ref, vdi.uuid, "vhd");
+
+                        if (m_shouldVerify)
+                        {
+                            Description = string.Format(Messages.EXPORTING_VDI_VERIFICATION, diskFilename);
+
+                            //TODO
```
AaronRobson, line 190: I presume the verification is planned to be included in a later change?

kc284: Yes, there is a separate ticket fo this.

XenModel/Actions/OvfActions/ImportImageAction.cs
kc284: Will correct it in the upcoming PR.

PR #2730, opened 2020-09-10
CA-343763: Corrected the handling of VMSS alert priority/severity according to CA-342874.

kc284:
- Distinguish between warnings and errors.
- Made PolicyAlert derive from MessageAlert to conform with other alert types.
- Replaced EventType column by a simple image column because it adds to clatter without really offering useful information.


AaronRobson (approved):

XenAdmin/Dialogs/ScheduledSnapshots/ScheduledSnapshotsDialog.cs
```diff
@@ -535,16 +520,16 @@ private void RefreshHistoryLabel()
 
         private class HistoryRow : DataGridViewRow
         {
-            private DataGridViewImageCell _expand = new DataGridViewImageCell();
-            private DataGridViewTextAndImageCell _result = new DataGridViewTextAndImageCell();
-            private DataGridViewTextBoxCell _dateTime = new DataGridViewTextBoxCell();
-            private DataGridViewTextBoxCell _description = new DataGridViewTextBoxCell();
+            private readonly DataGridViewImageCell _expand = new DataGridViewImageCell();
+            private readonly DataGridViewImageCell _image = new DataGridViewImageCell();
```
MihaelaStoica, line 524: With the switch to an image cell, we lost the ability to sort on that column (I don't know if it's possible to sort on Image cells)

MihaelaStoica (changes requested):

XenAdmin/Dialogs/ScheduledSnapshots/ScheduledSnapshotsDialog.cs
```diff
@@ -446,6 +446,23 @@ private void dataGridViewRunHistory_CellClick(object sender, DataGridViewCellEve
             }
         }
 
+        private void dataGridViewRunHistory_ColumnHeaderMouseClick(object sender, DataGridViewCellMouseEventArgs e)
+        {
+            if (e.ColumnIndex != ColumnImage.Index)
+                return;
```
MihaelaStoica, line 452: We also need to reset the `ColumnImage.HeaderCell.SortGlyphDirection` (to `SortOrder.None`) here when the other column headers are clicked, so that the sort arrow is being cleared from the Image column when we are sorting on other columns

MihaelaStoica (changes requested):

MihaelaStoica (approved):

PR #2733, opened 2020-09-22
CA-339233: Do not allow selection of networks while they're locked.

kc284:
Commits best viewed separately, but I recommend squashing them before merging.


XenAdmin/Controls/ComboBoxes/NetworkComboBox.cs
```diff
@@ -104,16 +103,110 @@ public void SelectItem(Func<NetworkComboBoxItem, bool> matchSelectionCriteria)
...
             public XenAPI.Network Network { get; }
+            
+            public bool Enabled
+            {
+                get
+                {
+                    if (Pif != null && Pif.Locked)
+                        return false;
+                    if (Network != null && Network.Locked)
+                        return false;
```
AaronRobson, line 202: Nit:
```
if (Pif?.Locked)
  return false;
if (Network?.Locked)
  return false;
return true
```
or even:
```
return !(Pif?.Locked || Network?.Locked);
```

AaronRobson (approved):

MihaelaStoica (approved):

PR #2737, opened 2020-10-22
XSI-772/CA-347407 and another couple of fixes

kc284:

### DIFF
--- a/XenAdmin/PasswordsRequest.cs
+++ b/XenAdmin/PasswordsRequest.cs
@@ -113,10 +113,6 @@ namespace XenAdmin
                     log.WarnFormat("Valid token, but for the wrong app. Will re-prompt");
                 }
             }
-            else
-            {
-                log.Warn("Missing or invalid token.");
-            }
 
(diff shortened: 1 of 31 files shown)


MihaelaStoica (approved):

XenAdmin/PasswordsRequest.cs
```diff
@@ -113,10 +113,6 @@ internal static void HandleRequest(string destdir)
                     log.WarnFormat("Valid token, but for the wrong app. Will re-prompt");
```
AaronRobson, line 113: Nit: This instance of `WarnFormat` should be `Warn` as it doesn't have any other parameters. It was already like that though, nothing you've changed.

AaronRobson (approved):

XenAdmin/PasswordsRequest.cs
kc284: A quick grep shows a few more instances in other files; also a few instances of the other related methods like DebugFormat etc. I'll raise a PR at some point to fix all of them.

PR #2740, opened 2020-10-26
Some refactoring around Commands to improve calculation efficiency

kc284:
Please review commits separately and with whitespace disabled.

The actual change making the difference in the calculations is the commit with the tooltip refactoring. As an example, I counted the times the method StartVMCommand.CanExecute(VM vm) is hit before and after this change. Previously, on VM selection and on cache update the method was hit 13 and 7 times respectively, after the change these numbers have gone down to 4 and 2. (Ideally we'd like them both to be 1, however this won't be possible unless we combine the CanExecute() and GetCantExecuteReasons() to one method CanExecute(out reasons), which is a big and scary change best left for later).

### DIFF
--- a/XenAdmin/Commands/AddHostCommand.cs
+++ b/XenAdmin/Commands/AddHostCommand.cs
@@ -70,7 +70,7 @@ namespace XenAdmin.Commands
         public AddHostCommand(IMainWindow mainWindow, Control parent)
             : base(mainWindow)
         {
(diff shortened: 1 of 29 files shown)


kc284:
I need to close and reopen the PR to trigger a new travis build.

AaronRobson (approved):

MihaelaStoica (approved):

PR #2741, opened 2020-10-29
 CP-15790: Implementation improvement aimed to make manual testing easier.

kc284:
- Ensure we read the updates.xml location anew from the registry, otherwise the application has to be relaunched to pick a registry change. Look in both hives HKCU and HKLM. To avoid performance impact on the UI, move the operation into the action's Run() body.
- Some refactoring to improve OOP design and fix XmlReader leak.


jijiang:
Looks good to me.

jijiang (approved):

AaronRobson (approved):

PR #2757, opened 2020-12-30
CA-350406: Exclude the Tools SR when plugging PBDs after a host reboot…

kc284:
… during an update process. Also, try plugging PBDs only for the host in question (other hosts are not necessary, as the PBDs are re-checked for any previously evacuated VMs before repatriation).

### DIFF
--- a/XenModel/XenAPI-Extensions/Host.cs
+++ b/XenModel/XenAPI-Extensions/Host.cs
@@ -1301,40 +1301,54 @@ namespace XenAPI
             bool allPBDsReady = false;
             int timeout = 120;
             log.DebugFormat("Waiting for PBDs on host {0} to become plugged", Name());
-            
-            do
+
+            while (timeout > 0)
             {
-                if (this.enabled)  // if the Host is not yet enabled, pbd.currently_attached may not be accurate: see CA-66496.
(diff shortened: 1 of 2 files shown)


AaronRobson (approved):

XenModel/XenAPI-Extensions/Host.cs
```diff
@@ -1301,40 +1301,54 @@ public void CheckAndPlugPBDs()
...
+
+                        if (pbd == null || pbd.currently_attached)
+                            continue;
+
+                        if (Helpers.StockholmOrGreater(this)) //CA-350406
                         {
-                            allPBDsReady = false;
-                            break;
+                            var sr = Connection.Resolve(pbd.SR);
+                            if (sr != null && sr.is_tools_sr)
```
AaronRobson, line 1320: For my understanding were tools SRs introduced in Stockholm?

kc284: The opposite: they were removed in stockholm, but on the server side the tools SR is deleted only after all the hosts have been upgraded, so we shouldn't try to plug it in.

MihaelaStoica (approved):

PR #2759, opened 2021-01-04
CA-345342 side effects.

kc284:
The next SDK drop will introduce http_actions with optional parameters; in the current implementation we call the http_actions via delegates, which do not cope very well with optional parameters. Removing those comes with a small cost of code repetition, mostly in relation to exception handling, but it's not too bad because some of the actions had custom exception handling already.

The other commits are some refactoring in the same area, the most notable being the addition of the optional parameter suppressFailures to task polling, which is meant to fix some cases when resources (e.g. uploaded hotfix VDIs or imported disks) are not always cleared after the action cancellation.

Commits best viewed separately and with whitespace diff off.


AaronRobson (approved):

XenModel/Actions/Pool_Patch/UploadPatchAction.cs
```diff
@@ -117,35 +117,42 @@ private Pool_patch UploadRetailPatch()
...
-                    File.Delete(retailPatchPath);
+                    if (ex.InnerException == null)
+                        throw;
+                    else
+                        throw ex.InnerException;
                 }
+
+                if (e is CancelledException || e is HTTP.CancelledException || e.InnerException is CancelledException)
+                {
+                    if (deleteFileOnCancel && File.Exists(retailPatchPath))
```
AaronRobson, line 149: By the looks of https://docs.microsoft.com/en-us/dotnet/api/system.io.file.delete?view=net-5.0#remarks the `File.Exists` check may be skipped.
"If the file to be deleted does not exist, no exception is thrown."

XenModel/Actions/AsyncAction.cs
```diff
@@ -268,7 +268,107 @@ public void DestroyTask()
...
+                        log.WarnFormat("Task {0} failed: {1}", RelatedTask.opaque_ref,
+                            task.error_info.Length > 0 ? task.error_info[0] : "Unknown failure");
+                        throw new Failure(task.error_info);
+                    }
+
+                case task_status_type.success:
+                    log.InfoFormat("Task {0} finished successfully", RelatedTask.opaque_ref);
+                    if (task.result != "") // Work around CA-6597
+                    {
+                        Match m = Regex.Match(Result, "<value>(.*)</value>");
```
AaronRobson, line 369: It's not ideal to parse XML with RegEx.
However I note that this is from the existing code in `TaskPoller.cs`, so it is probably safer to leave it as is.

AaronRobson (approved):

MihaelaStoica (approved):

XenModel/Actions/AsyncAction.cs
kc284: I agree; it can be revisited at a later time.

PR #2762, opened 2021-01-11
Fixed Next button state when re-landing on the page.

kc284:
Usability enhanements. Property modernisation.

### DIFF
--- a/XenAdmin/Wizards/ExportWizard/ExportOptionsPage.cs
+++ b/XenAdmin/Wizards/ExportWizard/ExportOptionsPage.cs
@@ -54,6 +54,7 @@ namespace XenAdmin.Wizards.ExportWizard
 		private bool m_isEncryptionOk = true;
 		private bool m_isSignatureOk = true;
         private bool m_buttonNextEnabled;
+        private bool _updating;
 
 		public ExportOptionsPage()
 		{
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/ExportWizard/ExportOptionsPage.cs
```diff
@@ -114,21 +114,21 @@ public ExportOptionsPage()
...
 		/// <summary>
 		/// Gets the value by which the help files section for this page is identified
 		/// </summary>
-		public override string HelpID { get { return "ExportOptions"; } }
+		public override string HelpID => "ExportOptions";
 
         protected override void PageLoadedCore(PageLoadedDirection direction)
 		{
-			SetButtonNextEnabled(true);
+			SetButtonNextEnabled(!IsDirty || m_isEncryptionOk && m_isSignatureOk);
```
AaronRobson, line 131: I'm a little confused by this I would have expected:
`SetButtonNextEnabled(IsDirty && m_isEncryptionOk && m_isSignatureOk);`

kc284: The very first time the user lands on the page they have not changed anything yet, hence IsDirty is false, however, it is valid to proceed without manifest or signature. Only if they have touched something need we check the other values.

AaronRobson (approved):

MihaelaStoica (approved):

PR #2772, opened 2021-01-26
Allow certain versions connect to hyperman

jijiang:


kc284:
You'll need to rebase your branch on the upstream's `feature/hyperman` because it contains commits we're alread merged.

XenModel/Messages.resx
```diff
@@ -6235,6 +6235,11 @@ Note: Any custom color selections will remain unchanged</value>
   <data name="GUEFI_SECURE_BOOT_MODE_LABEL" xml:space="preserve">
     <value>UEFI &amp;Secure Boot</value>
   </data>
+  <data name="GUI_NOT_COMPATIBLE" xml:space="preserve">
+    <value>This version of [XenCenter] can only connect to certain [XenServer] versions and cannot connect to {0}.
```
kc284, line 6239: I'm wondering whether we should list explicitly the versions we allow instead of saying vaguely "certain versions". What do the others think?

XenAdmin/MainWindow.cs
```diff
@@ -979,6 +979,34 @@ void connection_CachePopulated(IXenConnection connection)
                 LoadTasksAsMeddlingActions(connection);
             }
 
+            // Allow Citrix Hypervisor Center connect to Havana, Stockholm and cloud released CH only
```
kc284, line 982: I think this whole block should move to line 975. Regarding the if conditions, the first check is not needed because it is always true. The last check allows havana, but also ely and honolulu which we probably want to exclude (@MihaelaStoica could you please confirm), so let's try adding to it version checks from the Helpers class.

kc284 (changes requested):
As above.

XenAdmin/MainWindow.cs
jijiang: Fixed.

XenModel/Messages.resx
```diff
@@ -6235,6 +6235,11 @@ Note: Any custom color selections will remain unchanged</value>
   <data name="GUEFI_SECURE_BOOT_MODE_LABEL" xml:space="preserve">
     <value>UEFI &amp;Secure Boot</value>
   </data>
+  <data name="GUI_NOT_COMPATIBLE" xml:space="preserve">
+    <value>This version of [XenCenter] can only connect to [XenServer] 7.1.2 and 8.2+ versions and cannot connect to {0}.
```
MihaelaStoica, line 6239: The correct  product names and versions would be: XenServer 7.1 CU2 and Citrix Hypervison 8.2 and greater

XenAdmin/MainWindow.cs
```diff
@@ -972,6 +972,34 @@ void connection_CachePopulated(IXenConnection connection)
                     });
                     return;
                 }
+
+                // Allow Citrix Hypervisor Center connect to Havana, Stockholm and cloud released CH only
+                //
+                if ((int)API_Version.LATEST >= (int)API_Version.API_2_16
```
MihaelaStoica, line 978: `(int)API_Version.LATEST >= (int)API_Version.API_2_16` is not needed, as this will always be true

For checking the Havana version, we could use the existing `HavanaOrGreater` function which returns true for 7.1 CU2 (Havana) and any other future CUs we might have on 7.1

MihaelaStoica (changes requested):

XenAdmin/MainWindow.cs
jijiang: Regarding first comment, Tina is intending to share same source code for XC 8.2.x and hyperman, and doing some manual tests to assess the feasibility.
Second comment fixed.

XenModel/Messages.resx
jijiang: Fixed.

XenAdmin/MainWindow.cs
kc284: This check now allows anything from havana to stockholm. I think we need this:
```
current_version >= (int)API_Version.API_2_16 &&
server_min < (int)API_Version.API_2_15 &&
(server_min != (int)API_Version.API_2_6 || !Helpers.HavanaOrGreater(master))
```

jijiang: I think HavanaOrGreater() only hit 7.1 CU2 at this moment instead of all versions from Havana to Stockholm, because "!FalconOrGreater()" is included in HavanaOrGreater() function.

XenModel/Messages.resx
```diff
@@ -6235,6 +6235,11 @@ Note: Any custom color selections will remain unchanged</value>
   <data name="GUEFI_SECURE_BOOT_MODE_LABEL" xml:space="preserve">
     <value>UEFI &amp;Secure Boot</value>
   </data>
+  <data name="GUI_NOT_COMPATIBLE" xml:space="preserve">
+    <value>This version of [XenCenter] can only connect to {0} 7.1 CU2 and [XenServer] 8.2 and greater versions and cannot connect to {1}.
```
kc284, line 6239: The first sentence has too many ands. Can we perhaps say `This version of [XenCenter] can only connect to {0} 7.1 CU2, [XenServer] 8.2, and greater versions. It cannot connect to {1}.`

XenAdmin/MainWindow.cs
kc284: Ah true, I hadn't looked thoroughly into HavanaOrGreater() (perhaps it should probably be called something else because what it actually does is different from what it says). Ok, ignore.

MihaelaStoica: Yes, the name is confusing and the function is different from our other XOrGreater functions, but in this case it's exactly what we need

XenModel/Messages.resx
jijiang: Fixed.

kc284 (approved):

AaronRobson (approved):

MihaelaStoica (approved):

PR #2778, opened 2021-02-18
CP-35838: Add health check precheck into RPU wizard

jijiang:

### DIFF
--- a/XenAdmin/Diagnostics/Problems/PoolProblem/HealthCheckServiceProblem.cs
+++ b/XenAdmin/Diagnostics/Problems/PoolProblem/HealthCheckServiceProblem.cs
@@ -1,0 +1,83 @@
+﻿/* Copyright (c) Citrix Systems, Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer. 
+ * *   Redistributions in binary form must reproduce the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer in the documentation and/or other 
+ *     materials provided with the distribution. 
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+ * CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR 
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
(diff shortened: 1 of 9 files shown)


MihaelaStoica (changes requested):

XenModel/Messages.resx
```diff
@@ -2651,6 +2651,9 @@ Do you want to assign it to the schedule '{2}' instead?</value>
   <data name="CHECKING_HA_STATUS" xml:space="preserve">
     <value>Checking HA and WLB status</value>
   </data>
+  <data name="CHECKING_HEALTH_CHECK_SERVICE" xml:space="preserve">
+    <value>Checking pool health check service status</value>
```
kc284, line 2655: I think `Checking pool enrollment to the Health Check service` is more accurate.

XenModel/Messages.resx
```diff
@@ -10657,6 +10660,12 @@ Please reconnect the host and try again</value>
   <data name="PROBING_HBA_TITLE" xml:space="preserve">
     <value>Probing for LUNs on {0}</value>
   </data>
+  <data name="PROBLEM_HEALTH_CHECK_SERVICE_DESCRIPTION" xml:space="preserve">
```
kc284, line 10663: On the UI "Health Check" is used with capital initial letters as a term, so let's follow this convention here too (and in the messages below). Also, it needs the article.

XenAdmin/Diagnostics/Problems/PoolProblem/HealthCheckServiceProblem.cs
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+using XenAdmin.Core;
+using XenAdmin.Diagnostics.Checks;
+using XenAPI;
+
+
+namespace XenAdmin.Diagnostics.Problems.PoolProblem
+{
+    class HealthCheckServiceProblem : ProblemWithMoreInfo
```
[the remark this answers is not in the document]
jijiang: Fixed.

XenModel/Messages.resx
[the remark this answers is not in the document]
jijiang: Fixed.

jijiang: Fixed.

XenModel/Messages.resx
jijiang: Fixed.

kc284:
I downloaded the commit and had a look around. It works nice, however, I realised we have a usability issue. When the users RPU to an unknown version and see the warning, they have no way to disable health check because we don't show the dilalog and there is no action like the problem case. So I think we need to offer the option on the MoreInfo pop-up in case they need to disable it. Let's try adding a `public virtual AsyncAction LinkAction => null;` in WarningWithMoreInfo.cs, then in HealthCheckServiceWarning override it and also set LinkText, and in WarningWithMoreInfo.cs where we launch that warning dialog offer it as an alternative to the LinkData, i.e.
```
if (!string.IsNullOrEmpty(LinkText) && (!string.IsNullOrEmpty(LinkData) || LinkAction != null))
{
    dlg.LinkText = LinkText;
    if (!string.IsNullOrEmpty(LinkData))
        dlg.LinkData = LinkData;
    else
        dlg.LinkAction = LinkAction;
    dlg.ShowLinkLabel = true;
}
```

kc284:
A simpler alternative may be to override CreateAction() in the HealthCheckServiceWarning.cs and show a WarningDialog with the more info message and two buttons, DisableHealthCheck and Cancel.

XenModel/Messages.resx
```diff
@@ -10657,6 +10660,12 @@ Please reconnect the host and try again</value>
   <data name="PROBING_HBA_TITLE" xml:space="preserve">
     <value>Probing for LUNs on {0}</value>
   </data>
+  <data name="PROBLEM_HEALTH_CHECK_SERVICE_DESCRIPTION" xml:space="preserve">
+    <value>{0}: The Health Check service has been removed in [XenServer] {1}.</value>
```
kc284, line 10664: Could you please also change `[XenServer] {1}` to `{0} {1}` here and below, and format it with the BrandManager variables? We should start migrating from rebranding with sed.

AaronRobson (approved):

XenModel/Messages.resx
jijiang: Fixed in second commit.

kc284 (approved):

MihaelaStoica (approved):

PR #2781, opened 2021-02-25
CP-35898, CP-35899, CP-36202

kc284:
Commits best reviewed separately. Diff with whitespace off is recommended.

### DIFF
--- a/XenModel/XenAPI-Extensions/SR.cs
+++ b/XenModel/XenAPI-Extensions/SR.cs
@@ -120,7 +120,19 @@ namespace XenAPI
 
         public string FriendlyTypeName()
         {
-            return GetFriendlyTypeName(GetSRType(false));
+            var srType = GetSRType(false);
+
+            if (srType == SRTypes.unknown)
+            {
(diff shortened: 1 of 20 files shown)


XenModel/XenAPI-Extensions/SR.cs
```diff
@@ -120,7 +120,19 @@ public bool IsToolsSR()
 
         public string FriendlyTypeName()
         {
-            return GetFriendlyTypeName(GetSRType(false));
+            var srType = GetSRType(false);
```
kc284, line 123: Pro of this solution: if the vendor changes the details of the SM plugin (e.g. type change between prototyping and productization) XC will pick it without changes.
Con: There is no localization. It would have to be provided by the plugin (but we don't have such implementation yet).

kc284:
@edwintorok if you fancy looking at some C# code I'd be interested in your review too.

AaronRobson (approved):

MihaelaStoica (approved):

PR #2785, opened 2021-03-04
CP-35471: Update the legacy partition prechecks in the RPU

MihaelaStoica:


AaronRobson (approved):

XenAdmin/Diagnostics/Problems/HostProblem/HostNotSafeToUpgradeWarning.cs
```diff
@@ -1,87 +1,155 @@
...
+
+using XenAdmin.Core;
+using XenAdmin.Diagnostics.Checks;
+using XenAPI;
+
+namespace XenAdmin.Diagnostics.Problems.HostProblem
+{
+    public enum HostNotSafeToUpgradeReason { NotEnoughSpace, VdiPresent, UtilityPartitionPresent, LegacyPartitionTable, Default }
+
+    public class HostNotSafeToUpgradeWarning : WarningWithMoreInfo
```
kc284, line 40: One thing we could do to avoid repetition is remove the field _reason, replace the ShortMessage with a readonly field and the Message with an auto-property, and set both in the constructor, e.g.:
```
_host = host;
var newPartitionInfo = string.Format(...
string detail;

switch (reason)
{
    case HostNotSafeToUpgradeReason.NotEnoughSpace:
        _shortMessage = Messages.xxx
        detail = string.Format(...
        break;
    ...
}

Message = $"{_shortMessage}\n\n{newPartitionInfo}\n\n{detail}";
```

kc284:
We'll also need to update the RPU005 version in hotfix-map.json and the fileId in XenCenter.wxs.

XenAdmin/Diagnostics/Problems/HostProblem/HostNotSafeToUpgradeWarning.cs
MihaelaStoica: Good suggestion, it looks better now

MihaelaStoica:
I will add the new version of RPU005 once I have it signed and copied to the respective location

kc284 (approved):

PR #2786, opened 2021-03-09
CA-352455 + issue with action description while polling task

kc284:
- CA-352455: Parse the task result as xml document instead of using a regex.
- Do not use the task description as the AsyncAction description because in most cases it is not localised. Also, do not poll if the RelatedTask is null.

### DIFF
--- a/XenModel/Actions/AsyncAction.cs
+++ b/XenModel/Actions/AsyncAction.cs
@@ -141,7 +141,7 @@ namespace XenAdmin.Actions
         /// <summary>
         /// Prepare the action's task for exit by removing the XenCenterUUID.
         /// A call here just before exit will mean that the task will get picked 
-        /// up as a meddling action on restart of xencenter, and thus reappear in the log.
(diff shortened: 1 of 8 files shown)


XenModel/Actions/AsyncAction.cs
```diff
@@ -366,9 +367,15 @@ private bool Poll(double start, double finish)
...
-                        Match m = Regex.Match(task.result, "<value>(.*)</value>");
-                        if (m.Success)
-                            Result = m.Groups[1].Value;
+                        var doc = new XmlDocument();
+                        doc.LoadXml(task.result);
+                        var nodes = doc.GetElementsByTagName("value");
+
+                        foreach (XmlNode node in nodes)
+                        {
+                            Result = node.InnerText;
```
AaronRobson, line 378: Is the written `Result` even used for anything, it seems to be a local variable that doesn't get outputted anywhere?
Perhaps I've missed something with how async tasks work though.

AaronRobson (approved):

MihaelaStoica (approved):

PR #2787, opened 2021-03-10
CP-35345: Add pre-cloud connection warning

jijiang:

### DIFF
--- a/XenAdmin/TabPages/GeneralTabPage.cs
+++ b/XenAdmin/TabPages/GeneralTabPage.cs
@@ -459,6 +459,8 @@ namespace XenAdmin.TabPages
             }
             panel2.ResumeLayout();
             UpdateButtons();
+
+            SetupDeprecationBanner();
         }
 
         private void generateInterfaceBox()
@@ -2010,6 +2012,22 @@ namespace XenAdmin.TabPages
             finally
             {
                 panel2.ResumeLayout();
+            }
+        }
+
+        private void SetupDeprecationBanner()
+        {
+            if (Helpers.PostStockholm(xenObject.Connection) || (int)API_Version.LATEST < (int)API_Version.API_2_16)
+            {
+                Banner.Visible = false;
+            }
+            else
+            {
+                Banner.BannerType = DeprecationBanner.Type.Warning;
+                Banner.WarningMessage = string.Format(Messages.WARNING_PRE_CLOUD_VERSION_CONNECTION, BrandManager.BrandConsole, BrandManager.ProductBrand, BrandManager.ProductVersion82, BrandManager.LegacyConsole);
+                Banner.LinkText = Messages.PATCHING_WIZARD_WEBPAGE_CELL;
+                Banner.LinkUri = new Uri(InvisibleMessages.OUT_OF_DATE_WEBSITE);
+                Banner.Visible = true;
             }
         }
     }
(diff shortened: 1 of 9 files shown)


XenModel/Messages.resx
```diff
@@ -13839,6 +13839,11 @@ Schedule:
   <data name="WARNING_HEALTH_CHECK_SERVICE_INFO" xml:space="preserve">
     <value>The Health Check service has been removed in {0}. If you are going to upgrade the pool to this version, you must disable Health Check first.</value>
   </data>
+  <data name="WARNING_PRE_CLOUD_VERSION_CONNECTION" xml:space="preserve">
```
kc284, line 13842: I think we can make this message a bit shorter and be specific about the version since we disallow connection to others:
`<value>{0} is not recommended for general management of servers running {1} {2}. Please use {3} {2} or upgrade {1} using the Rolling Pool Upgrade wizard.</value>`
where the placeholders are BrandManager's `BrandConsole`, `ProductBrand`, `ProductVersion82`, `LegacyConsole` (this needs to be added)

XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -2012,5 +2014,20 @@ private void ToggleExpandedState(Func<PDSection, bool> expand)
...
+        {
+            Banner.Visible = false;
+            if (Helpers.PostStockholm(xenObject.Connection))
+                return;
+
+            if ((int)API_Version.LATEST >= (int)API_Version.API_2_16)
+            {
+                Banner.BannerType = DeprecationBanner.Type.Warning;
+                Banner.WarningMessage = Messages.WARNING_PRE_CLOUD_VERSION_CONNECTION;
+                Banner.LinkUri = new Uri(InvisibleMessages.OUT_OF_DATE_WEBSITE);
```
kc284, line 2028: I think we need to set the LinkText as well, because when I saw Details I thought it might launch a pop-up with more information. Maybe use Messages.PATCHING_WIZARD_WEBPAGE_CELL instead?

XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -2012,5 +2014,20 @@ private void ToggleExpandedState(Func<PDSection, bool> expand)
                 panel2.ResumeLayout();
             }
         }
+
+        private void SetupDeprecationBanner()
+        {
+            Banner.Visible = false;
```
kc284, line 2020: When the cache is updated, this method maybe called more than once, so always hiding it before checking whether to re-show it will cause flickering. How about this:
```
if (Helpers.PostStockholm(xenObject.Connection) || (int)API_Version.LATEST < (int)API_Version.API_2_16)
{
    Banner.Visible = false;
}
else
{
    ...
    Banner.Visible = true;
}
```

XenAdmin/TabPages/GeneralTabPage.cs
jijiang: Good suggestion. The deprecation banner in ballooning page could change to use "Go to web page" link text as well.

XenModel/Messages.resx
jijiang: Fixed.

XenAdmin/TabPages/GeneralTabPage.cs
jijiang: Fixed.

kc284 (approved):
It looks ok to me now. We may still want to discuss whether to show it on more places.

AaronRobson (approved):

MihaelaStoica (approved):

PR #2788, opened 2021-03-12
Update DEFAULT_NUM_VBDS_ALLOWED to 255

edwintorok:
All currently supported versions of the product support 255 VBDs,
see https://docs.citrix.com/en-us/xenserver/7-0/downloads/config-limits.pdf

When a VM is imported through XCM it may not have recommendations,
and XenCenter falls back to the default, which is 16.
Then adding another disk shows an error that the maximum number of disks was reached,
which contradicts our support statement of 255 disks.

Signed-off-by: Edwin Török <edvin.torok@citrix.com>


edwintorok:
No ticket number yet, when we have one I can update the commit message.

AaronRobson (approved):

kc284 (approved):

PR #2796, opened 2021-03-30
CA-352901, CA-352896, CA-352946 - release/stockholm/UPD-694

kc284:


jijiang (commented):
toolStripbuttonNew.Text should also been changed in ConversionDialog.ja.resx. See CA-352901 Description field item 3.

kc284:
ToolStripbuttonNew.Text in ConversionDialog.ja.resx already matches the documentation at https://docs.citrix.com/ja-jp/xencenter/current-release/conversion-manager.html#vmware. See
```
<data name="toolStripButtonNew.Text" xml:space="preserve"><value>新しい変換(&amp;N)...</value></data>
```
vs.
```
3. ［新しい変換］ をクリックします。
4. 新しい変換ウィザードで、VMwareサーバーの資格情報を入力します
```

XenAdmin/Controls/CustomDataGraph/DataSet.cs
```diff
@@ -132,14 +132,10 @@ public static DataSet Create(string uuid, IXenObject xo, bool show, string setty
...
+                    settype.StartsWith("write_latency_") || settype.StartsWith("vbd"))
                 {
-                    //if it's storage latency xapi units are in milliseconds
-                    dataSet.MultiplyingFactor = 1000000;
-                }
-                else if (settype.StartsWith("vbd"))
-                {
-                    //if it's vbd latency xapi units are in microseconds
+                    //if it's storage or vbd latency xapi units are in microseconds
                     dataSet.MultiplyingFactor = 1000;
```
AaronRobson, line 139: Does this change to the units at the bottom of the change here: https://github.com/xenserver/rrdd-plugins/commit/208550d95e2a28b27ea1e40a2c2dc1d22376a160#diff-94f636c01795015c97eb9d04ef1b1367138cb0f837f30b7193d9b6995e27f0f0R673 from `milliseconds` to `μs` make it through to us?

It so I'd recommend using those units that we are given to choose the right `dataSet.MultiplyingFactor` in case the units change in the future or (perhaps less likely) we need to support older toolstack versions.

kc284: Yes, this is planned for master (CP-36567).

AaronRobson: Fair enough, that was my only concern. I approve.

AaronRobson (approved):

MihaelaStoica (approved):

jijiang (approved):

PR #2813, opened 2021-04-28
CA-353524: Full copy of VM on the same SR is allowed.

kc284:
Moving and copying a VM should be treated differently, hence we need a different SrPickerItem subclass for each operation. Minor logic simplifications: I removed the methods ShowInVDISRList and CanCreateVmOn because in an occasion or two they were causing duplication of checks (this happens with other SR extension methods too, but it's out of scope to touch them at this point).

### DIFF
--- a/XenAdmin/Controls/SrPickerItem.cs
+++ b/XenAdmin/Controls/SrPickerItem.cs
@@ -139,21 +139,16 @@ namespace XenAdmin.Controls
     }
 
 
-    public class SrPickerMoveCopyItem : SrPickerItem
+    public class SrPickerCopyItem : SrPickerItem
     {
-        public SrPickerMoveCopyItem(SR sr, Host aff, VDI[] vdis)
+        public SrPickerCopyItem(SR sr, Host aff, VDI[] vdis)
             : base(sr, aff, vdis)
         {
         }
 
-        protected override bool CanBeEnabled
-        {
-            get
-            {
-                return !TheSR.IsDetached() && TheSR.SupportsVdiCreate() && !ExistingVDILocation() &&
-                       TheSR.VdiCreationCanProceed(DiskSize);
-            }
-        }
+        protected override bool CanBeEnabled =>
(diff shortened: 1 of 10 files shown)


XenAdmin/Controls/SrPickerItem.cs
```diff
@@ -269,17 +276,18 @@ private void Update()
...
-            if (ShowHiddenVDIs && !ExistingVDILocation() && CanBeEnabled)
+            if (CanBeEnabled)
             {
                 Description = string.Format(Messages.SRPICKER_DISK_FREE, Util.DiskSizeString(TheSR.FreeSpace(), 2),
                     Util.DiskSizeString(TheSR.physical_size, 2));
                 Enabled = true;
                 Show = true;
             }
-            else if (TheSR.PBDs.Count > 0 && TheSR.SupportsVdiCreate())
+            else
```
MihaelaStoica, line 290: This has the effect of showing a detached SR (no PBDs) in the list (correctly disabled), while previously detached SRs were not displayed. But I think this is OK

MihaelaStoica (approved):

XenAdmin/Controls/SrPickerItem.cs
kc284: Hmm, I've tried to simplify things, but the logic is still confusing. The properties in the SrPickerItem subclasses check for IsDetached() and also provide the disabled reason, so they give the imperssion that it's ok to show them. On the other hand, the same properties also check for SupportsVdiCeate(), which is not needed since we hide them here, but I was nervous about  removing it. Should I just move all the checks into the subclasses and merge the supported and CanBeEnabled logic into one tri-state property (returning enabe|disable|hide)?

MihaelaStoica: We could do the sugested improvement later, for now I am happy with these changes

AaronRobson (approved):

PR #2830, opened 2021-06-15
CA-353748 second issue: At the end of polling, the RelatedTask is des…

kc284:
…troyed; make a note of it for later use. Also: make a get_record where more than one object fields need to be queried.


AaronRobson (approved):

psafont (approved):
Although I'm not too familiar with C#, it looks like the session is returned from the function `ApplyFile` and is correctly used to check with `network.other_config[IMPORT_TASK]`

MihaelaStoica (approved):

PR #2833, opened 2021-06-28
DelegatedAsyncAction should only be used for simple API calls, hence …

kc284:
…created new class for retrieving WLB host evacuation recommendations. This allows removal of DelegatedAsyncAction.ResultObject, which might lead to complicated implementations. Also: some refactoring of the other WLB retrieve recommendations actions.


AaronRobson (approved):

CitrixChris (approved):
Code wise seems fine, I cant comment on overall functionality though

PR #2841, opened 2021-07-29
CA-349435 - Fixed tooltip display on Options and Properties dialogs

CitrixChris:


danilo-delbusso (approved):

XenAdmin/Dialogs/OptionsDialog.Designer.cs
```diff
@@ -59,8 +62,31 @@ private void InitializeComponent()
             // 
             // verticalTabs
             // 
-            resources.ApplyResources(this.verticalTabs, "verticalTabs");
             this.verticalTabs.Items.AddRange(new object[] {
+            this.confirmationOptionsPage1,
```
kc284, line 66: It seems additional copies of the tabs were added automatically to this list when work took place on the designer (you can see them on the left hand side panel if you run the application and bring up the Options dialog). I suspect that this has to do with the logic in the event handlers of ControlAdded/Removed in the VerticallyTabbedDialog, in all likelihood the check of whether the tab exists already is wrong. For the current PR let's just revert this addition. The VerticallyTabbedDialog should be fixed to prevent it from happening again, but feel free to put that commit in a different PR since it is not related to the tooltip (which now seems to work ok).

kc284 (changes requested):
I've just realised that I added the comment the other day and forgot to click the request changes button.

kc284:
I forgot to say that the commits need sign-off info.

kc284 (approved):

kc284:
The changes are ok now, but the commits still need amending to include sign-off info. Let me know if you need help to do this.

PR #2842, opened 2021-08-02
CA-350411: Add conditional page control refresh in Import Wizard

danilo-delbusso:
Page controls are now forcefully refreshed only when necessary.

I suggest removing whitespace diffs.

### DIFF
--- a/XenAdmin/Wizards/ImportWizard/ImportWizard.cs
+++ b/XenAdmin/Wizards/ImportWizard/ImportWizard.cs
@@ -81,7 +81,7 @@ namespace XenAdmin.Wizards.ImportWizard
         private IXenConnection TargetConnection { get; set; }
         #endregion
 
-        public ImportWizard(IXenConnection con, IXenObject xenObject, string filename, bool ovfModeOnly)
+        public ImportWizard(IXenConnection con, IXenObject xenObject, string filename, bool ovfModeOnly = false)
             : base(con)
         {
             InitializeComponent();
--- a/XenAdmin/Wizards/ImportWizard/ImportWizard.cs
+++ b/XenAdmin/Wizards/ImportWizard/ImportWizard.cs
@@ -32,18 +32,18 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 using XenAdmin.Actions;
 using XenAdmin.Actions.OvfActions;
+using XenAdmin.Actions.VMActions;
+using XenAdmin.Commands;
 using XenAdmin.Controls;
 using XenAdmin.Core;
 using XenAdmin.Mappings;
 using XenAdmin.Network;
 using XenAdmin.Wizards.GenericPages;
 using XenAPI;
-using System.Linq;
-using XenAdmin.Actions.VMActions;
-using XenAdmin.Commands;
 using XenOvf;
 using XenOvf.Definitions;
 using XenOvf.Utilities;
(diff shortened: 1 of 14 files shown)


CitrixChris (dismissed):

XenAdmin/Wizards/ImportWizard/ImportWizard.cs
```diff
@@ -251,17 +251,30 @@ protected override void UpdateWizardContent(XenTabPage page)
                     m_pageImportSource.DiskCapacity, m_pageImportSource.IsWIM, m_pageVMconfig.AdditionalSpace,
                     m_pageImportSource.FilePath, m_pageImportSource.ImageLength);
 
+				if (m_pageStorage.SelectedOvfEnvelope != m_envelopeFromVhd
```
kc284, line 254: An easier way to do this is assign the newly created envelope to a local variable and compare this with `m_envelopeFromVhd` before assigning to the latter:
```
var newEnv = OVF.Create(...);
if (newEnv != m_envelopeFromVhd) {...}
m_envelopeFromVhd = newEnv;
```

XenAdmin/Wizards/ImportWizard/ImportWizard.cs
```diff
@@ -251,17 +251,30 @@ protected override void UpdateWizardContent(XenTabPage page)
...
-				NotifyNextPagesOfChange(m_pageHost, m_pageStorage, m_pageNetwork);
 			}
 			else if (type == typeof(ImportSelectHostPage))
 			{
 			    TargetConnection = m_pageHost.ChosenItem == null ? null : m_pageHost.ChosenItem.Connection;
                 RemovePage(m_pageRbac);
                 ConfigureRbacPage(TargetConnection);
+
+				if (m_pageNetwork.Connection != TargetConnection
+					|| m_pageStorage.VmMappings != m_vmMappings
```
kc284, line 273: The m_vmMappings are set further down; at this point they have not changed yet, so the comparison with m_pageStorage.VmMappings is always true. However, I suggest using local variables for comparing the connection and mappings (like with the OvfEnvelope in the previous comment).

Also,  the removal/addition of the boot options page is not needed if the import type or connection have not changed.

kc284 (changes requested):
We can also add checks for the page types ImportSourcePage and GlobalSelectHost.

AaronRobson (dismissed):

kc284:
This one proved trickier than anticipated and there was also an additional bug whereby the Boot options selected by the user were not set correctly when one imported VM from image. I have submitted some corrections here https://github.com/danilo-delbusso/xenadmin/pull/9. The whole thing will need re-review once merged and a lot of testing because there is a risk of the changes introducing regressions.

kc284 (changes requested):
Setting it status to "changes requested" until we review and merge https://github.com/danilo-delbusso/xenadmin/pull/9

XenAdmin/Wizards/ImportWizard/NetworkPickerPage.cs
```diff
@@ -49,8 +49,8 @@ public partial class NetworkPickerPage : XenTabPage
 		#region Private fields
 		private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 		
-		private IXenConnection m_selectedConnection;
-        private Host m_selectedAffinity;
+		internal IXenConnection MSelectedConnection { get; set; }
```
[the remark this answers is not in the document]
danilo-delbusso: Thanks! I was wondering what that was 😂

AaronRobson (approved):

CitrixChris (approved):

PR #2844, opened 2021-08-02
CA-279581: Prevent panel headers from overlapping scrollbar

danilo-delbusso:
Because of this, the scrollbar is partially visible until these pixels are refreshed manually (either by resizing the window, moving to a different page, or by interacting with controls that occupy the same space.

In order to avoid this, the `AutoScroll` property is updated programmatically whenever the `panel2`'s `Paint` event is fired.

### DIFF
--- a/XenAdmin/TabPages/GeneralTabPage.cs
+++ b/XenAdmin/TabPages/GeneralTabPage.cs
@@ -1944,8 +1944,9 @@ namespace XenAdmin.TabPages
         }
 
         private void panel2_Paint(object sender, PaintEventArgs e)
-        {
-            panel2.AutoScroll = sections.Sum(s => s.Height) > panel2.Height;
+        {   
+            //Force scrollbar to be repainted to avoid occasional pixel glitch.
+            panel2.AutoScroll = sections.Sum(s => s?.Parent.Height ?? s.Height) > panel2.ClientRectangle.Height;
         }
 
         #endregion
--- a/XenAdmin/TabPages/GeneralTabPage.cs
+++ b/XenAdmin/TabPages/GeneralTabPage.cs
@@ -1943,6 +1943,11 @@ namespace XenAdmin.TabPages
             StartPutty( "env docker logs --tail=50 --follow --timestamps");
         }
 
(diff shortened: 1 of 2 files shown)


XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -1943,6 +1943,11 @@ private void buttonViewLog_Click(object sender, EventArgs e)
             StartPutty( "env docker logs --tail=50 --follow --timestamps");
         }
 
+        private void panel2_Paint(object sender, PaintEventArgs e)
+        {
+            panel2.AutoScroll = sections.Sum(s => s.Height) > panel2.Height;
```
kc284, line 1948: Each of the sections sits in a panel that has a top and bottom padding, so I think we need to sum the parents' height here rather than the sections'. Also, the sum should probably be compared with the height of panel2's client area which excludes borders, scrollbars etc. (even if we don't have horizontal scrollbar here).

XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -1943,6 +1943,12 @@ private void buttonViewLog_Click(object sender, EventArgs e)
             StartPutty( "env docker logs --tail=50 --follow --timestamps");
         }
 
+        private void panel2_Paint(object sender, PaintEventArgs e)
+        {   
+            //Force scrollbar to be repainted to avoid occasional pixel glitch.
+            panel2.AutoScroll = sections.Sum(s => s?.Parent.Height ?? s.Height) > panel2.ClientRectangle.Height;
```
kc284, line 1949: `s?.Parent.Height` is conflicting with `s.Height`. If s were null, then s.Height would be called, which would crash. I presume you probably meant to check for its parent's nullness, i.e. `s.Parent?.Height`

danilo-delbusso: Yes I did! Updated 👍

kc284 (approved):

CitrixChris (approved):

PR #2846, opened 2021-08-03
CA-339590: Add conditional page control refresh in HA Wizard

danilo-delbusso:
Only refresh when initially scanning the pool.

### DIFF
--- a/XenAdmin/Wizards/HAWizard.cs
+++ b/XenAdmin/Wizards/HAWizard.cs
@@ -159,8 +159,7 @@ namespace XenAdmin.Wizards
         {
             var pageType = senderPage.GetType();
 
-            if ((pageType == typeof(Intro) && !_rbacNeeded || pageType == typeof(RBACWarningPage)) &&
-                xenTabPageChooseSR.IsFirstLoad)
+            if (pageType == typeof(Intro) && !_rbacNeeded || pageType == typeof(RBACWarningPage))
             {
                 // Start HB SR scan
                 // If scan finds no suitable SRs ChooseSR will show sensible text and disallow progress.
--- a/XenAdmin/Wizards/HAWizard.cs
+++ b/XenAdmin/Wizards/HAWizard.cs
@@ -122,7 +122,13 @@ namespace XenAdmin.Wizards
 
         protected override void UpdateWizardContent(XenTabPage senderPage)
         {
-            if (senderPage.GetType() == typeof(AssignPriorities))
+            Type type = senderPage.GetType();
+
+            if(type == typeof(Intro))
+            {
+                NotifyNextPagesOfChange(xenTabPageChooseSR, xenTabPageAssignPriorities, xenTabPageHaFinish);
+            }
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/HAWizard.cs
```diff
@@ -122,7 +122,13 @@ protected override void OnShown(EventArgs e)
 
         protected override void UpdateWizardContent(XenTabPage senderPage)
         {
-            if (senderPage.GetType() == typeof(AssignPriorities))
+            Type type = senderPage.GetType();
```
kc284, line 125: This will still reset the AssignPriorities settings if the user goes all the way to the first page, but I don't think we need it as we probably want to retain the settings even if the user chooses to store the metadata on a different SR. I think it will be simpler if we don't use the IsDirty flag at all (in fact we should remove it from elsewhere at some point because it adds complexity to the wizard logic). The error here is probably the fact that the controls are populated in the PageLoadedCore overload, so we should remove this and just rename the method PopulatePageControls to PopulatePage (which is an override of the base class method and is called on FirstLoad).

kc284 (changes requested):

XenAdmin/Wizards/HAWizard.cs
```diff
@@ -159,7 +159,8 @@ protected override bool RunNextPagePrecheck(XenTabPage senderPage)
         {
             var pageType = senderPage.GetType();
 
-            if (pageType == typeof(Intro) && !_rbacNeeded || pageType == typeof(RBACWarningPage))
+            if ((pageType == typeof(Intro) && !_rbacNeeded || pageType == typeof(RBACWarningPage)) &&
```
kc284, line 162: I don't think you need this change at all. Unless I'm missing something, my dev testing shows that simply renaming PopulatePageControls->PopulatePage does the job.

XenAdmin/Wizards/HAWizard_Pages/AssignPriorities.cs
```diff
@@ -72,15 +72,16 @@ public new IXenConnection Connection
...
+        public override void PopulatePage()
         {
             Debug.Assert(connection != null, "Connection is null; set it to non-null before calling this method.");
-
             UpdateMenuItems();
             PopulateVMs();
             haNtolIndicator.Connection = Connection;
             haNtolIndicator.Settings = getCurrentSettings();
             StartNtolUpdate();
+            IsDirty = false;
```
kc284, line 83: You don't need this either, it won't be taken account of anyway since the page does not override ImplementsIsDirty() any more.

XenAdmin/Wizards/HAWizard.cs
danilo-delbusso: The `IsFirstLoad` check makes sure the control changes aren't reset if the user goes back to the intro page.

kc284 (approved):

CitrixChris (approved):

PR #2847, opened 2021-08-03
CA-310974: Hide GPU page if the host has no GPU capability

danilo-delbusso:
* Hid GPU page in PropertiesDialog
* Hid GPU page in NewVMWizard

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/NewVMWizard.cs
+++ b/XenAdmin/Wizards/NewVMWizard/NewVMWizard.cs
@@ -221,7 +221,7 @@ namespace XenAdmin.Wizards.NewVMWizard
 
                 RemovePage(pageVgpu);
                 gpuCapability = Helpers.GpuCapability(xenConnection) && selectedTemplate.CanHaveGpu();
-                if (gpuCapability && pageVgpu.GpusAvailable)
+                if (gpuCapability && Helpers.GpusAvailable(xenConnection))
                     AddAfterPage(page_5_CpuMem, pageVgpu);
 
                 RemovePage(page_1b_BiosLocking);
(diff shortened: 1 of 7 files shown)


kc284 (changes requested):
These changes will hide the page even if there are GPUs, the reason being that GpusAvailable is set in the Populate method which is hit when the page is shown (on the Properties dialog) or stepped into (on the NewVM wizard), i.e. after the check for page.GpusAvailable takes place.

I think what we need to do is remove completely the fields gpu_groups and gpusAvailable from the GpuEdit, move the gpusAvailable logic to a static Helpers method and use the latter in the condition checked on the dialog and the wizard.

Also, once the fields are gone, there will be some dead code that can be removed (e.g the one hiding all the controls and showing the current rubric).

XenAdmin/Wizards/NewVMWizard/NewVMWizard.cs
```diff
@@ -221,7 +221,7 @@ protected override void UpdateWizardContent(XenTabPage senderPage)
 
                 RemovePage(pageVgpu);
                 gpuCapability = Helpers.GpuCapability(xenConnection) && selectedTemplate.CanHaveGpu();
-                if (gpuCapability)
+                if (gpuCapability && Helpers.GpusAvailable(xenConnection))
```
kc284, line 224: I think the check Helpers.GpusAvailable(xenConnection) should move to the assignment of gpuCapability above, because it is used in the Finish() method of this wizard to decide wether we're passing vGpus in the creation action.

XenModel/Utils/Helpers.cs
```diff
@@ -1844,5 +1844,10 @@ where network.Key.StartsWith(string.Format("{0}/ip", device))
                     orderby network.Key
                     select network.Value.Split(new[] { "\n", "%n" }, StringSplitOptions.RemoveEmptyEntries)).SelectMany(x => x).Distinct().ToList();
         }
+
+        public static bool GpusAvailable(IXenConnection connection)
+        {
+            return connection?.Cache?.GPU_groups?.Any(g => g.PGPUs?.Count > 0 && g.supported_VGPU_types?.Count != 0) ?? false;
```
kc284, line 1850: The only strictly necessary null check is the one on the connection. The Cache and its properties (e.g. GPU_groups) are readonly and initialised as empty. The API object properties that are ref lists (g.PGPUs, g.supported_VGPU_types) are also initialised as empty and are demarshalled by the SDK as empty if they have no values.

kc284 (approved):

CitrixChris (approved):

PR #2848, opened 2021-08-04
CP-36567: Use the datasource units specified in the API.

kc284:
I have done some refactoring and renamed several Properties in order to compact the code and make it easier to read. I recommend reviewing each commit separately as it's easier to follow the changes.


XenAdmin/Controls/CustomDataGraph/DataSet.cs
```diff
@@ -57,10 +57,7 @@ public class DataSet : IComparable<DataSet>
         private bool _hide;
         private bool _deselected;
 
-        /// <summary>
-        /// A guess at the data type
-        /// </summary>
-        public DataType Type;
+        public DataType Category;
```
danilo-delbusso: Just checking that the comment's removal was on purpose here.

danilo-delbusso: Just checking that the comment's removal was done on purpose here. Ignore if that's the case.

XenAdmin/Controls/CustomDataGraph/DataSet.cs
```diff
@@ -58,10 +58,7 @@ public class DataSet : IComparable<DataSet>
         private bool _deselected;
 
         public DataType Category;
-        /// <summary>
-        /// What it really is
-        /// </summary>
-        public string TypeString;
+        public string DataSourceName;
```
danilo-delbusso, line 57: Similarly here.

XenAdmin/Controls/CustomDataGraph/DataSet.cs
kc284: I removed it because it's misleading. We're not guessing the type, just assigning a category so it's easier for the user to find the graphs (the category is shown on the Performance page > Actions button > New Graph dialog).

XenModel/Utils/Helpers.cs
```diff
@@ -959,31 +959,120 @@ public static void RemoveFromGuiConfig(Session session, IXenObject o, string key
...
+                case DataSourceCategory.Pvs:
+                    return Messages.DATATYPE_PVS;
+                default:
+                    return Messages.DATATYPE_CUSTOM;
+            }
+        }
+
         public static Regex CpuRegex = new Regex("^cpu([0-9]+)$");
-        public static Regex CpuAvgFreqRegex = new Regex("^CPU([0-9]+)-avg-freq$");
+        static Regex CpuAvgFreqRegex = new Regex("^CPU([0-9]+)-avg-freq$");
```
danilo-delbusso, line 1004: This is the only regex with uppercase lettering. So just double checking that's on purpose here.

XenModel/Utils/Helpers.cs
```diff
@@ -959,31 +959,120 @@ public static void RemoveFromGuiConfig(Session session, IXenObject o, string key
...
         static Regex GpuRegex = new Regex(@"^gpu_((memory_(free|used))|power_usage|temperature|(utilisation_(compute|memory_io)))_((([a-fA-F0-9]{4}\/)|([a-fA-F0-9]{8}\/))?[a-fA-F0-9]{2}\/[0-1][a-fA-F0-9].[0-7])$");
 
+        public static DataSourceCategory GetDataSourceCategory(string name)
+        {
+            if (CpuRegex.IsMatch(name) || CpuAvgFreqRegex.IsMatch(name) ||
+                CpuStateRegex.IsMatch(name) || CpuOtherRegex.IsMatch(name))
+                return DataSourceCategory.Cpu;
+
+            if (VcpuRegex.IsMatch(name))
+                return DataSourceCategory.Cpu;
```
danilo-delbusso, line 1039: These can be rewritten as a single if statement.

Same applies to line 1067 with

```csharp
if (XapiMemoryRegex.IsMatch(name))
                return DataSourceCategory.Memory;

if (name.StartsWith("memory"))
                return DataSourceCategory.Memory;
```

danilo-delbusso, line 1039: These can be rewritten as a single if statement.

Same applies to line 1067 with

```csharp
if (XapiMemoryRegex.IsMatch(name))
                return DataSourceCategory.Memory;

if (name.StartsWith("memory"))
                return DataSourceCategory.Memory;
```

XenModel/Utils/Helpers.cs
kc284: Yes, there are some host datasources that use uppercase lettering. You can open a shell to the host and check the names on the CLI: `xe host-data-source-list uuid=xxxx | grep name_label` (and similarly for a running VM with `vm-data-source-list`).

XenModel/Utils/Helpers.cs
kc284: Fixed.

danilo-delbusso (approved):

CitrixChris (approved):

PR #2849, opened 2021-08-05
CA-260319: Show warning in Health Check if connected with less privileged RBAC User

danilo-delbusso:


XenAdmin/Dialogs/HealthCheck/HealthCheckOverviewDialog.cs
```diff
@@ -279,15 +279,26 @@ public void UpdateAnalysisResult(HealthCheckSettings healthCheckSettings)
...
+            bool canAlterHealthCheck = Core.HealthCheck.PassedRbacChecks(pool.Connection);
+
             refreshLinkLabel.Visible =
                 editLinkLabel.Visible =
                 disableLinkLabel.Visible =
                 uploadRequestLinkLabel.Visible =
-                enrollNowLinkLabel.Visible = Core.HealthCheck.PassedRbacChecks(pool.Connection);
+                enrollNowLinkLabel.Visible = canAlterHealthCheck;
+
+            if (!canAlterHealthCheck)
```
kc284, line 293: We need an else block (or better reverse the if-check block so the condition is easier to grasp) to hide the message in the opposite case  (compare what happens now on the panel if you are connected to one pool as an unprivileged user and to another pool as a privileged one and you select the second pool afer the first one).

XenAdmin/Dialogs/HealthCheck/HealthCheckOverviewDialog.cs
```diff
@@ -279,15 +279,26 @@ public void UpdateAnalysisResult(HealthCheckSettings healthCheckSettings)
                 previousUploadPanel.Visible = false;
             }
         }
-
+        
         public void UpdateButtonsVisibility(Pool pool)
         {
+            bool canAlterHealthCheck = Core.HealthCheck.PassedRbacChecks(pool.Connection);
```
kc284, line 285: In order to write a better message (see my comment in Messages.resx) we should extend the PassedRbacChecks method to return the allowed roles for the operation. To do this replace in the latter method `CanPerform(RbacMethodList, IXenConnection)` with `public static bool CanPerform(RbacMethodList, IXenConnection, out List<Role>)`.

XenAdmin/Dialogs/HealthCheck/HealthCheckOverviewDialog.Designer.cs
```diff
@@ -284,6 +286,7 @@ private void InitializeComponent()
             // 
             resources.ApplyResources(this.notEnrolledPanel, "notEnrolledPanel");
             this.poolDetailsPanel.SetColumnSpan(this.notEnrolledPanel, 2);
+            this.notEnrolledPanel.Controls.Add(this.cannotAlterHealthCheckLabel, 0, 2);
```
kc284, line 289: We need to move the label outside this panel because it does not apply only to enrollment (see other operations that are hidden when RBAC check fails). In fact notEnrolledPanel is invisible when the pool is enrolled, so the message is not seen.

XenModel/Messages.ja.resx
```diff
@@ -14613,4 +14613,7 @@ VM のパフォーマンスを最適化するために、VCPU の数を物理 CP
   <data name="YOU_ARE_HERE" xml:space="preserve">
     <value>現在の位置</value>
   </data>
+  <data name="HEALTHCHECK_CANNOT_ALTER_SETTINGS" xml:space="preserve">
```
kc284, line 14616: Changing the l10n resx is not necessary at this stage, the translations are handled afterwards.

kc284 (changes requested):

XenAdmin/Dialogs/HealthCheck/HealthCheckOverviewDialog.Designer.cs
```diff
@@ -141,8 +143,9 @@ private void InitializeComponent()
             // 
             resources.ApplyResources(this.poolDetailsPanel, "poolDetailsPanel");
             this.tableLayoutPanel2.SetColumnSpan(this.poolDetailsPanel, 2);
-            this.poolDetailsPanel.Controls.Add(this.healthCheckStatusPanel, 0, 0);
-            this.poolDetailsPanel.Controls.Add(this.notEnrolledPanel, 0, 1);
+            this.poolDetailsPanel.Controls.Add(this.cannotAlterHealthCheckLabel, 0, 0);
```
kc284, line 146: The label now looks ok at the top. I suggest setting its bottom Margin to 10 to make it stand out a bit more from the sections after it, and maybe add a right and left Padding of 3 to match the other sections.

kc284 (approved):

CitrixChris (approved):

PR #2850, opened 2021-08-07
Tidy up config files

kc284:
Commits best reviewed separately, but can be squashed when merging. Note that the variables marked as todo are things we don't support at the moment and will be revisited in future.


AaronRobson (approved):

XenOvfApi/Security.cs
```diff
@@ -401,6 +392,52 @@ public static int CalculateStrength(string password)
...
+                case "des":
+                    return typeof(DESCryptoServiceProvider);
+                case "rc2":
+                    return typeof(RC2CryptoServiceProvider);
+                case "kw_aes128":
+                case "kw_aes256":
+                case "kw_aes192":
+                case "aes128_cbc":
+                case "aes256_cbc":
+                case "aes192_cbc":
```
danilo-delbusso, line 435: These are not used. I'd remove them unless they're here to aide future refactoring efforts.

danilo-delbusso (changes requested):

XenOvfApi/Security.cs
kc284: Indeed these can be removed because they fall into the default case, the only purpose they serve is to make explicit what algorithms one might expect. I think for the time being I'd ignore this because the list needs revisiting anyway (but it's not urgent since encryption is not supported yet).

danilo-delbusso (approved):

PR #2854, opened 2021-08-10
CA-339273: Add source server/pool info to descriptions of actions that move/copy/import resources

danilo-delbusso:

The actions that have been updated are:

### VMs

`VMMoveAction`
`VMCrossPoolMigrateAction`
`VMCloneAction`
`VMMigrateAction`
`VMCopyAction`
`ImportVmAction`

### VDIs

`MoveVirtualDiskAction`
`MigrateVirtualDiskAction`

Please advise if any should be added.

### DIFF
--- a/XenModel/Actions/VM/VMCloneAction.cs
+++ b/XenModel/Actions/VM/VMCloneAction.cs
@@ -37,9 +37,9 @@ namespace XenAdmin.Actions.VMActions
 {
     public class VMCloneAction : AsyncAction
     {
+        private readonly string _cloneName;
+        private readonly string _cloneDescription;
 
-        protected string _cloneName;
-        protected string _cloneDescription;
         public VMCloneAction(VM vm, string name, string description)
             : base(vm.Connection, string.Format(Messages.CREATEVM_CLONE, name, vm.NameWithLocation()))
         {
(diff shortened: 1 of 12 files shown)


XenModel/Actions/VM/VMCloneAction.cs
```diff
@@ -40,14 +41,13 @@ public class VMCloneAction : AsyncAction
         protected string _cloneName;
         protected string _cloneDescription;
         public VMCloneAction(VM vm, string name, string description)
-            : base(vm.Connection, string.Format(Messages.CREATEVM_CLONE, name, vm.Name()))
+            : base(vm.Connection, string.Format(Messages.CREATEVM_CLONE, name, vm.NameWithLocation()))
```
CitrixChris, line 44: Is name with location the best for this? Is the location the host? if so why not use name with host?

CitrixChris (changes requested):

XenModel/Actions/VM/VMCloneAction.cs
danilo-delbusso: The location is the host.
My rationale for this was:

- Method is there and is used for this purpose, so it's easier to maintain
- I would have add to modify the CREATEVM_CLOSE string resource to take into account a new variable for the `string.Format` call

AaronRobson (approved):

CitrixChris (approved):

kc284:
I have submitted some small corrections via https://github.com/danilo-delbusso/xenadmin/pull/10. Once the latter is reviewed and merged, we need a re-review of the total by someone else.

AaronRobson (approved):

CitrixChris (approved):

PR #2862, opened 2021-08-23
CA-357325: Add Rocky Linux icon

xihuan-citrix:


xihuan-citrix:
Hi @kc284  and @jijiang , could you help to review this PR?

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -841,21 +841,22 @@ public enum VmTemplateType
...
-            Turbo = 15,
-            Ubuntu = 16,
-            YinheKylin = 17,
-            Citrix = 18,
-            Solaris = 19,
-            Misc = 20,
-            Snapshot = 21,
-            SnapshotFromVmpp = 22,
-            Count = 23  //bump this if values are added
+            Rocky = 9,
```
AaronRobson, line 844: Given previous commits in this area I realise that these don't need to preserve ID numbers of existing items.
How do we choose the order of these items? Some of it seems roughtly alphabetical.

kc284: According to the comment, there is a certain order required only for the various template categories. I think it makes sense though to have alphabetical order within each category (when no other rules apply, e.g. in the case of Windows it makes sense to put legacy systems second). So, yes, it's probably best to move Rocky after RedHat.

kc284 (changes requested):
Please change the order as discussed above.

XenModel/XenAPI-Extensions/VM.cs
xihuan-citrix: Move rocky after redhat

kc284 (approved):

kc284:
@xihuan-citrix could you also ammend the commit message to use CP-38073 instead of CA-357325? Sorry, I forgot to say this earlier.

PR #2865, opened 2021-08-27
CP-36908: Remove support for xscontainer supp-pack in stockholm

kc284:


XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -1696,6 +1691,20 @@ private string hostInstalledSuppPacks(Host host)
...
+        {
+            if (!(xenObject is DockerContainer) || Helpers.ContainerCapability(xenObject.Connection))
+            {
+                Banner.Visible = false;
+            }
+            else
+            {
+                Banner.BannerType = DeprecationBanner.Type.Removal;
+                Banner.WarningMessage = string.Format(Messages.CONTAINER_MANAGEMENT_REMOVAL_WARNING, string.Format(Messages.XENSERVER_8_2, BrandManager.ProductVersion82));
+                Banner.Visible = true;
```
danilo-delbusso, line 1704: Slightly unrelated to the PR.

In the `DeprecationBanner`, if there is no value for the `Details` link, the link is still shown to the User as clickable. It would probably be best to hide it since clicking it doesn't do anything.

Since it's not related to the changes here I will be approving the PR. However if you have time I'd remove the visibility of the Details link when there's no specified link, like in this case. (Or even better, also link something here!)

danilo-delbusso (approved):

XenAdmin/TabPages/GeneralTabPage.cs
kc284: I made it point to the deprecated features page.

danilo-delbusso (approved):

MihaelaStoica (approved):

PR #2870, opened 2021-09-09
Prevent null reference when importing OVF files from web server

danilo-delbusso:
Prevents error while downloading OVF from a web server while importing a VM


XenAdmin/Wizards/ImportWizard/ImportSourcePage.cs
```diff
@@ -754,11 +754,14 @@ private void webclient_DownloadFileCompleted(object sender, AsyncCompletedEventA
                             int index = _uri.OriginalString.LastIndexOf('/') + 1;
                             var remoteDir = _uri.OriginalString.Substring(0, index);
 
-                            foreach (var file in envType.References.File)
+                            if (envType.References.File != null)
```
kc284, line 757: Can we add a null check on the References as well just in case.

kc284 (approved):

MihaelaStoica (approved):

PR #2875, opened 2021-09-21
Add GitHub action to update PR labels

danilo-delbusso:
Action can add/remove the following labels:

- One Approval label
- Two Approvals label
- Changes Requested label
- Updated PR Label

The labels have to exist in the repository already. The defaults of the action correspond to the labels present for `xenadmin` so there is no need to specify them.

Doesn't override existing labels. Only changes:

- 1 approval
- 2 approvals
- updated
- needs updating

### DIFF
--- a/.github/workflows/update-pr-labels.yml
+++ b/.github/workflows/update-pr-labels.yml
@@ -1,0 +1,16 @@
+name: Update PR Labels
+on:
+  pull_request_review:
+    types: [submitted]
+  pull_request:
+    types: [synchronize]
+
+jobs:
+  update-pr-labels:
+    runs-on: ubuntu-latest
+    name: Update PR Labels
+    steps:
+      - name: Update Labels
+        uses: danilo-delbusso/pr-review-labeller@v1.1.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


AaronRobson (approved):

CitrixChris (approved):

danilo-delbusso:
Needs:
1. Action permissions should be limited
2. 3 approvals should not remove other approval labels

danilo-delbusso:
Unfortunately, it seems that `GITHUB_TOKEN` cannot fetch write permissions to the PR if the PR has been opened from a fork. (Necessary for adding/removing labels)

While the `pull_request_target` event can achieve this level of permission (as opposed to the normal `pull_request`), it can only be use to add/remove the `updated` label.

For approval labels (`1 approval`, `2 approvals`, `needs changing`), GitHub will need to add a version of the `pull_request_review` event which runs within a scope that can use write permissions. Else we get this:
image

While there are some [workarounds](https://stackoverflow.com/a/67249854), they're just that: workarounds.

The only reasonable way is currently to:
- [Generate a token with the necessary permissions](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)
- [Add token as a repository variable](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository)
- Pass the token to the action as environment variable (much like `GITHUB_TOKEN` is being used now)

Another way is to also just wait for GitHub to catch up to their users (I'm surprised this doesn't exist yet), and put this on hold.


Thoughts?



________________________________

Some useful resources:

- [User reporting the issue, with an answer from GitHub](https://github.community/t/github-actions-are-severely-limited-on-prs/18179/2) (Before `pull_request_target` existed)
- [Issue raised to @actions](https://github.com/actions/first-interaction/issues/10)
- [Docs about `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)

danilo-delbusso:
I'll close this. Doesn't seem like there's an urgent need + GitHub hasn't announced a new event that would address this specific case. Feel free to re-open in case you think we should have more visibility.

PR #2884, opened 2021-09-28
Authentication for updates - part 2

kc284:


CFUValidator/CommandLineOptions/CommandLineParser.cs
```diff
@@ -46,7 +46,9 @@ public class CommandLineParser
             new CommandLineArgument(OptionUsage.Url, 'u', "URL of CFU XML. Cannot be used with -f flag."),
             new CommandLineArgument(OptionUsage.File, 'f', "Path to CFU XML file. Cannot be used with -u flag."),
             new CommandLineArgument(OptionUsage.ServerVersion, 's', "Server version to test, eg. 6.0.2."),
-            new CommandLineArgument(OptionUsage.Hotfix, 'p', "Space delimited list of patches/hotfixes to test, e.g. XS602E001.")
+            new CommandLineArgument(OptionUsage.Hotfix, 'p', "Space delimited list of patches/hotfixes to test, e.g. XS602E001."),
+            new CommandLineArgument(OptionUsage.Username, 'n', "Username for downloading hotfixes, e.g. john123"),
+            new CommandLineArgument(OptionUsage.ClientId, 'k', "ClientId to use for downloading hotfixes, eg. das1djs43ahgd5jgjh0")
```
danilo-delbusso, line 51: This is very small but SonarQube might complain, thinking this is an actual password. I'd use something less "realistic", or remove the example altogether

XenModel/Actions/Updates/TokenManager.cs
```diff
@@ -30,9 +30,191 @@
...
+
+                    case HttpStatusCode.Unauthorized:
+                        log.Error("Could not authenticate account (401 Unauthorized). Client ID may be invalid or revoked.");
+                        throw new Exception(string.Format(Messages.STRING_SPACE_STRING,
+                            Messages.FILESERVICE_AUTHENTICATE_ERROR, Messages.FILESERVICE_ERROR_401));
+
+                    case HttpStatusCode.InternalServerError:
+                        log.Error("Could not authenticate account (500 Internal server error). Could not get session token or configuration object was null.");
+                        throw new Exception(string.Format(Messages.STRING_SPACE_STRING,
+                            Messages.FILESERVICE_AUTHENTICATE_ERROR, Messages.FILESERVICE_ERROR_500));
```
danilo-delbusso, line 135: In this case, it is more accurate to say that `The authentication server has encountered a situation it doesn't know how to handle`, since 500 can be thrown for all kinds of reasons.

The issue is that it might delay future debugging in case a 500 is being thrown for reasons not related to the session token or the configuration object.

danilo-delbusso (changes requested):

danilo-delbusso (approved):

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -36,84 +36,228 @@
...
+            _patchUri = uri;
+            skipUnzipping = updateFileSuffixes.Any(item => _patchUri.ToString().EndsWith(item));
+        }
 
-        public DownloadAndUnzipXenServerPatchAction(string patchName, Uri uri, string outputFileName, bool suppressHist,
-            params string[] updateFileExtensions)
-            : base(null, uri == null
-                ? string.Format(Messages.UPDATES_WIZARD_EXTRACT_ACTION_TITLE, patchName)
-                : string.Format(Messages.DOWNLOAD_AND_EXTRACT_ACTION_TITLE, patchName), string.Empty, suppressHist) 
+        public override void RecomputeCanCancel()
```
CitrixChris, line 181: Dont see what the point in this function is, it assigns a variable that is never used

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -329,7 +338,35 @@ void client_DownloadFileCompleted(object sender, AsyncCompletedEventArgs e)
...
+                            patchDownloadError = new Exception(Messages.FILESERVICE_ERROR_403);
+                            break;
+
+                        case HttpStatusCode.NotFound:
+                            log.Error($"Could not download {updateName} (404 File not found).");
+                            patchDownloadError = new Exception(Messages.FILESERVICE_ERROR_404);
+                            break;
+
+                        default:
+                            patchDownloadError = e.Error;
```
CitrixChris, line 361: Shouldn't we log even if its not the above statuses

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
```diff
@@ -340,14 +377,43 @@ void client_DownloadFileCompleted(object sender, AsyncCompletedEventArgs e)
...
+        }
+
+        protected override void UpdateExtractionProgress(long bytesTransferred, long totalBytesToTransfer)
+        {
+            PercentComplete = (int)(100.0 * bytesTransferred / totalBytesToTransfer);
         }
 
+        public override void RecomputeCanCancel()
+        {
+            CanCancel = !Cancelling && !IsCompleted;
```
CitrixChris, line 416: again not sure of the point of this function that doesn't get called or assign anything that is used

CitrixChris (changes requested):

CFUValidator/Validators/ZipContentsValidator.cs
```diff
@@ -57,9 +58,16 @@ public ZipContentsValidator(List<XenServerPatchAlert> alerts)
 
         protected override void ValidateCore(Action<string> statusReporter)
         {
-            foreach (XenServerPatchAlert alert in alerts.OrderBy(a => a.Patch.Name))
+            try
             {
-                DownloadPatchFile(alert, statusReporter);
+                TokenManager.GetToken(_configProvider);
+
+                foreach (XenServerPatchAlert alert in alerts.OrderBy(a => a.Patch.Name))
```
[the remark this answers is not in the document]
kc284: Possibly, but I wouldn't address this right now because it's not code that has been changed by this PR, it was just rearranged to accommodate usage with a token.

XenAdmin/Core/Registry.cs
```diff
@@ -171,6 +171,24 @@ public static string GetCustomUpdatesXmlLocation()
                    ReadRegistryValue(RegistryHive.LocalMachine, XENCENTER_LOCAL_KEYS, CUSTOM_UPDATES_XML_LOCATION);
         }
 
+        public static string GetCustomFileServicePrefix()
```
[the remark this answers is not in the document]
kc284: I used this to match everything else in the file, note though that this is a static method and outside the class it's called as Registry.GetCustomFileServicePrefix(), so it's fairly obvious what it does.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
kc284: This is base class infrastructure used to decide whether an action is cancellable (e.g. if the action were shown on a dialog, it would decide whether to enable the cancel button for the user).

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
kc284: These are caught and logged further up (see method DownloadFile).
Throughtout XenModel actions you may see that some things are caught and logged within the Run() method (which is the action body) when we want to show a specialized message. What is not handled there is always caught within the base AsyncAction or ActionBase classes.

XenModel/Actions/Updates/DownloadAndUnzipXenServerPatchAction.cs
kc284: Same as above, it's base class infrastructure.

CitrixChris (approved):

PR #2891, opened 2021-10-13
A couple of corrections

kc284:


danilo-delbusso (approved):
Looks good, tested with a couple of non-fileservice URIs.

PR #2905, opened 2021-11-09
CA-343229: Changing roles when logged in as an AD user only takes effect after logging out/in

danilo-delbusso:
Best reviewed commit-by-commit. Whitespace differences are the automatic indent as completed by VS2019.

Took advantage to improve UX slightly:
- Grouped logging out actions into a `MultipleAction` to avoid one RBAc prompt per user disconnection
- Tell user that they're also going to disconnect other users when the selected users include the current one.

Users will still need to insert credentials for two RBAC prompts if they're changing roles to logged in users (one for changing the roles, one for disconnections if necessary). Not ideal but it makes the code tidier since we can reuse methods.

Tested with:

| Changing current user 	| Changing other user(s) 	| Other user(s) are(is) logged in 	| More than one user involved 	| Result 	|
|:---:	|:---:	|---	|---	|---	|
| NO 	| YES 	| NO 	| YES 	| ✅ 	|
| NO 	| YES 	| YES 	| NO 	| ✅ 	|
| NO 	| YES 	| YES 	| YES 	| ✅ 	|
| NO 	| YES 	| NO 	| NO 	| ✅ 	|
| YES 	| NO 	| NO 	| NO 	| ✅ 	|
| YES 	| NO 	| YES 	| NO 	| ✅ 	|
| YES 	| YES 	| YES 	| YES 	| ✅ 	|

### DIFF
--- a/XenAdmin/TabPages/AdPage.cs
+++ b/XenAdmin/TabPages/AdPage.cs
@@ -862,7 +862,7 @@ namespace XenAdmin.TabPages
                         {
                             subjectName = subjectName.Ellipsise(256);
                         }
-                        string msg = string.Format(entry.IsGroup ? Messages.AD_CONFIRM_SUICIDE_GROUP : Messages.AD_CONFIRM_SUICIDE,
+                        string msg = string.Format(entry.IsGroup ? Messages.AD_CONFIRM_LOGOUT_CURRENT_USER_GROUP : Messages.AD_CONFIRM_LOGOUT_CURRENT_USER,
                                     subjectName, Helpers.GetName(_connection).Ellipsise(50));
 
                         DialogResult r;
@@ -1081,10 +1081,10 @@ namespace XenAdmin.TabPages
                         }
                     }
 
-                    var suicide = false;
+                    var logoutCurrentSubject = false;
                     if (subjectsToLogout.Count > 0)
                     {
-                        if (!ConfirmLogout(subjectsToLogout.Contains(sessionUser), subjectsToLogout, out suicide))
+                        if (!ConfirmLogout(subjectsToLogout.Contains(sessionUser), subjectsToLogout, out logoutCurrentSubject))
                             return;
                     }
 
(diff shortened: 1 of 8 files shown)


XenAdmin/TabPages/AdPage.cs
```diff
@@ -1057,24 +1066,74 @@ private void ButtonChangeRoles_Click(object sender, EventArgs e)
...
+                    foreach (var loggedInSubject in selectedSubjects.Where(loggedInUsers.Contains))
+                    {
+                        var subjectRoles = loggedInSubject.roles.Select(_connection.Resolve).OrderBy(x => x).ToList();
+                        if (!selectedRoles.SequenceEqual(subjectRoles))
+                        {
+                            subjectsToLogout.Add(loggedInSubject);
+                        }
+                    }
+
+                    var suicide = false;
```
CitrixChris, line 1084: This feels like a negative language thing

CitrixChris (changes requested):

AaronRobson (approved):

CitrixChris (approved):

kc284 (changes requested):
Overall I agree with the solution provided, I have a few remarks:
1. The method LoggedInSubjects() in Helpers.cs is making a server call and is called on the UI thread. This practice should be avoided because it causes delays or freezing on the UI.
2. The warning prompts for the case of logging out after a role change need some clarification.
3. If the current logged in user is also logged in from other clients, logging the user out from the current application instance does not log out the other sessions. Note that this is an existing bug and is not caused by the changes in this PR, but we can try fixing it since we're here.

I've made the corrections in this PR https://github.com/danilo-delbusso/xenadmin/pull/11. The majority of changes are because of bullet point 3.

kc284 (dismissed):

CitrixChris (approved):

danilo-delbusso:
Force push related to trivial merge conflict in `XenModel/Utils/Helpers.cs`

kc284 (approved):

PR #2906, opened 2021-11-15
CA-360830: Do not allow the user to proceed without a client ID when …

kc284:
…applying updates after selecting new version ISO from disk.


CitrixChris (approved):

AaronRobson (approved):

danilo-delbusso (approved):
Tested with `xscontainer.iso` on 7.1 CU2 host

PR #2913, opened 2021-12-16
CA-361078: vCPUs count isn't exported/imported within OVFs

danilo-delbusso:
Didn't change other calls to `OVF.SetCPUs`, since the code is quite complex. If anyone can advise on which ones can be updated, I'll make those changes as well.

Change makes use of the `rasd:Limit` element of the `Item` tag representing vCPUs.
It is described by the [specification (see 7.7.11)](http://www.dmtf.org/standards/published_documents/DSP1041_1.1.pdf) as
>  `rasd:Limit`: A client may set the value of the Limit property to specify the maximum amount of host resource that is requested by the virtual resource allocation request.

### DIFF
--- a/XenOvfApi/OVF.cs
+++ b/XenOvfApi/OVF.cs
@@ -39,9 +39,9 @@ using System.Resources;
 using System.Security.Permissions;
 using System.Text;
 using System.Text.RegularExpressions;
+using XenCenterLib.Compression;
 using XenOvf.Definitions;
 using XenOvf.Utilities;
-using XenCenterLib.Compression;
 
 namespace XenOvf
 {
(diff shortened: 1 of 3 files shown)


CitrixChris (approved):

XenOvfApi/OVF.cs
```diff
@@ -4307,13 +4307,15 @@ public static VirtualHardwareSection_Type FindVirtualHardwareSectionByAffinity(E
         #endregion
 
         #region SETs
+
         /// <summary>
-        /// Set the count for the number of CPUs to assign to the virtual machine.
+        /// Set the count for the number of vCPUs and max vCPUs to assign to the virtual machine.
         /// </summary>
         /// <param name="ovfObj">EnvelopeType</param>
```
kc284, line 4314: Unrelated to this PR, but since we're here, the parameter name should change from `ovfObj` to `ovfEnv.

XenOvfApi/OVF.cs
```diff
@@ -4307,13 +4307,15 @@ public static VirtualHardwareSection_Type FindVirtualHardwareSectionByAffinity(E
...
-        /// Set the count for the number of CPUs to assign to the virtual machine.
+        /// Set the count for the number of vCPUs and max vCPUs to assign to the virtual machine.
         /// </summary>
         /// <param name="ovfObj">EnvelopeType</param>
         /// <param name="vsId">Virtual System Id</param>
-        /// <param name="cpucount">Number of CPUs</param>
-		public static void SetCPUs(EnvelopeType ovfEnv, string vsId, UInt64 cpucount)
+        /// <param name="cpucount">Number of vCPUs</param>
+        /// <param name="maxCpuCount">Number of max vCPUs, if available</param>
+        public static void SetCPUs(EnvelopeType ovfEnv, string vsId, UInt64 cpucount, ulong? maxCpuCount = null)
```
kc284, line 4318: Instead of making this nullable, you can define it as `ulong maxCpuCount = 100_000` and then further down it can be just `rasd.Limit = new cimUnsignedLong { Value = maxCpuCount };`

kc284 (changes requested):

kc284 (approved):

PR #2914, opened 2021-12-17
(Stockholm) CA-362018: Improve update notification dismissal message

danilo-delbusso:
Also removes the wording related to permanent dismissal, since we give users the option to restore update notifications.

Please advise on if the new wording does not correctly reflect XC's behaviour here.

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -12968,12 +12968,12 @@ Please check your settings and try again.</value>
     <value>{0}: Check skipped because WLB is enabled on pool {1}.</value>
   </data>
   <data name="UPDATE_DISMISS_ALL_CONTINUE" xml:space="preserve">
-    <value>You have applied filters to the list of updates. Do you want to dismiss all updates from every connected server, or only the updates you have chosen to view? In both cases the dismissed updates will be removed from the servers permanently.
+    <value>You have applied filters to the list of update notifications. Do you want to dismiss all update notifications for every connected server, or only the update notifications you have chosen to view?
 
 Note that if RBAC is enabled, only updates which you have privileges to dismiss will be affected.</value>
   </data>
   <data name="UPDATE_DISMISS_ALL_NO_FILTER_CONTINUE" xml:space="preserve">
-    <value>This operation will remove permanently all updates from every connected server. Do you want to continue?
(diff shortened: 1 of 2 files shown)


XenModel/Messages.resx
```diff
@@ -12968,20 +12968,20 @@ Please check your settings and try again.</value>
     <value>{0}: Check skipped because WLB is enabled on pool {1}.</value>
   </data>
   <data name="UPDATE_DISMISS_ALL_CONTINUE" xml:space="preserve">
-    <value>You have applied filters to the list of updates. Do you want to dismiss all updates from every connected server, or only the updates you have chosen to view? In both cases the dismissed updates will be removed from the servers permanently.
+    <value>You have applied filters to the list of updates. Do you want to dismiss all updates in {0} from every connected server, or only the updates you have chosen to view?
```
kc284, line 12971: I think we can keep the messages simple and avoid the client placeholder and string formatting by using the term "update notifications" instead of "updates". I.e. `You have applied filters to the list of update notifications. Do you want to dismiss all update notifications for every connected server, or only the update notifications you have chosen to view?` Then it's clear that we only touch the alerts and not the actual updates. Same in the two messages below.

kc284 (changes requested):

XenModel/Messages.resx
```diff
@@ -12968,20 +12968,20 @@ Please check your settings and try again.</value>
...
   </data>
   <data name="UPDATE_DISMISS_ALL_CONTINUE" xml:space="preserve">
-    <value>You have applied filters to the list of updates. Do you want to dismiss all updates from every connected server, or only the updates you have chosen to view? In both cases the dismissed updates will be removed from the servers permanently.
+    <value>You have applied filters to the list of update notifications. Do you want to dismiss all update notifications for every connected server, or only the update notifications you have chosen to view?
 
 Note that if RBAC is enabled, only updates which you have privileges to dismiss will be affected.</value>
   </data>
   <data name="UPDATE_DISMISS_ALL_NO_FILTER_CONTINUE" xml:space="preserve">
-    <value>This operation will remove permanently all updates from every connected server. Do you want to continue?
+    <value>This operation will dismiss all update notifications from every connected server. Do you want to continue?
```
kc284, line 12976: "for every"

CitrixChris (approved):

PR #2918, opened 2021-12-24
CP-38814: Add friendly names to new xapi alerts

danilo-delbusso:
This PR contains a temporary fix to the SDK. ~Commit can be removed if we instead prefer to update the entire SDK.~ 

🚨🚨🚨🚨🚨
SDK has been updated to `xapi 21.3.0/1.xs8`, **with a modification** needed to avoid issues related to CA-355432 as pointed out by @kc284.
🚨🚨🚨🚨🚨

The following additions have been made:

```
vm_migrated
vm_snapshotted
vm_snapshot_reverted
vm_checkpointed
vm_paused
vm_unpaused
```

### DIFF
--- a/XenModel/FriendlyNames.resx
+++ b/XenModel/FriendlyNames.resx
@@ -1682,11 +1682,20 @@
   <data name="Message.name-vif_qos_failed" xml:space="preserve">
     <value>VM Network Interface QoS Failed</value>
   </data>
+  <data name="Message.name-vm_checkpointed" xml:space="preserve">
+    <value>Checkpoint for virtual machine has been created</value>
+  </data>
   <data name="Message.name-vm_cloned" xml:space="preserve">
     <value>VM Cloned</value>
   </data>
   <data name="Message.name-vm_crashed" xml:space="preserve">
     <value>VM Crashed</value>
+  </data>
+  <data name="Message.name-vm_migrated" xml:space="preserve">
+    <value>Virtual Machine has been migrated</value>
+  </data>
+  <data name="Message.name-vm_paused" xml:space="preserve">
+    <value>Virtual machine has been paused</value>
   </data>
   <data name="Message.name-vm_rebooted" xml:space="preserve">
     <value>VM Rebooted</value>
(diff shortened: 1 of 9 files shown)


XenModel/FriendlyNames.resx
```diff
@@ -1211,6 +1211,24 @@
   <data name="Message.name-alarm" xml:space="preserve">
     <value>Unrecognized alarm</value>
   </data>
+  <data name="Message.name-vm_migrated" xml:space="preserve">
```
kc284, line 1214: Judging by other existing entries `Message.body-vm...`, I think we also need friendly versions with place holders for the message body. For those that have snapshots it would be good if we could parse the body to get the snapshot name, although the format is not very developer friendly (normally for cases like this we ask toolstack to provide the message body in xml, json would be another good idea; if it's too problematic/unreliable we could have the format changed on the xapi side).

kc284 (changes requested):

XenModel/FriendlyNames.resx
danilo-delbusso: Currently, this is the body of the message as returned by toolstack:

image


Could you guide me a bit on how you want me to reword/reformat them? I felt like they were quite well done.

FYI not sure how to replicate the migration one. The alert does not appear after/during migration, in either of the hosts involved.

CitrixChris (approved):

AaronRobson (approved):

XenModel/FriendlyNames.resx
danilo-delbusso: I'm adding the `under review` label until this point is resolved

danilo-delbusso:
Force pushed to solve conflicts (+ rebase onto master)

XenModel/FriendlyNames.resx
kc284: Indeed the friendly names have a tendency to avoid showing the uuid on the UI. But I suppose it doesn't hurt to leave it in even if it doesn't match other friendly names. I'm thinking now it's probably worse to try parse a body that might change without warning in future. Ok, ignore this.

kc284 (approved):

AaronRobson (approved):

CitrixChris (approved):

danilo-delbusso:
force push was a rebase on master to resolve conflicts

PR #2929, opened 2022-01-11
Improved logging and reporting percentage for XVA import + some efficiency corrections

kc284:


CitrixChris (approved):

AaronRobson (approved):

XenModel/Actions/OvfActions/ExportApplianceAction.cs
```diff
@@ -163,8 +163,7 @@ void UpdatePercentage(float x)
                 ManifestAndSign(env, appFolder, appFile);
 			}
 
-			PercentComplete = 100;
-			Description = Messages.COMPLETED;
+            Tick(100, Messages.COMPLETED);
 		}
```
danilo-delbusso, line 167: nit:
We have previously taken advantage of indentation mismatches like this one to update the entire file to using spaces instead of tabs (see https://github.com/xenserver/xenadmin/pull/2907#discussion_r760899610)

It'd probably be better to do the same here. At least to keep GitHub from showing it like this

danilo-delbusso (approved):

PR #2933, opened 2022-01-18
CP-31587: Add the capability to Citrix Hypervisor Center to automatically update itself to the latest version

CitrixChris:


danilo-delbusso (changes requested):
1. The colour of the button has a slightly different background, which doesn't blend with the rest of the top bar
image
2. Hovering away from the `Update` button doesn't remove the selection outline, which gets removed only when hovering over other elements on the top bar:
ezgif-6-22ebbd0759

This doesn't happen when the button is enabled.

3. Typo at the end of this warning message:
image

Also I would avoid using a question. Something on the lines of "Please make sure you've completed and saved your work."

4. For this, could we say "CHC could not validate the certificate associated with the installation file". The user probably doesn't need to know it was an .msi. At a minimum I'd switch `msi` to `the installation file`.
image

5. After the above error is show, the installation `.msi` is still in `%localappdata%/Temp/`. Is this by design? I can see you doing this to allow users to check the file but I'm not sure this is on purpose

XenAdmin/Program.cs
```diff
@@ -997,7 +997,9 @@ public static void OpenURL(string url)
         /// </summary>
         public static bool ForcedExiting = false;
 
-        public static Version Version => Assembly.GetExecutingAssembly().GetName().Version;
+        //public static Version Version => Assembly.GetExecutingAssembly().GetName().Version;
+        public static Version Version => new Version(1, 0, 0, 0);//Assembly.GetExecutingAssembly().GetName().Version;
+
```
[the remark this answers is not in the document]
CitrixChris: Yes that is not meant to be in the final version

XenAdmin/MainWindow.cs
```diff
@@ -3309,5 +3329,27 @@ private void statusLabelErrors_Click(object sender, EventArgs e)
...
+            updateAlert.Dismiss();
+            Program.Invoke(this, () => { updateClientToolStripMenuItem.Enabled = false; });
+        }
+
+        private void downloadInstallToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            var downloadAndInstallClientAction = new DownloadAndUpdateClientAction(updateAlert.Name, new Uri(updateAlert.NewVersion.Url), Path.Combine(Path.GetTempPath(), $"{updateAlert.Name}.msi"), true, updateAlert.Checksum);
+
+            DialogResult dialogResult = MessageBox.Show("In order to update your client will be closed. Is all your work done and saved?.", "Are you ready to update?", MessageBoxButtons.YesNo);
+            // Only start if user says yes.
```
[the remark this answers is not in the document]
CitrixChris: Yup, realized this after i opened the PR and have yet to make that change. Good spot tho

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+            var valid = false;
+            // Check digital signature of .msi
+            using (var basicSigner = X509Certificate.CreateFromSignedFile(outputPathAndFileName))
+            {
+                using (var cert = new X509Certificate2(basicSigner))
+                {
+                    try
+                    {
+                        valid = cert.Verify();
+                        valid = false;
```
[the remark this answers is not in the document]
CitrixChris: Left over testing code, will remove

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+            {
+                updateDownloadState = DownloadState.Cancelled;
+                log.DebugFormat("XenServer patch '{0}' download cancelled by the user", updateName);
+                return;
+            }
+
+            if (e.Error != null) //failure
+            {
+                updateDownloadError = e.Error;
+                log.DebugFormat("XenServer patch '{0}' download failed", updateName);
```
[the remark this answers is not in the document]
CitrixChris: Very true, this is CHC client update, will correct

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+                var hash = StreamUtilities.ComputeHash(stream, out var hashAlgorithm);
+                
+                //Convert to Hex string
+                StringBuilder sb = new StringBuilder();
+                foreach (var b in hash)
+                    sb.Append(b.ToString("x2"));
+             
+                var calculatedChecksum = sb.ToString();
+                
+                // Check if calculatedChecksum matches what is in chcupdates.xml
```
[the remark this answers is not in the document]
CitrixChris: probably dont need the comment, it is fairly simple. But honestly I cant see the typo, what am I missing?

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+        //If you consider increasing this for any reason (I think 5 is already more than enough), have a look at the usage of SLEEP_TIME_BEFORE_RETRY_MS in DownloadFile() as well.
+        private const int MAX_NUMBER_OF_TRIES = 5;
+
+        private readonly Uri address;
+        private readonly string outputPathAndFileName;
+        private readonly string updateName;
+        private readonly bool downloadUpdate;
+        private DownloadState updateDownloadState;
+        private Exception updateDownloadError;
+        private string checksum;
```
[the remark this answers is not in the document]
CitrixChris: Right oh, this was made from an existing action but can make the improvements to keep it simple/standardised

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+                LogDescriptionChanges = true;
+
+                if (IsCompleted || Cancelled)
+                    return;
+
+                if (Cancelling)
+                    throw new CancelledException();
+            }
+            
+            ValidateMsi();
```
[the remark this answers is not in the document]
CitrixChris: Because we need error messages based on the checksum or the valid cert check on the method throws exceptions that then appear as error messages in the action progress dialog. Returning true false doesn't give enough information as to what went wrong. I could split the method but its highly possibly once  Matthews feed back is in that this check will change.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+            Description = Messages.COMPLETED;
+            MarkCompleted();
+        }
+
+        private void ValidateMsi()
+        {
+
+            using (FileStream stream = new FileStream(outputPathAndFileName, FileMode.Open, FileAccess.Read))
+            {
+                var hash = StreamUtilities.ComputeHash(stream, out var hashAlgorithm);
```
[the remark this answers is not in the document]
CitrixChris: Indeed, i had originally been using the hashAlgorithm variable but had since seen it was not needed. Twas misses in tidy up

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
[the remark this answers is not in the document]
danilo-delbusso: that's on me, not sure why I saw it. apologies 😅

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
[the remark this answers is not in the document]
danilo-delbusso: ok then let's leave it. I agree that it's more important to give better messages to the user

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,304 @@
...
+            this.outputPathAndFileName = outputFileName;
+            this.checksum = checksum;
+        }
+
+        private WebClient client;
+
+        private void DownloadFile()
+        {
+            int errorCount = 0;
+            bool needToRetry = false;
```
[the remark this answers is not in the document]
CitrixChris: Actually this isn't true, you need to assign primatives otherwise it wont compile.

XenAdmin/MainWindow.cs
```diff
@@ -2666,16 +2666,27 @@ private void Updates_CollectionChanged(CollectionChangeEventArgs e)
...
+            Program.Invoke(this, () =>
+            {
                 updateAlert = Updates.UpdateAlerts.FirstOrDefault(update => update is ClientUpdateAlert) as ClientUpdateAlert;
-                updateClientToolStripMenuItem.Enabled = updateAlert != null; 
+                DisableAndHideUpdateClientToolStripMenuItem(updateAlert != null);
             });            
         }
 
+        /// <summary>
+        /// Hides and disables the update client button
```
danilo-delbusso, line 2678: super-nit: technically not what it does, more like "Changes state and visibility of the update client button"

XenAdmin/MainWindow.cs
```diff
@@ -2666,16 +2666,27 @@ private void Updates_CollectionChanged(CollectionChangeEventArgs e)
...
         private void UpdatesCheck_Completed(bool suceeded, string err)
         {
-            
-            Program.Invoke(this, () => {
+
+            Program.Invoke(this, () =>
+            {
                 updateAlert = Updates.UpdateAlerts.FirstOrDefault(update => update is ClientUpdateAlert) as ClientUpdateAlert;
-                updateClientToolStripMenuItem.Enabled = updateAlert != null; 
+                DisableAndHideUpdateClientToolStripMenuItem(updateAlert != null);
```
danilo-delbusso, line 2673: Could we check for updates on startup? Currently, if there are no updates, the update client button is shown. Then it disappears once I manually check for updates.

If we don't want to check for updates on startup, we should at least hide the button by default in my opinion

XenModel/Messages.resx
```diff
@@ -13048,6 +13048,12 @@ Please check your settings and try again.</value>
   <data name="UPDATES_WIZARD_WLB_ON_WARNING" xml:space="preserve">
     <value>{0}: Check skipped because WLB is enabled on pool {1}.</value>
   </data>
+  <data name="UPDATE_CLIENT_CONFIRMATION_MESSAGE" xml:space="preserve">
+    <value>In order to update Citrix Hypervisor Center will be closed. Please make sure you've completed and saved your work.</value>
+  </data>
+  <data name="UPDATE_CLIENT_CONFIRMATION_MESSAGE_TITLE" xml:space="preserve">
+    <value>Update Citrix Hypervisor Center</value>
+  </data>
```
danilo-delbusso, line 13056: We should use `BrandingManager` to fetch the name of the product, like we do in other parts of the app. So here it should be `in order to update {0}` and `Update {0}`

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,300 @@
...
+                    }
+                    catch (Exception e)
+                    {
+                        throw new Exception("CHC could not validate the certificate associated with the installation file.", e);
+                    }
+                }
+            }
+
+            if (!valid)
+                throw new Exception("Invalid digital signature on msi.");
```
danilo-delbusso, line 256: This one should also be a message that we save in Messages.resx, since it is shown to the user on failure.

danilo-delbusso (changes requested):
I understand you might still be working on this so please ignore any redundant comments here.

Just reminding you about points 4 and 5 of my previous review, and [this comment](https://github.com/xenserver/xenadmin/pull/2933#discussion_r786895562)

XenModel/Messages.resx
CitrixChris: Did not know that was a thing, will do

XenAdmin/MainWindow.cs
CitrixChris: this might be a missed setting on initial creation that check at startup as we do check at astart up

kc284:
@CitrixChris this PR contains commits that should not be there - which is probably what is causing the conflicts.

CitrixChris:
I believe that the commits are as they should be now

kc284 (changes requested):
I have made some corrections and tidy up in https://github.com/CitrixChris/xenadmin/pull/1

One thing I have not changed, but I think we should re-work/re-think is that we should probably close the application before launching the installer, instead or relying on the installer to close it. For one thing, closing the application will check whether there are any tasks running and allow the user to wait or cancel them first. But we can do this in a subsequent PR.

XenAdmin/TabPages/ManageUpdatesPage.cs
```diff
@@ -830,11 +830,10 @@ private List<ToolStripItem> GetAlertActionItems(Alert alert)
                 items.Add(fix);
             }
 
-            if (alert is ClientUpdateAlert)
+            if (alert is ClientUpdateAlert cua)
```
AaronRobson, line 833: Nit: this `cua` variable doesn't appear to be used.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,292 @@
...
+                {
+                    // Start the install process, it will handle closing of application.
+                    Process.Start(outputPathAndFileName);
+                    log.DebugFormat("Update {0} found and install started", updateName);
+                }
+            }
+            catch (Exception e)
+            {
+                if (File.Exists(outputPathAndFileName))
+                    File.Delete(outputPathAndFileName);
```
AaronRobson, line 206: I don't think that `File.Delete` raises an exception if the file had already been deleted.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,292 @@
...
+                if (Cancelling)
+                    throw new CancelledException();
+            }
+            
+            ValidateMsi();
+
+            // Install the msi            
+            try
+            {
+                if (File.Exists(outputPathAndFileName))
```
AaronRobson, line 196: Should we error out and/or log at this point if the `outputPathAndFileName` wasn't found?

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,292 @@
...
+            using (FileStream stream = new FileStream(outputPathAndFileName, FileMode.Open, FileAccess.Read))
+            {
+                var calculatedChecksum = string.Empty; 
+
+                var hash = StreamUtilities.ComputeHash(stream, out _);
+                if (hash != null)
+                    calculatedChecksum = string.Join("", hash.Select(b => $"{b:x2}"));
+
+                // Check if calculatedChecksum matches what is in chcupdates.xml
+                if (checksum != calculatedChecksum)
```
AaronRobson, line 227: Will the passed `checksum` have the same casing as the calculated one?

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
CitrixChris: I dont see how it would as it only deletes if the file exists.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            bool valid;
+            // Check digital signature of .msi
+            using (var basicSigner = X509Certificate.CreateFromSignedFile(outputPathAndFileName))
+            {
+                using (var cert = new X509Certificate2(basicSigner))
+                {
+                    try
+                    {
+                        valid = cert.Verify();
```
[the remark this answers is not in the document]
kc284: No, because the variable is needed further down, outside the scope of this block.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
kc284: @AaronRobson is right, but I think wherever this check occurs in XenCenter, it is there as an extra safety measure against an argument exception that Delete will throw if the parameter should ever be empty or null. Does not harm to keep it.

danilo-delbusso (approved):

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,286 @@
...
+                    if (updateDownloadState == DownloadState.Error)
+                    {
+                        needToRetry = true;
+
+                        // this many errors so far - including this one
+                        errorCount++;
+
+                        // logging only, it will retry again.
+                        log.ErrorFormat(
+                            "Error while downloading from '{0}'. Number of errors so far (including this): {1}. Trying maximum {2} times.",
```
AaronRobson, line 131: 👍 for making it clear that the log message error counts are inclusive of the current one.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,286 @@
...
+                            address, errorCount, MAX_NUMBER_OF_TRIES);
+
+                        if (updateDownloadError == null)
```
AaronRobson, line 134: Nit: This `if`...`else` could be replaced with:
```csharp
log.Error(updateDownloadError ?? "An unknown error occurred.");
```

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -0,0 +1,286 @@
...
+            using (FileStream stream = new FileStream(outputPathAndFileName, FileMode.Open, FileAccess.Read))
+            {
+                var calculatedChecksum = string.Empty; 
+
+                var hash = StreamUtilities.ComputeHash(stream, out _);
+                if (hash != null)
+                    calculatedChecksum = string.Join("", hash.Select(b => $"{b:x2}"));
+
+                // Check if calculatedChecksum matches what is in chcupdates.xml
+                if (checksum.ToLower() != calculatedChecksum.ToLower())
```
AaronRobson, line 221: The culture specific casing rules can be unexpected sometimes, although hex digits are probably pretty safe.
If you want to be really paranoid about avoiding cultural differences in string comparison I'd recommend going for either:
```csharp
if (checksum.ToLowerInvariant() != calculatedChecksum.ToLowerInvariant())
```
Or
```csharp
if (!checksum.Equals(calculatedChecksum, StringComparison.InvariantCultureIgnoreCase))
```
But I wouldn't block the PR on it.

AaronRobson (approved):

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
AaronRobson: See also: http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
CitrixChris: except updateDownloadError is not a string but an exception so not comparable

PR #2935, opened 2022-01-24
CA-362709, CA-362711: Reworked the logic for allowing full disk copy …

kc284:
…when creating a VM from a snapshot or custom template.

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
+++ b/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
@@ -53,26 +53,19 @@ namespace XenAdmin.Wizards.NewVMWizard
         private Host _affinity;
         private bool _loadRequired = true;
         private bool loading;
+        private SR _fullCopySR;
+        private bool _canCreateFullCopy;
 
         public Page_Storage()
         {
             InitializeComponent();
         }
 
-        public override string Text
-        {
-            get { return Messages.NEWVMWIZARD_STORAGEPAGE_NAME; }
-        }
+        public override string Text => Messages.NEWVMWIZARD_STORAGEPAGE_NAME;
 
-        public override string PageTitle
-        {
-            get { return Messages.NEWVMWIZARD_STORAGEPAGE_TITLE; }
-        }
+        public override string PageTitle => Messages.NEWVMWIZARD_STORAGEPAGE_TITLE;
 
-        public override string HelpID
-        {
-            get { return "Storage"; }
-        }
+        public override string HelpID => "Storage";
 
         protected override void PageLoadedCore(PageLoadedDirection direction)
         {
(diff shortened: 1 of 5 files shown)


danilo-delbusso (commented):
I have tested all (non physical/virtual related) scenarios for VM created with a custom template (snapshot one incoming, also see tickets for more info on the scenarios I have tested).

Only sticking one I want to double check is:

CoreOS template:
2xSRs (both on Local Storage SR), Local Storage SR is full

New VM:
When opening the Storage page of the New VM wizard, we find both SRs have been substituted for other options (as Local Storage is full, and can't be used for creating a New VM).

However, we expect the Fast Clone checkbox to be disabled and checked (since both destination SRs here are not the original ones), but it is **enabled** and checked when testing.

I may have misinterpreted what the default checkbox option should be so feel correct me.

image

kc284:
First of all, the wizard by design will try placing the new VM's disks on the pool's default SR and only look for alternatives if the default is full. On the above screenshot, it looks like the default is an iSCSI, but it's probably full, hence it picked the next available, which is NFS (please confirm that on hovering the tooltip implies that the warning is shown because the default iSCSI in unavailable).

The major change of this PR is how we decide whether full copy is available to default-template/custom-template/snapshot. Previously for custom templates and snapshots it was a bit arbitrary, now it depends on whether the disks support the operation.

The logic of picking the target SR for full copy has not changed that much, except adding the space check, so:
- if both of the new VM's disks are on the same target SR and it has space, full copy is available, hence the checkbox is enabled.
- if the new VM's disks are on different target SRs, full copy will be available if one these SRs is the same as the source SR where the template's disks live, and has space.
On either case, if the checkbox is enabled we check it because clone is preferable for being faster, but the user can opt for full copy instead. If full copy is not possible, then the checkbox is disabled and obviously checked since we'll be cloning.
I think the above should match what you see?

danilo-delbusso:
> please confirm that on hovering the tooltip implies that the warning is shown because the default iSCSI in unavailable

Yes that's what it says 👍 

__________________

As for the rest of the comment, thank you for the clarifications. It all makes sense now. I was suspicious that I had missed something in the requirements and it seems like I did.

Since this was the only sticking point for me (i.e. why the checkbox state changed when we had different vs same target SRs), I'll approve and update the test results on the ticket.

danilo-delbusso (approved):

XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
```diff
@@ -412,15 +434,22 @@ private void EditButton_Click(object sender, EventArgs e)
...
+            UpdateStatusForEachDisk();
+            UpdateEnablement();
         }
 
 
         private void DisksGridView_SelectionChanged(object sender, EventArgs e)
         {
-            UpdateEnablement();
+            EditButton.Enabled = DisksGridView.SelectedRows.Count > 0;
+            DeleteButton.Enabled = DisksGridView.SelectedRows.Count > 0 && ((DiskGridRowItem)DisksGridView.SelectedRows[0]).CanDelete;
```
AaronRobson, line 447: Can we have `SelectedRows.Count` of 2 or higher? If so please check the `CanDelete` falg for all of them.

kc284: The DataGridView has MultiSelect=false, however, this is not a bad point, we can make the code safer by checking Count==1 rather than >0.

kc284: Updated.

CitrixChris (approved):

PR #2938, opened 2022-01-28
CP-36392: Fix miscellaneous code smells

danilo-delbusso:
Best reviewed one commit at a time.

This change set is quite small, but I wanted to sneak it in.

There might be conflicts as we merge other PRs, because of the `using` commit. I'll solve and rebase when necessary. 

Also, I'm not 100% sure the solution in d5b7e2266c8101c5782ffc0c06b028ab3ca2de30 is the best one, so feel free to weigh in.

### DIFF
--- a/XenAdmin/Controls/VerticalTabs.cs
+++ b/XenAdmin/Controls/VerticalTabs.cs
@@ -33,6 +33,7 @@ using System;
 using System.Windows.Forms;
 using System.Drawing;
 using System.Drawing.Drawing2D;
+using System.Reflection;
 using XenAdmin.Core;
 
 
@@ -46,24 +47,13 @@ namespace XenAdmin.Controls
             string SubText { get; }
             Image Image { get; }
         }
-
         public VerticalTabs()
         {
-            base.ItemHeight = 40;
-            IntegralHeight = false; 
-            DrawMode = DrawMode.OwnerDrawFixed;
+            IntegralHeight = false;
         }
 
-        public override int ItemHeight
-        {
-            get
-            {
-                return base.ItemHeight;
-            }
-            set
-            {
-            }
-        }
+        public override int ItemHeight => 40;
+        public override DrawMode DrawMode => DrawMode.OwnerDrawFixed;
 
         public Func<Rectangle, Rectangle> AdjustItemTextBounds;
 
--- a/XenAdmin/Commands/Command.cs
+++ b/XenAdmin/Commands/Command.cs
@@ -32,12 +32,9 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Drawing;
-using System.Linq;
 using System.Windows.Forms;
 using XenAdmin.Actions;
-using XenAdmin.Core;
 using XenAdmin.Dialogs;
-using XenAdmin.Network;
 using XenAPI;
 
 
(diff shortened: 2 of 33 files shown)


kc284:
Regarding the Dispose() fixes, might it not be a better if those classes actually implemented IDisposable?

Regarding the empty setters, do we need to throw the exceptions within DEBUG directives? What if we implement them and ensure we don't try to set them anywhere.

danilo-delbusso:
> Regarding the Dispose() fixes, might it not be better if those classes actually implemented IDisposable?

Sure I'll look into it.

> Regarding the empty setters, do we need to throw the exceptions within DEBUG directives? What if we implement them and ensure we don't try to set them anywhere.

My opinion is that we should try to enforce a restriction like this. Now, I'm not sure why this `set { } ` pattern was added to these properties, but I think we shouldn't add a potential source of issues just because SonarQube complains about the `value` not being present in the setter. I'm also happy to flag the corresponding smells as "won't fix". The idea behind throwing an exception while in `DEBUG` is that whatever developer is using the setter will hopefully notice while making the changes.

AaronRobson (dismissed):

CitrixChris (approved):

XenAdmin/Controls/VerticalTabs.cs
```diff
@@ -48,17 +48,32 @@ public interface IVerticalTab
...
             DrawMode = DrawMode.OwnerDrawFixed;
         }
 
-        public override int ItemHeight
+        public sealed override int ItemHeight
         {
             get => base.ItemHeight;
-            set => throw new NotSupportedException($"Invalid set call for '{MethodBase.GetCurrentMethod()?.Name}', value: '{value}'");
+            set
+            {
```
kc284, line 64: Isn't this more complicated than the original code? It even qualifies for a comment.

danilo-delbusso: that's true but it uses `value`, which is the original issue brought up by SonarQube.

The only other options are to:
1. Disable the rule
2. Review every instance of the rule being broken in the codebase, and flag it as "won't fix" on SonarQube

AaronRobson (approved):

XenAdmin/Controls/VerticalTabs.cs
kc284: Why don't we just throw the exception? Plus that, if the idea is not to use the setter, line 55 will cause confusion.

danilo-delbusso: Even with this format:
```csharp
public VerticalTabs()
        {
            IntegralHeight = false; 
            DrawMode = DrawMode.OwnerDrawFixed;
        }

        public sealed override int ItemHeight
        {
            get => DEFAULT_ITEM_HEIGHT;
            set => throw new NotSupportedException($"Invalid set call for '{MethodBase.GetCurrentMethod()?.Name}', value: '{value}'");
            
        }
```

The exception is still thrown:
image

(Open `Tools > Options`)

You're right, the line 55 is not necessary since that's the default `ItemHeight`, anyway. I can remove it

kc284: I think we should just remove setting the virtual members from the constructor and use the a partial override instead:
```
public override int ItemHeight => 40;
public override DrawMode DrawMode => DrawMode.OwnerDrawFixed;
```

danilo-delbusso: sounds good to me. Will do shortly

kc284 (approved):

PR #2939, opened 2022-01-28
XSI-1177/CA-362931: A number of new version pre-checks did not run when in automated updates mode.

kc284:


danilo-delbusso (approved):
Pre-checks successfully block update on both update methods while updating 8.2.0 to 8.2 LTSR (with 1 PV guest on one of the hosts in a pool).

CitrixChris (approved):

PR #2940, opened 2022-01-29
CP-38893: Make internal roles unavailable for selection. Some refactoring of Role extension method overloads.

kc284:
Commits best viewed separately.


AaronRobson (approved):

danilo-delbusso (approved):
Works both in 8.2.0 and 8.2.50 with toolstack changes. 👍

PR #2942, opened 2022-02-04
CP-38526/XSI-1055: Improved import performance for OVF/OVA packages with .vhd disks.

kc284:


XenModel/Actions/OvfActions/Import.cs
```diff
@@ -405,18 +414,18 @@ private XenRef<VDI> ImportFile(string diskName, string pathToOvf, string filenam
                     Host = Connection.Hostname,
                     Port = Connection.Port,
                     Path = "/import_raw_vdi",
-                    Query = string.Format("session_id={0}&task_id={1}&vdi={2}",
-                        Connection.Session.opaque_ref, taskRef.opaque_ref, vdiuuid)
+                    Query = string.Format("session_id={0}&task_id={1}&vdi={2}&format={3}",
```
danilo-delbusso, line 417: Looks like we'd be passing an empty query parameter here if the disk isn't `vhd`. Is that by design?

kc284: Yes, in this case it is imported as a raw vdi.

danilo-delbusso (approved):

CitrixChris (approved):

PR #2944, opened 2022-02-09
CP-38542: Remove option to disable live patching.

kc284:


danilo-delbusso (approved):

CitrixChris (approved):

kc284:
I rebased on master because of conflicts. Re-review is not necessary as the conflicting file was LivePatchingEditPage.ja.resx which is being removed.

PR #2953, opened 2022-02-17
Update alert test fixes

CitrixChris:


XenAdminTests/UnitTests/AlertTests/AlertClassUnitTestData - Copy.cs
```diff
@@ -0,0 +1,38 @@
+﻿/* Copyright (c) Citrix Systems, Inc.
```
kc284, line 1: Generally we would want a file named `UpdateAlertClassUnitTestData.cs` rather than `AlertClassUnitTestData - Copy.cs`, but because it is such a small class I'd just put it in the same file as `AlertClassUnitTestData.cs`, it keeps the solution size smaller and makes navigation easier for the developer.

kc284 (changes requested):

AaronRobson (approved):

AaronRobson (approved):

kc284 (approved):

PR #2957, opened 2022-02-21
(master) CA-364152 + a regression in DR

kc284:


danilo-delbusso (approved):
tested with 1 disconnected host

danilo-delbusso (approved):

XenModel/Actions/ParallelAction.cs
```diff
@@ -175,6 +175,9 @@ void EnqueueAction(AsyncAction action, ProduceConsumerQueue queue, List<Exceptio
 
         protected override void RecalculatePercentComplete()
         {
+            if (actionsCount == 0)
+                return;
```
AaronRobson, line 179: Would it make sense to make the `PercentComplete` to be `100` or something in this case?
(as I said mentioned the other PR: https://github.com/xenserver/xenadmin/pull/2956#discussion_r812805961)

AaronRobson (approved):

XenModel/Actions/ParallelAction.cs
kc284: It shouldn't be needed, the base class will set it to 100 on completion, which will be immediately since there are no sub-actions.

PR #2959, opened 2022-03-02
Changes the confimation dialog to one that actually shows background …

CitrixChris:
…tasks

Signed-off-by: Christophe25 <christopher.lancaste1@citrix.com>

### DIFF
--- a/XenAdmin/Alerts/Types/ClientUpdateAlert.cs
+++ b/XenAdmin/Alerts/Types/ClientUpdateAlert.cs
@@ -34,6 +34,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Windows.Forms;
 using XenAdmin.Actions;
+using XenAdmin.Actions.GUIActions;
 using XenAdmin.Core;
 using XenAdmin.Dialogs;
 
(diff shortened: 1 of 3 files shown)


XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -106,15 +107,24 @@ public override bool Equals(Alert other)
 
         public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IWin32Window parent)
         {
-            using (var dialog = new WarningDialog(
-                           string.Format(Messages.UPDATE_CLIENT_CONFIRMATION_MESSAGE, BrandManager.BrandConsole),
```
danilo-delbusso, line 110: We should also remove `UPDATE_CLIENT_CONFIRMATION_MESSAGE_TITLE` and `UPDATE_CLIENT_CONFIRMATION_MESSAGE` from `Messages.resx`. Doesn't seem they're used anywhere else now:
image

danilo-delbusso (changes requested):
Works great! Just one minor issue.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
CitrixChris: sure thing

danilo-delbusso (approved):

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -106,15 +107,24 @@ public override bool Equals(Alert other)
 
         public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IWin32Window parent)
         {
-            using (var dialog = new WarningDialog(
-                           string.Format(Messages.UPDATE_CLIENT_CONFIRMATION_MESSAGE, BrandManager.BrandConsole),
-                           ThreeButtonDialog.ButtonOK, ThreeButtonDialog.ButtonCancel)
-                       {WindowTitle = string.Format(Messages.UPDATE_CLIENT_CONFIRMATION_MESSAGE_TITLE, BrandManager.BrandConsole)})
+            bool currentTasks = false;
```
kc284, line 110: If we went with this solution, we would also need to modify slightly the blurb and buttons of the CloseXenCenterWarningDialog to mention that we're exiting because we're updating.

However, a more important issue I see is that launching this dialog should be the last thing to happen before exiting, which is not the case because we haven't downloaded the update yet. On a slow network this might take a while and it's not certain that no other actions will start in the meantime, plus that during that interval the user might have the time to wait for their actions to finish instead of cancelling them (even if no one is forcing them to do so).

Maybe we should leave the original warning dialog as it was, and if we make the change I'm mentioning in the other comment, check and launch the CloseXenCenterWarningDialog (and then we don't need to change any wording) in that block immediately before calling the msi process as a last line of defence.

kc284 (changes requested):

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
danilo-delbusso: > Maybe we should leave the original warning dialog as it was, and if we make the change I'm mentioning in the other comment, check and launch the CloseXenCenterWarningDialog (and then we don't need to change any wording) in that block immediately before calling the msi process as a last line of defence.

I agree that this is the best option, as users might find frustrating having to start over just because they forgot to close a couple of actions.

kc284 (changes requested):
I've put some corrections here https://github.com/CitrixChris/xenadmin/pull/3

kc284 (approved):

PR #2961, opened 2022-03-02
CP-39142: Add RPU pre-check in CHCenter to warn of DMC removal. Renamed some methods.

kc284:

### DIFF
--- a/XenAdmin/Diagnostics/Checks/DmcCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/DmcCheck.cs
@@ -1,0 +1,103 @@
+﻿/* Copyright (c) Citrix Systems, Inc. 
+ * All rights reserved. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer. 
+ * *   Redistributions in binary form must reproduce the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer in the documentation and/or other 
+ *     materials provided with the distribution. 
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+ * CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR 
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR 
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, 
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING 
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE 
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF 
+ * SUCH DAMAGE.
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Windows.Forms;
+using XenAdmin.Core;
+using XenAdmin.Diagnostics.Hotfixing;
+using XenAdmin.Diagnostics.Problems;
+using XenAdmin.Diagnostics.Problems.HostProblem;
+using XenAdmin.Diagnostics.Problems.PoolProblem;
+using XenAPI;
+
+namespace XenAdmin.Diagnostics.Checks
+{
+    internal class DmcCheck : PoolCheck
+    {
+        private readonly Pool _pool;
+        private readonly Dictionary<string, string> _installMethodConfig;
+        private readonly bool _manualUpgrade;
+        private readonly Control _control;
+
+        public DmcCheck(Control control, Pool pool, Dictionary<string, string> installMethodConfig, bool manualUpgrade)
+            : base(pool)
+        {
+            _control = control;
+            _pool = pool;
+            _installMethodConfig = installMethodConfig;
+            _manualUpgrade = manualUpgrade;
+        }
+
+        public override bool CanRun()
+        {
+            return !Helpers.Post82X(_pool.Connection) && !Helpers.FeatureForbidden(_pool.Connection, Host.RestrictDMC);
+        }
+
+        protected override Problem RunCheck()
+        {
+            var coordinator = Helpers.GetCoordinator(Pool.Connection);
+
+            if (!_manualUpgrade)
+            {
+                var hotfix = HotfixFactory.Hotfix(coordinator);
+                if (hotfix != null && hotfix.ShouldBeAppliedTo(coordinator))
+                    return new HostDoesNotHaveHotfixWarning(this, coordinator);
+            }
+
+            var vms = _pool.Connection.Cache.VMs.Where(v =>
+                (v.power_state == vm_power_state.Running ||
+                 v.power_state == vm_power_state.Paused ||
+                 v.power_state == vm_power_state.Suspended) &&
+                (v.memory_static_min > v.memory_dynamic_min || //corner case, probably unlikely
+                 v.memory_dynamic_min != v.memory_dynamic_max ||
+                 v.memory_dynamic_max != v.memory_static_max)).ToList();
+
+            if (vms.Count == 0)
+                return null;
+
+            string upgradePlatformVersion = null;
+
+            if (_installMethodConfig != null)
+                Host.TryGetUpgradeVersion(coordinator, _installMethodConfig, out upgradePlatformVersion, out _);
+
+            if (string.IsNullOrEmpty(upgradePlatformVersion))
+                return new PoolHasVmsWithDmcWarning(_control, this, _pool, vms);
+                
+            if (Helpers.Post82X(upgradePlatformVersion))
+                return new PoolHasVmsWithDmcProblem(_control, this, _pool, vms);
+
+            return null;
+        }
+
+        public override string Description => Messages.DMC_CHECK_ENABLED;
+    }
+}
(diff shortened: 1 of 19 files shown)


XenAdmin/Wizards/BallooningWizard_Pages/MemorySettings.cs
```diff
@@ -64,7 +65,11 @@ protected override void PageLoadedCore(PageLoadedDirection direction)
             if (!_changed)
                 return;
 
-            if (VMs != null && VMs.Count > 0 && VMs[0].advanced_ballooning())
+            var connection = VMs?.FirstOrDefault()?.Connection;
+            var ballooningInUse = VMs?.FirstOrDefault()?.UsesBallooning();
```
AaronRobson, line 69: Do we not need to check if Any VMs have `UsesBallooning` set?

XenAdmin/Diagnostics/Checks/DmcCheck.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            var vms = _pool.Connection.Cache.VMs.Where(v =>
+                (v.power_state == vm_power_state.Running ||
+                 v.power_state == vm_power_state.Paused ||
+                 v.power_state == vm_power_state.Suspended) &&
+                (v.memory_static_min > v.memory_dynamic_min || //corner case, probably unlikely
+                 v.memory_dynamic_min != v.memory_dynamic_max ||
+                 v.memory_dynamic_max != v.memory_static_max)).ToList();
+
+            if (vms.Count == 0)
```
AaronRobson, line 84: Nit: I've seen recommendations (from Resharper or SonarQube I forget which) that this is better:
```csharp
if (!vms.Any())
```

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeWizardPrecheckPage.cs
```diff
@@ -247,6 +247,15 @@ where check.CanRun()
             if (iloChecks.Count > 0)
                 groups.Add(new CheckGroup(Messages.CHECKING_POWER_ON_MODE_GROUP, iloChecks));
 
+            //Checking DMC
+            var dmcChecks = (from Pool pool in SelectedPools
+                let check = new DmcCheck(this, pool, InstallMethodConfig, ManualUpgrade)
+                where check.CanRun()
+                select check as Check).ToList();
+
+            if (dmcChecks.Count > 0)
```
AaronRobson, line 256: Nit: Similar here:
```csharp
if (dmcChecks.Any())
```

XenAdmin/Diagnostics/Checks/DmcCheck.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        protected override Problem RunCheck()
+        {
+            var coordinator = Helpers.GetCoordinator(Pool.Connection);
+
+            if (!_manualUpgrade)
+            {
+                var hotfix = HotfixFactory.Hotfix(coordinator);
+                if (hotfix != null && hotfix.ShouldBeAppliedTo(coordinator))
+                    return new HostDoesNotHaveHotfixWarning(this, coordinator);
```
danilo-delbusso, line 73: Question:
Why are we not also checking if supporter hosts need an hotfix?

XenAdmin/Diagnostics/Checks/DmcCheck.cs
```diff
@@ -0,0 +1,103 @@
...
+            string upgradePlatformVersion = null;
+
+            if (_installMethodConfig != null)
+                Host.TryGetUpgradeVersion(coordinator, _installMethodConfig, out upgradePlatformVersion, out _);
+
+            if (string.IsNullOrEmpty(upgradePlatformVersion))
+                return new PoolHasVmsWithDmcWarning(_control, this, _pool, vms);
+                
+            if (Helpers.StockholmOrGreater(upgradePlatformVersion))
+                return new PoolHasVmsWithDmcProblem(_control, this, _pool, vms);
```
danilo-delbusso, line 96: Should this check be `Helpers.Post82X`?

XenAdmin/Diagnostics/Problems/PoolProblem/PoolHasVmsWithDmcProblem.cs
```diff
@@ -0,0 +1,146 @@
...
+using XenAdmin.Actions;
+using XenAdmin.Commands;
+using XenAdmin.Core;
+using XenAdmin.Diagnostics.Checks;
+using XenAdmin.Dialogs;
+using XenAPI;
+
+namespace XenAdmin.Diagnostics.Problems.PoolProblem
+{
+    class PoolHasVmsWithDmcProblem : ProblemWithMoreInfo
```
danilo-delbusso, line 45: Nit-picky code style ReSharper suggestions (since it's a new file I'd say let's keep things consistent):
- both problem classes can be `internal`
- `_control` should be named `Control` since it's not private

XenAdmin/Diagnostics/Problems/PoolProblem/PoolHasVmsWithDmcProblem.cs
```diff
@@ -0,0 +1,146 @@
...
+                string.Format(Messages.STRING_SPACE_STRING, BrandManager.ProductBrand, BrandManager.ProductVersionPost82),
+                BrandManager.BrandConsole);
+
+        protected override AsyncAction CreateAction(out bool cancelled)
+        {
+            Program.Invoke(_control, delegate
+            {
+                using (var dlg = new ErrorDialog(Message,
+                           new ThreeButtonDialog.TBDButton(Messages.DMC_DISABLE, DialogResult.OK, ThreeButtonDialog.ButtonType.ACCEPT, true),
+                           ThreeButtonDialog.ButtonCancel)
```
danilo-delbusso, line 76: I'd set `Cancel` as the default button since the action shuts down VMs.

image
```suggestion
                using (var dlg = new ErrorDialog(Message,
                           new ThreeButtonDialog.TBDButton(Messages.DMC_DISABLE, DialogResult.OK, ThreeButtonDialog.ButtonType.ACCEPT),
                           new ThreeButtonDialog.TBDButton(Messages.CANCEL, DialogResult.Cancel, ThreeButtonDialog.ButtonType.CANCEL, true))
```

or at least remove pre-selection

```suggestion
                using (var dlg = new ErrorDialog(Message,
                           new ThreeButtonDialog.TBDButton(Messages.DMC_DISABLE, DialogResult.OK, ThreeButtonDialog.ButtonType.ACCEPT),
                           ThreeButtonDialog.ButtonCancel)
```

danilo-delbusso (changes requested):
Warning/Blocking Warning show as expected on manual/automated RPUs from 8.2 Host ✔️

XenAdmin/Wizards/BallooningWizard_Pages/MemorySettings.cs
kc284: FirstOrDefault() is essentially the same as Any(), the difference being that it also returns the first item found rather than just that it has been found. But I suppose Any() makes for more elegant code here - I'll change it 👍

XenAdmin/Diagnostics/Checks/DmcCheck.cs
kc284: Any() is definitely preferable to linq's Count() method, so I expect a warning for this, but does it warn for the Count property too? The latter is more efficient as it does not recalculate.

XenAdmin/Wizards/BallooningWizard_Pages/MemorySettings.cs
kc284: I've just realised that my comment above was not relevant to what you meant. My reply was about `VMs?.FirstOrDefault(v=>v.UsesBallooning())` vs `VMs?.Any(v=>v.UsesBallooning())`. 

`VMs?.FirstOrDefault()?.UsesBallooning()` is slightly better that the original `VMs[0].UsesBallooning` code, but I'm afraid I don't know why the original code was only checking for the first VM, I agree that `VMs?.Any(v=>v.UsesBallooning())` makes more sense.

kc284: It is likely that `VMs?.All(v=>v.UsesBallooning())` is the most correct answer of all. But given the uncertainty that exists in this area, I think I'll leave it as is for now because it is closer to the original code. I'll revisit this within CP-39141, because this correction is outside the scope of CP-39142.

XenAdmin/Diagnostics/Checks/DmcCheck.cs
kc284: 🤦 😵

XenAdmin/Diagnostics/Checks/DmcCheck.cs
kc284: Further down Host.TryGetUpgradeVersion needs one host as a parameter, so the coordinator will do. Note though that when the RPU hotfix is applied, it is installed on the supporters too.

XenAdmin/Diagnostics/Problems/PoolProblem/PoolHasVmsWithDmcProblem.cs
kc284: Internal is the default if nothing is specified (see https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers). But I agree that it's better to have it explicitly written.

XenAdmin/Diagnostics/Problems/PoolProblem/PoolHasVmsWithDmcProblem.cs
```diff
@@ -0,0 +1,146 @@
...
+                       {
+                           LinkText = Messages.LEARN_MORE,
+                           LinkData = InvisibleMessages.DEPRECATION_URL,
```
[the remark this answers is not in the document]
kc284: 🤦

XenAdmin/Diagnostics/Checks/DmcCheck.cs
danilo-delbusso: Could there be an instance where the coordinator has a hotfix, but not the supporters?

danilo-delbusso (approved):
🚀 LGTM

XenAdmin/Diagnostics/Checks/DmcCheck.cs
kc284: Now that I'm thinking about it, this check derives from PoolCheck, not HostCheck - I need to double check as I'm not sure any more whether it will be applied to all hosts. If there are further down HostChecks that need it though, it will be flagged by the pre-checks.

kc284: I had a look at the code again: so this is a waring, which means it does not provide an action to install it. However, there is the  HostHasHotfixCheck at the beginning of the prechecks, which will show a blocking error for all the hosts to be upgraded and will offer the installation action. After that, the only way we could end up with a supporter not having the hotfix would be if installation failed or if a supporter was added to the pool after installation.

danilo-delbusso: Sounds good, Thank you for looking into it!

XenAdmin/Diagnostics/Checks/DmcCheck.cs
AaronRobson: Fair point, yes the efficiency of linq's `Count()` was what I was concerned with.
I will mark the thread as resolved.

AaronRobson (approved):

PR #2962, opened 2022-03-02
CP-38895: Check version annotations when using API fields or methods.

kc284:


danilo-delbusso (approved):
Looks good and I've tested a couple of commands. FQP should be taking care of the rest. ✔️

AaronRobson (approved):

PR #2963, opened 2022-03-08
CA-364273: Do not log out the current user if they cancel the operation from the elevated session dialog.

kc284:


danilo-delbusso (approved):
Tested with CH 8.2

CitrixChris (approved):

PR #2972, opened 2022-03-17
CA-365297: Correction when changing a VM's memory settings:

kc284:
- Corrected the logic deducing whether shutting down and restart is needed.
- Fixed crash when starting a VM without a home host.
- Also, renamed some fields for better compliance with C# naming conventions.

Note that when the VM is suspended or paused, its power state is not restored after the shutdown, but the VM is restarted. This is an older issue, but I would like to avoid fixing it now as it involves more changes and more testing. So I'm only addressing what's in the cited ticket.


kc284:
Ooops, got the tickets wrong. Closing this PR.

PR #2973, opened 2022-03-17
CP-39382: Lock msi file while verifying and launching install process

CitrixChris:


CitrixChris:
This should be merged after #2959

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -211,24 +212,27 @@ public void InstallMsi()
                 log.Error("Exception occurred when starting the installation process.", e);
                 throw;
             }
+            finally
+            {
+                MsiStream.Dispose();
```
danilo-delbusso, line 217: Nit: To avoid refactoring errors, I'd do `MsiStream?.Dispose()`

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -190,14 +191,14 @@ protected override void Run()
...
             if (!File.Exists(outputPathAndFileName))
                 throw new Exception(Messages.DOWNLOAD_CLIENT_INSTALLER_MSI_NOT_FOUND);
 
             // Install the msi            
             try
             {
+                ValidateMsi();
                 // Start the install process, it will handle closing of application.
                 Process.Start(outputPathAndFileName);
                 log.DebugFormat("Update {0} found and install started", updateName);
```
danilo-delbusso, line 204: Nit, not related to this PR. To avoid refactoring errors, I'd do `OnInstall?.Invoke()`

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -61,7 +61,11 @@ public class DownloadAndUpdateClientAction : AsyncAction, IByteProgressAction
         private string checksum;
         private WebClient client;
```
danilo-delbusso, line 62: very nit: if you're making other changes, could you rename all private fields here using the `_variableName` convention?

This will also allow you to remove `this.` calls in the constructor
image

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -106,21 +107,49 @@ public override bool Equals(Alert other)
...
             {
-                dlg.ShowDialog(parent);
+                if (new Dialogs.WarningDialogs.CloseXenCenterWarningDialog(true).ShowDialog() != DialogResult.OK)
+                {
+                    action.Cancel();
+                    return;
+                }
             }
+
+            action.InstallMsi();
```
danilo-delbusso, line 147: nit, for avoiding issues in the future, we should null-check `action`.

```csharp
if (sender is DownloadAndUpdateClientAction action)
            {
                if (currentTasks)
                {
                    if (new Dialogs.WarningDialogs.CloseXenCenterWarningDialog(true).ShowDialog() != DialogResult.OK)
                    {
                        action.Cancel();
                        return;
                    }
                }
                action.InstallMsi();
            }
```

danilo-delbusso (dismissed):
Works great! Just some style related comments
image

One related comment + question. There still is a time gap between when the file is downloaded and when the read lock is acquired. It's only two "steps" before, so I think it's a reasonable compromise.
However, I was wondering if you could acquire the lock before even checking if the file exists.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
danilo-delbusso: can be rewritten slightly, but you get the gist 😅

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
CitrixChris: Is that within the ValidateMsi function?

CitrixChris:
> 

Yes indeed there is a gap and this is fine, the lock just needs to be in place between verifying -> installing. It doesnt matter if the lock is earlier but it cant be after its verified. It just needs to exists while verifying and launching.

danilo-delbusso:
> > One related comment + question. There still is a time gap between when the file is downloaded and when the read lock is acquired. It's only two "steps" before, so I think it's a reasonable compromise.
> > However, I was wondering if you could acquire the lock before even checking if the file exists.
> 
> @danilo-delbusso Yes indeed there is a gap and this is fine, the lock just needs to be in place between verifying -> installing. It doesnt matter if the lock is earlier but it cant be after its verified. It just needs to exists while verifying and launching.

sounds good, thanks for looking into it! I will remove the label and await the rebase before I review

kc284 (changes requested):
Something has gone wrong with the rebase, it still contains commits it should not.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -181,35 +184,56 @@ protected override void Run()
...
-            }
+            var hash = StreamUtilities.ComputeHash(MsiStream, out _);
+            if (hash != null)
+                calculatedChecksum = string.Join("", hash.Select(b => $"{b:x2}"));
+
+            // Check if calculatedChecksum matches what is in chcupdates.xml
+            if (!_checksum.Equals(calculatedChecksum, StringComparison.InvariantCultureIgnoreCase))
+                throw new Exception(Messages.UPDATE_CLIENT_INVALID_CHECKSUM );
 
             bool valid = false;
```
danilo-delbusso, line 232: _**very**_ nit: you don't need to initialise `bool valid = false`. `bool valid;` works too since you're not using the `false` value in any execution path.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -133,28 +133,14 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
                 break;
             }
 
-            if (currentTasks)
+            if (!currentTasks)
```
danilo-delbusso, line 136: I think this one is supposed to be `if(currentTasks)`

This way we're prompted even when there are no actions:
image

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -51,36 +53,37 @@ public class DownloadAndUpdateClientAction : AsyncAction, IByteProgressAction
...
-        private Exception updateDownloadError;
-        private string checksum;
-        private WebClient client;
+        private readonly Uri _address;
+        private readonly string _outputPathAndFileName;
+        private readonly string _updateName;
+        private readonly bool _downloadUpdate;
+        private DownloadState _updateDownloadState;
+        private Exception _updateDownloadError;
+        private string _checksum;
```
danilo-delbusso, line 62: nit: `_checksum` can be readonly

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -31,9 +31,11 @@
 
 using System;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Net.Mime;
```
danilo-delbusso, line 38: `using System.Net.Mime;` isn't used

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -31,9 +31,11 @@
```
danilo-delbusso, line 31: nit: the class should probably be in the `XenAdmin.Actions.Updates` namespace. I know this is wildly inconsistent in the rest of the codebase, but it'd be nice to start keeping track of this when editing files

The same applies for `ClientUpdateAlert.cs`

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -181,35 +184,56 @@ protected override void Run()
...
-                    calculatedChecksum = string.Join("", hash.Select(b => $"{b:x2}"));
+            var calculatedChecksum = string.Empty; 
 
-                // Check if calculatedChecksum matches what is in chcupdates.xml
-                if (!checksum.Equals(calculatedChecksum, StringComparison.InvariantCultureIgnoreCase))
-                    throw new Exception(Messages.UPDATE_CLIENT_INVALID_CHECKSUM );
-            }
+            var hash = StreamUtilities.ComputeHash(MsiStream, out _);
+            if (hash != null)
+                calculatedChecksum = string.Join("", hash.Select(b => $"{b:x2}"));
```
danilo-delbusso, line 226: **very nit:** can use `string.Empty`, if you're making other changes

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -133,28 +133,14 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
                 break;
```
danilo-delbusso, line 133: `using System.Diagnostics;` isn't used (GH doesn't let me comment on that line 🤦 )

danilo-delbusso (changes requested):
Mostly minor style comments.

I have tried out the lock and it seems to behave correctly. No writing permissions

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
CitrixChris: woops that was a bit of test  cdoe thanks for the spot

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
CitrixChris: While i dont need to I find it helps (mostly me) see what default values are rather than _knowing_ what they are when looking at code

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
CitrixChris: The action I agree with, the alert not so much as it is not an action. It is also intertwinned with many more classes and dont want to break it.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -181,35 +183,56 @@ protected override void Run()
...
-            if (!File.Exists(outputPathAndFileName))
+        public void InstallMsi()
+        {
+            if (!File.Exists(_outputPathAndFileName))
                 throw new Exception(Messages.DOWNLOAD_CLIENT_INSTALLER_MSI_NOT_FOUND);
 
+            // Install the msi            
+            try
+            {
+                ValidateMsi();
```
kc284, line 196: I don't think it's a good idea to move the validation outside the action. For one thing  if it fails, the exception will be propagated upwards and caught by xencenter as an unhandled exception without showing a meaningful error.

kc284 (changes requested):
I've put some modifications here https://github.com/CitrixChris/xenadmin/pull/4 mostly because one or two of these issues were introduced by me when we worked on the previous PR.

kc284:
This needs rebasing.

danilo-delbusso (approved):

AaronRobson (approved):

kc284 (approved):

PR #2980, opened 2022-03-29
CA-322672: Do not stop polling for RDP capabilities on VMs

danilo-delbusso:
### I have also tidied up some code, so go commit by commit while reviewing (the first commit is the one with the fix).

**Feel free to comment on the design of the solution. This one unfortunately increases resource usage, since it keeps polling VMs every 30 seconds just for this use case.**

The bug was caused because polling for RDP changes (as performed by `connectionPoller`) only started if the VM had RDP capabilities when first connecting to it with the console. This meant that if the RDP support was added after the console connection opened (by, for instance, installing VM Tools), CHC would not check again to see if the VM could use RDP. Even after a reboot, since the check was performed when creating the console screen.

These changes keep polling open console connections every 30 seconds per VM, to check the status of RDP capabilities.

Without this, changes in the `guest_metrics` aren't spotted. Consequently, the user is not shown the option to connect with RDP until they disconnect (or the console session is disconnected), even if the VM is capable.

To replicate the bug this PR is fixing:
1. Create a brand new W10 VM on an host
2. Install Windows VM Tools from [Citrix](https://citrix.com/downloads/citrix-hypervisor)
3. Reboot
Sometimes it won't show you the option to use RDP

With the changes here, CHC should prompt you to use RDP for the connection after the reboot (but not before! since it shouldn't work. this is why we also need to check for the presence of data in `networks` in `guest_metrics`)

### DIFF
--- a/XenAdmin/ConsoleView/VNCTabView.cs
+++ b/XenAdmin/ConsoleView/VNCTabView.cs
@@ -1093,7 +1093,7 @@ namespace XenAdmin.ConsoleView
             {
                 log.DebugFormat("RDP detected for VM '{0}'", source == null ? "unknown/null" : source.name_label);
                 toggleToXVNCorRDP = RDP;
-                
+
                 if (vncScreen.UseVNC)
                     toggleConsoleButton.Text = CanEnableRDP() ? enableRDP : UseRDP;
 
--- a/XenModel/XenAPI-Extensions/VM.cs
+++ b/XenModel/XenAPI-Extensions/VM.cs
@@ -348,7 +348,8 @@ namespace XenAPI
             var metrics = Connection.Resolve(this.guest_metrics);
             if (metrics == null)
                 return false;
-
+            // feature-ts is the feature flag indicating the toolstack
+            // can enable RDP remotely (by writing to control/ts)
             return 0 != IntKey(metrics.other, "feature-ts", 0);
         }
 
@@ -357,7 +358,7 @@ namespace XenAPI
             var vmMetrics = Connection.Resolve(this.guest_metrics);
             if (vmMetrics == null)
                 return false;
-
+            // data/ts indicates the VM has RDP enabled
             return 0 != IntKey(vmMetrics.other, "data-ts", 0);
         }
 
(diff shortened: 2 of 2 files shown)


XenAdmin/ConsoleView/VNCTabView.cs
```diff
@@ -1099,11 +1093,11 @@ private void OnDetectRDP_()
             {
                 log.DebugFormat("RDP detected for VM '{0}'", source == null ? "unknown/null" : source.name_label);
                 toggleToXVNCorRDP = RDP;
-
+
```
AaronRobson, line 1096: Nit: Trailing whitespace.

danilo-delbusso: That's the default behaviour from VS2022, i.e. if you add an empty line, it'll add whitespace that matches the indentation level

XenAdmin/ConsoleView/XSVNCScreen.cs
```diff
@@ -295,8 +295,6 @@ private void PollRDPPort(Object Sender)
         {
             if (hasRDP)
             {
-                if (connectionPoller != null)
```
kc284, line 298: Is it necessary to leave the connectionPoller ticking? StartPolling does get called when the VM guest metrics change, perhaps we should be watching for property `other` additionally to networks? After removing the `if (hasRDP)` from line 776, I would think that now the method PollRDPPort will be called because a new connectionPoller will be created (new creation may not be necessary, but this is a different issue - I'm trying out some changes on the side to address this in a subsequent PR).

danilo-delbusso: The `connectionPoller` keeps ticking because a user can uninstall the VM tools. This would disable the capability of using RDP, but the button would still show. From my testing, XC wouldn't actually spot this change (or at least not reliably), because the guest metrics change event is fired only when XC tries to read it. So if XC isn't reading the a VM guest tools value, it doesn't know that anything has changed (no event is fired). Been a while since I opened this PR so the details might be slightly different.

Currently, if the VM tools are uninstalled and the user clicks on the RDP button, the user would be shown nothing.
To replicate (using `master`)
1. Install VM tools and reboot
2. Disconnect -> reconnect (this forces XC to read `guest_metrics` and similar)
3. Uninstall VM tools
4. Should be able to press that button

We could address this by try/catching the error, and possibly updating the RDP button to disabled (or hide it?) if pressing it fails. I think it fails (silently) when it looks for IPs in `networks`, and doesn't find them. We could expand that code to alert the user.
This would check for RDP capability every time the button is pressed, which would be used to alert the user that they cannot use RDP anymore.

kc284 (approved):

PR #2984, opened 2022-04-04
CA-365905 (XSI-1215): Create a temporary file in the target download …

kc284:
…folder instead of the user's temp folder.

Let's have this reviewed and tested, and if all ok I'll port it to master and raise relevant PRs to the SDK repos.


danilo-delbusso (approved):
tested a CoreOS export to between two W10 VMs on the same network. No data added on the exporting VM's local disk as expected 👍

PR #2985, opened 2022-04-06
CP-36392: Improvement of sonarqube score

kc284:
Please review each commit separately. This PR contains also some related refactoring that I've been meaning to do for some time (these are the commits that do not cite a ticket and they are more straight-forward to review).


XenModel/Actions/AsyncAction.cs
```diff
@@ -179,7 +179,7 @@ public void RunAsync(SudoElevationResult sudoElevationResult)
         /// use that session for the action: it is then the responsibility of the calling function
         /// to make sure the session has the appropriate privileges and tidy it up afterwards.
         /// </summary>
-        public void RunExternal(Session session)
+        public void RunSync(Session session)
```
danilo-delbusso, line 182: nit: we could set `session = null` as a default. It's used in a few places.

Even nit-tier: The comment says "run this action non-async". We could just write "synchronously"

XenAdmin/Dialogs/OptionsPages/SaveAndRestoreOptionsPage.cs
```diff
@@ -54,44 +57,33 @@ public SaveAndRestoreOptionsPage()
...
             {
-                // we need to save stuff but without a password
                 Properties.Settings.Default.SaveSession = true;
                 Properties.Settings.Default.RequirePass = false;
 
                 Program.MainPassword = null;
             }
             else
             {
-                // password protect stuff
```
danilo-delbusso, line 82:
```csharp
// password protect stuff
```
not particularly enlightening indeed 😂

XenModel/Actions/ActionBase.cs
```diff
@@ -250,6 +249,11 @@ public ActionBase(string title, string description, bool suppressHistory, bool c
                 NewAction?.Invoke(this);
         }
 
+        public static ActionBase CreateDummyAction(string title, string description, bool completeImmediately = true, string error = null)
+        {
+            return new ActionBase(title, description, false, completeImmediately, error);
+        }
+
```
danilo-delbusso, line 256: This solution is definitely an improvement in terms of developer experience, but SonarQube is likely still going to complain. We're not using the return value of `CreateDummyAction` now.

Ideally we'd have a separate method invoke the actions we create, and we'd need to call it everywhere. Constructors shouldn't be performing these kinds of side effects.

something like `new ActionBase().Run()`

XenModel/XenAPI/Session.cs
```diff
@@ -30,6 +30,7 @@
```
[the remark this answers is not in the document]
AaronRobson: On this basis I'd recommend excluding the directory in the call to SonarQube.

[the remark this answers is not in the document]
kc284: I don't see why, it **is** being fixed; there will follow a relevant PR for the SDK repo later.

XenAdmin/Controls/CustomGridView/GridRow.cs
```diff
@@ -38,38 +38,29 @@ namespace XenAdmin.Controls.CustomGridView
 {
     public class GridRow : IComparable<GridRow>
     {
-        public Dictionary<string, GridItemBase> Items = new Dictionary<string, GridItemBase>();
-        public List<GridRow> Rows = new List<GridRow>();
+        public Dictionary<string, GridItemBase> Items { get; }= new Dictionary<string, GridItemBase>();
```
AaronRobson, line 41: Nit: should have a space before the `=`.

AaronRobson (approved):

XenAdmin/Wizards/PatchingWizard/PlanActions/PatchPrecheckOnHostPlanAction.cs
```diff
@@ -63,7 +63,7 @@ protected override void RunWithSession(ref Session session)
         {
             var coordinator = Helpers.GetCoordinator(Connection);
 
-            var mapping = (from HostUpdateMapping hum in mappings
+            HostUpdateMapping mapping = (from HostUpdateMapping hum in mappings
                 let xpm = hum as XenServerPatchMapping
                 where xpm != null && xpm.Matches(coordinator, xenServerPatch)
                 select xpm).FirstOrDefault();
```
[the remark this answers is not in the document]
kc284: I think the way it is now it is more efficient because the casting is performed only on the found element, while if we put it in the linq expression it is performed on all mappings that were selected.

XenModel/Actions/AsyncAction.cs
kc284: Will fix the comment. However, I don't see RunSync being called anywhere with a null session.

danilo-delbusso: To be fair, it's only used three times. That's why it's "nit":
image

kc284:
Please review the new commits separately. Most of the changes are code style issues )with the exception of the commit with the DummyAction), but I expect they will fix some reported code smells too.

XenAdmin/Dialogs/Wlb/WlbCredentialsDialog.cs
```diff
@@ -194,38 +194,34 @@ private void textboxWlbUrl_TextChanged(object sender, EventArgs e)
...
-                    // adopt UriBuilder as a quick validator
-                    UriBuilder ub = new UriBuilder(string.Format("http://user:password@{0}:80/", addr));
-                    return ub.Host == addr;
-                }
-                catch
-                {
-                    return false;
-                }
+                // use UriBuilder as a quick validator
+                var ub = new UriBuilder($"http://{addr}:80/");
```
danilo-delbusso, line 219: original was `http://user:password@{0}:80/`

should this be `http://user:password@{addr}:80/` ?

danilo-delbusso (changes requested):
Great job! Just one small comment

danilo-delbusso (approved):

PR #2990, opened 2022-04-19
CA-365905 (XSI-1215): Reworked previous solution:

kc284:
Create a temporary file in the target download folder only if the latter is in a different drive from the default temporary folder.

I thought I'd rework this because there are many occasions where the user performs smaller downloads to their local drive and they may find it strange to see a randomly named file in the target folder when it's not strictly necessary since the default temp folder can be used in this case.

### DIFF
--- a/XenModel/XenAPI/HTTP.cs
+++ b/XenModel/XenAPI/HTTP.cs
@@ -771,16 +771,24 @@ namespace XenAPI
         public static void Get(DataCopiedDelegate dataCopiedDelegate, FuncBool cancellingDelegate,
             Uri uri, IWebProxy proxy, string path, int timeoutMs)
         {
-            var dir = Path.GetDirectoryName(path);
-            if (dir == null)
-            {
-                if (path == null)
-                    throw new ArgumentNullException(nameof(path));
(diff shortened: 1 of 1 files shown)


XenModel/XenAPI/HTTP.cs
```diff
@@ -771,16 +771,20 @@ public static Stream HttpGetStream(Uri uri, IWebProxy proxy, int timeoutMs)
...
-                    throw new ArgumentNullException(nameof(path));
+            if (string.IsNullOrWhiteSpace(path))
                 throw new ArgumentException(nameof(path));
-            }
 
-            var tmpFile = Path.Combine(dir, Path.GetRandomFileName());
-            File.Delete(tmpFile);
+            var tmpFile = Path.GetTempFileName();
+
+            if (Path.GetPathRoot(path) != Path.GetPathRoot(tmpFile))
```
danilo-delbusso, line 779: I'd add a comment to show why we're doing this. Might not be very clear on why we did this in a few years

danilo-delbusso (changes requested):

danilo-delbusso (approved):

CitrixChris (approved):

PR #2996, opened 2022-04-25
Added code owners (default reviewers).

kc284:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

Let's see if it works and then I'll make the change in our other repos too.


danilo-delbusso (commented):
Would it be better to use the team instead of maintaining the list of individuals?

kc284:
I was thinking about that, but then a team can have more people that are not full-time members or dedicated reviewers, in which case we'll have to be removing entries manually from the PR.

danilo-delbusso (approved):

CitrixChris (approved):

PR #2997, opened 2022-04-26
CA-329538: Tell the user to ensure tools are installed if a VM cannot be migrated via drag-and-drop

kc284:
Also, show the name of the VM in the various failure messages as drag-and-drop may operate on multiple VMs.

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -8677,16 +8677,16 @@ Do you want to continue?</value>
     <value>Please eject your CD</value>
   </data>
   <data name="MIGRATION_NOT_ALLOWED" xml:space="preserve">
-    <value>This VM may not migrate at the moment</value>
+    <value>'{0}' may not migrate at the moment. Please ensure you have installed {1} on it</value>
   </data>
   <data name="MIGRATION_NOT_ALLOWED_CPU_FEATURES" xml:space="preserve">
-    <value>This VM may not migrate because the destination host does not have some of the CPU features that the VM is currently using</value>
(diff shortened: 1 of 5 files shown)


XenModel/Messages.resx
```diff
@@ -8677,16 +8677,16 @@ Do you want to continue?</value>
     <value>Please eject your CD</value>
   </data>
   <data name="MIGRATION_NOT_ALLOWED" xml:space="preserve">
-    <value>This VM may not migrate at the moment</value>
+    <value>'{0}' may not migrate at the moment. Have you installed {1} on it?</value>
```
danilo-delbusso, line 8680: Not sure I like the wording. Would something on the lines of `Please make sure you have installed {1} on the VM` work better?

danilo-delbusso (changes requested):
Since you're making changes to this code, a suggestion (though this could be achieved at another time, in a different PR)

I have noticed that the user can drag a VM to an offline pool, and isn't shown a message. This then results in the Migrate VM wizard being shown.

While this is correct, it'd probably be better if we didn't allow the user to drag the VM to offline host/pools in the first place. 
Is this an easy addition?

Animation

kc284:
I've updated the message. The drag-drop issue you mentioned above is easy to fix, but I'll send a new PR for it because I discovered one more problem in the same functionality: when the host is live, the UI freezes until the cross-pool migrate wizard is launched.

danilo-delbusso (approved):

CitrixChris (approved):

PR #2998, opened 2022-04-27
CA-293683: Prevent automatic scrolling in RPU wizard if user has scrolled manually

danilo-delbusso:
### 🚨 Please review each commit separately
Only the first one contains the fix.

The others are to tidy some aspects of the code.

Feel free to weigh in on the naming and/or location of `AppTextBox`, I wasn't sure what to choose for them.

_______

The basic idea here is that we're not automatically keeping the `TextBox` scrolled all the way down if the user has moved the scrollbar.
If they move it back to the bottom, it'll go back to scrolling automatically.

I have also made the event return the `newPosition` of the bar, and added _ad-hoc_ logic for the horizontal scrollbar. This is just so this event is ready for other use cases that might arise in the future.

### DIFF
--- a/XenAdmin/Controls/SmartScrollTextBox.cs
+++ b/XenAdmin/Controls/SmartScrollTextBox.cs
@@ -46,8 +46,19 @@ namespace XenAdmin.Controls
     /// </summary>
     internal class SmartScrollTextBox : TextBox
     {
-        protected internal bool IsVerticalScrollBarAtBottom =>
-            string.IsNullOrEmpty(Text) || GetPositionFromCharIndex(Text.Length).Y < Height;
+        private bool _scrolledProgrammatically;
+
+        protected internal bool IsVerticalScrollBarAtBottom => GetPositionFromCharIndex(Text.Length - 1).Y < Height;
+
+        public void SetTextWithoutScrolling(string text)
+        {
+            var oldVerticalPosition = Win32.GetScrollBarPosition(Handle, Win32.ScrollBarConstants.SB_VERT);
+            Text = text;
+            Win32.SetScrollPos(Handle, (int)Win32.ScrollBarConstants.SB_VERT, oldVerticalPosition, true);
+
+            _scrolledProgrammatically = true;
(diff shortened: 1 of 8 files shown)


kc284 (changes requested):

kc284:
Once you make the above tweaks, would it be possible to re-arrange the commits (it will probably take some conflict resolving) so that the actual bug fix is still in one commit separate from the code tidy? Because the bug fix will need porting to the lcm stream. If it is too much hassle leave it - porting can be done as re-implementation of the changes.

AaronRobson (dismissed):

danilo-delbusso:
> e tweaks, would it be possible to re-arrange the commits (it will probably take some conflict resolving) so that the actual bug fix is still in one commit separate from the code tidy? Because the bug fix will need porting to the lcm stream. If it is too much hassle leave it - porting can be done as re-implementation of the changes.

All done, though GitHub doesn't cope well with this so you'll need to re-review the whole PR I'm afraid.

Most changes are just tidy up work, anyway. Happy to help with any queries

kc284:
Thanks for the corrections.
I have now checked out the code to try it out (in fact, I cherry-picked the first commit to the lcm branch where installing updates is possible) and it does not seem to work for me. When the output exceeded the window height I scrolled upwards and expected it to stay there until I scrolled down, but when new output arrived it still scrolled to the end on its own.

danilo-delbusso:
> Thanks for the corrections. I have now checked out the code to try it out (in fact, I cherry-picked the first commit to the lcm branch where installing updates is possible) and it does not seem to work for me. When the output exceeded the window height I scrolled upwards and expected it to stay there until I scrolled down, but when new output arrived it still scrolled to the end on its own.

Thanks, I'll flag the PR as `needs updating` while I look into this

danilo-delbusso:
Remarks about the new changes.

The issue brought up by @kc284 was the missing `-1` in `IsVerticalScrollBarAtBottom`.

While making this change I also noticed another issue. `TextBox` automatically scrolls to the top when updating its `Text` value. Hence, I added a custom method that prevents this.

kc284:
Works ok for me now.

kc284 (approved):

CitrixChris (approved):

kc284:
Reminder that the first and fourth commit need porting to release/stockholm/lcm.

PR #3000, opened 2022-05-03
Some layout and logic changes so the `DRFailoverWizardRecoverPage` looks and behaves more like the other wizard pages with progress bars

kc284:
Normally we don't change the blurb at the top of the page, only the status text at the bottom. I also moved the progress bar to the bottom, so now the only wizard page where the progress bar is not at the bottom is the `AutomatedUpdates` page.


kc284:
I discovered I have forgotten one commit outside the PR - some minor tidy up. I pushed it in now.

danilo-delbusso (approved):

CitrixChris (approved):

PR #3002, opened 2022-05-06
CP-39720: Enable auth tokens

CitrixChris:


kc284 (approved):

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -102,6 +105,13 @@ private void DownloadFile()
 
                     //start the download
                     _updateDownloadState = DownloadState.InProgress;
+
+                    if (!string.IsNullOrEmpty(_authToken))
+                    {
+                        NameValueCollection myQueryStringCollection = new NameValueCollection();
+                        myQueryStringCollection.Add("burritoToken", _authToken);
```
AaronRobson, line 112: Please consider naming it using the `INTERNAL_STAGE_AUTH_TOKEN` constant (i.e. `"InternalStageAuthToken"`) instead of mentioning the code name of burrito.

CitrixChris: surething

kc284 (changes requested):
Let's implement @AaronRobson 's suggestion.

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
```diff
@@ -102,6 +105,13 @@ private void DownloadFile()
 
                     //start the download
                     _updateDownloadState = DownloadState.InProgress;
+
+                    if (!string.IsNullOrEmpty(_authToken))
+                    {
+                        NameValueCollection myQueryStringCollection = new NameValueCollection();
+                        myQueryStringCollection.Add("InternalStageAuthToken", _authToken);
```
kc284, line 112: You may want to use the constant from the Registry.cs instead of hardcoding, so in future you'll have to change it only in one place.

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -330,12 +332,19 @@ protected virtual XmlDocument FetchCheckForUpdatesXml()
...
                 var proxy = XenAdminConfigManager.Provider.GetProxyFromSettings(Connection, false);
 
                 using (var webClient = new WebClient())
                 {
                     webClient.Proxy = proxy;
                     webClient.Headers.Add("User-Agent", _userAgent);
+                    if (!string.IsNullOrEmpty(authToken))
+                    {
+                        NameValueCollection myQueryStringCollection = new NameValueCollection();
+                        myQueryStringCollection.Add("burritoToken", authToken);
```
kc284: This should need changing too - hence better use the constant from Registry.cs

XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs
CitrixChris: Ah thats a good idea

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
CitrixChris: Realised i missed that one seconds before this comment!

AaronRobson (approved):
I'm happy, assuming the constant is used in both places as @kc284 requested.

PR #3007, opened 2022-05-11
CP-36392: Fixes to sonarqube score.

kc284:


kc284:
Please do not review yet, I've just realised it has broken a couple of unit tests.

kc284:
I reverted the padding changes - zero padding is unreliable. I think the encryption algorithm may need to change completely. Will revisit it at a later time.

AaronRobson (approved):

danilo-delbusso (approved):

PR #3008, opened 2022-05-11
L10n: Resx resync and trnslations updates

menganikishore:
Signed-off-by: kishore Mengani <kishore.mengani@citrix.com>


danilo-delbusso (approved):

AaronRobson (approved):

kc284:
There were some issues in this PR that broke the build, no need to submit corrections as I've already submitted them in https://github.com/xenserver/xenadmin/pull/3010.

In that PR I have pointed out the lines that were breaking the build via comments. The key thing to notice is that Windows forms do not allow flexible syntax like the CSS, so Location/Size are always a (int, int) pair and Margin/Padding are always a (int,int,int,int) quadruplet.

PR #3012, opened 2022-05-23
CP-38687 + issue with version checking

kc284:

The first commit is the fix to ticket CP-38687.

The second is simply copying a number of methods to a new file and capitalizing the first letter of method ProductVersionCompare().

The third commit is a fix to an issue I noticed whereby the recent version checks added via CP-38895 were not giving the correct result because the key software_version.xapi is a 2 digit version and the original ProductVersionCompare() considered 1.2 being different from 1.2.0 (which is the same thing the framework's method Version.CompareTo() does, so the original code was practically reinventing the wheel), however, in CH we generally don't consider 1.2 and 1.2.0 as being different. Also, the use of the 2-digit key software_version.xapi may cause us issues in future if the change we check for may have happened on a 3rd digit bump, hence I think the best is to check the key software_version.xapi_build, which is an RPM version, hence I rewrote the method. I added some unit tests as well to ensure the method can be used safely for the legacy comparison on the key software_version.platform_version.


XenModel/XenAPI/JsonRpcClient.cs
```diff
@@ -12205,6 +12205,20 @@ public void message_destroy(string session, string _message)
             Rpc("message.destroy", new JArray(session, _message ?? ""), serializer);
         }
 
+        public void message_destroy_many(string session, List<XenRef<Message>> _messages)
+        {
+            var converters = new List<JsonConverter> {new XenRefListConverter<Message>()};
+            var serializer = CreateSerializer(converters);
+            Rpc("message.destroy_many", new JArray(session, _messages == null ? new JArray() : JArray.FromObject(_messages, serializer)), serializer);
+        }
```
danilo-delbusso, line 12213: Do we need to add the non-async version? It should be added the next time we update the SDK, anyway

XenModel/Utils/Helpers.Versions.cs
```diff
@@ -29,89 +29,144 @@
  * SUCH DAMAGE.
  */
 
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using XenAdmin.Network;
 using XenAPI;
+using XenCenterLib;
```
danilo-delbusso, line 37:
```suggestion

````

File: XenModel/Utils/Helpers.Versions.cs [hunk 3]
```diff
@@ -29,89 +29,144 @@
...
 
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using XenAdmin.Network;
 using XenAPI;
+using XenCenterLib;
+
 
 namespace XenAdmin.Core
```
danilo-delbusso, line 40: nit: since you've created a new file, can we also update the namespace? to `XenModel.Utils`

danilo-delbusso (changes requested):

XenModel/XenAPI/JsonRpcClient.cs
kc284: I just added both because when I was first trying out the new call I was using the sync version, but then I decided to switch to the async one (so the session is not blocked in case xapi takes too long to dismiss a huge number of alerts on a real system).

XenModel/Utils/Helpers.Versions.cs
kc284: Then I should also change the namespace in Helpers.cs because the class is partial, which means many more changes because the class is used all over the place. I don't disagree with the change, but I would leave it for a different PR.

XenAdmin/Actions/GUIActions/DismissAlertsAction.cs
```diff
@@ -71,11 +73,44 @@ protected override void Run()
...
+                    RelatedTask = Message.async_destroy_many(Session, msgRefs);
+                    PollToCompletion(0, 50);
+                    Alert.RemoveAlert(a => msgAlerts.Contains(a));
+
+                    for (var i = 0; i < otherAlerts.Count; i++)
+                    {
+                        Description = string.Format(Messages.ACTION_REMOVE_ALERTS_PROGRESS_DESCRIPTION, i, _alerts.Count);
+                        _alerts[i].Dismiss();
+                        PercentComplete = 50 + i * 50 / otherAlerts.Count;
+                    }
```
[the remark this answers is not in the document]
kc284: I've removed the Description from that line completely. That code path can use the Description as set in the constructor.

CitrixChris (approved):

danilo-delbusso (approved):

kc284:
Rebased on master to resolve conflicts.

PR #3018, opened 2022-06-06
CP-39818: Auto-launch app after installation is finalized

danilo-delbusso:
Example:

https://user-images.githubusercontent.com/32554698/172155939-48d66a4c-2bc8-4384-bd92-73f3e816311d.mp4

The background of the checkbox cannot be changed without worsening UX:

see [issue in GitHub](https://github.com/wixtoolset/issues/issues/1687)
or [thread here](https://wix-users.narkive.com/Ss5GWl5A/checkbox-drawn-with-different-coloured-background)
or [StackOverflow thread](https://stackoverflow.com/questions/31178681/how-can-i-make-the-background-of-an-checkbox-transparent-in-wix)

### DIFF
--- a/WixInstaller/XenCenter.wxs
+++ b/WixInstaller/XenCenter.wxs
@@ -271,13 +271,13 @@
             <Custom Action="ClearAllUsers" After="FindRelatedProducts">NOT Installed AND PERMACHINEINSTALL</Custom>
             <ExecuteAction Sequence="1102" />
         </InstallUISequence>
-         <UI>
+        <UI>
             <Publish Dialog="ExitDialog" 
                 Control="Finish" 
                 Event="DoAction" 
-                Value="LaunchApplication">WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed
-			</Publish>
-         </UI>
+                Value="LaunchApplication">WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 AND NOT (WixUI_InstallMode="Remove")
+            </Publish>
+        </UI>
         <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="!(loc.XenCenter_Launch) $(var.BrandConsole)" />
         <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOX" Value="1" />
         <Property Id="WixShellExecTarget" Value="[#XenCenterEXE]" />
@@ -286,6 +286,6 @@
             ExeCommand="/X"  
             Impersonate="yes"
             Return="asyncNoWait" />
-		<Icon Id="XenCenterICO" SourceFile="$(env.RepoRoot)\Branding\Images\AppIcon.ico" />
+        <Icon Id="XenCenterICO" SourceFile="$(env.RepoRoot)\Branding\Images\AppIcon.ico" />
     </Product>
 </Wix>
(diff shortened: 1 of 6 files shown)


danilo-delbusso:
Flagging as ready for review since this is a valid solution, and the checkbox background can be fixed by changing the banner bitmap

WixInstaller/XenCenter.wxs
```diff
@@ -271,6 +271,20 @@
...
+			</Publish>
+         </UI>
+        <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Launch $(var.BrandProduct)" />
+        <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOX" Value="1" />
+        <Property Id="WixShellExecTarget" Value="[#XenCenterEXE]" />
+        <CustomAction Id="LaunchApplication" 
+            BinaryKey="WixCA" 
+            DllEntry="WixShellExec"
+            Impersonate="yes" />
+		<Icon Id="XenCenterICO" SourceFile="$(env.RepoRoot)\Branding\Images\AppIcon.ico" />
```
kc284, line 288: Indentation.

WixInstaller/XenCenter.wxs
```diff
@@ -271,6 +271,20 @@
             <Custom Action="ClearAllUsers" After="FindRelatedProducts">NOT Installed AND PERMACHINEINSTALL</Custom>
             <ExecuteAction Sequence="1102" />
         </InstallUISequence>
-        <Icon Id="XenCenterICO" SourceFile="$(env.RepoRoot)\Branding\Images\AppIcon.ico" />
+         <UI>
+            <Publish Dialog="ExitDialog"
```
kc284, line 275: When the application in already installed and the user wants to repair, the checkbox does not appear. I think we should show it in this case too.

WixInstaller/XenCenter.wxs
```diff
@@ -271,6 +271,20 @@
...
+                Value="LaunchApplication">WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed
+			</Publish>
+         </UI>
+        <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Launch $(var.BrandProduct)" />
+        <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOX" Value="1" />
+        <Property Id="WixShellExecTarget" Value="[#XenCenterEXE]" />
+        <CustomAction Id="LaunchApplication" 
+            BinaryKey="WixCA" 
+            DllEntry="WixShellExec"
+            Impersonate="yes" />
```
kc284, line 287: I believe we should also add `Return="asyncNoWait"` so the installer won't have to wait until the process starts.

WixInstaller/XenCenter.wxs
```diff
@@ -271,6 +271,20 @@
...
         </InstallUISequence>
-        <Icon Id="XenCenterICO" SourceFile="$(env.RepoRoot)\Branding\Images\AppIcon.ico" />
+         <UI>
+            <Publish Dialog="ExitDialog" 
+                Control="Finish" 
+                Event="DoAction" 
+                Value="LaunchApplication">WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed
+			</Publish>
+         </UI>
+        <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Launch $(var.BrandProduct)" />
```
kc284, line 281: `$(var.BrandConsole)`

kc284: In fact this text will need localization.

kc284:
Regarding the checkbox background issue, because transparency is at the moment only supported for text controls, a possible workaround might be to use a combination of checkbox without text and a text control to achieve the desired result.

kc284 (changes requested):
As above.

AaronRobson (approved):

CitrixChris (approved):

kc284 (approved):

PR #3027, opened 2022-07-18
CA-368811, CA-368812: Issues on the import wizard.

kc284:
Please review each commit separately. Commits 6b1d658 and c93c4c7 are a bit risky;  c93c4c7 may not even be the real solution to the issue it's attempting to fix, but any further changes are riskier.

### DIFF
--- a/XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
+++ b/XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
@@ -47,9 +47,9 @@ namespace XenAdmin.Wizards.ImportWizard
 		public ImageVMConfigPage()
 		{
 			InitializeComponent();
-            m_upDownMemory.Minimum = m_upDownMemory.Increment = VM.DEFAULT_MEM_MIN_IMG_IMPORT;
-            m_upDownMemory.Maximum = VM.DEFAULT_MEM_ALLOWED;
-            m_ctrlError.HideError();
+            m_upDownMemory.Minimum = m_upDownMemory.Increment = (decimal)VM.DEFAULT_MEM_MIN_IMG_IMPORT / Util.BINARY_MEGA;
+            m_upDownMemory.Maximum = (decimal)VM.DEFAULT_MEM_ALLOWED / Util.BINARY_MEGA;
+			m_ctrlError.HideError();
 		}
 
 		#region Base class (XenTabPage) overrides
--- a/XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
+++ b/XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
@@ -31,6 +31,7 @@
 
 using System;
 using XenAdmin.Controls;
+using XenAPI;
 using XenModel;
 
 namespace XenAdmin.Wizards.ImportWizard
@@ -46,6 +47,8 @@ namespace XenAdmin.Wizards.ImportWizard
 		public ImageVMConfigPage()
 		{
 			InitializeComponent();
+            m_upDownMemory.Minimum = m_upDownMemory.Increment = VM.DEFAULT_MEM_MIN_IMG_IMPORT;
+            m_upDownMemory.Maximum = VM.DEFAULT_MEM_ALLOWED;
             m_ctrlError.HideError();
 		}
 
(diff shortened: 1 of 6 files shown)


danilo-delbusso (dismissed):

XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
```diff
@@ -47,9 +47,9 @@ internal partial class ImageVMConfigPage : XenTabPage
 		public ImageVMConfigPage()
 		{
 			InitializeComponent();
-            m_upDownMemory.Minimum = m_upDownMemory.Increment = VM.DEFAULT_MEM_MIN_IMG_IMPORT;
-            m_upDownMemory.Maximum = VM.DEFAULT_MEM_ALLOWED;
-            m_ctrlError.HideError();
+            m_upDownMemory.Minimum = m_upDownMemory.Increment = (decimal)VM.DEFAULT_MEM_MIN_IMG_IMPORT / Util.BINARY_MEGA;
+            m_upDownMemory.Maximum = (decimal)VM.DEFAULT_MEM_ALLOWED / Util.BINARY_MEGA;
+			m_ctrlError.HideError();
```
AaronRobson, line 52: Nit: tabs make the indentation look off, please use spaces.

XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
```diff
@@ -47,9 +47,9 @@ internal partial class ImageVMConfigPage : XenTabPage
 		public ImageVMConfigPage()
 		{
 			InitializeComponent();
-            m_upDownMemory.Minimum = m_upDownMemory.Increment = VM.DEFAULT_MEM_MIN_IMG_IMPORT;
-            m_upDownMemory.Maximum = VM.DEFAULT_MEM_ALLOWED;
-            m_ctrlError.HideError();
+            m_upDownMemory.Minimum = m_upDownMemory.Increment = (decimal)VM.DEFAULT_MEM_MIN_IMG_IMPORT / Util.BINARY_MEGA;
+            m_upDownMemory.Maximum = (decimal)VM.DEFAULT_MEM_ALLOWED / Util.BINARY_MEGA;
```
AaronRobson, line 51: I assume the value given to the `m_upDownMemory` control is already in Megabytes and this is to correct the Min and Max to be in those same units.

AaronRobson (approved):

XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
danilo-delbusso: you're exactly right

best to do it here than to refactor `m_upDownMemory` to use bytes :)

kc284: When the page is populated the value is set to the minimum. which is in  MB. Then when the user changes it, the value is passed down as is and added to the OVF envelope with unit MB. See https://github.com/xenserver/xenadmin/blob/7c873334399a63f4293497b10c4d89ddda736ebb/XenAdmin/Wizards/ImportWizard/ImportWizard.cs#L272 and https://github.com/xenserver/xenadmin/blob/7c873334399a63f4293497b10c4d89ddda736ebb/XenOvfApi/OVF.cs#L1410

XenAdmin/Wizards/ImportWizard/ImageVMConfigPage.cs
kc284: @danilo-delbusso has normalized the whitespace for the whole file, but this will be included in a future PR.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -57,6 +57,7 @@ public partial class VM : IComparable<VM>
         private const int DEFAULT_NUM_VIFS_ALLOWED = 7;
         private const int DEFAULT_NUM_VBDS_ALLOWED = 255;
         public const long DEFAULT_MEM_ALLOWED = 1 * Util.BINARY_TERA;
+        public const long DEFAULT_MEM_MIN_IMG_IMPORT = 256 * Util.BINARY_MEGA;
```
CitrixChris, line 60: What is the reason for setting the minimum memory to 256MB? Does this still cause issues?

CitrixChris (changes requested):

XenModel/XenAPI-Extensions/VM.cs
kc284: You just need a minimum memory value to assign to the VM that will be created (with 0 it won't start).

PR #3039, opened 2022-08-12
CP-39938: Flag a partially applied hotfix with red only if homogeneous application is required.

kc284:


XenAdmin/TabPages/GeneralTabPage.cs
```diff
@@ -1724,59 +1768,6 @@ private static string HVMBootMode(VM vm)
...
-        {
-            Pool_patch[] patches = xenObject.Connection.Cache.Pool_patches;
-
-            List<String> output = new List<String>();
-
-            foreach (Pool_patch patch in patches)
-                if (predicate(patch))
-                    output.Add(patch.name_label);
-
-            output.Sort(StringUtility.NaturalCompare);
```
danilo-delbusso, line 1758: nit: only detail is that it looks like we're losing this ordering. could it be an issue in terms of UX?

kc284: There is ordering applied, see lines 621, 627, 634, is it not working?

danilo-delbusso: Looks like I missed it! 😅

danilo-delbusso (approved):

CitrixChris (approved):

PR #3046, opened 2022-08-19
vTPM original implementation

kc284:


XenModel/Messages.resx
```diff
@@ -2891,6 +2891,15 @@ Do you want to assign these VMs to the schedule '{0}' instead?</value>
   <data name="COLLAPSE_CHILDREN" xml:space="preserve">
     <value>&amp;Collapse Children</value>
   </data>
+  <data name="COMMAND_RESET_VTPM_WARNING" xml:space="preserve">
+    <value>Resetting the vTPM that is attached to a VM to its original state may result in deactivation of security features on the VM. Do you want to continue?</value>
```
danilo-delbusso, line 2895:
```suggestion
<value>Resetting the vTPM that is attached to a VM to its original state may result in deactivation of security features on the VM. Do you want to continue?</value>

Comment (danilo-delbusso) [inline XenModel/Messages.resx:2895]:
```suggestion
    <value>Resetting the vTPM that is attached to a VM to its original state may result the in deactivation of some security features on the VM. Do you want to continue?</value>
```

XenModel/Messages.resx
```diff
@@ -2891,6 +2891,15 @@ Do you want to assign these VMs to the schedule '{0}' instead?</value>
   <data name="COLLAPSE_CHILDREN" xml:space="preserve">
     <value>&amp;Collapse Children</value>
   </data>
+  <data name="COMMAND_RESET_VTPM_WARNING" xml:space="preserve">
+    <value>Resetting the vTPM that is attached to a VM to its original state may result in deactivation of security features on the VM. Do you want to continue?</value>
+  </data>
+  <data name="COMMAND_VTPM_DISABLED_NON_UEFI" xml:space="preserve">
+    <value>TPM is supported only for VMs using the UEFI or UEFI secure boot modes.</value>
```
danilo-delbusso, line 2898:
```suggestion
    <value>vTPMs are only supported for VMs using the UEFI or UEFI secure boot modes.</value>
    ```

File: XenModel/Messages.resx [hunk 3]
```diff
@@ -14018,6 +14027,48 @@ Schedule:
...
+  <data name="VTPM_RBAC_RESTRICTION_RESET" xml:space="preserve">
+    <value>A {0} user does not have sufficient permissions to reset the vTPM that is attached to a VM to its original state. Please login using an account with one of the following roles:
+
+{1}</value>
+  </data>
+  <data name="VTPM_REMOVE" xml:space="preserve">
+    <value>Remove vTPM</value>
+  </data>
+  <data name="VTPM_REMOVE_WARNING" xml:space="preserve">
+    <value>Removing the vTPM may result in deactivation of security features on the VM and may even prevent it from starting. Do you want to continue?</value>
```
danilo-delbusso, line 14070:
```suggestion
    <value>Removing the vTPM may result in the deactivation of some security features on the VM and prevent it from starting. Do you want to continue?</value>
```

XenAdmin/Dialogs/VtmpManagementDialog.cs
```diff
@@ -0,0 +1,333 @@
...
+                return;
+
+            if (e.PropertyName != "power_state" && e.PropertyName != "VTPMs")
+                return;
+
+            Program.Invoke(this, RefreshTabs);
+        }
+
+
+        private void splitContainer_Panel1_Resize(object sender, EventArgs e)
```
danilo-delbusso, line 251: very nit, but there's a few methods here that are separated by two empty lines instead of one. If you're making other changes, it'd be nice to also normalise this.

danilo-delbusso (changes requested):
Great work! :rocket: Just a few changes.

- In the TPM Manager, the `Add vTPM` button has a black border after it's selected. A small UX issue, but if it's easy to fix, it'd be nice to have.
- The vertical sections of the TPM Manager don't have min widths and they can be resized too much. If we limit it, users can drag all the way to one side and still be able to tell what's what. The left section suffers from this more.
- I know this would result in a lot of renaming but I think it'd be better if we use the "destroy / create" verbs fot vTPM actions. This way, they match with the xapi terms, and avoid possible confusion.

XenAdmin/MainWindow.resx
```diff
@@ -2311,130 +2311,139 @@
...
     <value>&amp;Export...</value>
   </data>
+  <data name="toolStripSeparator32.Size" type="System.Drawing.Size, System.Drawing">
+    <value>261, 6</value>
+  </data>
+  <data name="toolStripMenuItemVtpm.Size" type="System.Drawing.Size, System.Drawing">
+    <value>264, 22</value>
+  </data>
+  <data name="toolStripMenuItemVtpm.Text" xml:space="preserve">
+    <value>Trusted Plat&amp;form Module Manager...</value>
```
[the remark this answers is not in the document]
kc284: The wording for various management dialogs is a bit mixed in XC, e.g. we use "Manage vApps..." and "Conversion Manager...". I went with "TPM Manager..." because Win10 uses "TPM Management" as the title of the relevant dialog, although "Manage vTPMs..." may be more descriptive of what our dialog does. Ok, I'll revisit this, will probably check with the server team too.

XenModel/Messages.resx
kc284: I don't see why this indentation should change.

XenModel/Messages.resx
kc284: My intention was to differentiate between vTPM as instance of virtual TPM and TPM as a feature - I think in this context we need the latter.

XenAdmin/Dialogs/VtmpManagementDialog.cs
```diff
@@ -0,0 +1,333 @@
...
+
+        private void VtpmManagementDialog_HelpButtonClicked(object sender, CancelEventArgs e)
+        {
+            HelpManager.Launch("VtpmManagementDialog");
+        }
+
+        private void VtpmManagementDialog_HelpRequested(object sender, HelpEventArgs hlpevent)
+        {
+            HelpManager.Launch("VtpmManagementDialog");
+        }
```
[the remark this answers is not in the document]
kc284: Hmm, that's copy-paste of bad OO implementation. It needs fixing in more than one places.

XenModel/Messages.resx
danilo-delbusso: I think this one was a mistake from my part, ignore the comment :)

kc284:
Regarding the border of the Add button and the panel resizing, these issues exist on all dialogs deriving from VerticallyTabbedDialog, so I think they can be addressed outside of the scope of this PR.

Regarding the verbs for add/remove, we don't need to use the same verbs as the server side, but rather pick verbs that are user friendlier and consistent with the rest of the UI.

danilo-delbusso (approved):

CitrixChris (approved):

PR #3064, opened 2022-11-14
CP-41108: Replace the C++ splash screen with a C# one. Program.cs refactoring.

kc284:

### DIFF
--- a/XenAdmin/SplashScreenContext.cs
+++ b/XenAdmin/SplashScreenContext.cs
@@ -46,8 +46,6 @@ namespace XenAdmin
         private MainWindow _mainWindow;
         private readonly string[] _args;
 
-        public static MainWindow MainWindow { get; private set; }
-
         public SplashScreenContext(params string[] args)
         {
             _args = args;
@@ -95,7 +93,7 @@ namespace XenAdmin
         {
             if (_mainWindow == null)
             {
-                MainWindow = _mainWindow = new MainWindow(_args);
+                Program.MainWindow = _mainWindow = new MainWindow(_args);
                 _mainWindow.CloseSplashRequested += _mainWindow_CloseSplashRequested;
                 _mainWindow.FormClosed += mainWindow_FormClosed;
                 _mainWindow.Show();
--- a/XenAdmin/Controls/Folders/FolderListItem.cs
+++ b/XenAdmin/Controls/Folders/FolderListItem.cs
@@ -81,7 +81,6 @@ namespace XenAdmin.Controls
                 // This is an old platform without FontSmoothingType support.
                 TransparentUsually = Color.Transparent;
             }
-
         }
 
         public FolderListItem(string path, AllowSearch allowSearch, bool changeButton)
@@ -97,7 +96,7 @@ namespace XenAdmin.Controls
 
         public Color ForeColor { get; set; } = SystemColors.ControlText;
 
-        private Font Font => Program.DefaultFont;
+        private static Font Font => Program.DefaultFont;
 
         public int MaxWidth { get; set; }
 
--- a/XenAdmin/Program.cs
+++ b/XenAdmin/Program.cs
@@ -70,7 +70,7 @@ namespace XenAdmin
         public static Font DefaultFontItalic;
         public static Font DefaultFontHeader;
 
-        public static MainWindow MainWindow => SplashScreenContext.MainWindow;
+        public static MainWindow MainWindow;
 
         public static CollectionChangeEventHandler ProgramInvokeHandler(CollectionChangeEventHandler handler)
         {
--- a/XenAdmin/WinformsXenAdminConfigProvider.cs
+++ b/XenAdmin/WinformsXenAdminConfigProvider.cs
@@ -48,10 +48,10 @@ namespace XenAdmin
 {
     public class WinformsXenAdminConfigProvider : IXenAdminConfigProvider
     {
-        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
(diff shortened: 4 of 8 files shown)


XenAdmin/SplashScreenContext.cs
```diff
@@ -0,0 +1,120 @@
...
+            ThreadPool.QueueUserWorkItem(obj =>
+            {
+                try
+                {
+                    Settings.Load();
+                    Settings.ReconfigureProxyAuthenticationSettings();
+                    Settings.ConfigureExternalSshClientSettings();
+                    Settings.Log();
+
+                    _splashScreen.AllowToClose = true;
```
danilo-delbusso, line 71: wouldn't it be better to get rid of the timer and field, and have a method named `StartMainWindow` or similar?

danilo-delbusso (changes requested):
ok so the vast majority of the comments are nit picky changes. There's enough of them to warrant a cleanup commit I think. 

Good job! 👏

XenAdmin/MainWindow.cs
```diff
@@ -716,73 +711,80 @@ private void ShowAboutDialogOnStartup()
...
-                        Activate();
-                    }
-                    break;
-                case ArgType.Passwords:
-                    Trace.Assert(false);
-                    break;
+                        XenConnectionUI.BeginConnect(connection, true, null, false);
+                        break;
+                    default:
+                        log.Warn("CLI: Wrong syntax or unknown command line options.");
```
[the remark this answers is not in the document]
kc284: We can't do this for security reasons: it may be a highly improbable case, but if a user passed in `password connect server username` instead of `connect server username password`, the password would end up showing in the logs. This is the reason why logging around the CLI commands is sparse in general.

XenAdmin/SplashScreenContext.cs
kc284: The reason why I added both is that I wanted the SplashScreen to close after at least 2 seconds have elapsed or the settings have been loaded, whichever first. I couldn't think of a simpler way to do it.

XenAdmin/MainWindow.cs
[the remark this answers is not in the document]
danilo-delbusso: oh right! makes sense 👍

XenAdmin/SplashScreenContext.cs
danilo-delbusso: Quick idea, haven't tested or fleshed it out: what if we have an event in Settings, fired when they have completed loading. Then we can close the splash screen when that event is fired.

XenAdmin/Wizards/RollingUpgradeWizard/RollingUpgradeExtrasPage.resx
```diff
@@ -382,6 +382,6 @@
     <value>RollingUpgradeExtrasPage</value>
```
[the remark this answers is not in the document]
kc284: When I changed `[XenCenter_No_Space]Main` to `[XenCenter_No_Space]`, SelectMultipleVMDestinationPage.resx had `XenCenterMain`, hence was not changed. When I raised the current PR though, I rebased on master where SelectMultipleVMDestinationPage.resx had changed in the mean time to `[XenCenter_No_Space]Main`. I'll change this particular instance to be consistent. Note there seem to exist many instances of `XenCenterMain` left there since the very first time we replaced the brand names with placeholders, but they don't seem to cause any issues.

XenAdmin/SplashScreenContext.cs
```diff
@@ -0,0 +1,120 @@
...
+{
+    internal class SplashScreenContext : ApplicationContext
+    {
+        private static readonly log4net.ILog _log = log4net.LogManager.GetLogger(MethodBase.GetCurrentMethod().DeclaringType);
+
+        private readonly SplashScreen _splashScreen;
+        private MainWindow _mainWindow;
+        private readonly string[] _args;
+
+        public static MainWindow MainWindow { get; private set; }
```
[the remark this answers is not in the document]
kc284: I know it's not that great, but I wanted to keep the non-static `_mainWindow` field in place so as to have symmetry with the `_splashscreen` field. An alternative solution is to remove `SplashScreenContext.MainWindow` completely and instead set the `Program.MainWindow` in line 96 - but to do this the latter will have to become `public set` and there is a risk that people may start setting it all over the place. Or we can replace the 492 usages of `Program.MainWindow` with `SplashScreenContext.MainWindow` 😃 
Secretly I'm hoping that one day in the remote future we might be able to get rid of all these static instances because they complicate the logic.

XenAdmin/SplashScreenContext.cs
kc284: Hmmm, loading entails calling all four of these methods (2 of which may additionally try a save operation which may last a bit), where would the event be fired?

XenAdmin/Settings.cs
```diff
@@ -31,15 +31,17 @@
 
 using System;
 using System.Collections.Generic;
+using System.Configuration;
```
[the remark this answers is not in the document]
kc284: This is used by variables in GetUserConfigPath().

XenAdmin/SplashScreenContext.cs
kc284: Plus that if loading settings happens to finish too quickly, I still think we should give the full 2 seconds before showing the main window, otherwise the splash won't show for too long.

danilo-delbusso: tried to flesh out my idea out but it but turned out to be more complex than the current implementation. This works 👍

danilo-delbusso (approved):
LGTM

CitrixChris (approved):

PR #3069, opened 2022-12-02
(LCM) CP-41048: reworked the previous solution

kc284:
I recommend reviewing/testing this separately because the content is not exactly the same as https://github.com/xenserver/xenadmin/pull/3065, I couldn't cherry pick all of the code as it was.

### DIFF
--- a/XenAdmin/Controls/SrPicker.cs
+++ b/XenAdmin/Controls/SrPicker.cs
@@ -29,6 +29,7 @@
  * SUCH DAMAGE.
  */
 
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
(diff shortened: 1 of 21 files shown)


danilo-delbusso (approved):
I really did try to break it but no luck 😅. Good job!

danilo-delbusso:
Actually I do have a comment: scanning creates one event for each SRs, is this a bit spam-y?
image

kc284:
Yeah, it is spammy, however, if construct the SrRefresh action with suppressHistory=true, it won't be added to the History collection and then I have no way of checking that a scan is already in progress for this SR when the control is launched. At the moment I think we'll have to live with it, but at some point later maybe we should investigate changing the meaning of suppressHistory from "do not add to History" to "do not show on EventsTab".

CitrixChris (approved):

PR #3072, opened 2022-12-12
CP-41432: Load home page as an emdedded single html file.

kc284:
Please do not merge before the branding PR is merged.


kc284:
I force-pushed because I rebased on master.

XenAdmin/TabPages/HomePage.Designer.cs
```diff
@@ -34,16 +34,19 @@ private void InitializeComponent()
             // 
             // webBrowser
             // 
+            this.webBrowser.AllowWebBrowserDrop = false;
             resources.ApplyResources(this.webBrowser, "webBrowser");
+            this.webBrowser.IsWebBrowserContextMenuEnabled = false;
             this.webBrowser.Name = "webBrowser";
+            this.webBrowser.ScriptErrorsSuppressed = true;
```
danilo-delbusso, line 41: Not sure why but the scrollbar is now visible, `ScrollBarsEnabled = false` addressed this issue

danilo-delbusso (changes requested):

XenAdmin/TabPages/HomePage.Designer.cs
kc284: At the minimum main window size and the default position of the vertical splitter the home page should look ok - at least I don't see scrollbars in XC (there do exist in CHC though because that homepage is bigger; the old size will be restored with CP-41433). Unfortunately we can't disable the scrollbars completely because if the user moves the splitter to the right the home page will be truncated.

danilo-delbusso (approved):

XenAdmin/TabPages/HomePage.cs
```diff
@@ -42,41 +42,59 @@ public partial class HomePage : DoubleBufferedPanel, IControlWithHelp
...
+                webBrowser.DocumentText = _html;
             }
             catch (Exception ex)
             {
-                log.Error(string.Format("Failed to load the HomePage. Url = {0}", Location), ex);
+                log.Error($"Failed to load the HomePage for {Program.CurrentLanguage}", ex);
             }
-
-            initializing = false;
+
```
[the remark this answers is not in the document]
kc284: I'll make a note and fix it in the next PR.

CitrixChris (approved):

PR #3074, opened 2022-12-19
CP-41281: Exposed pool option for migration compression

kc284:


XenModel/Messages.resx
```diff
@@ -4623,6 +4626,9 @@ Plugging in untrustworthy USB devices to your computer may put your computer at
   <data name="DISABLED_CLUSTERING_ON_POOL" xml:space="preserve">
     <value>Disabled clustering on {0}.</value>
   </data>
+  <data name="DISABLED_MIGRATION_COMPRESSION" xml:space="preserve">
+    <value>Compression during VM migration disabled</value>
```
danilo-delbusso, line 4630: Text is truncated too much when opening dialog with default size. I understand the suggested text is not as accurate, but it's slightly more informative for users that don't resize the section. Thoughts? (Applies to the enabled version, too).

```suggestion
    <value>Compression migration disabled</value>
````

kc284: You probably mean "migration compression", but indeed the term is not accurate. I think a better solution might be to increase the default width of the section? Because for other pages the subtext may be equally long or even longer (some pages concatenate and show several settings).

danilo-delbusso: > You probably mean "migration compression"
correct, sorry

> increase the default width of the section

well let's see how that looks, might be easier

kc284: I moved the splitter a bit to the right and also made the messages slightly shorter. The disabled case is still a bit longer than the splitter width, but the advantage of the new text is that one does not need to move the splitter to see the value because they can see `Don't compress` immediately.

danilo-delbusso (approved):

PR #3079, opened 2023-01-09
CA-374160: Check for prerequisite updates shows the current version found, not the one required.

kc284:


XenAdmin/Diagnostics/Checks/PatchPrecheckCheck.cs
```diff
@@ -239,7 +239,7 @@ private Problem FindProblem(string errorcode, string param1, string param2, stri
                     return new ConflictingUpdatePresent(this, param2, Host);
 
                 case "UPDATE_PRECHECK_FAILED_PREREQUISITE_MISSING":
-                    return new PrerequisiteUpdateMissing(this, param2, Host);
```
danilo-delbusso, line 242: seems pretty straightforward, just one detail: do we know for a fact that `<required/>` is always `param3`? (not sure how to check this, but I can see that as a possible reason this bug was introduced)

kc284: Yeah, it looks a bit random, but I believe it's fairly guaranteed by line 183 further up in this file.

danilo-delbusso (approved):

CitrixChris (approved):

PR #3086, opened 2023-01-24
CP-41737: Add FCoE deprecation warnings

danilo-delbusso:
There's some tidy-up code here, I suggest you review one commit at a time.

Warnings added to
- RPU wizard from pre-82x
- New SR wizard
- General tab of SR

Note that the XC warning for pre-cloud connections is superseded by this new one. This is by design as there are planned changes to that warning. This is temporary as we will be making changes to that piece of logic, but should enable testing of the feature.

I am redirecting to a vanity link, this will be updated by the docs team once the deprecation has its own page.

### DIFF
--- a/XenAdmin/Diagnostics/Checks/PoolHasDeprecatedSrsCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/PoolHasDeprecatedSrsCheck.cs
@@ -1,6 +1,5 @@
-﻿/* Copyright (c) Citrix Systems, Inc. 
- * All rights reserved. 
- * 
+﻿/* Copyright (c) Cloud Software Group, Inc. 
+ *
  * Redistribution and use in source and binary forms, 
  * with or without modification, are permitted provided 
  * that the following conditions are met: 
--- a/XenAdmin/Diagnostics/Checks/PoolHasDeprecatedSrsCheck.cs
+++ b/XenAdmin/Diagnostics/Checks/PoolHasDeprecatedSrsCheck.cs
@@ -46,13 +46,11 @@ namespace XenAdmin.Diagnostics.Checks
         {
         }
 
-        public override bool CanRun() => !Helpers.Post82X(Host.Connection);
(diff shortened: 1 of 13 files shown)


CitrixChris (approved):

XenAdmin/Diagnostics/Checks/PoolHasDeprecatedSrsCheck.cs
```diff
@@ -0,0 +1,58 @@
...
+        public PoolHasDeprecatedSrsCheck(Host host)
+            : base(host)
+        {
+        }
+
+        public override bool CanRun() => !Helpers.Post82X(Host.Connection);
+
+        protected override Problem RunHostCheck()
+        {
+            var hasFCoESr = Host.Connection.Cache.SRs.Any(sr => sr.GetSRType(true) == SR.SRTypes.lvmofcoe);
```
kc284, line 53: We need to check the target version and decide accordingly what to show. The check is done with the upgrade plugin for automated mode. For now we'll show a warning only for >= Post82X. In manual mode or when the aforementioned plugin has failed and we don't know the target version, we need to show a conditional warning. (In future we'll add a blocking error if IsYOrGreater in automated mode and keep the warning for manual; this will also depend on the upgrade compatibility matrix, but we don't need to worry about it now).

kc284 (changes requested):

XenAdmin/Diagnostics/Checks/PoolHasDeprecatedSrsCheck.cs
danilo-delbusso: image

CitrixChris (approved):

danilo-delbusso:
Rebased on master as requested by @kc284

kc284:
Changing label until this one is reviewed https://github.com/danilo-delbusso/xenadmin/pull/21

kc284 (approved):

CitrixChris (approved):

PR #3091, opened 2023-02-01
CP-40157: Updates to `GraphDetailsDialog`

### LINKED ISSUE #3080: Performance Counter IntelliCache Cache hits + misses not handled
I don't know if it can happen in a real environment, but I think if IntelliCache is available and one selects a server and it's performance tab it throws exceptions here:

https://github.com/xenserver/xenadmin/blob/master/XenAdmin/Controls/CustomDataGraph/DataSet.cs#L87

Because the following two cases are not handled:

```
case "hits/s":
case "misses/s":
```

It just happened when I tested it against a VM and thought I would let you know.

IntelliCache_Cache_Hits

danilo-delbusso:
- Expose some hidden data sources to UI 
- Update friendly name for xapi_open_fds data source
    - Also apply string sorting
- Add data source description as column in GraphDetailsDialog
- Increase default GraphDetailsDialog width to accomodate new column

Also:

Tidy up source in GraphDetailsDialog
- Rename fields
- Reorder methods
- Rename methods
- Add `sealed` to `DataSourceGridViewRow`

_________________

This PR also fixes #3080

### DIFF
--- a/XenAdmin/Controls/CustomDataGraph/DataSet.cs
+++ b/XenAdmin/Controls/CustomDataGraph/DataSet.cs
@@ -62,11 +62,7 @@ namespace XenAdmin.Controls.CustomDataGraph
         public DataSet(IXenObject xo, bool hide, string datasourceName, List<Data_source> datasources)
         {
             XenObject = xo;
-            Hide = datasourceName == "xapi_open_fds" ||
-                   datasourceName == "pool_task_count" ||
-                   datasourceName == "pool_session_count" ||
-                   datasourceName == "memory" ||
-                   datasourceName == "memory_total_kib" || hide;
+            Hide = datasourceName == "memory" || datasourceName == "memory_total_kib" || hide;
 
             DataSourceName = datasourceName;
 
--- a/XenAdmin/Controls/CustomDataGraph/DataSet.cs
+++ b/XenAdmin/Controls/CustomDataGraph/DataSet.cs
@@ -97,6 +97,8 @@ namespace XenAdmin.Controls.CustomDataGraph
                     CustomYRange = new DataRange(1, 0, 1, Unit.MegaHertz, RangeScaleMode.Auto);
                     break;
                 case "requests/s":
+                case "hits/s":
+                case "misses/s":
                 case "err/s":
                     CustomYRange = new DataRange(1, 0, 1, Unit.CountsPerSecond, RangeScaleMode.Auto);
                     break;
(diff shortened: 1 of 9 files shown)


danilo-delbusso:
Not sure what's happening with the checks:
```log
The content of D:\a\xenadmin\xenadmin\XenModel\FriendlyNames.resx isn't sorted alphabetically.
Please sort it using the script in scripts/sort_strings.ps1:
./sort_strings.ps1 -PATHS D:\a\xenadmin\xenadmin\XenModel\FriendlyNames.resx
```

But I have run the following locally:

```bash
 .\scripts\check_strings_sorting.ps1 -PATHS .\XenModel\FriendlyNames.resx -NOISY -CHECK_LOCALIZED
```
and all three files are shown as sorted.

Similarly, re-applying the sorting results in no changes.

Issue here might be that GitHub is running the action on the PR before the force push? (when there was indeed a sorting issue).

I'd ignore it unless I'm missing something

kc284 (changes requested):
Maybe move the description column immedaitely after the name as this is how we have it in several other gridviews. Also, because the description may be of unknown and potentially longth length, Isuggest settting its sizing mode to Fill with a minimum width of 100 (i.e. exactly like the name column).

XenAdmin/Controls/CustomDataGraph/DataSet.cs
```diff
@@ -62,11 +62,7 @@ public class DataSet : IComparable<DataSet>
         public DataSet(IXenObject xo, bool hide, string datasourceName, List<Data_source> datasources)
         {
             XenObject = xo;
-            Hide = datasourceName == "xapi_open_fds" ||
-                   datasourceName == "pool_task_count" ||
-                   datasourceName == "pool_session_count" ||
-                   datasourceName == "memory" ||
-                   datasourceName == "memory_total_kib" || hide;
+            Hide = datasourceName == "memory" || datasourceName == "memory_total_kib" || hide;
```
kc284, line 65: There is one more place where they need to be shown, see class DataSourceItemList
 in GraphHelper.cs.

kc284:
Since you're here, do you think you could add a couple of lines to fix this issue https://github.com/xenserver/xenadmin/issues/3080
The reporter has already suggested the fix, just add
```
case "hits/s":
case "misses/s":
```
under
```
case "requests/s":
case "err/s":
```

kc284 (approved):

CitrixChris (approved):

PR #3096, opened 2023-02-15
(master) CP-42155: Removal of HealthCheck service leftovers

kc284:
This is mostly deleted code plus a new warning when the user first connects to a pool with HealthCheck enrollment to disable it.


danilo-delbusso (changes requested):
So the dialog is a bit odd to me. We ask users to please disable the HealthCheck service before continuing, but nothing happens if they don't press Yes. Also it's not clear what we're continuing (we just logged in, really).

Ideally we wouldn't give users a choice, and pressing "No" or "Cancel" would disconnect the pool. What worries me with that approach is that if the action fails, users can't use the pool at all. Thoughts?

_________

Also one small thing I did notice that would be nice to address now (since we will likely forget about it):

`ZipStatusReportAction` is now only used by the Server Status Report wizard. We should put it inside the `StatusReportAction` file, like the others (or even better: put each in its own file under a `StatusReport/` folder).

CitrixChris (approved):

kc284:
> So the dialog is a bit odd to me. We ask users to please disable the HealthCheck service before continuing, but nothing happens if they don't press Yes. Also it's not clear what we're continuing (we just logged in, really).
> Ideally we wouldn't give users a choice, and pressing "No" or "Cancel" would disconnect the pool. What worries me with that approach is that if the action fails, users can't use the pool at all. Thoughts?

I'm implying "before proceeding to manage the pools", but if it's not obvious I'll try to rephrase. The reason nothing happens if they press no is that I don't think the problem is too serious, if they don't disable it, they will be nudged again next time they connect until they disable it. I suppose I could remove the option and only allow them to say Ok, after all why keep it if it's not working, but I don't think we should disconnect the pool for this.

XenAdmin/MainWindow.cs
```diff
@@ -1045,18 +1025,27 @@ private void connection_CachePopulated(IXenConnection connection)
             if(licenseTimer != null)
                 licenseTimer.CheckActiveServerLicense(connection, false);
 
-            if (Registry.GetBrandOverride() == "XenCenter" || BrandManager.BrandConsole == "XenCenter")
+            if ((Registry.GetBrandOverride() == "XenCenter" || BrandManager.BrandConsole == "XenCenter") &&
```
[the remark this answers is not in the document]
kc284: Hmm, now that I'm thinking about it, `DEBUG` is not going to work if someone builds and installs the application locally without having overriden the branding placeholders - unless they have changed the scripts to build in debug mode. The registry override sounds more flexible.

XenAdmin/MainWindow.cs
```diff
@@ -1025,19 +1025,18 @@ private void connection_CachePopulated(IXenConnection connection)
...
                     {
-                        using (var dlg = new WarningDialog(string.Format(Messages.PROBLEM_HEALTH_CHECK_ON_CONNECTION, pool.Name()),
-                                   new ThreeButtonDialog.TBDButton(Messages.PROBLEM_HEALTH_CHECK_HELP, DialogResult.Yes),
-                                   new ThreeButtonDialog.TBDButton(Messages.CANCEL, DialogResult.No)))
+                        using (var dlg = new InformationDialog(
+                                   string.Format(Messages.PROBLEM_HEALTH_CHECK_ON_CONNECTION, pool.Name(), BrandManager.BrandConsole),
+                                   ThreeButtonDialog.ButtonOK))
                         {
-                            if (dlg.ShowDialog() == DialogResult.Yes)
+                            if (dlg.ShowDialog() == DialogResult.OK)
```
danilo-delbusso, line 1039: nit, but I think worth addressing: Users can still use the `x` button on the dialog to close it. I think some may expect it to work in the same way as a "Cancel" button. Can we remove it? Or maybe, we should remove enrollment as soon as the pool connects, and then tell the user we just did.

danilo-delbusso (changes requested):

XenAdmin/MainWindow.cs
kc284: We can't remove enrollment silently - for one thing if the user does not have enough permissions the operation will fail and they will get an RBAC error out of nowhere. I don't think it's a good idea in terms of UX to remove the X button either. If the changed text does not make things clearer (due to the finality in the statement) maybe we should restore the Cancel button.

kc284: I think the bit `XC will disable Health Check and clear the enrollment before you can proceed to manage your system` does do the trick, because it doesn't sound like the user has an option which could be activated if they hit Esc, the latter rather dismisses the dialog and XC proceeds to do what it said it will.

danilo-delbusso (approved):
All good, only needs rebasing

PR #3102, opened 2023-02-21
(LCM) CA-375173: Create the VDI and VBD records before closing the new disk dialog

kc284:
Also fixed issue where the SR was not preselected when editing an existing disk and no scans were taking place.

### DIFF
--- a/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
+++ b/XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
@@ -379,7 +379,7 @@ namespace XenAdmin.Wizards.NewVMWizard
                 if (dialog.ShowDialog() != DialogResult.OK)
                     return;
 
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/NewVMWizard/Page_Storage.cs
```diff
@@ -409,7 +409,7 @@ private void EditButton_Click(object sender, EventArgs e)
                 if (dialog.ShowDialog(ParentForm) != DialogResult.OK)
                     return;
 
-                selectedItem.Disk = dialog.NewDisk();
```
kc284, line 412: Although I could not reproducde this, I suspect that what happened is that the SrPicker.SR was null when NewDisk() was called here, hence the XenRef<SR> constructor within this method crashed,hence I changed where the disk is created in the NewDiskDialog.

danilo-delbusso (approved):
Not sure how we're going to test this (if at all?) but the changes look good and they don't seem to have broken anything 🤷

CitrixChris (approved):

PR #3107, opened 2023-02-24
CA-375668: Report bytes uncompressed instead of percentage when uncompressing a gzip.

kc284:
The original code was crashing because we switched to .NET framework's compression classes and GZipStream is a bit basic and does not support Position.

I tried to find an alternative method of reporting progress, however, I could not see a reliable way of knowing the uncompressed file size, particularly for the huge gzips handled by the application. The compression ratio depends on the data being compressed and the method and library that was used, which are generally unknown. One solution I thought of was to try to approximate the progress by assuming a compression ratio between 2:1 and 5:1 (typically offered by zlib), but in the end I thought best to just display bytes uncompressed without showing percentage. Let me know what you think.

### DIFF
--- a/XenAdmin/Wizards/ImportWizard/ImportSourcePage.cs
+++ b/XenAdmin/Wizards/ImportWizard/ImportSourcePage.cs
@@ -532,7 +532,8 @@ namespace XenAdmin.Wizards.ImportWizard
                     _filesToDownload.Enqueue(file);
 
                     LongProcessWrapper(() => _webClient.DownloadFileAsync(_uri, downloadedPath, file),
-                        string.Format(Messages.IMPORT_WIZARD_DOWNLOADING, Path.GetFileName(downloadedPath).Ellipsise(50)));
+                        string.Format(Messages.IMPORT_WIZARD_DOWNLOADING, Path.GetFileName(downloadedPath).Ellipsise(50)),
+                        ProgressBarStyle.Blocks);
                     return m_textBoxFile.Text == downloadedPath;
                 }
                 return false;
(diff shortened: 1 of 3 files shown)


XenAdmin/Wizards/ImportWizard/ImportSourcePage.cs
```diff
@@ -630,18 +632,15 @@ private void _unzipWorker_DoWork(object sender, DoWorkEventArgs e)
...
 
-                            int percentage = (int)Math.Floor((double)bzis.Position * 100 / length);
-
-                            if (percentage < 0)
-                                _unzipWorker.ReportProgress(0);
-                            else if (percentage > 100)
-                                _unzipWorker.ReportProgress(100);
-                            else
-                                _unzipWorker.ReportProgress(percentage);
+                            if (totalBytesRead % (128 * Util.BINARY_MEGA) == 0)
```
danilo-delbusso, line 642: This only reports progress if the total bytes hits a multiple of 128 MB (`134,217,728`) I think we should not report it like this. Our buffer is `4096` so let's maybe report based on that. Probably shouldn't be every `4096`, but a better multiple of it? If you were trying to avoid the bar flashing in and out immediately, I'd just use minimum size of bytes read before we use show the bar.

Also, we are likely never hitting 100%. I understand users might not see even see the progress bar at 100%, but this is just not accurate.

danilo-delbusso: Also I was playing with this and the original `.gz` file is removed. I understand we're telling the user but is it hard to not remove the original file? What is the idea behind doing so?

kc284: Given the disks contained in the packages are frequently huge, I think it's space concerns and also avoidance of re-extraction if the user needs to import it again.

danilo-delbusso (approved):

kc284:
I force-pushed to update the commit descirption with the ticker number. No other change.

CitrixChris (approved):

PR #3108, opened 2023-02-27
CP-41775: Fix using Hanja/Hangeul keys over VNC

rosslagerwall:


rosslagerwall:
While this is not dependent on the QEMU changes, we should probably wait until the QEMU side is merged before merging this  just in case the upstream reviewers decide on a different approach.

danilo-delbusso (approved):

kc284 (approved):

CitrixChris (approved):

rosslagerwall:
I think this is ready to go in now.

danilo-delbusso:
Tested and works. Thanks @rosslagerwall 🚀 

This hasn't broken anything for existing hosts, as expected.

PR #3120, opened 2023-03-22
Dead/obsolete code removal and some issues on the AddServerDialog

kc284:
My original intention was to just submit commit `Removed obsolete code/hidden functionality...`
After I did it though, for some reason I could not type at all in the AddServerDialog, it kept crashing with overflow symptoms. The problem insisted even when I reset on master and only stopped when I wrote commit `Use the server combobox's item list for autocompletion suggestions`.
While doing this I noticed some issues, which lead to the rest of the commits in this PR.
Please review them separately and if possible see if you have any issues with typing in the AddServerDialog after the changes.


XenAdmin/Dialogs/AddServerDialog.resx
danilo-delbusso, line 1: nit picky but I will never be able to un-see it:

_"Incorrect username and/or password"_ isn't center-aligned with the ❌ icon anymore
before: 
image
after:
image

danilo-delbusso (changes requested):
All good, just one small comment :)

XenAdmin/Dialogs/AddServerDialog.resx
kc284: Hmm, actually it's wrong on either case, the icon should be aligned with the textbox, not the label - at least this is how it is everywhere else in the application. Alternatively the message should be a bit further down and the icon aligned with the Username/Password labels.

danilo-delbusso: Great stuff, ty :D 

image

danilo-delbusso (approved):

CitrixChris (approved):

PR #3121, opened 2023-03-23
CP-42286: Add friendly name for `pool_total_session_count`

danilo-delbusso:
The description has also been updated with this new definition:
https://github.com/xapi-project/xen-api/pull/4942

### DIFF
--- a/XenAdmin/Controls/CustomDataGraph/DataSet.cs
+++ b/XenAdmin/Controls/CustomDataGraph/DataSet.cs
@@ -99,6 +99,7 @@ namespace XenAdmin.Controls.CustomDataGraph
                 case "hits/s":
                 case "misses/s":
                 case "err/s":
+                case "sessions/s":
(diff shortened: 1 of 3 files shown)


danilo-delbusso:
Converting to draft as this actually has an issue that should be addressed from the xapi side

XenAdmin/Controls/CustomDataGraph/DataSet.cs
```diff
@@ -99,6 +99,7 @@ public DataSet(IXenObject xo, bool hide, string datasourceName, List<Data_source
                 case "hits/s":
                 case "misses/s":
                 case "err/s":
+                case "sessions/s":
```
kc284, line 102: I noticed that Toolstack have set the units to `sessions per second`. I've asked them to change it to be consistent with the cases above.

danilo-delbusso: Already changed it myself and the pr has been merged :)

kc284 (approved):

PR #3123, opened 2023-03-24
SrAction RBAC fix + refactoring. Icon fix on DR config dialog

kc284:
The oiginal intention was to fix a regression in the RBAC permissions of the unplug and destroy PBDs action which we missed in the PR removing the PureAsyncAction, but then I thought that the SrAction design could be greatly simplified if split into four actions.

The truncated icon issue is something I noticed accidentally when examining the uses of SrAction: 
image
Unfortunately we don't have a 32x32 info icon to match the warning icons on the same dialog, but I don't think it matters very much since they don't appear at the same time. A correctly displayed small icon is better than the truncated one.


XenModel/Actions/SR/SrAction.cs
```diff
@@ -35,139 +35,80 @@
...
-                    ApiMethodsToRoleCheck.Add("SR.async_forget");
-                    break;
-                case SrActionKind.SetAsDefault:
-                    ApiMethodsToRoleCheck.AddRange(
-                        "pool.set_crash_dump_SR",
-                        "pool.set_default_SR",
-                        "pool.set_suspend_image_SR");
-                    break;
-                case SrActionKind.UnplugAndDestroyPBDs:
-                    ApiMethodsToRoleCheck.AddRange("PBD.async_unplug", "SR.async_destroy");
```
kc284, line 76: Here is the regression I mentioned at the descirption. It should be PBD.async_destroy, not SR. The practical result is the same because both calls are available to pool operators, but we should be using the correct call.

CitrixChris (approved):

danilo-delbusso (approved):

PR #3124, opened 2023-03-27
Do not pre-pend key to `InternalStageAuthToken` query parameters

danilo-delbusso:
Use of `InternalStageAuthToken=foo` is not needed anymore. Furthermore, we can now provide any token format.

### DIFF
--- a/XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
+++ b/XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
@@ -29,6 +29,7 @@
  */
 
 using System;
+using System.Web;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
@@ -43,6 +44,8 @@ namespace XenAdmin.Actions
 {
     public class DownloadUpdatesXmlAction : AsyncAction
     {
+        private static readonly log4net.ILog Log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod()?.DeclaringType);
+
         private const string ClientVersionsNode = "versions";
         private const string XenServerVersionsNode = "serverversions";
         private const string PatchesNode = "patches";
@@ -338,14 +341,17 @@ namespace XenAdmin.Actions
 
         protected virtual XmlDocument FetchCheckForUpdatesXml()
         {
-            var xdoc = new XmlDocument();
+            var checkForUpdatesXml = new XmlDocument();
             var checkForUpdatesUrl = XenAdminConfigManager.Provider.GetCustomUpdatesXmlLocation() ?? BrandManager.UpdatesUrl;
             var authToken = XenAdminConfigManager.Provider.GetInternalStageAuthToken();
-            var uri = new Uri(checkForUpdatesUrl);
+            var uriBuilder = new UriBuilder(checkForUpdatesUrl);
 
+            uriBuilder.Query = AddAuthTokenToQueryString(authToken, uriBuilder.Query);
+
+            var uri = uriBuilder.Uri;
             if (uri.IsFile)
             {
-                xdoc.Load(checkForUpdatesUrl);
+                checkForUpdatesXml.Load(checkForUpdatesUrl);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -338,14 +341,17 @@ private void GetXenServerVersions(XmlDocument xdoc)
...
         protected virtual XmlDocument FetchCheckForUpdatesXml()
         {
-            var xdoc = new XmlDocument();
+            var checkForUpdatesXml = new XmlDocument();
             var checkForUpdatesUrl = XenAdminConfigManager.Provider.GetCustomUpdatesXmlLocation() ?? BrandManager.UpdatesUrl;
             var authToken = XenAdminConfigManager.Provider.GetInternalStageAuthToken();
-            var uri = new Uri(checkForUpdatesUrl);
+            var uriBuilder = new UriBuilder(checkForUpdatesUrl);
+
+            uriBuilder.Query = AddAuthTokenToQueryString(authToken, uriBuilder.Query);
```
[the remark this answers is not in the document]
CitrixChris: I've also made similar modifications in my desensitise PR

CitrixChris (changes requested):

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -356,20 +362,53 @@ protected virtual XmlDocument FetchCheckForUpdatesXml()
...
+                    query.Add(HttpUtility.ParseQueryString(existingQueryString));
+                }
+
+                var tokenQueryString = HttpUtility.ParseQueryString(authToken);
+
+                query.Add(tokenQueryString);
+
+                queryString = string.Join("&",
+                    query.AllKeys
+                        .Where(key => key != null)
```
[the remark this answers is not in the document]
kc284: I'm not talking about the value `query[key]`, I'm talking about the `key` itself, i.e. `.Where(key => !string.IsNullOrEmpty(key))`. A query like `https://blah.com?x=a&y=b&=c` doesn't look right.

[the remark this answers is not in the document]
danilo-delbusso: oh! silly me. Will take care of it :)

kc284 (dismissed):

CitrixChris (changes requested):
There is also an auth token in XenModel/Actions/Updates/DownloadAndUpdateClientAction.cs 
So that will need to be updated.

CitrixChris (approved):

XenModel/Utils/Helpers.cs
```diff
@@ -1537,5 +1539,46 @@ public static bool ConnectionRequiresRbac(IXenConnection connection)
...
         }
+
+        /// <summary>
+        /// Adds the specified authentication token to the existing query string and returns
+        /// the modified query string.
+        /// </summary>
+        /// <param name="authToken">The authentication token to add to the query string.</param>
+        /// <param name="existingQueryString">The existing query string to add the token to.</param>
+        /// <returns>The modified query string.</returns>
+        public static string AddAuthTokenToQueryString(string authToken, string existingQueryString)
```
CitrixChris, line 1550: While this works it is a little overkill, we have a preformatted auth token that you pull apart and then put back together.

danilo-delbusso: I could add it with simple string concatenation but we wouldn't have any sanity check mechanism to go with it. You could argue it's fine since we know right now what's being used etc, but it's not really future proofed. I went for future proofing this time and so I sticked with suggestions in the Microsoft docs

danilo-delbusso: So basically I agree with you, but I usually lean towards adding a bit more checks to avoid potential issues in the future, even if unlikely and with low impact, like in this case

XenModel/Actions/Updates/DownloadUpdatesXmlAction.cs
```diff
@@ -338,38 +338,32 @@ private void GetXenServerVersions(XmlDocument xdoc)
...
                     webClient.Proxy = proxy;
                     webClient.Headers.Add("User-Agent", _userAgent);
-                    if (!string.IsNullOrEmpty(authToken))
-                    {
-                        NameValueCollection myQueryStringCollection = new NameValueCollection();
-                        myQueryStringCollection.Add(XenAdminConfigManager.Provider.GetInternalStageAuthTokenName(), authToken);
-                        webClient.QueryString = myQueryStringCollection;
-                    }
-
                     using (var stream = new MemoryStream(webClient.DownloadData(uri)))
```
kc284, line 361: What value does uri have here? Isn't still the one from line 345? Don't we need the updated one including the query from line 353?

danilo-delbusso:
Commit bfd99e1 is needed as the strings aren't re-encoded by `WebClient`, as found by @CitrixChris

kc284 (approved):

CitrixChris (approved):

PR #3127, opened 2023-03-31
CP-40407: Removal of IXmlRpcProxy related classes and simulation infrastructure

kc284:
Since XmlRpc usage has been dropped from the SDK and the related classes have also been removed, their use had to be removed from the application too, or any new SDK drop would not compile. These classes were employed by the testing infrastructure; I have been brutal and removed that one too; in future we should replace it by the Moq framework (and generally it's not ideal to carry too much mocking/testing code within the product).

This PR retains only the functionality of loading xapi databases, which is meant only for visualisation of data and not operations on them. It's best to review each commit separately. Note that a small part of the code that has been deleted in the second commit, it has been restored slightly modified in a different file in the third commit - this involves the code in Marshalling.cs and Db.cs of the third commit.

Marshalling.cs will have to change in xen-api repo too as it's part of the SDK. Note that this code is used by the UpdateFrom(hashtable) methods, which are currently used in the product only for two things: loading the xapi databases (internal only functionality), and loading xenserver settings from OVF metadata during package import (externally facing functionality).

Care needs to be taken when reviewing the file XenConnection.cs in the third commit. I think a lot of the throuble I had with simulating a successful connection to a xapi database was due to the fact that certain statements were happening in what I deemed as incorrect order. Also there were some errors that were silenced and I think that was the worng thing to do. Please let me know if something is not clear, I'm happy to look at the PR together if necessary.


kc284:
The forced push was because I had forgotten to add the ticket number in the 2nd commit's description.

CitrixChris (approved):

danilo-delbusso (approved):

PR #3136, opened 2023-04-21
CA-376480, CA-374325, & CA-378489: Prevent two instances from starting after running installer; Add desktop shortcut option; Remove existing shortcuts

danilo-delbusso:
When an instance of XenCenter is already running, the installer shows the `MsiRMFilesInUse` fragment. Pressing OK within that fragment with the `WixUIRMOption` property set to `UseRM` means that the `RMShutdownAndRestart` event kicks in at the end of the installation.

This happens at the same time as the existing "Launch XenCenter" event (if the matching checkbox is checked), causing in the executable being started twice at about the same time. My guess is that the pipe system is not initialised at that point, and that's how we get two applications.

To avoid this, this commit adds a new custom property: `XS_WixUIRMPressedOk`. This is set to `1` when the user clicks OK on the `WixUIRMOption`. It allows us to know whether or not that dialog was shown at all, as checking `WixUIRMOption` is not possible (it's already set to `UseRM` by default). By checking the value of this property we prevent the "Launch XenCenter" option to be shown if `MsiRMFilesInUse` was shown to the user (and they clicked OK). This applies both in case of a repair and an upgrade from a previous version.

### DIFF
--- a/WixInstaller/XenCenter.wxs
+++ b/WixInstaller/XenCenter.wxs
@@ -149,7 +149,7 @@
                 <Component Id="ProgramFilesShortcut" Guid="63b44222-b5e8-4ab1-8f68-baed79abbf36">
                     <Condition>INSTALLSHORTCUT</Condition>
                     <Shortcut Id="ApplicationDesktopShortcut" Name="$(var.BrandConsole)" Target="[INSTALLDIR]$(var.BrandConsoleNoSpace).exe" WorkingDirectory="INSTALLDIR"/>
-                    <RemoveFolder Id="DesktopFolder" On="uninstall"/>
+                    <RemoveFolder Id="DesktopFolder" On="both"/>
                     <RegistryValue Root="HKCU" Key="Software\$(var.BrandProduct)\$(var.BrandConsoleNoSpace)" Name="installed" Type="integer" Value="1" KeyPath="yes" />
                 </Component>
             </Directory>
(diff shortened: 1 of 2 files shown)


WixInstaller/XenCenter.wxs
```diff
@@ -238,7 +238,7 @@
             <Publish Dialog="ExitDialog" 
                 Control="Finish" 
                 Event="DoAction" 
-                Value="LaunchApplication">WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 AND NOT (WixUI_InstallMode="Remove")
+                Value="LaunchApplication">WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 AND NOT (WixUI_InstallMode="Remove") AND XS_WixUIRMPressedOk="0"
```
danilo-delbusso, line 241: FYI. Technically, if `(WixUI_InstallMode="Remove") AND XS_WixUIRMPressedOk="0"` is `true`, the checkbox won't show at all. This is just an extra check

kc284:
As I had written on CA-374325, I still think the page where we specify the installation location is cluttered. The last page has more space - and there are several other installers that have the desktop shortcut option on the last page. Could we give it a go and see how it looks?

danilo-delbusso:
Reminder to also check installer icon post-uninstall

WixInstaller/wix_src.patch
```diff
@@ -998,18 +997,18 @@ diff -ru wixlib/WixUI_zh-CN.wxl wixlib/WixUI_zh-CN.wxl
...
 -                    <Condition Action="show">WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT AND NOT Installed</Condition>
-+                <Control Id="OptionalCheckBox" Type="CheckBox" X="135" Y="190" Width="220" Height="40" Hidden="yes" Property="WIXUI_EXITDIALOGOPTIONALCHECKBOX" CheckBoxValue="1"  Text="[WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT]">
-+                    <Condition Action="show">WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT AND NOT (WixUI_InstallMode="Remove") AND XS_WixUIRMPressedOk="0"</Condition>
++                <Control Id="InstallShortcutCheckbox" Type="CheckBox" X="135" Y="185" Width="290" Height="10" Property="INSTALLSHORTCUT" CheckBoxValue="1" Text="!(loc.InstallDirDlgShortCut)" />
++                <Control Id="OptionalCheckBox" Type="CheckBox" X="135" Y="200" Width="220" Height="10" Hidden="yes" Property="WIXUI_EXITDIALOGOPTIONALCHECKBOX" CheckBoxValue="1" Text="[WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT]">
++                    <Condition Action="show">WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT AND NOT (WixUI_InstallMode="Remove") AND XS_WixUIRMPressedOk="0"</Condition>
```
kc284, line 1006: Line 1006 ends in LF instead of CRLF and may cause the patch to fail to apply.

kc284 (changes requested):
1 line ending needs to change, otherwise it looks ok to me.

kc284 (approved):

PR #3137, opened 2023-04-24
CA-375803: Catch exceptions when creating a new template from snapshot

danilo-delbusso:
Both when user cancels and when action fails.

Also completed action in `TODO`


CitrixChris (dismissed):

XenAdmin/Commands/NewTemplateFromSnapshotCommand.cs
```diff
@@ -65,15 +68,16 @@ public NewTemplateFromSnapshotCommand(IMainWindow mainWindow, VM snapshot)
...
 
             // Generate a unique suggested name for the new template
-            string defaultName = Helpers.MakeUniqueName(String.Format(Messages.TEMPLATE_FROM_SNAPSHOT_DEFAULT_NAME, snapshot.Name()), takenNames);
+            var defaultName = Helpers.MakeUniqueName(string.Format(Messages.TEMPLATE_FROM_SNAPSHOT_DEFAULT_NAME, snapshot.Name()), takenNames);
 
-            using (var dialog = new InputPromptDialog {
+            using (var dialog = new InputPromptDialog
+            {
                 Text = Messages.SAVE_AS_TEMPLATE,
                 PromptText = Messages.NEW_TEMPLATE_PROMPT,
```
kc284, line 82: Not caused by this PR, but since we're here can we add articles to this message: `Enter a &amp;name for the new template:`

XenAdmin/Commands/NewTemplateFromSnapshotCommand.cs
```diff
@@ -82,34 +86,38 @@ protected override void RunCore(SelectedItemCollection selection)
...
         }
 
-        void action_Completed(ActionBase sender)
+        private static void action_Completed(ActionBase sender)
         {
-            AsyncAction action = (AsyncAction)sender;
-            var vm = action.Connection.Resolve(new XenRef<VM>(action.Result));
+            var action = (AsyncAction)sender;
+            VM vm = null;
+            try
```
kc284, line 100: This is one of the cases where the exception does not need catching, but rather prevention.
When the action is cancelled or fails, the getter of action.Result throws that cancellation or failure exception (it's debatable whether this is a good design, but this is a different story), so we shoud not try to get the Result unless the action has succeeded. Instead of the try-catch, all this handler needs is a check at the beginning:
```
if (!sender.Succeeded || !(sender is AsyncAction action))
    return;
```
I've seen this once before, generally we do check before getting the Result, but there might be some forgotten case somewhere, an audit might be a good idea.

kc284 (changes requested):

XenAdmin/Controls/Wlb/WlbOptimizePool.cs
```diff
@@ -594,39 +594,38 @@ private void BuildRecList()
         /// </summary>
         protected void OptRecRetrieveAction_Completed(ActionBase sender)
         {
-            AsyncAction action = (AsyncAction)sender;
-            if (action.IsCompleted)
-            {
-                action.Completed -= OptRecRetrieveAction_Completed;
+            if (!(sender is AsyncAction asyncAction) || !action.IsCompleted)
```
kc284, line 597: Why not check directly if sender is WlbRetrieveRecommendationsAction?

XenAdmin/Dialogs/GraphDetailsDialog.cs
```diff
@@ -208,7 +208,7 @@ private void EnableDataSource()
 
         private void getDataSourcesAction_Completed(ActionBase sender)
         {
-            if (!(sender is GetDataSourcesAction action))
+            if (!(sender is GetDataSourcesAction action) || !action.Succeeded)
```
kc284, line 211: What happens on the UI if the action fails?

XenModel/Messages.resx
```diff
@@ -8989,7 +8989,7 @@ It is strongly recommended that you Cancel and apply the latest version of the p
     <value>Edit tags...</value>
   </data>
   <data name="NEW_TEMPLATE_PROMPT" xml:space="preserve">
-    <value>Enter &amp;name for new template:</value>
+    <value>Enter a &amp;name for new template:</value>
```
kc284, line 8992: "the new template"

danilo-delbusso: 🤦 oops

kc284:
Regarding my questions above on PerformancePage.cs and GraphDetailsDialog.cs, and because the oriiginal code did not check if the action succeeded, I'm trying to establish whether we have forgotten the failure case, or whether indeed we don't need to check (because the action result values, which are presumably null/empty in this case, may be needed to clear the UI).

XenAdmin/Dialogs/GraphDetailsDialog.cs
danilo-delbusso: It shouldn't happen because I think all users can access the list of data sources, but it keeps loading indefinitely.

We should handle this better. We can add some error state.

image

kc284: Hmm, I suspected so because the spinner at least is not hidden elsewhere.
Simplest solution: revert the Succeeded check in both files. We're not accessing `Result` there anyway, so no need to change it.

kc284 (approved):

CitrixChris (approved):

PR #3139, opened 2023-04-24
CP-39687+CP-42691: XVA export for VMs with vTPMs is now supported; use allowed_operations for enabling vTPM add/remove

kc284:


XenModel/Utils/Helpers.Versions.cs
```diff
@@ -450,6 +450,12 @@ public static bool XapiEqualOrGreater_22_33_0(IXenConnection conn)
             return coordinator == null || ProductVersionCompare(coordinator.GetXapiVersion(), "22.33.0") >= 0;
         }
 
+        public static bool XapiEqualOrGreater_23_10_0(IXenConnection conn)
+        {
+            var coordinator = GetCoordinator(conn);
+            return coordinator == null || ProductVersionCompare(coordinator.GetXapiVersion(), "23.10.0") >= 0;
+        }
+
```
danilo-delbusso, line 464: Not relevant to this PR but maybe for the future: looks like we're just replicating code here. We should probably have `XapiEqualOrGreater(IXenConnection connection, XapiVersion version)` with a list of version enums encoding all versions used (or something similar). This would still allow us to easily refactor existing calls. Can be done another time, just wanted to add my 2 cents on the matter.

XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -1797,9 +1797,12 @@ public bool CanAddVtpm(out string cannotReason)
         {
             cannotReason = null;
 
-            if (VTPMs.Count >= VM.MAX_ALLOWED_VTPMS)
+            if (Helpers.XapiEqualOrGreater_23_10_0(Connection))
+                return allowed_operations.Contains(vm_operations.create_vtpm);
```
danilo-delbusso, line 1801: Here and in other pages we're not returning a `cannotReason`. I'd add message saying that the host doesn't support the operation.

danilo-delbusso (changes requested):

XenModel/XenAPI-Extensions/VM.cs
kc284: Yes, I did remark on the ticket that this change affects the UX. I was wondering about adding a message like this, but I was not sure whether it would add any value because it doesn't really explain why the host doesn't support it. I can add it though if you think that it's better for the user to see it.

XenModel/Utils/Helpers.Versions.cs
kc284: I agree, things will get out of hand soon.

XenModel/XenAPI-Extensions/VM.cs
danilo-delbusso: We can just add something like "creating vtpms is not a supported operation for this host" or something. It's better than nothing, I'd definitely want to see it

danilo-delbusso (approved):

CitrixChris (approved):

PR #3140, opened 2023-04-26
Updates legacy url to updates.hypervisor.cloud.com

CitrixChris:


danilo-delbusso:
shouldn't we just get rid of the legacy option? Is this a "just in case" approach?

danilo-delbusso (changes requested):
We also need to bump the branding version in the `Jenkinsfile` to `v2.8`

Branding/Branding.resx
```diff
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="LEGACY_UPDATES_URL" xml:space="preserve">
-    <value>https://updates.xensource.com/XenServer/updates.xml</value>
+    <value>https://updates.hypervisor.cloud.com/xenserver/updates.xml</value>
```
kc284, line 121: This now exposes the URL we didn't want to expose. Please revert it and add the value in the internal repo, and a line replacing https://updates.xensource.com/XenServer/updates.xml with the value in re-branding.sh.

kc284 (changes requested):

Branding/Branding.resx
kc284: Please also add the ticket number in the commit message for tracking.

PR #3147, opened 2023-05-02
Add missing `reads/s` unit mapping

danilo-delbusso:


XenAdmin/Controls/CustomDataGraph/DataSet.cs
```diff
@@ -100,6 +100,7 @@ public DataSet(IXenObject xo, bool hide, string datasourceName, List<Data_source
                 case "misses/s":
                 case "err/s":
                 case "sessions/s":
+                case "reads/s":
```
kc284, line 103: As previously discussed, could you please also check wethere there are any RRDs with units `writes/s` which we might be also missing?

kc284 (approved):

CitrixChris (approved):

PR #3148, opened 2023-05-03
CA-376887: Fix threading management for `ArchiveMaintainer`

danilo-delbusso:
I have split this PR in many commits because it contains a lot of tidying up. I highly suggest you read through the PR one commit at a time.

The underlying issue that is being addressed here is that `ArchiveMaintainer` wasn't stricly tied to its XenObject before. While fetching operations were running the XenObject could change. A similar issue happened with other private fields such as `FirstTime`. The latter was the cause for one the issues described in the original ticket description.

This refactor removes the manual handling of the `Thread` and instead uses `CancellationToken`s to manage ongoing requests. The new `ArchiveMaintainer` is related to **one** `XenObject`. If a user navigates away from the `XenObject`'s Performance tab the operation will spawn a new thread while the old one is cancelled. The new thread operates under a separate `ArchiveMaintainer` and has its own state. If the user goes back to the original `XenObject`, its `ArchiveMaintainer` spawns a new thread. If the previous one was still cancelling, that will still happen in the background. Before the changes, there was only one `Thread` and one `ArchiveMaintainer`.

Furthermore, the use of `CancellationToken` has been injected all the way into the `while (reader.Read())` of the data fetching requests. This avoids useless hanging of threads that are waiting on large data to dowload even though we don't need it anymore.

`PerformancePage` is now the handler of all the `ArchiveMaintainer`s, and it is also responsible for telling the UI controls which `ArchiveMaintainer` is the one they should be using.

### DIFF
--- a/XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
+++ b/XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
@@ -533,6 +533,24 @@ namespace XenAdmin.Controls.CustomDataGraph
             }
         }
 
+
+        public static ArchiveInterval NextArchiveDown(ArchiveInterval current)
+        {
+            switch (current)
+            {
+                case ArchiveInterval.FiveSecond:
+                    return ArchiveInterval.None;
+                case ArchiveInterval.OneMinute:
+                    return ArchiveInterval.FiveSecond;
+                case ArchiveInterval.OneHour:
+                    return ArchiveInterval.OneMinute;
+                case ArchiveInterval.OneDay:
+                    return ArchiveInterval.OneHour;
+                default:
+                    return ArchiveInterval.None;
+            }
+        }
+
         #endregion
 
         #region Helpers
@@ -586,21 +604,5 @@ namespace XenAdmin.Controls.CustomDataGraph
 
         #endregion
 
-        public static ArchiveInterval NextArchiveDown(ArchiveInterval current)
-        {
-            switch (current)
-            {
-                case ArchiveInterval.FiveSecond:
-                    return ArchiveInterval.None;
-                case ArchiveInterval.OneMinute:
-                    return ArchiveInterval.FiveSecond;
-                case ArchiveInterval.OneHour:
-                    return ArchiveInterval.OneMinute;
-                case ArchiveInterval.OneDay:
-                    return ArchiveInterval.OneHour;
-                default:
-                    return ArchiveInterval.None;
-            }
-        }
     }
 }
--- a/XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
+++ b/XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
@@ -67,11 +67,6 @@ namespace XenAdmin.Controls.CustomDataGraph
         private const int HOURS_IN_ONE_WEEK = 168;
         private const int DAYS_IN_ONE_YEAR = 366;
 
-        private static readonly TimeSpan FiveSeconds = TimeSpan.FromSeconds(5);
-        private static readonly TimeSpan OneMinute = TimeSpan.FromMinutes(1);
-        private static readonly TimeSpan OneHour = TimeSpan.FromHours(1);
-        private static readonly TimeSpan OneDay = TimeSpan.FromDays(1);
-
         private const int SLEEP_TIME = 5000;
 
         private static readonly log4net.ILog Log =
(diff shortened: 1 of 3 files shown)


XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
```diff
@@ -169,7 +180,8 @@ private void Update()
             {
                 var xenObject = XenObject;
 
-                var serverWas = ServerNow; // get time before updating so we don't miss any 5 second updates if getting the past data
+                var serverWas =
+                    ServerNow; // get time before updating so we don't miss any 5 second updates if getting the past data
```
kc284: Line breaks like this make the code less readable. I'd suggest moving ServerNow back to the previous line and move the commnet underneath.

danilo-delbusso: agreed but it looks like it's been addressed in other commits in this PR 👍

CitrixChris (approved):

kc284 (changes requested):

XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
```diff
@@ -116,67 +156,65 @@ private void StartUpdateLoop(object _)
             var serverWas = ServerNow;
             InitialLoad(serverWas);
 
-            while (!_cancellationTokenSource.Token.IsCancellationRequested)
+            while (!RequestedCancellation)
             {
-                if (serverWas - LastFiveSecondCollection > TimeSpan.FromSeconds(5))
+                if (ServerNow - LastFiveSecondCollection > TimeSpan.FromSeconds(5))
```
kc284, line 161: Was the replacement of serverWas with ServerNow intentional? Every thime this is called, it calls DateTime.UtcNow, which returns a different time. Don't we need to go through this if blocks with the same time?

XenAdmin/TabPages/PerformancePage.cs
```diff
@@ -82,11 +78,7 @@ protected override void Dispose(bool disposing)
             if (_disposed)
                 return;
 
-            _archiveMaintainers.ForEach(a =>
-            {
-                a.Stop();
-                a.ArchivesUpdated -= ArchiveMaintainer_ArchivesUpdated;
-            });
+            _archiveMaintainer.Dispose();
```
kc284, line 81: Unregister events before disposing.

XenAdmin/TabPages/PerformancePage.cs
```diff
@@ -175,34 +167,32 @@ private get
...
+                _xenObject = value;
+                RegEvents();
+
+                var newArchiveMaintainer = new ArchiveMaintainer(value);
+                newArchiveMaintainer.ArchivesUpdated += ArchiveMaintainer_ArchivesUpdated;
+                GraphList.ArchiveMaintainer = newArchiveMaintainer;
+                DataPlotNav.ArchiveMaintainer = newArchiveMaintainer;
+                try
+                {
+                    _archiveMaintainer?.Dispose();
```
kc284, line 181: Unregister events before disposing.

XenAdmin/TabPages/PerformancePage.cs
```diff
@@ -333,12 +323,7 @@ private void DeregEvents()
         public override void PageHidden()
         {
             DeregEvents();
-
-            if (ArchiveMaintainer != null)
-            {
-                ArchiveMaintainer.ArchivesUpdated -= ArchiveMaintainer_ArchivesUpdated;
-                ArchiveMaintainer.Stop();
-            }
+            _archiveMaintainer?.Dispose();
```
kc284, line 326: Unregister events before disposing.

XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
danilo-delbusso: oops I was playing with it and forgot about it. needs tweaking

XenAdmin/TabPages/PerformancePage.cs
danilo-delbusso: which events do you mean?

XenAdmin/Controls/CustomDataGraph/ArchiveMaintainer.cs
```diff
@@ -158,36 +158,37 @@ private void StartUpdateLoop(object _)
...
                 {
                     Get(ArchiveInterval.OneHour, UpdateUri, RRD_Update_InspectCurrentNode, XenObject);
 
                     if (RequestedCancellation)
                         break;
 
-                    LastOneHourCollection = ServerNow;
+                    LastOneHourCollection = serverWas;
                     Archives[ArchiveInterval.OneHour].Load(_setsAdded);
                 }
```
kc284, line 194: What about ths one?

XenAdmin/TabPages/PerformancePage.cs
```diff
@@ -78,6 +78,7 @@ protected override void Dispose(bool disposing)
             if (_disposed)
                 return;
 
+            DeregisterEvents();
```
kc284, line 81: I missed this earlier. I believe these two line should be within the `if (disposing)` block.

kc284 (approved):

PR #3151, opened 2023-05-09
It was not possible to load a xapi database from the CLI.

kc284:


XenAdmin/MainWindow.cs
```diff
@@ -729,28 +729,18 @@ internal void ProcessCommand(params string[] args)
...
+                        else
+                            XenConnectionUI.BeginConnect(connection, true, this, false);
                         break;
                     default:
                         log.Warn("CLI: Wrong syntax or unknown command line options.");
                         break;
                 }
             }
 
-            if (_launched)
```
kc284, line 746: I think `_launcehd` is not needed because this method is called only when the MainWindow has been created.

danilo-delbusso (approved):

CitrixChris (approved):

PR #3153, opened 2023-06-05
CP-40844: Adds download source action that gets the latest source code in the production stage

CitrixChris:
… in the production stage.


XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -146,5 +147,50 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
             }
         }
+
+        public static void DownloadSource(IWin32Window parent)
+        {
+            var saveSourceDialog = new SaveFileDialog();
+            saveSourceDialog.FileName = "XenCenter-source";
+            saveSourceDialog.DefaultExt = "zip";
+            saveSourceDialog.Filter = "Zip files (*.zip)|*.zip|All files (*.*)|*.*";
+            saveSourceDialog.InitialDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),"Downloads");
```
danilo-delbusso, line 157: - I would use the `BrandingManager`'s version of `XenCenter` to name the file.
- Also the filter should be localised since it's visible to customers. Possibly even the `-source` bit in the file name. If you don't want to localise it, we can remove "zip files" but ideally it would. Maybe check what we do in other similar cases.
- For the `InitialDirectory`, using "Downloads" as a name for the folder doesn't necessarily work for all locales. The server status report uses `Win32.GetKnownFolderPath(Win32.KnownFolders.Downloads);` to fetch the path. While I'm not the biggest fan of calling the win32 APIs directly, we already have the methods so might as well use them.
- I'd also use the object initialiser, I think it looks neater. But that's just preference so fell free to ignore:
```csharp
 var saveSourceDialog = new SaveFileDialog
            {
                FileName = "XenCenter-source",
                DefaultExt = "zip",
                Filter = "Zip files (*.zip)|*.zip|All files (*.*)|*.*",
                InitialDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),"Downloads")
            };
```

XenModel/Actions/Updates/DownloadSourceAction.cs
```diff
@@ -0,0 +1,265 @@
...
+        private readonly bool _downloadSource;
+        private DownloadState _sourceDownloadState;
+        private Exception _sourceDownloadError;
+        private readonly string _authToken;
+        private WebClient _client;
+        
+        public string ByteProgressDescription { get; set; }
+
+        public DownloadSourceAction(string sourceName, Uri uri, string outputFileName)
+            : base(null, /*string.Format(Messages.DOWNLOAD_SOURCE_ACTION_TITLE,*/ sourceName,
```
danilo-delbusso, line 64: Leftover commented out code here

danilo-delbusso (changes requested):
While making the changes for the source info to be fetched as part of the client updates XML, there's some low hanging fruits that can be changed even before that 👇

XenModel/Actions/Updates/DownloadSourceAction.cs
```diff
@@ -0,0 +1,265 @@
+﻿/* Copyright (c) Cloud Software Group, Inc.
```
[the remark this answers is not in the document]
danilo-delbusso: Mh I can see this being a big pain when the merge feature branch is merged though. not sure what the best way forward is, though

https://github.com/xenserver/xenadmin/pull/3152/files/194e61b988469087b94622112b399d22bf412dc0..23cea7e250aa9c64837832e82baae49f9a643eb9

XenModel/BrandManager.cs
```diff
@@ -31,6 +31,7 @@
 using System.Diagnostics;
 using System.Reflection;
 using System.Resources;
+using System.Web.UI;
```
kc284, line 34: The changes in BrandManager.cs and AssemblyInfo.cs are not needed since we're not introducing any new branded values. If I remember well we said we would be adding two new entries `source_url` and `source_checksum` in the XCupdates.xml?

XenAdmin/MainWindow.cs
```diff
@@ -3276,5 +3277,15 @@ private void toolStripMenuItemCfu_Click(object sender, EventArgs e)
         {
             Updates.CheckForUpdates(true);
         }
+
```
kc284, line 3280: Aplogies if I'm missing something, but why have we added two items?

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -146,5 +147,50 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
                     dlg.ShowDialog(parent);
             }
         }
+
+        public static void DownloadSource(IWin32Window parent)
+        {
+            var saveSourceDialog = new SaveFileDialog();
```
kc284, line 153: It's a goog practice to construct and launch modal dialogs within a using directive so it gets disposed after use and releases OS resources.

XenModel/Actions/Updates/DownloadSourceAction.cs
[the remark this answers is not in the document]
kc284: 👍 
Or you could make the DownloadFileAction an abstract base class containing the shared bits and derive the other two from it.

kc284 (changes requested):
As above.

XenModel/BrandManager.cs
CitrixChris: source_url yes but i tought we werent bothered with the checksum on the source as it is just the source. We can though

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +136,66 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+
+                if (saveSourceDialog.ShowDialog() != DialogResult.OK)
+                {
+                    return;
+                }
+                outputPathAndFileName = saveSourceDialog.FileName;
+            }
+
+            Console.WriteLine(Updates.ClientVersions.Count);
+            var clientVersion = Updates.ClientVersions.First();
```
danilo-delbusso, line 170: probably meant `FirstOrDefault` here?

> The [First<TSource>(IEnumerable<TSource>)](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.first?view=netframework-4.8.1#system-linq-enumerable-first-1(system-collections-generic-ienumerable((-0)))) method throws an exception if source contains no elements. To instead return a default value when the source sequence is empty, use the [FirstOrDefault](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.firstordefault?view=netframework-4.8.1) method.

```suggestion
            var clientVersion = Updates.ClientVersions.FirstOrDefault();
```

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -31,13 +31,15 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Web.UI.WebControls;
```
danilo-delbusso, line 35: `using` here is not required 👀 

```suggestion
```

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +136,66 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+                downloadAndInstallClientAction.ReleaseDownloadedContent(true);
 
                 using (var dlg = new ErrorDialog(Messages.UPDATE_CLIENT_FAILED_INSTALLER_LAUNCH))
                     dlg.ShowDialog(parent);
             }
         }
+
+        public static void DownloadSource(IWin32Window parent)
+        {
+            var outputPathAndFileName = string.Empty;
```
danilo-delbusso, line 154: super nit picky but we don't use `string.Empty`:


```suggestion
            string outputPathAndFileName;
```

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +136,66 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+                    continue;
+
+                currentTasks = true;
+                break;
+            }
+
+            if (currentTasks && new Dialogs.WarningDialogs.CloseXenCenterWarningDialog(true).ShowDialog(parent) != DialogResult.OK)
+            {
+                downloadSourceAction.ReleaseDownloadedContent();
+                return;
```
danilo-delbusso, line 197: again nit picky but we can get rid of `return`;

```suggestion
```

danilo-delbusso (changes requested):
Gonna leave it at this for now, I'd rather wait for the merge conflicts to be addressed first as they're not super straight forward and I might just be commenting on stuff that's already been fixed on `master`

XenModel/Messages.resx
```diff
@@ -4995,9 +4995,45 @@ Plugging in untrustworthy USB devices to your computer may put your computer at
   <data name="DOWNLOAD_ELLIPSES" xml:space="preserve">
     <value>Download...</value>
   </data>
+  <data name="DOWNLOAD_FILE_ACTION_PROGRESS_DESCRIPTION" xml:space="preserve">
```
kc284, line 4998: There are a lot of duplicate keys here, this can't compile.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +135,64 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+
+                if (saveSourceDialog.ShowDialog() != DialogResult.OK)
+                {
+                    return;
+                }
+                outputPathAndFileName = saveSourceDialog.FileName;
+            }
+
+            var clientVersion = Updates.ClientVersions.FirstOrDefault();
+            if (clientVersion == null) { return; }
```
kc284, line 169: Trivial: return on a new line is clearer.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +135,64 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+            var sourceUrl = clientVersion.SourceUrl;
+            var downloadSourceAction = new DownloadSourceAction(clientVersion.Name, clientVersion.Version, new Uri(sourceUrl), outputPathAndFileName);
+
+            using (var dlg = new ActionProgressDialog(downloadSourceAction, ProgressBarStyle.Continuous))
+            {
+                dlg.ShowCancel = true;
+                dlg.ShowDialog(parent);
+            }
+
+            if (!downloadSourceAction.Succeeded)
```
kc284, line 179: I don't think the rest of this method is needed; we only save the zip, we don't need to close the application or disconnect, hence no need to check for unfinished tasks.

XenAdmin/MainWindow.resx
```diff
@@ -3183,6 +3183,12 @@
   <data name="pluginItemsPlaceHolderToolStripMenuItem8.Text" xml:space="preserve">
     <value>PluginItemsPlaceHolder</value>
   </data>
+  <data name="downloadLatestSourceToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
+    <value>186, 22</value>
+  </data>
+  <data name="downloadLatestSourceToolStripMenuItem.Text" xml:space="preserve">
+    <value>Download v{0} source.</value>
```
kc284, line 3190: No fullstop needed at the end.

XenAdmin/MainWindow.cs
```diff
@@ -2713,7 +2714,15 @@ private void SetClientUpdateAlert()
         {
             updateAlert = Updates.ClientUpdateAlerts.FirstOrDefault();
             if (updateAlert != null)
+            {
                 relNotesToolStripMenuItem.Text = string.Format(Messages.MAINWINDOW_UPDATE_RELEASE, updateAlert.NewVersion.Version);
+                downloadLatestSourceToolStripMenuItem.Text = string.Format(Messages.DOWNLOAD_SOURCE, updateAlert.NewVersion.Version);
```
kc284, line 2719: I think these two are the other way round, it's downloadSourceToolStripMenuItem that's associated with the alert, right?

Also, downloadLatestSourceToolStripMenuItem is always there even if the check does not run automatically, so I believe we need to format it (in the construcotr alongside other items) using the current version.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +135,64 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+                saveSourceDialog.InitialDirectory = Win32.GetKnownFolderPath(Win32.KnownFolders.Downloads);
+
+                if (saveSourceDialog.ShowDialog() != DialogResult.OK)
+                {
+                    return;
+                }
+                outputPathAndFileName = saveSourceDialog.FileName;
+            }
+
+            var clientVersion = Updates.ClientVersions.FirstOrDefault();
```
kc284, line 168: This method is also called from the item under the help menu when alerts are absent, in which case it won't download the source. Maybe make the clientVersion a parameter and set it accordingly in the item click handlers.

kc284 (changes requested):

XenAdminTests/XenAdminTests.csproj
```diff
@@ -163,7 +163,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\XenAdmin\XenAdmin.csproj">
-      <Project>{70BDA4BC-F062-4302-8ACD-A15D8BF31D65}</Project>
+      <Project>{70bda4bc-f062-4302-8acd-a15d8bf31d65}</Project>
```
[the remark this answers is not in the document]
CitrixChris: No idea, Visual studio must have deemed it necessary, ill revert it.

scripts/xenadmin-build.sh
```diff
@@ -204,6 +204,7 @@ echo "INFO: Generating XCUpdates.xml"
 
 # UPDATE_URL points at the updates XML, we need to point to the MSI
 msi_url="${XC_UPDATES_URL/XCUpdates.xml/$BRANDING_BRAND_CONSOLE_NO_SPACE.msi}"
+source_url="${XC_UPDATES_URL/XCUpdates.xml/$BRANDING_BRAND_CONSOLE_NO_SPACE-source.zip}"
```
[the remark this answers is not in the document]
CitrixChris: Ah no worries good spot

XenAdmin/MainWindow.cs
CitrixChris: Good spot on the alert-associated tooltip. 

Now that I've swapped them though I don't understand what you mean about its always there. It will be there if there is updates.clientversions has something in it. It always should be the latest version of XC so if I have the latest version I should show the button in the help menu so I can download that source. It will be hidden if the xcupdates.xml fails to download or doesn't exist.

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
CitrixChris: It will as the clientVersions is the pre-processed list of updates, it only becomes an update alert if the installed version < update version

XenModel/BrandManager.cs
```diff
@@ -52,6 +53,7 @@ static BrandManager()
             ProductVersionPost82 = customBranding.ProductVersionText;
             XcUpdatesUrl = customBranding.XcUpdatesUrl;
             CfuUrl = customBranding.CfuUrl;
+            SourceUrl = XcUpdatesUrl.Replace("XCUpdates.xml", "XenCenter-source.zip");
```
danilo-delbusso, line 56: We're not using this value in the `BrandManager`. Should we remove it?

XenAdmin/MainWindow.resx
danilo-delbusso, line 1: there's an awful lot of automated changes here that are hard to decipher like this. Is this VS22 doing its thing?

danilo-delbusso (changes requested):

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
```diff
@@ -134,17 +135,50 @@ public static void DownloadAndInstallNewClient(ClientUpdateAlert updateAlert, IW
...
+            if (clientVersion == null) 
+                return;
+            var sourceUrl = clientVersion.SourceUrl;
+            var downloadSourceAction = new DownloadSourceAction(clientVersion.Name, clientVersion.Version, new Uri(sourceUrl), outputPathAndFileName);
```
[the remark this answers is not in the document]
CitrixChris: Interesting as that is true current but it shouldn't ever be a problem once this is merged but i should handle it better just incase

XenAdmin/MainWindow.cs
```diff
@@ -2713,7 +2714,15 @@ private void SetClientUpdateAlert()
         {
             updateAlert = Updates.ClientUpdateAlerts.FirstOrDefault();
             if (updateAlert != null)
+            {
```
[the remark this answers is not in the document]
CitrixChris: We currently pull the url from the XCupdates.xml file so we would need to fetch that before we can download anything. 
So this is to cover the event that update checks are not automatic, so the menu item would need to trigger a fetch of the updates file and pull the source url. It would default to "Download latest source" instead of having the version number auto populated, I think this is doable.

You mentioning this made me notice a bug that if the auto check for updates is disabled the menu item under help is visible without a version number so I'll fix that to be by default invisible.

XenModel/BrandManager.cs
CitrixChris: probably

XenAdmin/Alerts/Types/ClientUpdateAlert.cs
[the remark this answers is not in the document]
CitrixChris: In fact no as the same would be true for clientversion.Name clientVersion.Version etc. Future XMLs will have a sourceURL value with this PR.

XenModel/Actions/Updates/DownloadFileAction.cs
```diff
@@ -52,32 +52,38 @@ public class DownloadAndUpdateClientAction : AsyncAction, IByteProgressAction
         //If you consider increasing this for any reason (I think 5 is already more than enough), have a look at the usage of SLEEP_TIME_BEFORE_RETRY_MS in DownloadFile() as well.
         private const int MAX_NUMBER_OF_TRIES = 5;
 
+        protected Uri Address => _address;
+        protected string OutputPathAndFileName => _outputPathAndFileName;
```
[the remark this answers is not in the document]
CitrixChris: This was done to reduce exposure of the class but still allow access.

CitrixChris:
Needs rebasing onto master after https://github.com/xenserver/xenadmin/pull/3223  is merged

kc284 (changes requested):
I have made one or two more corrections and some refactorging. This is the commit: https://github.com/kc284/xenadmin/commit/094bd183ef373de0fbe2a9f198a30a63678881a7

If you add me as collaborator to your fork, I can send a PR directly and also add comments to explain the change.

kc284 (approved):
I suggest we squash this when we merge it.

danilo-delbusso (approved):

PR #3154, opened 2023-06-05
Prevent `long` overflow when VM has a very large amount of memory

danilo-delbusso:
Resulted in XenCenter hanging with a significant memory leak.

I know what you're going to ask: "How could this possibly happen, anyway?"

Well it happened to me, so here's a PR to fix it 🤷 

To repro the issue "just" edit an OVF to have 10 exabytes of memory, import it, then open its Memory tab (or the host's)

### DIFF
--- a/XenAdmin/Controls/Ballooning/ShinyBar.cs
+++ b/XenAdmin/Controls/Ballooning/ShinyBar.cs
@@ -263,7 +263,7 @@ namespace XenAdmin.Controls.Ballooning
                         Size size = Drawing.MeasureText(g, label, Program.DefaultFont);
                         Rectangle rect = new Rectangle(new Point(pos - size.Width/2, text_top), size);
 
-                        if (LabelShouldBeShown(max, label, x))
+                        if (LabelShouldBeShown(max, x))
                         {
                             Drawing.DrawText(g, label, Program.DefaultFont, rect, Grid, Color.Transparent);
                         }
@@ -282,7 +282,7 @@ namespace XenAdmin.Controls.Ballooning
         /// <param name="label"></param>
         /// <param name="x"></param>
         /// <returns></returns>
-        private static bool LabelShouldBeShown(double max, string label, double x)
+        private static bool LabelShouldBeShown(double max, double x)
         {
             return max <= Util.BINARY_GIGA  || (x % (0.5 * Util.BINARY_GIGA)) == 0;
         }
--- a/XenAdmin/Controls/Ballooning/ShinyBar.cs
+++ b/XenAdmin/Controls/Ballooning/ShinyBar.cs
@@ -30,6 +30,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Drawing2D;
 using System.Drawing.Imaging;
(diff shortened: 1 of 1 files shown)


XenAdmin/Controls/Ballooning/ShinyBar.cs
```diff
@@ -276,7 +282,7 @@ protected void DrawGrid(Graphics g, Rectangle barArea, double bytesPerPixel, dou
         /// <param name="label"></param>
         /// <param name="x"></param>
         /// <returns></returns>
-        private static bool LabelShouldBeShown(double max, string label, long x)
+        private static bool LabelShouldBeShown(double max, string label, double x)
```
kc284, line 285: Since you're here, the label param is unused.

XenAdmin/Controls/Ballooning/ShinyBar.cs
```diff
@@ -244,7 +250,7 @@ protected void DrawGrid(Graphics g, Rectangle barArea, double bytesPerPixel, dou
             using (Pen pen = new Pen(Grid))
             {
                 bool withLabel = true;
-                for (long x = 0; x <= max; x += incr)
+                for (double x = 0; x <= max; x += incr)
                 {
                     // Tick
                     int pos = barArea.Left + (int)((double)x / bytesPerPixel);
```
kc284, line 256: Unnecessary cast of `x` to double. Same further up in line 246 with `incr` (but it won't let me add a comment there).

XenAdmin/Controls/Ballooning/ShinyBar.cs
```diff
@@ -222,6 +222,12 @@ private static GraphicsPath GetRoundedPath(RectangleF rect, float radius)
 
         protected void DrawGrid(Graphics g, Rectangle barArea, double bytesPerPixel, double max)
         {
+            // prevent max overflows from creating an infinite loop
+            if (max <= 0)
```
kc284, line 226: Is this check needed? `max` is always xapi values, if it ever came here as negative, we'd rather want to know about it instead of masking it. If you must, add perhaps a Debug assertion instead?

kc284 (changes requested):

XenAdmin/Controls/Ballooning/ShinyBar.cs
danilo-delbusso: The check is here just because if the xapi values are wrong, XenCenter loops indefinitely. They obviously shouldn't be wrong, but it can happen (happened to me 😅, so it's possible)

XenAdmin/Controls/Ballooning/ShinyBar.cs
danilo-delbusso: If `max` is a double, the same should apply to `x` and `incr` since `long`s cannot handle as large a set of numbers.

If anything, I think `incr` should be a `double` - that was my intention at least, I just missed it

XenAdmin/Controls/Ballooning/ShinyBar.cs
kc284: I believe it was x that overflowed rather than max, I don't see how max can have overflown, if xapi values were negative, XC would have probably crashed before getting there. I'm not against defensive programming, it's just that in this case the check might cause worse confusion.

danilo-delbusso: I see your point, it's not very clear to someone that hasn't seen this before. I'll stick with Debug.Assert then 👍

kc284 (approved):

CitrixChris (approved):

PR #3155, opened 2023-06-05
Fixed two issues when saving pool properties

kc284:
I noticed these issues when working on the `merge` branch and I had to fix them now because they are affecting me when dev-testing. However, they are issues on `master` and need to be considered separately, so I'm raising this PR against `master` and we can pull `master` into `merge` at a later time.

The issue with the unlocked object is another regression from #3077 , whereby on some occasions saving the pool properties was failing with the error "the object is not locked".

The other one is an older issue whereby when one opened the pool Properties dialog and clicked OK without changing anything, the Save actions was still run. I suppose I could have fixed this with fewer changes, but the original implementation was too complicated and could be simplified. The designer and resx changes were autogenerated when I switched the default language to English (for some reason it was set to Japanese).


XenAdmin/SettingsPanels/PerfmonAlertOptionsPage.cs
```diff
@@ -62,117 +63,112 @@ public PerfmonAlertOptionsPage()
...
-
-                return EmailAddressTextBox.Text;
-            }
+                IsBalloon = true,
+                ToolTipIcon = ToolTipIcon.Warning,
+                ToolTipTitle = Messages.INVALID_PARAMETER
+            };
+
+            MailLanguageComboBox.Items.Add(new ToStringWrapper<string>(Messages.MAIL_LANGUAGE_ENGLISH_CODE, Messages.MAIL_LANGUAGE_ENGLISH_NAME));
+            MailLanguageComboBox.Items.Add(new ToStringWrapper<string>(Messages.MAIL_LANGUAGE_CHINESE_CODE, Messages.MAIL_LANGUAGE_CHINESE_NAME));
```
CitrixChris, line 74: Do we need these if we have dropped localisation?

CitrixChris (changes requested):

XenAdmin/Dialogs/PropertiesDialog.cs
```diff
@@ -115,25 +114,25 @@ private void Build()
...
             bool is_pool_or_standalone = is_pool || (is_host && !is_in_pool);
 
-            bool wlb_enabled = (Helpers.WlbEnabledAndConfigured(xenObjectCopy.Connection));
+            bool wlb_enabled = (Helpers.WlbEnabledAndConfigured(_xenObjectCopy.Connection));
 
-            bool is_VM_appliance = xenObjectCopy is VM_appliance;
+            bool is_VM_appliance = _xenObjectCopy is VM_appliance;
 
-            bool is_VMSS = xenObjectCopy is VMSS;
+            bool is_VMSS = _xenObjectCopy is VMSS;
```
danilo-delbusso, line 135: nit picky but since you've renamed variables already let's get rid of snake case here for all of these. ReSharper should make it painless

danilo-delbusso (changes requested):

CitrixChris (approved):

XenAdmin/Dialogs/PropertiesDialog.cs
kc284: Done;  I also made the checks slightly more efficient.

PR #3159, opened 2023-06-16
CA-378482: Fix cancellation and multi-thread support to `ConversionWizard` `BackgroundWorker`s

danilo-delbusso:
- `DoWorkEventArgs.Cancel` was never actually set to `true`, so `RunWorkerCompletedEventArgs.Cancelled` was always false. This adds correct implementation of cancellation mechanisms for `BackgroundWorker`s
- Ensure no UI changes are made if a `BackgroundWorker` completes after a page has been disposed (by for instance closing the wizard)
- Moving back and fourth between pages sometimes caused issues because `BackroundWorker.RunWorkerAsync` calls were made before the previous run finished. This "enqueues" another run if one hasn't completed
- Update accessibility letter in `SelectMultipleNetworkPage` from `P` to `M` since it clashed with the `Previous` button

I have used the same pattern for every page. One could argue that we need a `XenTabPage` subclass with multithread support to load things on load but that was just a bigger refactor that was going to take ages to review and we have bigger fish to fry at the moment I'm afraid

### DIFF
--- a/XenAdmin/Wizards/ConversionWizard/CredentialsPage.cs
+++ b/XenAdmin/Wizards/ConversionWizard/CredentialsPage.cs
@@ -44,6 +44,8 @@ namespace XenAdmin.Wizards.ConversionWizard
         private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
         private bool _buttonNextEnabled;
+        private bool _runWorkerAgain;
+        private readonly object _workerRunLock = new object();
 
         public CredentialsPage()
         {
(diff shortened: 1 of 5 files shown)


CitrixChris (approved):

kc284 (changes requested):
This solution has some issues and I think it addresses only partially the cause of the problem.

First, there are locks in places that are not needed: the event handlers of a wizard page's LoadedCore/LeaveCore are not called by various threads, they are all on one thread (the main one); the code in backgroundWorker's DoWork is on a different thread, however, its Completed event is on the main thread (there may be occurences in XC where invoke is called in the event handler, but this doesn't mean they're correct).

Second, I think the main cause for the crash is the background workers do not have their events unregistered and are not disposed. One should not have to check if the parent page is disposed in the backgroundWorker's Completed event.  And we may be missing cancellations when moving backwards or checks when moving forwards; if that worked ok we might not need to check an additional bool (_runWorkerAgain).

I'd keep the checks within DoWork for pending cancellations. I'm not very confident about the change in the use of Populate in SelectMultipleVMNetworkPage because it's used in other wizards with slightly different logic. But it may not be needed if the above changes are made.

PR #3164, opened 2023-06-28
CP-43070 adds leaf coalesce alert

CitrixChris:


XenModel/Messages.resx
```diff
@@ -7616,6 +7616,24 @@ SR UUID: {1}</value>
   <data name="LAUNCH_LICENSE_MANAGER" xml:space="preserve">
     <value>License Manager</value>
   </data>
+  <data name="LEAF_COALESCE_COMPLETED_DESCRIPTION" xml:space="preserve">
+    <value>Data merge for VM {0} completed</value>
+  </data>
+  <data name="LEAF_COALESCE_COMPLETED_TITLE" xml:space="preserve">
```
kc284, line 7622: From what I understood from the task's description, we don't want to show the leaf coalescing text as is, which means we either need to make the alert's Title return the same as the Description, or ask @MarkSymsCtx whether it night be useful for the user to also see the VDI's name in the detailed description (e.g. something like "Data is being merged for VM x on disk y"?).

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        public override string Description
+        {
+            get
+            {
+                switch (LeafType)
+                {
+                    case MessageType.LEAF_COALESCE_START_MESSAGE:
+                        return string.Format(Messages.LEAF_COALESCE_START_DESCRIPTION, vmID);
```
kc284, line 75: As mentioned above, `vm.Name()` is user friendlier than the uuid.

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
```diff
@@ -0,0 +1,112 @@
...
+            MapVdiToVm(msg);
+        }
+
+        private void MapVdiToVm(Message msg)
+        {            
+            var obj = msg.GetXenObject();
+            
+            if (obj is VDI vdi && vdi.Connection != null)
+            {
+                foreach (VBD vbd in vdi.Connection.ResolveAll(vdi.VBDs))
```
kc284, line 56: I know code calling ResolveAll is used all over the application, however, a loop like this is  more effcient:
```
foreach (var vbdRef in vdi.VBDs)
{
    var vbd = vdi.Connection.Resolve(vbdRef);
    
    if (vbd != null && vbd.currently_attached)
    {
        VM vm = vbd.Connection.Resolve(vbd.VM);
        
        if (vm != null)
        {
            _vm = vm;
            break;
        }
    }
}
```
because it resolves only as many VBDss as needed (instead of all), and avoids resolving the VM for detached VDIs.

kc284 (changes requested):

XenModel/FriendlyNames.resx
```diff
@@ -1586,6 +1586,15 @@
   <data name="Message.help-host_sync_data_failed" xml:space="preserve">
     <value>Help on HA</value>
   </data>
+  <data name="Message.help-leaf_coalesce_completed" xml:space="preserve">
```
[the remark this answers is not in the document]
CitrixChris: Arent they needed to avoid debug assert that x doesn't exist in Friendly names?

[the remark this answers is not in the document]
kc284: They would be needed if that code were hit, but it won't be hit since you have handled these cases already in Title and Description using entries from Messages.resx.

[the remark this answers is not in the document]
CitrixChris: wonderful understood

XenModel/Messages.Designer.cs
```diff
@@ -21874,6 +21874,33 @@ public class Messages {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized string similar to Data merge for VDI &apos;{0}&apos; of VM &apos;{1}&apos; has completed..
```
kc284, line 21878: - This file is autogenerated from Messages.resx. Any changes to the keys and their values should be made in the resx file, because next time something is added in the latter, the compilation will wipe out the manual changes in the Designer.cs.
- Since we still have the `*_TITLE` and `*_DESCRIPTION` keys, I would suggest using them, i.e. add the VDI name in the description only to make it more detailed than the title.

XenModel/FriendlyNames.resx
```diff
@@ -2006,4 +2006,4 @@
   <data name="VM.TemplateDescription-Other install media" xml:space="preserve">
     <value>If the operating system you plan to use is not listed, you may be able to install it by selecting this template.  We recommend that only advanced users attempt to use this template, and you should note that our products have been tested running only the supported distributions and specific versions covered by the standard supplied templates.</value>
   </data>
-</root>
+</root>
```
kc284, line 2009: Very trivial, but since you'll be making other changes, you can also revert this file, this change is not needed.

kc284 (changes requested):

XenModel/Messages.Designer.cs
CitrixChris: I removed those keys as they were now going to be the same. Happy to revert tho and make this change :)

CitrixChris: Not sure how ive achieved this. That was what I had intended but clearly something went awry.

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
```diff
@@ -0,0 +1,114 @@
...
+using static XenAPI.Message;
+
+namespace XenAdmin.Alerts
+{
+    public class LeafCoalesceAlert : MessageAlert
+    {
+        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
+
+        private readonly VM _vm;
+        private readonly VDI _vDI;
```
danilo-delbusso, line 41: super nit picky so only if you're making other change: should we rename `_vDI` -> `_vdi`?

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        public override string HelpID
+        {
+            get
+            {
+                return "LeafCoalesceAlert";
+            }
+        }
```
danilo-delbusso, line 112: again very nit picky so only if you're making other changes:


```suggestion
      public override string HelpID => "LeafCoalesceAlert";
```

danilo-delbusso (approved):

XenModel/FriendlyNames.Designer.cs
```diff
@@ -4463,6 +4463,33 @@ public class FriendlyNames {
             }
         }
         
+        /// <summary>
```
kc284, line 4466: This file needs to be reverted too since the resx was reverted.

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
```diff
@@ -0,0 +1,114 @@
...
+                switch (Message.Type)
+                {
+                    case MessageType.LEAF_COALESCE_START_MESSAGE:
+                        return string.Format(Messages.LEAF_COALESCE_START_TITLE, _vm.Name());
+                    case MessageType.LEAF_COALESCE_COMPLETED:
+                        return string.Format(Messages.LEAF_COALESCE_COMPLETED_TITLE, _vm.Name());
+                    case MessageType.LEAF_COALESCE_FAILED:
+                        return string.Format(Messages.LEAF_COALESCE_FAILED_TITLE, _vm.Name());
+                    default:
+                        return base.Title; ;
```
kc284, line 101: My VS cannot compile the code, it's complaining about the double semi-colon here.

kc284 (changes requested):

kc284:
This PR has a lot of writing and re-writing; GIven we need one more commit and the total line diff count is relatively small, I suggest squashing all the commits into the first one to clean the history a bit.

kc284:
After testing, I discovered a few more issues, which I'm attaching here as a patch - I tried to raise a PR to your fork, but for some reason it doesn't turn up on the list, have you by any chance disabled PRs?

To summarise, we need to override also FixLinkAction and HelpLinkText (now I understand that you had added FriendlyNames for those two properties, but that was not the right solution). Also, I noticed that if the VDI is detached or if it's deleted or if the VM is deleted, the messages are not removed from the server. Not sure whether this is a server bug or not, but in order to avoid delays, let's handle it here.

Other than these it looks ok to me, I didn't see more breakages, so let's apply the attached, have Danilo also review it, and we should be good to go.
[xenadmin-cea44b4-Some more corrections.patch](https://github.com/xenserver/xenadmin/files/12187620/xenadmin-cea44b4-Some.more.corrections.patch)

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
```diff
@@ -28,6 +28,7 @@
  * SUCH DAMAGE.
  */
 
+using System;
```
danilo-delbusso, line 31: nit but do we need this now?

danilo-delbusso (approved):

XenAdmin/Alerts/Types/LeafCoalesceAlert.cs
CitrixChris: Yes as it is needed for the Action type on line 111

kc284 (approved):

PR #3167, opened 2023-07-03
CA-379402: Fix EUA check & add hotfix check to `HostMemoryPostUpgradeCheck`

danilo-delbusso:
- b834a7e842b3042caa716c92c33b491027bc7769 ensures we do not run EUA check for upgrades for pre-82X
- 0dfcc3852abdba193ce0bb5dcf304b2931eb5f6c adds a missing hotfix presence check for `HostMemoryPostUpgradeCheck`

### DIFF
--- a/XenAdmin/Diagnostics/Checks/UpgradeRequiresEUA.cs
+++ b/XenAdmin/Diagnostics/Checks/UpgradeRequiresEUA.cs
@@ -49,12 +49,15 @@ namespace XenAdmin.Diagnostics.Checks
         private readonly Control _control;
         private readonly HashSet<string> _euas;
         private readonly HashSet<IXenObject> _hostsFailedToFetchEua;
-        public UpgradeRequiresEua(Control control, List<Host> hosts, IReadOnlyDictionary<string, string> installMethodConfig)
+        private readonly Dictionary<string, string> _installMethodConfig;
+
+        public UpgradeRequiresEua(Control control, List<Host> hosts, Dictionary<string, string> installMethodConfig)
             : base(hosts)
         {
             if (installMethodConfig == null || !installMethodConfig.TryGetValue("url", out var uriText))
                 return;
 
+            _installMethodConfig = installMethodConfig;
             _control = control;
 
             try
@@ -70,7 +73,7 @@ namespace XenAdmin.Diagnostics.Checks
             _hostsFailedToFetchEua = new HashSet<IXenObject>();
         }
 
-        public override bool CanRun() => _targetUri != null && Hosts.Any(Helpers.YangtzeOrGreater);
+        public override bool CanRun() => _targetUri != null;
 
         private void FetchHostEua(Host host)
         {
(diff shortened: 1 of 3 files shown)


XenAdmin/Diagnostics/Checks/UpgradeRequiresEUA.cs
```diff
@@ -90,15 +93,31 @@ private void FetchHostEua(Host host)
...
                 if (hotfix != null && hotfix.ShouldBeAppliedTo(host))
                     return new HostDoesNotHaveHotfixWarning(this, host);
             }
 
+            string upgradePlatformVersion = null;
+            if (_installMethodConfig != null)
+                Host.TryGetUpgradeVersion(Hosts.FirstOrDefault(), _installMethodConfig, out upgradePlatformVersion, out _);
+
+            // There's no EUA for Pre-82X versions
+            if (!Helpers.Post82X(upgradePlatformVersion))
```
jijiang, line 116: Is it checking source XS version or destination XS version? The test failed on RPU from Yangtze to Master, will this use case return null here?

danilo-delbusso: This new check is for the destination XS. 

The test failed from Yangtze to Master because XenCenter could not fetch the EUA (was not present), that's not a XenCenter issue.

This is addressing upgrades to non-Master. e.g. Stockolhm to Yangtze.

I do need to make a further tweak to support this, however. Will send in a new change

danilo-delbusso: True but we should still add a version check, and while we're at it prevent EUA from showing when we know it's not there, anyway

jijiang (approved):

XenAdmin/Diagnostics/Checks/UpgradeRequiresEUA.cs
```diff
@@ -70,7 +73,7 @@ public UpgradeRequiresEua(Control control, List<Host> hosts, IReadOnlyDictionary
             _hostsFailedToFetchEua = new HashSet<IXenObject>();
         }
 
-        public override bool CanRun() => _targetUri != null && Hosts.Any(Helpers.YangtzeOrGreater);
+        public override bool CanRun() => _targetUri != null;
```
kc284, line 76: I noticed two more issues, the code is in fact in RollingUpgradeWizardPrecheckPage.cs
```
if (!ManualUpgrade)
{
    // EUA check
    var euaCheck = GetPermanentCheck("EUA",
        new UpgradeRequiresEua(this, hostsToUpgrade, InstallMethodConfig));
    var euaChecks = new List<Check> { euaCheck };
    groups.Add(new CheckGroup(Messages.ACCEPT_EUA_CHECK_GROUP_NAME, euaChecks));

    var safeToUpgradeChecks = (from Host host in hostsToUpgrade
        let check = new SafeToUpgradeCheck(host, InstallMethodConfig)
        where check.CanRun()
        select check as Check).ToList();

    if (safeToUpgradeChecks.Count > 0)
        groups.Add(new CheckGroup(Messages.CHECKING_SAFE_TO_UPGRADE, safeToUpgradeChecks));

    var prepareToUpgradeChecks = (from Host host in hostsToUpgrade
        select new PrepareToUpgradeCheck(host, InstallMethodConfig) as Check).ToList();

    if (prepareToUpgradeChecks.Count > 0)
        groups.Add(new CheckGroup(Messages.CHECKING_PREPARE_TO_UPGRADE, prepareToUpgradeChecks));
}
```

- In the above code where we construct the UpgradeRequiresEua we don't check if it CanRun(), so it runs successfully even if the Uri is invalid (because the constructor just logs the error quietly in this case).
- This is an older issue: I find a bit strange that the PrepareToUpgradeCheck, which checks for invalid installer URLs, comes last  in this block. I think it should be first.

kc284 (changes requested):

XenAdmin/Diagnostics/Checks/UpgradeRequiresEUA.cs
danilo-delbusso: > In the above code where we construct the UpgradeRequiresEua we don't check if it CanRun(), so it runs successfully even if the Uri is invalid (because the constructor just logs the error quietly in this case).

Actually `CanRun` runs as part of `RunAllChecks()`, which is what actually runs `RunCheck()`.

However, there is a commment I missed that points out that the check should not be added at all if `CanRun` isn't valid, so this needs changing anyway

> This is an older issue: I find a bit strange that the PrepareToUpgradeCheck, which checks for invalid installer URLs, comes last in this block. I think it should be first.

sure! 👍

kc284 (approved):

PR #3169, opened 2023-07-11
CA-379640, CA-356178, CA-370866: Do not reset remote consoles if they're still attempting connection to a VM

danilo-delbusso:
I split into commits so it's easier to review them, since the PR contains a lot of tidying up for `XSVNCScreen`.

If you want to reproduce the issue this is fixing, just add multiple calls to `ThreadPool.QueueUserWorkItem(TryToConnectRDP);` in `VNCTabView.cs:1178`:

```csharp
                    if (vncScreen.RdpIp == null)
                        toggleConsoleButton.Enabled = false;
                    ThreadPool.QueueUserWorkItem(TryToConnectRDP);
                    ThreadPool.QueueUserWorkItem(TryToConnectRDP);
                    ThreadPool.QueueUserWorkItem(TryToConnectRDP);
                    ThreadPool.QueueUserWorkItem(TryToConnectRDP);
                    ThreadPool.QueueUserWorkItem(TryToConnectRDP);
```

The changes in 2f8dde440622a70da699906e60f64398bf427cd8 prevent `AccessViolationException`s being called, as they're hit when two instances of `AxMSTSCLib.AxMsRdpClient9` call `Connect` on the same IP + Port combination, which results in the same memory being accessed in unmanaged code.

Also as part of this PR:
- [Add IRdpClient to abstract some common functionality of RDP clients](https://github.com/xenserver/xenadmin/commit/7067629b29e194d3842887c472b3bfee3e704c5f)

    The reason this wasn't added to the `AxMsRdpClientX` classes is that I thought it best to not touch `AxMSTSCLib` classes as they are likely generated.

- [Catch HRESULT E_FAIL exceptions for IRDPClient.Connect calls](https://github.com/xenserver/xenadmin/commit/9ea0a53447d4a4bcd5fb1db7890467cfe52ec31d)

   The `Connect` method returns `E_FAIL` if it is called while the control is already connected or in the connecting state. This can be hit when a lot of connections are being opened at the same time, and it's there as a failsafe.

   Also adds `IsAttemptingConnection` as a new field

- [Dispose of timer when creating new port polling Timer](https://github.com/xenserver/xenadmin/pull/3169/commits/60a5580d207d7b1df4fdf1b57ad5715f66b0317b)
  
  While testing I found that there were a couple of ways whereby multiple `Timer`s could be running indefinitely (unless `XSVNCScreen.cs` was disposed of). This happened because we were not disposing of the `Timer` in `StartPolling`.

### DIFF
--- a/XenAdmin/ConsoleView/RdpClient.cs
+++ b/XenAdmin/ConsoleView/RdpClient.cs
@@ -97,6 +97,11 @@ namespace XenAdmin.ConsoleView
             rdpControl.Resize += resizeHandler;
         }
 
+        private bool _connecting;
+        private bool _authWarningVisible;
+
+        public bool IsAttemptingConnection => _connecting || _authWarningVisible; 
+
         private void RDPConfigure(Size currentConsoleSize)
         {
             rdpControl.BeginInit();
@@ -104,7 +109,7 @@ namespace XenAdmin.ConsoleView
             rdpControl.Dock = DockStyle.None;
             rdpControl.Anchor = AnchorStyles.None;
             rdpControl.Size = currentConsoleSize;
-            RDPAddOnDisconnected();
+            AddRDPEventHandlers();
             rdpControl.Enter += RdpEnter;
             rdpControl.Leave += rdpClient_Leave;
             rdpControl.GotFocus += rdpClient_GotFocus;
@@ -123,15 +128,28 @@ namespace XenAdmin.ConsoleView
             }
         }
 
-        private void RDPAddOnDisconnected()
+        private void AddRDPEventHandlers()
         {
             if (rdpControl == null)
                 return;
 
-            if (rdpClient9 == null)
-                rdpClient6.OnDisconnected += rdpClient_OnDisconnected;
-            else
-                rdpClient9.OnDisconnected += rdpClient_OnDisconnected;
+            var rdpClient = (IRdpClient)rdpClient9 ?? rdpClient6;
+            if (rdpClient == null)
+            {
+                return;
+            }
+
+            rdpClient.OnDisconnected += (_, e) =>
+            {
+                Program.AssertOnEventThread();
+                OnDisconnected?.Invoke(this, EventArgs.Empty);
+            };
+            rdpClient.OnConnected += (_1, _2) => _connecting = false;
+            rdpClient.OnConnecting += (_1, _2) => _connecting = true;
+            rdpClient.OnDisconnected += (_1, _2) => _connecting = _authWarningVisible = false;
+            rdpClient.OnAuthenticationWarningDisplayed += (_1, _2) => _authWarningVisible = true;
+            rdpClient.OnAuthenticationWarningDismissed += (_1, _2) => _authWarningVisible = false;
+
         }
 
         private void RDPSetSettings()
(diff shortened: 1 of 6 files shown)


CitrixChris (changes requested):

CitrixChris (approved):

XenAdmin/ConsoleView/XSVNCScreen.cs
```diff
@@ -474,100 +474,118 @@ private void SetPendingVNCConnection(Stream s)
...
+            var wasFocused = false;
+            Controls.Clear();
             //console size with some offset to accomodate focus rectangle
-            Size currentConsoleSize = new Size(this.Size.Width - CONSOLE_SIZE_OFFSET, this.Size.Height - CONSOLE_SIZE_OFFSET);
-                
-            // Stop the old client.
-            if (RemoteConsole != null)
+            var currentConsoleSize = new Size(Size.Width - CONSOLE_SIZE_OFFSET, Size.Height - CONSOLE_SIZE_OFFSET);
+
+            lock (_rdpConnectionLock)
```
kc284, line 527: Why is the lock needed? The method is always called on the main thread (there is an assertion at the beginning of the method), so it's not likely that two threads will be accessing it.

XenAdmin/ConsoleView/RdpClient.cs
```diff
@@ -123,15 +128,28 @@ public Point rdpLocationOffset
...
-                rdpClient9.OnDisconnected += rdpClient_OnDisconnected;
+            var rdpClient = (IRdpClient)rdpClient9 ?? rdpClient6;
+            if (rdpClient == null)
+            {
+                return;
+            }
+
+            rdpClient.OnDisconnected += (_, e) =>
+            {
+                Program.AssertOnEventThread();
```
kc284, line 144: Not your change, but why is an assertion needed here? Is it likely that it will be called from a thread other than the UI? Alternatively, if that were likely, then you'd probably want to make _connecting and _authWarningvisible volatile.

XenAdmin/ConsoleView/XSVNCScreen.cs
danilo-delbusso: I thought it might prevent future issues. In general my approach was to try and limit multi-thread access to `RemoteConsole` so this is just part of that.

XenAdmin/ConsoleView/RdpClient.cs
danilo-delbusso: > Not your change, but why is an assertion needed here? Is it likely that it will be called from a thread other than the UI? Alternatively, if that were likely, then you'd probably want to make _connecting and _authWarningvisible volatile (because that might be the case for the rest of the events).

I actually remember thinking the same. Wasn't going to mess with it though 😂 

I'd leave it since in the future we'd like to refactor this? 🤞

XenAdmin/ConsoleView/XSVNCScreen.cs
danilo-delbusso: > If a different thread accesses this code, the assertion will throw a Fatal error before you get to accessing the RemoteConsole.
good point 👍 will remove it

kc284 (approved):

PR #3177, opened 2023-07-19
CA-380218: It was not possible to create VMs with VDIs smaller than the max disk…

kc284:
… size when the summed size exceeded the latter limit. The limitation of 2TB for non-GFS2 SRs is on the size of individual VDIs, not the total.


danilo-delbusso (approved):
pretty straight forward

CitrixChris (approved):

PR #3178, opened 2023-07-20
CP-43264: Handle new alert for CSS agreement expiry

kc284:
The commit with the Repositories is needed for the following PRs, but I squeezed it in this PR (which is smaller) to divide the review load.

### DIFF
--- a/XenAdmin/Alerts/Types/CssExpiryAlert.cs
+++ b/XenAdmin/Alerts/Types/CssExpiryAlert.cs
@@ -1,0 +1,108 @@
+﻿/* Copyright (c) Cloud Software Group, Inc. 
+ * 
+ * Redistribution and use in source and binary forms, 
+ * with or without modification, are permitted provided 
+ * that the following conditions are met: 
+ * 
+ * *   Redistributions of source code must retain the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer. 
+ * *   Redistributions in binary form must reproduce the above 
+ *     copyright notice, this list of conditions and the 
+ *     following disclaimer in the documentation and/or other 
+ *     materials provided with the distribution. 
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND 
+ * CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, 
+ * INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF 
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE 
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR 
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, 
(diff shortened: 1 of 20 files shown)


XenAdmin/Help/HelpManager.resx
```diff
@@ -270,6 +270,9 @@
   <data name="CrossPoolMigrateWizard_TransferNetworkPane" xml:space="preserve">
     <value>vms-relocate</value>
   </data>
+  <data name="CssExpiryAlert" xml:space="preserve">
+    <value></value>
```
kc284, line 274: This is empty as I don't know the value yet.

XenAdmin/Alerts/Types/CssExpiryAlert.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                if (nodes.Count > 0 && Util.TryParseIso8601DateTime(nodes[0].InnerText, out var result))
+                    _cssExpiryDate = result;
+            }
+            catch
+            {
+                //ignore
+            }
+
+            switch (Message.Type)
```
danilo-delbusso, line 62: nit do you think we should add a `default`? there's a few `Message.Type` we're not supporting here

XenModel/Messages.resx
```diff
@@ -1834,6 +1834,27 @@ This alarm is set to be triggered when the total throughput exceeds {4}.</value>
   <data name="ALERT_CPUS_SUB_TEXT" xml:space="preserve">
     <value>When CPU usage exceeds {0}% for {1} min(s)</value>
   </data>
+  <data name="ALERT_CSS_EXPIRED_DESCIRPTION" xml:space="preserve">
+    <value>Your Customer Success Services (CSS) agreement expired on {0}.</value>
+  </data>
+  <data name="ALERT_CSS_EXPIRED_LINK_TEXT" xml:space="preserve">
+    <value>Renew CSS Agreement</value>
```
danilo-delbusso, line 1841: should we then change it to the wording used in the license changes PR, or am I changing it there?

danilo-delbusso (approved):
approved because only nit picky comments

also we should figure out what's going into `ALERT_CSS_EXPIRED_LINK_TEXT` before merging

XenModel/Network/Cache.cs
CitrixChris, line 1: While I can see there are a couple of new lines in this file a lot seem to have simply been moved around and I cant understand why. I'm assuming it has something to do with line 48/49 "Keep Sorted please". 
What is sorting method?

CitrixChris (changes requested):

XenModel/Network/Cache.cs
kc284: The fields are entered manually, I don't think there is a particular reason to keep them sorted other than to help developers find easily whether a new API class of objects is cached already. Messy files do tend to lead to duplication of code and occasionally bugs.

CitrixChris (approved):

XenModel/Messages.resx
kc284: I understood that in the case of expiry the license manager should also read `Renew CSS agreement`.

kc284:
@danilo-delbusso in your comment above do you mean `CssExpiryAlert` rather than `ALERT_CSS_EXPIRED_LINK_TEXT`?

XenAdmin/Alerts/Types/CssExpiryAlert.cs
kc284: Static analysers will flag this, in reality though it is not needed because the constructor of the class is only called for the listed message types, hence  we don't expect any surprises (compare with the ParseMessage method calling the constructor, where indeed we need the default clause for unknown messages).

PR #3179, opened 2023-07-24
CP-43265: Updates config dialog. CP-42066: Rebranding tweaks. Retrieve available updates. Reflect status on icons.

kc284:
Each commit is self contained, please review them separately.


danilo-delbusso (changes requested):
found only one small thing that needs changing

CitrixChris (approved):

kc284:
After testing, I discovered a few more issues, which I'm attaching here as a patch - I tried to raise a PR to your fork, but for some reason it doesn't turn up on the list, have you by any chance disabled PRs?

To summarise, we need to override also FixLinkAction and HelpLinkText (now I understand that you had added FriendlyNames for those two properties, but that was not the right solution). Also, I noticed that if the VDI is detached or if it's deleted or if the VM is deleted, the messages are not removed from the server. Not sure whether this is a server bug or not, but in order to avoid delays, let's handle it here.

Other than these it looks ok to me, I didn't see more breakages, so let's apply the attached, have Danilo also review it, and we should be good to go.
[xenadmin-af1f9dd-Some more corrections.patch](https://github.com/xenserver/xenadmin/files/12186482/xenadmin-af1f9dd-Some.more.corrections.patch)

danilo-delbusso (approved):

PR #3180, opened 2023-07-25
CA-380355 & CA-380545: Fix creation of Desktop shortcut in installer

danilo-delbusso:
As discussed with the team, the shortcut has now been moved back to the CustomDialog dialog.

Furthermore, I have removed `RemoveFolder` for the Desktop folder, as it's only really meant for deleting empty folders, and causes permission issues if users without write access to delete the folder use the installer.

Thank you @edwintorok for the help with this!


kc284:
>  as it's only really meant for deleting empty folders
What exactly was this deleting, subfolders of the user's Desktop that contained the shortcut in question? Not sure I understand how they were created and what happens now that the option is removed.

danilo-delbusso:
> What exactly was this deleting, subfolders of the user's Desktop that contained the shortcut in question? Not sure I understand how they were created and what happens now that the option is removed.

Its function is to delete the specified folder if that folder is empty.

It didn't matter in the past because deleting the Desktop folder isn't allowed AFAIK? We just noticed now because the `RemoveFolder` action now runs both on install and uninstall, while in the past it was only running on uninstall.

It just makes no sense for Desktop, so I have removed it.

CitrixChris (approved):

neatleygit (approved):

kc284 (changes requested):
There are dos line endings in diff metadata lines, the build may fail, better keep them as unix line endings.

kc284 (changes requested):
As above.

kc284:
I noticed one more thing, but this may have been introduced at a previous commit, and not sure if it's a real issue since I'm testing with all sorts of hacked builds: I had on a VM a build from the feature/cdn-updates branch, then installed on top a build from this PR opting in for the desktop shortcut, and it created a second entry in the Control panel > Programs.

kc284 (approved):
The top blurb has now moved dangerously close to the top border, but since it's not cropped I'm ok with merging it.

PR #3184, opened 2023-07-31
CA-375900: Prepend `\\?\` to file paths when creating streams for archive generation

danilo-delbusso:
You'll need a host/pool with GFS2 storage to replicate this, as it creates large paths in `%temp%` when generating the Server Status Report.

The string works to enable creation of files with paths larger than 260 characters. This solution avoids us having to update registry keys which is an unwanted side effect.

7zip does this already so bypassing Windows restrictions has precedents. Plus this restriction isn't even relevant for our uses. We only really use it to build the archive.

Resources:
- [Naming Files, Paths, and Namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation)
- [Maximum Path Length Limitation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry)
- [CreateFileA](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea)


XenCenterLib/StringUtility.cs
```diff
@@ -218,5 +218,21 @@ public static bool IsIPAddress(string s)
...
+        /// <summary>
+        /// To be used to format file paths for File streams. <br />
+        /// Prepends \\?\ to path if it is longer than the default MAX_PATH in Windows. <br /><br />
+        /// See <see href='https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea'>CreateFileA</see> and <see href='https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry'>Maximum Path Length Limitation</see> for more info.
+        /// </summary>
+        /// <param name="inputPath">The input file path</param>
+        /// <returns></returns>
+        public static string FormatFilePathForFileStream(string inputPath)
+        {
+            if (string.IsNullOrEmpty(inputPath) || inputPath.Length <= 260 || inputPath.StartsWith("\\\\?\\"))
```
kc284, line 231: - Maybe rename this method to `ToLongPath" or something so it's more obvious what it does.
- Wouldn't it be easier on the eye if we used a verbatim string @""? :dizzy_face:
- I've been testing this and I believe it should be `inputPath.Length < 260` - the equality crashes for me (I think it's because of the `<nul>` end character). It's also giving me an additional restriction when I'm trying to create a directory to write a file: in `directory.Lengh < 248`, can we check whether need to take this into account too when adding directories.
- There are some unit tests around the ArchiveIterator/Writer; given the nature of the change, it may be a good idea to see if we can add some cases.

kc284 (changes requested):

XenCenterLib/StringUtility.cs
CitrixChris: +1 on name change

XenCenterLib/Archive/ArchiveWriter.cs
```diff
@@ -52,29 +52,51 @@ public virtual void SetBaseStream(Stream outputStream)
...
+            // AllDirectories options because in .NET 4.8 they do not enumerate all elements if they
+            // have paths longer than 260 characters (240 for directories).
+            // If moving to .NET Core, please consider reverting this .
+            AddFiles(StringUtility.ToLongWindowsPath(pathToArchive), cancellingDelegate);
+            var directories = Directory.GetDirectories(StringUtility.ToLongWindowsPath(pathToArchive));
+            for (var j = 0; j < directories.Length; j++)
+            {
+                string dirPath = directories[j];
+                cancellingDelegate?.Invoke();
+                AddDirectory(CleanRelativePathName(pathToArchive, dirPath), Directory.GetCreationTime(dirPath));
```
kc284, line 68: I think these methods can be compacted a bit to one recursive method. Could we perhaps rename the existing abstarct AddDirectory to AddDirectoryCore and then create a new private one which will be calling AddDirectoryCore (upon condition, as I understand that we don't want to include the top directory), then the content of AddFiles, then the recursive AddDirectory on each subfolder. The cancelling delegate should be easy to call in there, although calculating the progress may be slightly trickier.

CitrixChris (approved):

CitrixChris (approved):

CitrixChris (approved):

kc284 (approved):

PR #3192, opened 2023-08-08
Update `InvisibleMessages`

danilo-delbusso:
Also removed `FCOE_SR_DEPRECATION_URL` as it's now simply pointing to `DEPRECATION_URL`


kc284 (approved):
Presumably you're just cleaning the localised resx fils and not providing full localisation for all entries because not all en-us links are overriden, but this is ok since all resx will need to be reviewed anyway if we ever do localisation again.

danilo-delbusso:
> Presumably you're just cleaning the localised resx fils and not providing full localisation for all entries because not all en-us links are overriden, but this is ok since all resx will need to be reviewed anyway if we ever do localisation again.

yup ideally they'd all have overrides. at this point I was more concerned with keeping differences we know of, and cleaning the rest

CitrixChris (approved):

PR #3195, opened 2023-08-11
CA-381102 and other corrections

kc284:
Please review each commit separately. The one about the supp-packs is very invasive.


CitrixChris (approved):

danilo-delbusso (approved):
All good just wanted to mention it'd be great if we could pre-select hosts in the Configure Updates dialog when it's opened from their respective context menu

PR #3199, opened 2023-08-15
CA-381500: Fix `misc` instances of Unlicensed wording showing for post Nile host/pools

danilo-delbusso:
Also rename section to "Expiry status" to better reflect its meaning

### DIFF
--- a/XenAdmin/TabPages/GeneralTabLicenseStatusStringifier.cs
+++ b/XenAdmin/TabPages/GeneralTabLicenseStatusStringifier.cs
@@ -47,8 +47,8 @@ namespace XenAdmin.TabPages
         {
             get
             {
-                if (Status != null && Status.LicensedHost != null && Status.LicenseExpiresIn != null
-                    && !LicenseStatus.IsInfinite(Status.LicenseExpiresIn))
+                if (Status?.LicensedHost != null && Status.LicenseExpiresIn != new TimeSpan() 
+                                                 && !LicenseStatus.IsInfinite(Status.LicenseExpiresIn))
                 {
                     return HelpersGUI.DateTimeToString(Status.LicensedHost.LicenseExpiryUTC().ToLocalTime(),
                         Messages.DATEFORMAT_DMY_LONG, true);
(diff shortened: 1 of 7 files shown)


XenAdmin/TabPages/GeneralTabLicenseStatusStringifier.cs
```diff
@@ -74,20 +74,20 @@ public string ExpiryStatus
                         return Messages.LICENSE_UNLICENSED;
 
                     if (Status.CurrentState == LicenseStatus.HostState.Free)
-                        return Messages.LICENSE_UNLICENSED;
+                        return Helpers.NileOrGreater(Status.LicensedHost) ? Messages.GENERAL_UNKNOWN : Messages.LICENSE_UNLICENSED;
```
kc284, line 77: Is this unkown or never?

danilo-delbusso: oh what about naming the field "License Status" instead of "Expiry Status"?


That way I can keep "Licensed" and "Not Found", and use "Trial License" for post Nile?

kc284: I presume the reason why the original key was "Status" is because it's already under the licence details title, so it's known we refer to Licence status. Also note that it's Status on the licence manager too, so it's probably good to use uniform terminology.

"Not found" might imply "error trying to deduce the status" - Trial License sounds better, is the change too big/risky?

danilo-delbusso: > I presume the reason why the original key was "Status" is because it's already under the licence details title, so it's known we refer to Licence status. 

good point. will revert to Status

> "Not found" might imply "error trying to deduce the status" - Trial License sounds better, is the change too big/risky?

No I'll just add a new string for post Nile, rest will stay the same 👍

kc284 (approved):

neatleygit (approved):

PR #3200, opened 2023-08-15
CA-381502: Show the disabled reason in a new column. Extended the messages shown.

kc284:
Please review each commit separately. I recommend disabling whitespace diffs.

### DIFF
--- a/XenModel/Messages.resx
+++ b/XenModel/Messages.resx
@@ -10486,6 +10486,15 @@ This will cancel the update process and may leave your system in an unstable sta
   <data name="PATCHINGWIZARD_SELECTSERVERPAGE_CANNOT_INSTALL_UPDATE_COORDINATOR_POST_7_0" xml:space="preserve">
     <value>Cannot install updates on this host because the coordinator is running a version higher than {0}</value>
   </data>
+  <data name="PATCHINGWIZARD_SELECTSERVERPAGE_CDN_NOT_SYNCHRONIZED" xml:space="preserve">
+    <value>You have never synchronized with the update channel.</value>
+  </data>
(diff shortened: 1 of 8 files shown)


XenModel/Messages.resx
```diff
@@ -10486,6 +10486,15 @@ This will cancel the update process and may leave your system in an unstable sta
...
     <value>Cannot install updates on this host because the coordinator is running a version higher than {0}</value>
   </data>
+  <data name="PATCHINGWIZARD_SELECTSERVERPAGE_CDN_NOT_SYNCHRONIZED" xml:space="preserve">
+    <value>You have never synchronized with the update channel.</value>
+  </data>
+  <data name="PATCHINGWIZARD_SELECTSERVERPAGE_CDN_REPOS_NOT_CONFIGURED" xml:space="preserve">
+    <value>You have not configured an update channel.</value>
+  </data>
+  <data name="PATCHINGWIZARD_SELECTSERVERPAGE_CDN_UPDATES_APPLIED" xml:space="preserve">
+    <value>All updates available at the last synchronization have been applied.</value>
```
danilo-delbusso, line 10496: WB just `All available updates have been applied.`? In general I'd shorten it because with default sizes the message is always ellipsized

danilo-delbusso: I understand adding "at the last synchronization" is more accurate but maybe we don't need to be that verbose here

kc284: I think it is important to make the user understand what we mean by available updates. It does appear ellipsized, but it probably doesn't matter very much because they can expand the column or the window to see the rest of the message - at least now they can see something is there instead of on an invisible tooltip.

danilo-delbusso: cool well happy to keep it this way anyway 👍

danilo-delbusso (approved):

neatleygit (approved):

PR #3202, opened 2023-08-17
Wording change; corrections for supp-pack installation.

kc284:
The request for the wording change came from docs.
The other two commits are issues I noticed while fixing CA-381225 (a little of it was caused by the changes in #3195).
Please review each commit separately and with whitespace diff disabled.


danilo-delbusso (approved):
nice though I don't understand why we call it `UpdateType.Legacy`. 😕

neatleygit (approved):

PR #3204, opened 2023-08-18
CA-381212, CA-381664, CA-381215, CP-44618

kc284:


kc284:
I added one more commit. It's irrelevant to updates from CDN, but it was needed because the  space requiremetns for the HA hearbeat SR have now increased and the info was outdated. I removed it completely, so we don't have to update it again. The user can always find more details in the docs.

CitrixChris (approved):

XenModel/Actions/Updates/CheckForCdnUpdatesAction.cs
```diff
@@ -67,15 +78,39 @@ protected override void Run()
...
+                             "no_repository_enabled or invalid_updateinfo_xml or invalid_repomd_xml. " +
+                             "See the server side logs for details.");
+                }
+                else if (ex.Message.Contains("Received error code HTTP/1.1 500 Internal Server Error\r\n from the server"))
+                {
+                    log.Warn("Failed to retrieve available updates.");
+                    log.Warn(ex.Message);
+                    log.Warn("One of the following errors has occurred: " +
+                             "get_host_updates_failed or updates_require_sync. " +
+                             "See the server side logs for details.");
```
danilo-delbusso, line 109: should we avoid listing errors? they could change in the future. I'd just leave "see the server side logs for details". same for 404s above

danilo-delbusso (changes requested):

XenModel/Actions/Updates/CheckForCdnUpdatesAction.cs
kc284: I have raised CP-44822 for a proper fix, so I epxect this to change aagian at some point. Until then, I only listed them lest someone complains the application logs show nothing. If you think it is not necessary, I can remove them, up to you.

danilo-delbusso: >  I only listed them lest someone complains the application logs show nothing.

I think the warnings in general are very helpful and we should have them in the XC logs. I just want to remove the list of possible errors. They could become outdated and there's no way we'll remember to update them. Let's just point people to xensource.log

danilo-delbusso (approved):

PR #3211, opened 2023-08-29
CP-44767, CP-44766 & CP-44765: Refactor usage of VM restrictions and add `reference_label` to OVFs

danilo-delbusso:
Please read the following before reviewing 👇 

I suggest you review this PR this way:

1. Tidy up commits, one at a time. They should be fairly straightforward.
2. For the others, still one at a time but please hold comments until the last one, since I refactored quite a few things then. You could review them together but I moved methods a bit in e9d3c9555d7c28e9f023cc478e9ba8d72c5ad3b4 so it won't be super easy I'm afraid.

What this PR is effectively doing is ignoring the VM's `restrictions` in its `recommendations` parameter. It is switching to checking the value in the VM templates for the host we're connected to. 

If possible, it returns the value of the template that matched the VM (using `reference_label`). If that cannot be done, it will either return a default value or fetch the max/min one from the list of all templates.

> Why from all templates?

We need to set a default fall-back value for these fields. We could just update the constants in `VM.cs`. However, it was pointed out to me that the max or min values across all templates are usually what we would define as the fall-back values, anyway.

Also, this means that these defaults will change according to which host the VM is residing on, which is an extra added benefit of doing it this way.

### DIFF
--- a/XenModel/XenAPI-Extensions/VM.cs
+++ b/XenModel/XenAPI-Extensions/VM.cs
@@ -284,7 +284,11 @@ namespace XenAPI
         }
 
         /// <summary>
-        /// Fetch a restriction value from all template restrictions,
+        /// Attempt to fetch a restriction value from a VM template with the same <see cref="VM.reference_label"/>
+        /// as the input VM. This ensures that restriction values are not limiting guest capabilities when restrictions change
+        /// across server upgrades. See CP-44766 for more info.
+        /// <br />
+        /// If the a matching template can't be found, we get the value from all template restrictions,
         /// falling back to a default  if none is found.
         /// This can be used to fetch limits for resource counts on the VM&apos;s host.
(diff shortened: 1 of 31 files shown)


XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -158,7 +158,7 @@ public enum VmTemplateType
         ///     List of distros that we treat as Linux/Non-Windows (written in the VM.guest_metrics
         ///     by the Linux Guest Agent after evaluating xe-linux-distribution)
         /// </summary>
-        private static readonly string[] _linuxDistros =
+        private static readonly string[] LinuxDistros =
```
CitrixChris, line 158: These change is against our coding standard

CitrixChris (changes requested):

kc284 (approved):

PR #3224, opened 2023-09-11
CA-382850: Catch exception when listing pipe names with invalid characters

danilo-delbusso:
Pipes can be created with invalid characters in their names (such as `|`). This results in an exception being thrown when those files are accessed via standard API calls such as `Directory.GetFiles`. Other processes might create these pipes and inadvertently prevent XenCenter from starting altogether.

### DIFF
--- a/XenCenterLib/NamedPipes.cs
+++ b/XenCenterLib/NamedPipes.cs
@@ -266,7 +266,38 @@ namespace XenCenterLib
             public static bool ExistsPipe(string pipePath)
             {
                 log.Debug($"Checking {pipePath} exists");
-                return Directory.GetFiles(@"\\.\pipe\").Contains(pipePath);
+
+                // CA-382850: We iterate manually in order to catch ArgumentExceptions
+                // when listing files. Pipes can be created with invalid characters in 
+                // their names. This throws exception when those files are accessed.
+                // Other processes might create these pipes and inadvertently prevent
+                // XenCenter from starting
+
+                var e = Directory.EnumerateFiles(@"\\.\pipe\");
+                using (var enumerator = e.GetEnumerator())
+                {
+                    while (true)
+                    {
+                        try
+                        {
+                            if (!enumerator.MoveNext())
+                            {
+                                break;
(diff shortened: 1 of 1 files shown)


XenCenterLib/NamedPipes.cs
```diff
@@ -266,7 +266,39 @@ public Pipe(string path)
...
+                // XenCenter from starting
+
+                var e = Directory.EnumerateFiles(@"\\.\pipe\");
+                using (var enumerator = e.GetEnumerator())
+                {
+                    while (true)
+                    {
+                        try
+                        {
+                            if (!enumerator.MoveNext())
```
CitrixChris, line 283: Does movenext() cause the invalid characters exception to trigger? 
If not then you can change the loop so its while(enumerator.movenext())

CitrixChris (changes requested):

XenCenterLib/NamedPipes.cs
danilo-delbusso: yeah it throws it when the enumerator yields back the file, that's why I couldn't do a simple `foreach` + character check here :(

XenCenterLib/NamedPipes.cs
```diff
@@ -266,7 +266,38 @@ public Pipe(string path)
...
                 log.Debug($"Checking {pipePath} exists");
-                return Directory.GetFiles(@"\\.\pipe\").Contains(pipePath);
+
+                // CA-382850: We iterate manually in order to catch ArgumentExceptions
+                // when listing files. Pipes can be created with invalid characters in 
+                // their names. This throws exception when those files are accessed.
+                // Other processes might create these pipes and inadvertently prevent
+                // XenCenter from starting
+
+                var e = Directory.EnumerateFiles(@"\\.\pipe\");
```
kc284, line 276: Doesn't this throw ArgumentException if the files have invalid characters?

kc284 (approved):

CitrixChris (approved):

PR #3242, opened 2023-10-24
CA-384191: Fix missing values for min VM restrictions

danilo-delbusso:
- Min and max should also include the defaults when looking across all templates
- Ensure exception isn't hit unnecessarily when parsing restriction value for VMs
- Allow use of `GetRestrictionValueFromMatchingTemplate` with template objects
- Parallelize `GetRestrictionValueAcrossTemplates` call

### DIFF
--- a/XenModel/XenAPI-Extensions/VM.cs
+++ b/XenModel/XenAPI-Extensions/VM.cs
@@ -260,7 +260,10 @@ namespace XenAPI
         /// <returns>The value if found. If not found, null is returned instead</returns>
         private static T? GetRestrictionValueFromMatchingTemplate<T>(VM vm, string field, string attribute) where T : struct
         {
-            if (!vm.is_a_template && !string.IsNullOrEmpty(vm.reference_label))
+            if(vm.is_a_template)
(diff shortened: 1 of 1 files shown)


XenModel/XenAPI-Extensions/VM.cs
```diff
@@ -343,6 +354,7 @@ public int MinVCPUs()
                 return (T) value;
 
             var templateValues = GetRestrictionValueAcrossTemplates<T>(vm, field, "max");
+            templateValues.Add(defaultValue);
             return templateValues.Count == 0 ? defaultValue : templateValues.Max();
```
CitrixChris, line 358: surely templateValues will never be none if you just added to it. Same with line 377

CitrixChris (changes requested):

XenModel/XenAPI-Extensions/VM.cs
danilo-delbusso: good point 😓

CitrixChris (approved):

kc284 (approved):

PR #3243, opened 2023-10-24
CP-45226: Added friendly name for new datasource 'hostload'

kc284:


XenModel/Utils/Helpers.cs
```diff
@@ -671,6 +671,7 @@ public static string ToStringI18N(this DataSourceCategory category)
         static Regex XapiMemoryRegex = new Regex("^xapi_(allocation|free_memory|live_memory|memory_usage)_kib$");
         static Regex StatefileLatencyRegex = new Regex("^statefile/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}/latency$");
         static Regex LoadAvgRegex = new Regex("loadavg");
+        static Regex HostloadRegex = new Regex("hostload");
```
danilo-delbusso, line 674: nit: `HostLoadRegex`

CitrixChris (approved):

XenModel/Utils/Helpers.cs
kc284: Fixed.

XenModel/FriendlyNames.resx
```diff
@@ -1856,6 +1856,9 @@
   <data name="Message.name-wlb_vm_relocation" xml:space="preserve">
     <value>WLB VM Relocation</value>
   </data>
+  <data name="OPERATIONAL_METRICS_HOSTLOAD" xml:space="preserve">
+    <value>CPU Load</value>
```
DeliZhangX, line 1860: Looks this will show on UI. So I suggest to use "Host Load" or "Host CPU Load" that keeps same style to `Control Domain Load` for `loadavg`

XenModel/Utils/Helpers.cs
```diff
@@ -706,6 +707,9 @@ public static DataSourceCategory GetDataSourceCategory(string name)
             if (LoadAvgRegex.IsMatch(name))
                 return DataSourceCategory.LoadAverage;
 
+            if (HostloadRegex.IsMatch(name))
```
danilo-delbusso, line 710: 👀 also here and on line 906 please

danilo-delbusso (changes requested):

XenModel/FriendlyNames.resx
kc284: Yeah, I was wondering about that, then I thought I'd drop the "Host" bit because we don't use it in the other host CPU datasources, but I'll add it back. Thanks.

XenModel/Utils/Helpers.cs
kc284: I left a diff out 🤦‍♀️

XenModel/FriendlyNames.resx
kc284: Since it's similar to `loadavg` I also changed the category type shown by the UI.

PR #3244, opened 2023-10-25
CA-379971: Ensure `ParallelAction` is not waiting on `_lock` when no actions are running

danilo-delbusso:
One could also fix this by not running actions as soon as the `EnqueueAction` is called (or maybe calling `Monitor.Wait` before the actions have started), but I didn't want to touch a lot of this code and potentially add new issues.

_______

The issue we're fixing here is that there is a chance that all actions have completed before we hit the `Wait` call, we need to make sure we don't hit a deadlock.

This can happen if for instance there is only one action, and it is a "dummy" action, such as the one used in the EUA check.

Also I have removed the compound assignment for the `volatile _completedActionsCount` since Visual Studio was flagging it as a "suspicious usage of a volatile variable". I personally don't think it's a problem but better safe than sorry.

Contains minor whitespace fixes, too

### DIFF
--- a/XenModel/Actions/ParallelAction.cs
+++ b/XenModel/Actions/ParallelAction.cs
@@ -57,7 +57,7 @@ namespace XenAdmin.Actions
         private readonly int _maxNumberOfParallelActions;
         private readonly int _actionsCount;
         private readonly object _lock = new object();
-        private volatile int _completedActionsCount ;
+        private volatile int _completedActionsCount;
 
         /// <summary>
         /// Use this constructor to create a cross connection ParallelAction.
@@ -124,7 +124,10 @@ namespace XenAdmin.Actions
 
             lock (_lock)
             {
-                Monitor.Wait(_lock);
+                // CA-379971: There is a chance that all actions have completed before we
+                // hit the Wait call, we need to make sure we don't hit a deadlock
+                if (_completedActionsCount != _actionsCount)
+                    Monitor.Wait(_lock);
             }
         }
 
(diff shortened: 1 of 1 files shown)


XenModel/Actions/ParallelAction.cs
```diff
@@ -180,7 +183,7 @@ private void action_Completed(ActionBase sender)
             sender.Completed -= action_Completed;
             lock (_lock)
             {
-                _completedActionsCount++;
+                _completedActionsCount = _completedActionsCount + 1;
```
kc284, line 186: I think this is overzealous flagging on behalf of VS, possibly because it does not detect that the assignement is within a lock. The ++ incerement is indeed a compound operation because first it reads the value, then it adds one, then writes it, but since it is happening within the lock, I don't see what can interrupt it. `_completedActionsCount = _completedActionsCount + 1;` is the same thing, it only differs syntactically. The real correction would be to use Interlocked.Increment, but as I said, I don't see why it is needed since we have locked and the line cannot be accessed by another thread.

danilo-delbusso: 100, completely agree. I just made the change to remove the warning, since it's not actually changing the logic in any way.

Either way, it's only visual. Either we have slightly more verbose code, or we get a warning in the editor. I prefer having the former and getting rid of the latter, but it's just preference ofc.

kc284 (approved):

CitrixChris (approved):

PR #3248, opened 2023-11-01
CP-43651, one inconsistency, layout tweaks, and minor refactoring on MultipleDvdIsoList.

kc284:
The inconsistency is in commit 81a269522c66c774afdb2b52a6fec5a51ffc6945
Commits best reviwed separately.

### DIFF
--- a/XenAdmin/Controls/MultipleDvdIsoList.cs
+++ b/XenAdmin/Controls/MultipleDvdIsoList.cs
@@ -169,33 +169,14 @@ namespace XenAdmin.Controls
                 }
             }
 
-            if (comboBoxDrive.Items.Count == 0)
-            {
-                comboBoxDrive.Visible = false;
-                cdChanger1.Visible = false;
-                labelSingleDvd.Visible = false;
-                linkLabelEject.Visible = false;
-                newCDLabel.Visible = VM != null && !VM.is_control_domain;
-                
-            }
-            else if (comboBoxDrive.Items.Count == 1)
-            {
-                comboBoxDrive.Visible = false;
-                cdChanger1.Visible = true;
+            labelSingleDvd.Visible = comboBoxDrive.Items.Count == 1;
+            if (labelSingleDvd.Visible)
                 labelSingleDvd.Text = comboBoxDrive.Items[0].ToString();
-                labelSingleDvd.Visible = true;
-                tableLayoutPanel1.ColumnStyles[0].Width = labelSingleDvd.Width;
-                newCDLabel.Visible = false;
-                linkLabelEject.Visible = true;
-            }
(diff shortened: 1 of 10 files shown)


CitrixChris (approved):

XenAdmin/Controls/MultipleDvdIsoList.cs
danilo-delbusso, line 1: I wouldn't usually bother with something like this but since the PR contains layout tweaks:

While controls such as the Eject link are properly aligned within their row, the DVD drive row is not vertically aligned in respect to the blue bar anymore (top padding looks smaller than bottom paddin), so we just need to adjust that I'd say.

Before:
image

After:
image

XenAdmin/Controls/MultipleDvdIsoList.cs
```diff
@@ -251,17 +251,26 @@ private void comboBoxDrive_SelectedIndexChanged(object sender, EventArgs e)
...
 
-                using (var dlg = new ActionProgressDialog(createDriveAction, ProgressBarStyle.Marquee))
-                    dlg.ShowDialog(this);
+            if (VM.IsHVM())
+            {
+                using (var dialog = new WarningDialog(
+                           string.Format(Messages.NEW_DVD_DRIVE_CREATE_CONFIRMATION, VM.Name()),
+                           new ThreeButtonDialog.TBDButton(Messages.NEW_DVD_DRIVE_CREATE_YES_BUTTON, DialogResult.Yes, ThreeButtonDialog.ButtonType.ACCEPT, true),
+                           ThreeButtonDialog.ButtonNo))
+                    if (dialog.ShowDialog(Program.MainWindow) != DialogResult.Yes)
```
danilo-delbusso, line 246: I understood that we wanted the confirmation dialog in all cases here. I thought we were adding it to avoid users inadvertently creating things in general. Not sure if the hot-plugging issue applies to VMs without `HVM_boot_policy` but we can just show a generic confirmation dialog anyway there. thoughts?

danilo-delbusso (changes requested):

XenAdmin/Controls/MultipleDvdIsoList.cs
kc284: I added the check because in CreateCdDriveAction the VBD is created with `empty=true` and later VbdCreateAndPlugAction does nothing if VBD.empty and the VM is not HVM - to my understanding if it's not HVM it won't try to hot plug it, so no need to say it can't be hot unplugged. Note also that in recent servers all VMs are HVMs, so it is less likely that the user will have a non-HVM.

danilo-delbusso (approved):

PR #3249, opened 2023-11-01
CA-384579: Call previous NRPE plugin xapi to resolve certs permission issue

BengangY:
There is a bug caused by [CA-384572](https://issues.citrite.net/browse/CA-384572). NRPE service can't start after RPU from Yangtze to XS 8 Stream.
The certificate used by NRPE which is generated in Yangtze has no permission. This leads the NRPE operation command (start, restart, etc) failed.
A workaround is that XenCenter sends 'control' command before NRPE enabled. ('control' will modify the permission of certificate)

### DIFF
--- a/XenModel/Actions/NRPE/NRPEUpdateAction.cs
+++ b/XenModel/Actions/NRPE/NRPEUpdateAction.cs
@@ -28,9 +28,9 @@
  * SUCH DAMAGE.
  */
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using XenAPI;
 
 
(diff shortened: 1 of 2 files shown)


XenModel/Actions/NRPE/NRPEUpdateAction.cs
```diff
@@ -104,6 +104,23 @@ private void SetNRPEConfigureForHost(IXenObject o)
...
         {
+            if (enableNRPE)
+            {
+                Dictionary<string, string> enableDict = new Dictionary<string, string>
+                {
+                    { "enable", "true" }
+                };
+                try
+                {
+                    string controlResult = Host.call_plugin(host.Connection.Session, host.opaque_ref, NRPEHostConfiguration.XAPI_NRPE_PLUGIN_NAME,
```
danilo-delbusso, line 115: I can see future devs wondering why this was added. Let's make their lives easier by adding a comment here with the ticket number to tell them this is needed without having to `git blame`.

something like

```
// CA-384579: If the certificate used by NRPE was generated in Yangtze it has no permissions. 
// We prevent command (start, restart, etc) failures by calling control, which modifies certificate permissions.
```

danilo-delbusso (changes requested):
only nit picky one

XenModel/Actions/NRPE/NRPEUpdateAction.cs
BengangY: Added it.

danilo-delbusso: wasn't aware that we were planning to remove it in the future! maybe we didn't need the comment, then. but doesn't hurt. ty!

danilo-delbusso (approved):

XenModel/Actions/NRPE/NRPEUpdateAction.cs
```diff
@@ -104,6 +104,28 @@ private void SetNRPEConfigureForHost(IXenObject o)
 
         private void SetNRPEStatus(IXenObject host, bool enableNRPE)
         {
+            // Workaround for CA-384579 (caused by CA-384572):
+            // If the certificate used by NRPE was generated in Yangtze it has no permissions. 
+            // We prevent command (start, restart, etc) failures by calling control, which modifies certificate permissions.
+            // After CA-384572 is resolved we will remove this workaround code.
```
kc284, line 110: This would probably need a version check instead of removal, so it works also for those that delay installing the updates with the fix to said ticket.

kc284:
Adding the "blocked" label until we decide whether the fix is needed.

kc284 (approved):

PR #3252, opened 2023-11-02
Small corrections to repo syncing and the behaviour of some gridview rows

kc284:

### DIFF
--- a/XenAdmin/Core/Updates.cs
+++ b/XenAdmin/Core/Updates.cs
@@ -120,6 +120,17 @@ namespace XenAdmin.Core
             CdnUpdateInfoChanged?.Invoke(connection);
         }
 
+        public static SyncWithCdnAction CreateSyncWithCdnAction(Pool pool)
+        {
+            var syncAction = new SyncWithCdnAction(pool);
+            syncAction.Completed += a =>
+            {
+                if (a.Succeeded)
+                    CheckForCdnUpdates(a.Connection);
+            };
+            return syncAction;
+        }
+
         public static void CheckForCdnUpdates(IXenConnection connection, bool runSynchronous = false)
         {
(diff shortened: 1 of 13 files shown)


CitrixChris (approved):

XenAdmin/Core/Updates.cs
```diff
@@ -120,6 +120,17 @@ public static void RemoveCdnInfoForConnection(IXenConnection connection)
...
 
+        public static SyncWithCdnAction CreateSyncWithCdnAction(Pool pool)
+        {
+            var syncAction = new SyncWithCdnAction(pool);
+            syncAction.Completed += a =>
+            {
+                if (a.Succeeded)
+                    CheckForCdnUpdates(a.Connection);
+            };
+            return syncAction;
```
danilo-delbusso, line 131: Is there an instance whereby we sync and not check for updates? Since we always run `CheckForCdnUpdates` after `SyncWithCdnAction`, should we just be adding `CheckForCdnUpdatesAction` in `SyncWithCdnAction:Run`, or is that a side effect?

Usually I'd be against having one action do two things, but I don't see why we'd run sync without the check afterwards, so it sort of is the same action.

danilo-delbusso (approved):
I'm approving because in general the PR can go in like this, my only comment is more of a nit-picky question so feel free to ignore

XenAdmin/Core/Updates.cs
kc284: The only reason is that `SyncWithCdnAction` has to call `CheckForCdnUpdates`, not `CheckForCdnUpdatesAction`, and the former uses concepts (the Registry) which belong to XenAdmin, not XenModel, and it is not straight forward, if feasible, to move it because it blurrs the interface between business and UI logic (which is already not very clean in some areas).

PR #3256, opened 2023-11-11
IH-414: Add warning in XCM management dialog to notice unsupported guest

BengangY:
New XCM version (8.3.1 or later):
image
image
image

Old XCM version:
image

### DIFF
--- a/XenAdmin/Wizards/ConversionWizard/VmSelectionPage.cs
+++ b/XenAdmin/Wizards/ConversionWizard/VmSelectionPage.cs
@@ -40,10 +40,12 @@ namespace XenAdmin.Wizards.ConversionWizard
 {
     public partial class VmSelectionPage : XenTabPage
     {
-        private static readonly log4net.ILog log = log4net.LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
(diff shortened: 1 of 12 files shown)


XenAdmin/Wizards/ConversionWizard/VmSelectionPage.cs
```diff
@@ -139,17 +166,72 @@ private void BulkCheck(bool check)
...
+            showOnlySupportedGuestCheckBox.Enabled = enabled;
+        }
+
+        private bool IsSupportedGuest(VmInstance vm)
+        {
+            if (vm.SupportedXSVersions == null)
+            {
+                return true;
+            }
+            foreach (string supportedVersion in vm.SupportedXSVersions)
```
kc284, line 211: This has confused me a bit: `SupportedXSVersions` are the XenServer versions where the guest is supported, right? Then, should we not compare it to the version of the _Pool_ where the VPX is running instead of the VPX version?

kc284 (changes requested):

XenAdmin/Wizards/ConversionWizard/VmSelectionPage.cs
BengangY: It should be compared XenServer version instead of the VPX version. Have already fixed.

PR #3257, opened 2023-11-13
CA-368247 and CA-379225

kc284:


XenCenterLib/Win32.cs
```diff
@@ -508,36 +508,6 @@ public enum ScrollState
         [DllImport("Kernel32", SetLastError = true, CharSet = CharSet.Unicode)]
         public static extern bool GetDiskFreeSpaceEx(string lpszPath, ref long lpFreeBytesAvailable, ref long lpTotalNumberOfBytes, ref long lpTotalNumberOfFreeBytes);
 
-        /// <summary>
-        /// Will return null if the disk parameters could not be determined.
-        /// </summary>
-        /// <param name="path">An absolute path</param>
-        /// <exception cref="Exception">May be thrown</exception>>
-        public static DiskSpaceInfo GetDiskSpaceInfo(string path)
```
kc284, line 513: I moved this method closer to the file it is callwed from because I don't think it can be used in a wider context (attempt to use it in ExportSelectVMsPage.cs instead of Win32.GetDiskFreeSpaceEx causes a crash).

PR #3262, opened 2023-12-04
CA-386317: Add localization support for data-source descriptions

danilo-delbusso:
Also contains cleanup. In general, I tried to only have one method to override data source names, so it's in `Helpers`.

### DIFF
--- a/XenModel/Utils/Helpers.cs
+++ b/XenModel/Utils/Helpers.cs
@@ -719,12 +719,34 @@ namespace XenAdmin.Core
             return DataSourceCategory.Custom;
         }
 
+        private static string GetDataSourceId(string originalId)
+        {
+            switch (originalId)
+            {
+                case "memory_free_kib":
+                    return "memory_used_kib";
+                case "memory_internal_free":
+                    return "memory_internal_used";
+                default:
+                    return originalId;
+            }
+        }
+
+        public static string GetFriendlyDataSourceDescription(string name, IXenObject iXenObject)
+        {
+            var id = GetDataSourceId(name);
+            // suppressing the assert check since we don't support localization anymore, so there's no need to override all descriptions.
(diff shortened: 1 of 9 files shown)


XenModel/Utils/Helpers.cs
```diff
@@ -719,12 +719,34 @@ public static DataSourceCategory GetDataSourceCategory(string name)
             return DataSourceCategory.Custom;
         }
 
+        private static string GetDataSourceId(string originalId)
+        {
+            switch (originalId)
+            {
+                case "memory_free_kib":
+                    return "memory_used_kib";
```
CitrixChris, line 727: I think these are the wrong way around

CitrixChris (changes requested):

XenModel/Utils/Helpers.cs
danilo-delbusso: The original was performing the same overrides, unless I'm missing something :/ 

`memory_free_kib` 👉 `memory_used_kib`
`memory_internal_free` 👉 `memory_internal_used`

```csharp
if (dataSource.name_label == "memory_free_kib" && xenObject is Host)
                    friendlyName = Helpers.GetFriendlyDataSourceName("memory_used_kib", xenObject);
                else if (dataSource.name_label == "memory_internal_free" && xenObject is VM)
                    friendlyName = Helpers.GetFriendlyDataSourceName("memory_internal_used", xenObject);
                else
                    friendlyName = Helpers.GetFriendlyDataSourceName(dataSource.name_label, xenObject);
```

and

```csharp
if (datasourceName == "memory_free_kib")
                FriendlyName = Helpers.GetFriendlyDataSourceName("memory_used_kib", xo);
            else if (datasourceName == "memory_internal_free")
                FriendlyName = Helpers.GetFriendlyDataSourceName("memory_internal_used", xo);
            else
                FriendlyName = Helpers.GetFriendlyDataSourceName(datasourceName, xo);
```

